PR #34, opened 2020-08-19
Use single browser per process

tulioag:
Grid tests are failing because grid was using a custom initialized browser. 
To avoid the increased complexity in the merge script, I created a [pull-request](https://github.com/vaadin/vaadin-grid-flow/pull/1101) to vaadin-grid-flow making the necessary changes. If that is merged, this pull-request can be simplified.


tulioag:
https://github.com/vaadin/vaadin-grid-flow/pull/1101 is now merged.

PR #43, opened 2020-09-01
Update the script to include the plugin for license bundle

ZheSun88:


scripts/templates/pom-flow-pro.xml
```diff
@@ -0,0 +1,130 @@
...
+                        <phase>process-classes</phase>
+                        <goals>
+                            <goal>manifest</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <instructions>
+                        <Bundle-Name>${project.artifactId}</Bundle-Name>
+                        <Bundle-SymbolicName>${project.groupId}.flow.component.crud</Bundle-SymbolicName>
```
manolo, line 67: there is some problem here, component is used above and in other poms for replacements, and here is crud, I guess it should be something more generic to avoid conflicts

scripts/updateJavaPOMs.js
```diff
@@ -59,7 +83,11 @@ async function consolidatePomParent() {
   });
 }
 async function consolidatePomFlow() {
-  consolidate('pom-flow.xml', `${mod}/${name}-flow/pom.xml`)
+  if (proComponents.includes(componentName)){
```
manolo, line 86: I'd rather have a function `isProComponent` that returns true by opening the LICENSE file and looking for some magic words

manolo (changes requested):

scripts/templates/pom-flow-pro.xml
ZheSun88: yes, updated with using `proComponent`

scripts/updateJavaPOMs.js
ZheSun88: so let us ignore this for now.

manolo (approved):

PR #50, opened 2020-09-07
Align changes with V14 changes

manolo:


build.sh
```diff
@@ -35,37 +37,35 @@ tcStatus() {
...
-$cmd || exit 1
-
-cmd="mvn install -DskipTests -Drelease -B -T C$processors"
-tcLog "Installing flow components - $cmd"
-$cmd || exit 1
-
-if [ -n "$BUILD" ]
+cmd="mvn install -Drelease -B -q -T C$processors"
+tcLog "Unit-Testing and Installing flow components - $cmd"
+sleep 10
```
tulioag, line 54: Do we need to sleep here?

build.sh
```diff
@@ -86,9 +84,38 @@ then
...
 then
+  mode="-Dfailsafe.forkCount=$processors -Dcom.vaadin.testbench.Parameters.testsInParallel=$parallel"
   ### Run IT's in merged module
-  cmd="mvn verify -Drun-it -Drelease -Dcom.vaadin.testbench.Parameters.testsInParallel=1 $args -pl integration-tests"
-  tcLog "Running merged ITs - mvn verify -Drun-it -Drelease -pl integration-tests ..."
+  cmd="mvn verify -B -Drun-it -Drelease $mode $args -pl integration-tests"
+  tcLog "Running merged ITs - mvn verify -B -Drun-it -Drelease -pl integration-tests ..."
   echo $cmd
-  $cmd
+  $cmd 2>&1 | egrep --line-buffered -v \
```
tulioag, line 92: Won't suppressing the output make the build harder to debug?

build.sh
```diff
@@ -35,37 +37,35 @@ tcStatus() {
...
+tcLog "Unit-Testing and Installing flow components - $cmd"
+sleep 10
+$cmd
+if [ $? != 0 ]
 then
-  cmd="mvn test -B -Drun-it -T C$processors"
-  tcLog "Unit-Testing - $cmd"
-  $cmd
+  tcLog "Unit-Testing and Installing flow components (2nd try) - $cmd"
+  sleep 30
```
tulioag, line 59: What are we waiting 30 seconds for?

tulioag (approved):

build.sh
manolo: As said above, without sleep the second time can fail, but the sleep makes the build pass

build.sh
manolo: Failed tests are run a second time, there we have the full output if it fails again.
Removed stuff in this grep is just for removing lines that does not say so much, but just noise.

build.sh
manolo: In this case it seems that does not help, because for some reason this build fails frequently the first time.
But sleeping in the retry seems to help. Removing this.

PR #51, opened 2020-09-08
Update any branch by computing origin branch from versions.json

manolo:


scripts/getVersions.js
```diff
@@ -0,0 +1,46 @@
...
+
+const https = require("https");
+const xml2js = require('xml2js');
+const fs = require('fs');
+
+async function main() {
+  const parentJs = await xml2js.parseStringPromise(fs.readFileSync('pom.xml', 'utf8'));
+  const version = parentJs.project.version[0];  
+  const branch = version.replace(/^(\d+\.\d+).*$/, '$1');
+
```
tulioag, line 19: I  suggest we test if the branch exists by making a HEAD request to `https://api.github.com/repos/vaadin/platform/branches/${version)`. A 404 indicates that it does not exist, so we should use master. That way we don't need to hard code the master value and remember to update it manually.

tulioag (approved):

scripts/getVersions.js
manolo: Done

tulioag (approved):

PR #58, opened 2020-09-09
Set versions for the 17.0 branch

manolo:


ZheSun88 (approved):

ZheSun88 (changes requested):
https://github.com/vaadin/vaadin-flow-components/blob/17.0/vaadin-flow-components-shared/pom.xml#L7

this place needs to be updated too.

ZheSun88 (approved):

PR #59, opened 2020-09-09
Apply changes for the 14.4 branch from 14.3

manolo:


ZheSun88 (commented):
https://github.com/vaadin/vaadin-flow-components/blob/17.0/vaadin-flow-components-shared/pom.xml#L7

this file should be updated too?

PR #71, opened 2020-09-11
Add script for updating @NpmPackage annotation

ZheSun88:


scripts/updateNpmVer.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+grep -r @NpmPackage ../* >> info.json
```
manolo, line 3: This is wrong, all scripts should be run from the parent folder `./scripts/updateNpmVer.sh` also we can reduce more the folders to look for java files `vaadin*/vaadin*-flow/src/main/java`

manolo (changes requested):

scripts/updateNpmVer.sh
manolo: Also redirection should be one `>` to reset content in case the file was left there by accident in a previous run

manolo: Done

manolo (approved):

PR #83, opened 2020-09-14
 Keep versions in 14.4

manolo:


scripts/mergeITs.js
```diff
@@ -60,8 +60,9 @@ async function createPom() {
     const prev = await prevP;
     const id = name.replace('-flow-parent', '');
     // Add component-flow and component-testbench dependencies
-    addDependency(prev, 'com.vaadin', `${id}-flow`, '${project.version}');
-    addDependency(prev, 'com.vaadin', `${id}-testbench`, '${project.version}', 'test');
+    const componentVersion = /^14\.[3-4]/.test(version) ? `\$\{${id.replace(/-/g, '.')}.version\}` : '${project.version}'
```
tulioag, line 63: Should we pass the 14 as a parameter somehow and try to keep this script the same both in master and in this branch?

tulioag (approved):
Minor issue:
- README.md is missing information regarding sauce labs

tulioag (approved):

scripts/mergeITs.js
manolo: This script is going to be run for a couple of weeks, and then we can delete it once that origin branches are frozen. The decision is already that only 14.4 will have maintenance releases based on old setup. I think it is not worth

tulioag (approved):

PR #137, opened 2020-09-26
ci: perform a fast build if the [skip ci] flag is given

### LINKED ISSUE #136: run fast checks when  '[skip ci]' fragment is added to commit message
Run TC fast builds when
- PR title has the `[skip ci]` 
- All commit messages have the `[skip ci]`

manolo:
Fixes #136


build.sh
```diff
@@ -79,6 +84,19 @@ then
   $cmd || tcStatus 1 "Unit-Testing failed"
 fi
 
+tcLog "Checking for skip-ci labels"
+computeFastBuild
+if [ -n "$FAST_BUILD" ]
```
tulioag, line 89: I think it would be more idiomatic if `computeFastBuild` returned `0` for a fast build or `1` otherwise. Then the test would become
```sh
if computeFastBuild
then
...
```
To achieve that, the easiest way in my opinion would be to replace the last line of the function with
```sh
echo "$prMessages" | grep -q  '\[skip ci\]' && return 0 || return 1
```

tulioag (changes requested):

tulioag (approved):

PR #146, opened 2020-09-29
Take DataView concept into use

mshabarov:
Fixes https://github.com/vaadin/vaadin-combo-box-flow/issues/355


manolo (changes requested):
Since this is already being reviewed in https://github.com/vaadin/vaadin-combo-box-flow/pull/380/files I think it's better to fix tests there, then merge it, discard this PR, and eventually let the script that brings changes from original repos to multi-repo to make the modifications that tests might need.

PR #168, opened 2020-10-06
fix: propagate theme attribute to list box on overlay

TatuLund:
Fixes: https://github.com/vaadin/vaadin-select/issues/235


tomivirkki:
Note that this doesn't fix the original issue as the theme attribute isn't propagated to the `<vaadin-item>`'s whose `part="content"` you wanted to style. Please add a test that ensures that the correct thing gets fixed.

yuriy-fix:
Closing as stale.

PR #190, opened 2020-10-07
fix: remove the debouncing of ensureSubCacheForScaledIndex

yuriy-fix:
Web-component: grid

Fixes: https://github.com/vaadin/vaadin-grid/issues/1809

Details: Flow Grid's connector overrides the ensureSubCacheForScaledIndex function, and adds a debouncer for the logic, the data request gets made asynchronously when expanding the Grids item. If the request for recalculateColumnWithds is done on expand listener, the width calculation might occasionally happen before the node gets expanded on the client-side, thus resulting into columns with undesired widths.


manolo:
You need to rebase master so as those unrelated tests pass

manolo:
Seems that these tests always fail with this PR

Screenshot 2020-10-08 at 19 50 10

yuriy-fix:
Closed in favor of https://github.com/vaadin/vaadin-flow-components/pull/241.

PR #228, opened 2020-10-13
Prototype: preserve expansions

tulioag:
fixes vaadin/vaadin-grid#1820


tulioag:
Closed in favor of #363

PR #248, opened 2020-10-14
fix: Keep HelperComponent after reset

tulioag:
Fixes: vaadin/vaadin-checkbox#191

Web-component: checkbox


alvarezguille (approved):
LGTM
@tulioag could you check if RadioButtonGroup would need the same fix and create an issue or PR for it?

PR #275, opened 2020-10-19
fix: override setValue to set proper value when detached

yuriy-fix:
Web-component: radio-button

Fixes: https://github.com/vaadin/vaadin-radio-button/issues/166

Details: RBG automatically adds RadioButtons inside of itself. If the RBG's value gets changed (setValue) while the component isn't in the DOM, the internal RB's remain unchanged. And once they're added back to DOM, Flow synchronizes the old RB values which causes the RBG to revert back to the old value. In order to fix it buttons will be reset on detach.

### DIFF
--- a/vaadin-radio-button-flow-parent/vaadin-radio-button-flow/src/main/java/com/vaadin/flow/component/radiobutton/RadioButtonGroup.java
+++ b/vaadin-radio-button-flow-parent/vaadin-radio-button-flow/src/main/java/com/vaadin/flow/component/radiobutton/RadioButtonGroup.java
@@ -245,6 +245,7 @@ public class RadioButtonGroup<T>
         	dataProviderListenerRegistration = null;
         }
         super.onDetach(detachEvent);
+        reset();
     }
 
     /**
--- a/vaadin-radio-button-flow-parent/vaadin-radio-button-flow/src/main/java/com/vaadin/flow/component/radiobutton/RadioButtonGroup.java
+++ b/vaadin-radio-button-flow-parent/vaadin-radio-button-flow/src/main/java/com/vaadin/flow/component/radiobutton/RadioButtonGroup.java
@@ -231,6 +231,12 @@ public class RadioButtonGroup<T>
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


vaadin-radio-button-flow-parent/vaadin-radio-button-flow/src/main/java/com/vaadin/flow/component/radiobutton/RadioButtonGroup.java
```diff
@@ -245,6 +245,7 @@ protected void onDetach(DetachEvent detachEvent) {
         	dataProviderListenerRegistration = null;
         }
         super.onDetach(detachEvent);
+        reset();
```
tomivirkki, line 248: Using `reset()` seems to also `clear()` the component's value. So on detach, the RBG would have its value set to `null`.

tomivirkki (changes requested):

vaadin-radio-button-flow-parent/vaadin-radio-button-flow/src/main/java/com/vaadin/flow/component/radiobutton/RadioButtonGroup.java
yuriy-fix: You are right, probably I will also add test for default `detach-attach` without `PolymerTemplate` to test this usecase.

yuriy-fix: Added test to cover this usecase as well.

tomivirkki: Screenshot 2020-10-22 at 10 03 49

Not sure why the existing buttons/radio group weren't re-used for the new test. Nothing blocking, but generally the test pages should also be easy to understand.

tomivirkki (approved):

PR #349, opened 2020-11-02
test: reenable ignored tests

manolo:
Related-to: vaadin/vaadin-grid-flow#635 vaadin/vaadin-grid-flow#122 vaadin/flow#4268

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -54,11 +54,12 @@ tcStatus() {
 
 saveFailedTests() {
   try=$1
-  failed=`egrep '<<< ERROR|<<< FAILURE' integration-tests/target/failsafe-reports/*txt | perl -pe 's,.*/(.*).txt:.*,$1,g' | sort -u`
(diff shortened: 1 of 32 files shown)


scripts/build.sh
```diff
@@ -18,6 +20,21 @@ then
   done
 fi
 
+## compute modules that were modified in this PR
+if [ -z "$modules" -a -n "$PR" ]
+then
+  modified=`curl -s https://api.github.com/repos/vaadin/vaadin-flow-components/pulls/$PR/files \
+    | jq -r '.[] | .filename' | grep 'vaadin.*parent' | perl -pe 's,^vaadin-(.*)-flow-parent.*,$1,g' | sort -u`
+  if [ `echo "$modules" | wc -w` -lt 5 ]
```
tulioag, line 28: Won't `$modules` be empty here?

manolo: Because modules is empty, and it's wrapped in quotes it will output zero here which is correct try this : `echo "" | wc -w`

manolo: you're right, it should be `$modified` fixed

tulioag (approved):

PR #468, opened 2020-12-01
fix: Prevent IllegalArgumentException when sortable columns are remov…

TatuLund:
…ed and recreated

Fixes: https://github.com/vaadin/vaadin-grid/issues/1969

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
@@ -2932,6 +2932,7 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
         Objects.requireNonNull(column, "column should not be null");
 
         ensureOwner(column);
+        column.setSortable(false);
         removeColumnAndColumnGroupsIfNeeded(column);
         column.destroyDataGenerators();
         keyToColumnMap.remove(column.getKey());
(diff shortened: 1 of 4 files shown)


tomivirkki:
Missing tests

tomivirkki:
The issue is likely caused by https://github.com/vaadin/vaadin-grid/issues/1691 which means that the `<vaadin-grid> 's` `_sorters` array keeps references to sorters of removed column instances. If that's the case, we should fix the root cause of the problem (in the Web Component)

tomivirkki:
Well...this change might actually still be required to fully fix the issue. If a sorted column is removed from a detached Flow Grid, there's no active Web Component to notify of the change so you might end up with a wrongly synchronized server-side state if the sortOrder isn't updated here.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -2929,6 +2929,8 @@ public void removeColumn(Column<T> column) {
         Objects.requireNonNull(column, "column should not be null");
 
         ensureOwner(column);
+        List<GridSortOrder<T>> order = new ArrayList<>();
+        setSortOrder(order, false);
```
tomivirkki, line 2933: Wouldn't this reset all sorting from the Grid whenever a single column gets removed?

TatuLund: Yes, and I think it is the correct behavior.

hindog:
FYI, I'm using Vaadin 14.4.7 (NPM) and currently hitting this issue.  Is there a work-around for this?  

One attempt I had tried was to try 'hiding' instead of 'removing' the columns with `column.setVisible(false)` followed by`grid.sort(null)` to clear the sort.  Let's say I have column "A" and column "B":

- Sort column "A"
- "Remove" column "A" by doing: `columnA.setVisible(false); grid.clear(null)`
- Sort column "B"
- My data provider will receive sorts for "A" and "B", when I would only expect it for "B" because I did `grid.clear(null)` and hid the column.

tulioag (approved):

PR #478, opened 2020-12-02
fix: Fix setting value to lazy loading ComboBox

### LINKED ISSUE #391: Cannot set a value for a lazy loaded ComboBox
Vaadin 18.0.0.rc1
Vaadin 18.0.0.rc2

`ComboBox<Person> comboBox = new ComboBox<>();`
`ComboBoxLazyDataView<Person> dataView = comboBox.setItems(query -> {`
`	return service.fetch(query).stream();`
`});`
`Person person = service.findById(1);`
`comboBox.setValue(person);`

`comboBox.setValue` throws:
"java.lang.IllegalStateException: Cannot set a value for a ComboBox without items. Use setItems to populate items into the ComboBox before setting a value.
	at com.vaadin.flow.component.combobox.ComboBox.setValue(ComboBox.java:371)"

pleku:
Things should only throw when no items/callback has been set to the CB.
There is a possibility that the value is set but the backend doesn't have
this item available, but that is an edge case and something that should
not happen and should be possible to prevent by resetting the data and
thus a responsibility of the application developer.

Fixes #391

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/ComboBoxTest.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/ComboBoxTest.java
@@ -15,8 +15,6 @@
  */
 package com.vaadin.flow.component.combobox;
 
-import static org.junit.Assert.assertEquals;
-
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -47,6 +45,7 @@ import com.vaadin.flow.server.VaadinSession;
 import com.vaadin.flow.shared.Registration;
 
 import elemental.json.Json;
+import static org.junit.Assert.assertEquals;
 
 public class ComboBoxTest {
 
@@ -305,6 +304,16 @@ public class ComboBoxTest {
                 "Cannot set a value for a ComboBox without items.");
         ComboBox<String> combo = new ComboBox<>();
         combo.setValue("foo");
+    }
+
+    // https://github.com/vaadin/vaadin-flow-components/issues/391
+    @Test
+    public void setValueWithLazyItems_doesntThrow() {
+        final ComboBox<String> comboBox = new ComboBox<>();
+        comboBox.setItems(query -> Stream.of("foo", "bar"));
+        comboBox.setValue("foo");
+
+        Assert.assertEquals("foo", comboBox.getValue());
     }
 
     @Test
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridTest.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridTest.java
@@ -17,7 +17,9 @@
 package com.vaadin.flow.component.grid;
 
 import java.util.Arrays;
+import java.util.stream.Stream;
 
+import org.junit.Assert;
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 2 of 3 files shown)


vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/ComboBoxTest.java
```diff
@@ -307,6 +306,16 @@ public void setValueWithoutItems_throw() {
         combo.setValue("foo");
     }
 
+    // https://github.com/vaadin/vaadin-flow-components/issues/391
+    @Test
+    public void setValueWithLazyItems_doesntThrow() {
+        final ComboBox<String> comboBox = new ComboBox<>();
+        comboBox.setItems(query -> Stream.of("foo", "bar"));
```
mshabarov, line 313: As now we have a `setValue` tests for unknown size and in-memory, is that reasonable to add a similar test for lazy loading with defined size?

Another point is to check that if `setValue` is invoked with "absent-item", it won't throw anything if the items are set to combobox, so let's add a corresponding test.

mshabarov (changes requested):

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridTest.java
```diff
@@ -51,4 +53,13 @@ public void dataViewForFaultyDataProvider_throwsException() {
         grid.getListDataView();
     }
 
+    @Test
+    public void selectItem_lazyDataSet_selectionWorks() {
```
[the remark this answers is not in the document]
pleku: It was just to make sure that grid doesn't suffer the same issue, and I don't mind having a test for that to make sure it never will either

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/ComboBoxTest.java
pleku: >As now we have a setValue tests for unknown size and in-memory, is that reasonable to add a similar test for lazy loading with defined size?

I don't see how the defined/undefined size makes any difference to the `setValue` code, for completeness sake I can do it if there are other changes required

>Another point is to check that if setValue is invoked with "absent-item", it won't throw anything if the items are set to combobox, so let's add a corresponding test.

To be clear, you're referring to the case where the set item is not fetched on the first range ? Or what ? The item is not at all available ?

mshabarov: > The item is not at all available ?

Yes, I consider this case.

mshabarov: Just to verify that no exception is thrown when the the item "baz" is not present

pleku: >Just to verify that no exception is thrown when the the item "baz" is not present

I tried to add what I'd consider a meaningful test case for this, but failed because I cannot figure out a nice way to set the request range for the ComboBox, which is needed if I want to get a fetch triggered. Otherwise it I don't think it is meaningful. I don't want to spend any more time on this as it is irrelevant for the bug in question. Someone (else) can do that in another PR.

mshabarov (approved):

PR #501, opened 2020-12-07
fix: update versions for non vaadin components

manolo:


ZheSun88 (approved):

ZheSun88:
just a reminder: with 14.X branch there is another webjarupdate file. probably also need to be updated.

PR #512, opened 2020-12-08
fix: Fix client side validation overriding

pleku:
The workaround for overriding the client side validation for certain
components was causing an unnecessary roundtrip to the server side,
which also caused two warnings to be logged by Flow each time a disabled
or a readonly component was added to the UI. The warning happened twice
since two executeJS calls were enqueued: one immediately, and another
when attached. Apparently there was no good reason for this behavior.

This fix will only apply the executeJS once when the component is
attached and it will not require a separate roundtrip to workaround the
client side validation overriding the invalid value because the JS
execution is enqueued on the "beforeClientResponse" phase when the
server side invalid value has been resolved and can be used in the JS
execution.

Further improvement would be to move the FieldValidationUtil to its own
module and reused from there, but that is left for later (?).

Fixes vaadin/flow#8848


vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow/src/main/java/com/vaadin/flow/component/datetimepicker/FieldValidationUtil.java
```diff
@@ -31,36 +29,41 @@ private FieldValidationUtil() {
...
-            }
-        });
+        if (component.isInvalid()) {
+            /*
+             * By default the invalid flag is set to false. Workaround the case
+             * where the client side validation overrides the invalid state
+             * before the validation function itself is overridden above.
+             */
+            expression.append("this.invalid = true;");
+            expression.append("this.validate();");
```
DiegoCardoso, line 65: Are these calls to `this.validate()` needed? They were not before this change.

pleku: I added that to basically make sure the invalid state is applied correctly in the component, regardless of it is needed or not.
I have no idea how redundant it is as I don't know the web component(s) that well. This is basically "just in case" setting the attribute would not be enough in some certain case or for some component.

I can remove the part if it should never be needed or add a comment that clarifies that this is "just in case and maybe reduntant" ?

DiegoCardoso: As the patched `validate` only returns `checkValidity` value and this method doesn't produce any side effect, my understanding is that it's not needed.

pleku: Done

pleku:
I have no idea how the build failures (CRUD) could be related to this change 😢

DiegoCardoso:
It's a bit weird, yes. The [build](https://teamcity.vaadin.com/viewLog.html?buildId=70310&buildTypeId=VaadinFlowComponents_PullRequestValidation&tab=buildResultsDiv) for the commit (cb61957) where you made the changes is green.

DiegoCardoso:
I guess I found the source of all evil:

I think [this change](https://github.com/vaadin/vaadin-flow-components/commit/8b5b88893fdf1954ee4eadf83c262d560a0415e6#diff-255f48ed31d114d33fd62d5f44b260202e622606b1f6ffb8fa39fbbc96306a84R355) is breaking `CrudElement` because on `openRowForEditing` method it calls `this.$("vaadin-crud-edit").all().get(row).click();` and the flush is making the `vaadin-crud-edit` that goes inside `caption#sizer` be stamped before the buttons on the rows, so it won't open the editor.

DiegoCardoso (approved):

manolo:
@pleku seems that you didn't sign license agreement.

DiegoCardoso:
@manolo I think CLA is drunk. You can see on previous commits that it was green. Need a way to force the recheck

PR #515, opened 2020-12-09
fix: remove redundant sorter renderer

yuriy-fix:
Fixes https://github.com/vaadin/vaadin-grid/issues/1841.


tulioag (changes requested):
Can we have a test checking for the presence of `vaadin-grid-sorter` ?

yuriy-fix:
There is a test for the presence of `vaadin-grid-sorter` in `SortingIT`. I have adjusted the `SortingPage` in order to cover the case described in the issue.

tulioag (approved):

PR #541, opened 2020-12-18
fix: use the next row for getting crud-edit

yuriy-fix:
Details: with the `_observer.flush` [update](https://github.com/vaadin/vaadin-flow-components/commit/8b5b88893fdf1954ee4eadf83c262d560a0415e6) to the grid's connector, caption#sizer is stamped before the rows in vaadin-crud. It has the first edit button, so we need to get `row + 1` crud-edit when selecting all of those in the crud.

### DIFF
--- a/vaadin-crud-flow-parent/vaadin-crud-testbench/src/main/java/com/vaadin/flow/component/crud/testbench/CrudElement.java
+++ b/vaadin-crud-flow-parent/vaadin-crud-testbench/src/main/java/com/vaadin/flow/component/crud/testbench/CrudElement.java
@@ -19,6 +19,7 @@ package com.vaadin.flow.component.crud.testbench;
 
 import com.vaadin.flow.component.button.testbench.ButtonElement;
 import com.vaadin.flow.component.grid.testbench.GridElement;
+import com.vaadin.flow.component.grid.testbench.GridTHTDElement;
 import com.vaadin.flow.component.textfield.testbench.TextFieldElement;
 import com.vaadin.testbench.ElementQuery;
 import com.vaadin.testbench.TestBenchElement;
@@ -26,6 +27,7 @@ import com.vaadin.testbench.elementsbase.Element;
 
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 /**
  * TestBench element for the vaadin-crud element
@@ -81,7 +83,12 @@ public class CrudElement extends TestBenchElement {
         if (isEditOnClick()) {
             this.getGrid().getCell(row, 0).click();
         } else {
-            this.$("vaadin-crud-edit").all().get(row).click();
+            GridTHTDElement editCell = getGrid().getAllColumns().stream()
+                    .map(column -> getGrid().getRow(row).getCell(column))
+                    .filter(cell -> cell.getInnerHTML()
+                            .contains("vaadin-crud-edit"))
+                    .collect(Collectors.toList()).get(0);
+            editCell.$("vaadin-crud-edit").get(0).click();
         }
     }
 
--- a/vaadin-crud-flow-parent/vaadin-crud-testbench/src/main/java/com/vaadin/flow/component/crud/testbench/CrudElement.java
+++ b/vaadin-crud-flow-parent/vaadin-crud-testbench/src/main/java/com/vaadin/flow/component/crud/testbench/CrudElement.java
@@ -20,6 +20,7 @@ package com.vaadin.flow.component.crud.testbench;
 import com.vaadin.flow.component.button.testbench.ButtonElement;
 import com.vaadin.flow.component.grid.testbench.GridElement;
 import com.vaadin.flow.component.grid.testbench.GridTHTDElement;
+import com.vaadin.flow.component.grid.testbench.GridTRElement;
 import com.vaadin.flow.component.textfield.testbench.TextFieldElement;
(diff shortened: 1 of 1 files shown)


vaadin-crud-flow-parent/vaadin-crud-testbench/src/main/java/com/vaadin/flow/component/crud/testbench/CrudElement.java
```diff
@@ -81,7 +81,7 @@ public void openRowForEditing(int row) {
         if (isEditOnClick()) {
             this.getGrid().getCell(row, 0).click();
         } else {
-            this.$("vaadin-crud-edit").all().get(row).click();
+            this.$("vaadin-crud-edit").all().get(row + 1).click();
```
tomivirkki, line 84: This is flaky, as it relies on the order in which the "vaadin-crud-edit" elements appear in the light DOM of grid and it is in no way deterministic! Even the previous solution had that flaw.

tomivirkki: I wonder if there would be a cleaner solution to first accessing the correct row by the given index and then the edit element on that row.

What really makes me wonder the most is that why the "sizer" row has the edit element in the first place. All the cells in the sizer should be empty. Verified this in the crud [demo](https://cdn.vaadin.com/vaadin-crud/1.3.0/demo/#crud-basic-demos).

DiegoCardoso: Checking the [Java demos](https://vaadin.com/components/vaadin-crud/java-examples), there might be some difference how it works with Flow.

The `GridElement#getRow` method also relies on the DOM to get the `<tr>` element: 
```js
var rowsInDom = grid.$.items.children;
var rowInDom = Array.from(rowsInDom).filter(function(row) { 
  return !row.hidden && row.index == rowIndex;
})[0];
```

tomivirkki: That doesn't rely on the DOM index, but it actually checks the `row.index` property, which is the correct approach

tomivirkki: > Checking the Java demos, there might be some difference how it works with Flow

Seems so. The sizer cells should never have any content.

DiegoCardoso: Ah, okay. My bad.

Couldn't find a nice way to go from the row, since from `GridElement#getRow` we are on the Grid's shadow DOM and need to go back to the light DOM to find the `vaadin-crud-edit` element:
```js
// $0 is the <tr> element
Array.from($0.cells)
  .map(cell => cell.firstElementChild.assignedNodes()[0].children[0])
  .filter(el => el && el.localName == 'vaadin-crud-edit')[0]
```

yuriy-fix: Have tried to go with cleaner solution to get `vaadin-crud-edit`. It has became a bit longer, but now it's based on `row.index`.

As for the sizer cells content, I guess we can create another issue 🤔

tomivirkki: > As for the sizer cells content, I guess we can create another issue

Sounds good. It is a separate bug.

vaadin-crud-flow-parent/vaadin-crud-testbench/src/main/java/com/vaadin/flow/component/crud/testbench/CrudElement.java
```diff
@@ -81,7 +83,12 @@ public void openRowForEditing(int row) {
         if (isEditOnClick()) {
             this.getGrid().getCell(row, 0).click();
         } else {
-            this.$("vaadin-crud-edit").all().get(row).click();
+            GridTHTDElement editCell = getGrid().getAllColumns().stream()
+                    .map(column -> getGrid().getRow(row).getCell(column))
```
DiegoCardoso, line 87: Not sure if this is an premature optimization, but doesn't it make sense to store `getGrid().getRow(row)` in a variable?

yuriy-fix: Done it this way, so it will be obvious what's wrong if you are getting `NoSuchElementException`.

tomivirkki (approved):

vaadin-crud-flow-parent/vaadin-crud-testbench/src/main/java/com/vaadin/flow/component/crud/testbench/CrudElement.java
yuriy-fix: Updated now.

DiegoCardoso (approved):

PR #588, opened 2021-01-25
Fix: Clear subchildren when refreshing item

tulioag:
#Fixes https://github.com/vaadin/vaadin-grid/issues/1819

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -720,21 +720,16 @@ import { ItemCache } from '@vaadin/vaadin-grid/src/vaadin-grid-data-provider-mix
         const generateRange = size => [...Array(size).keys()];
 
         function findDescendants(pkey,pages) {
-          const result = [];
           const parentCache = cache[pkey];
 
-          if(!parentCache) return result;
+          if(!parentCache) return [];
 
           pages = pages ? pages:generateRange(parentCache.length);
-          for(const currentPage of pages) {
-            const items = parentCache[currentPage];
-            if (items) {
-              result.push(...items);
-              const descendants = items.flatMap(item => findDescendants(item.key));
-              result.push(...descendants);
-            }
-          }
-          return result;
+
+          return pages.flatMap(currentPage => {
+            const items = parentCache[currentPage] || [];
+            return items.concat(items.flatMap(item => findDescendants(item.key)));
+          });
         }
 
         const {cache:gridCache, scaledIndex} = grid._cache.getCacheAndIndexByKey(pkey);
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -763,10 +763,8 @@ import { ItemCache } from '@vaadin/vaadin-grid/src/vaadin-grid-data-provider-mix
         const end = length + index;
         const childCache = cache.itemCaches[scaledIndex];
         for(let itemIndex = index; itemIndex < end; itemIndex++) {
-          const key = childCache.items[itemIndex];
           delete childCache.items[itemIndex];
           delete childCache.itemCaches[itemIndex];
-          delete childCache.itemkeyCaches[key];
         }
         cache.updateSize();
       }
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -732,19 +732,24 @@ import { ItemCache } from '@vaadin/vaadin-grid/src/vaadin-grid-data-provider-mix
           });
         }
 
-        const {cache:gridCache, scaledIndex} = grid._cache.getCacheAndIndexByKey(pkey);
-
-        const firstPage = Math.floor(index / grid.pageSize);
         const updatedPageCount = Math.ceil(length / grid.pageSize);
-
+        const firstPage = Math.floor(index / grid.pageSize);
         const pages = generateRange(updatedPageCount).map(i => firstPage + i);
-        const affectedItems = findDescendants(pkey,pages);
-        grid.$connector.doDeselection(affectedItems.filter(item => selectedKeys[item.key]));
         for(const page of pages) {
           delete cache[pkey][page];
         }
 
+        // Use grid.clearCache if clearing the root cache.
+        if(!parentKey) {
+          grid.clearCache();
+          return;
+        }
+
+        const affectedItems = findDescendants(pkey,pages);
+        grid.$connector.doDeselection(affectedItems.filter(item => selectedKeys[item.key]));
+
         const keys = affectedItems.map(item => item.key);
+        const {cache:gridCache, scaledIndex} = grid._cache.getCacheAndIndexByKey(pkey);
         clearSubcache(gridCache,scaledIndex,keys);
       });
 
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -718,24 +718,8 @@ import { ItemCache } from '@vaadin/vaadin-grid/src/vaadin-grid-data-provider-mix
         }
         // Generates an array of integers from 0 to size-1
         const generateRange = size => [...Array(size).keys()];
-        function clearSubcache(cache,scaledIndex,childKeys,page) {
-          const keySet = new Set(keys);
-          Array.from(grid.$.items.children)
-            .flatMap((row) => Array.from(row.children))
-            .forEach((cell) => {
-              const item = cell._instance && cell._instance.item
-              if(item && keySet.has(item.key)) {
-                // Force data system to pick up subproperty changes
-                cell._instance && cell._instance._setPendingProperty('item', {}, false);
-              }
-            });
 
-          delete cache.itemCaches[scaledIndex];
-          delete cache.items[scaledIndex];
-          delete cache.itemkeyCaches[scaledIndex];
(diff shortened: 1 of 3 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -716,23 +716,59 @@ import { ItemCache } from '@vaadin/vaadin-grid/src/vaadin-grid-data-provider-mix
...
 
-        for (let i = 0; i < updatedPageCount; i++) {
-          let page = firstPage + i;
-          let items = cache[pkey][page];
-          grid.$connector.doDeselection(items.filter(item => selectedKeys[item.key]));
+          if(!parentCache) return [];
+
+          pages = pages ? pages:generateRange(parentCache.length);
+
+          return pages.flatMap(currentPage => {
```
web-padawan, line 729: Note: this would need to be refactored when cherry-picking to Vaadin 14 as IE11 does not support `flatMap`.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -716,23 +716,59 @@ import { ItemCache } from '@vaadin/vaadin-grid/src/vaadin-grid-data-provider-mix
...
+        }
+
+        const updatedPageCount = Math.ceil(length / grid.pageSize);
+        const firstPage = Math.floor(index / grid.pageSize);
+        const pages = generateRange(updatedPageCount).map(i => firstPage + i);
+
+        // Uses cache, so needs to run before deleting
+        const affectedItems = findDescendants(pkey,pages);
+
+        for(const page of pages) {
```
web-padawan, line 742: This would need to be refactored when cherry-picking to Vaadin 14 as IE11 does not support `for .. of`.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
tulioag: Replaced with `map` and `reduce`

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
tulioag: Replaced with `forEach`

tomivirkki (changes requested):

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -716,23 +716,62 @@ import { ItemCache } from '@vaadin/vaadin-grid/src/vaadin-grid-data-provider-mix
...
+        // Uses cache, so needs to run before deleting
+        const affectedItems = findDescendants(pkey,pages);
+
+        pages.forEach(page => {
+          delete cache[pkey][page];
+        });
+
+        // Use grid.clearCache if clearing the root cache.
+        if(!parentKey) {
+          grid.clearCache();
```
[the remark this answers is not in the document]
tulioag: Removed

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -716,23 +716,62 @@ import { ItemCache } from '@vaadin/vaadin-grid/src/vaadin-grid-data-provider-mix
...
+      function clearSubcache(cache,scaledIndex, index, length) {
+        const end = length + index;
+        const childCache = cache.itemCaches[scaledIndex];
+        for(let itemIndex = index; itemIndex < end; itemIndex++) {
+          delete childCache.items[itemIndex];
+          delete childCache.itemCaches[itemIndex];
+        }
+        cache.updateSize();
+      }
+
```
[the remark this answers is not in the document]
tulioag: It seemed to make sense removing things that don't exist anymore from both caches. But you are right, it is not necessary. I simplified the code.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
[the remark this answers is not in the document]
tomivirkki: You still need to cover the root level case (other tests break because without it). The root level cache is just a regular item cache so you can use the same logic (clearSubcache) for clearing it.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -731,8 +731,24 @@ import { ItemCache } from '@vaadin/vaadin-grid/src/vaadin-grid-data-provider-mix
...
+       * items.
+       */
+      function clearSubcache(cache,scaledIndex, index, length) {
+        const end = length + index;
+        const childCache = cache.itemCaches[scaledIndex];
+        for(let itemIndex = index; itemIndex < end; itemIndex++) {
+          delete childCache.items[itemIndex];
+          delete childCache.itemCaches[itemIndex];
+        }
+        cache.updateSize();
```
tomivirkki, line 749: This will update the parent cache's effective size, but that's only one level upwards from the child cache. It works in case your grid only has 2 levels but with any deeper hierarchy, the root level effective size would not get updated.

If you want to keep the cache size update here in the same function, you'll also need to pass the root level cache as a parameter.

tomivirkki: Could be cleaner to just update it at the very end of `clear(index, length, parentKey)` (but would it make this function incomplete?)

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -731,8 +731,24 @@ import { ItemCache } from '@vaadin/vaadin-grid/src/vaadin-grid-data-provider-mix
...
         }
+        const cacheAndIndex = grid._cache.getCacheAndIndexByKey(pkey);
+        clearSubcache(cacheAndIndex.cache, cacheAndIndex.scaledIndex, index, length);
       });
 
+      /*
+       * Deletes items from grid cache, including the descendants of the affected
+       * items.
+       */
+      function clearSubcache(cache,scaledIndex, index, length) {
```
tomivirkki, line 742: Due to the comment below, you don't need to pass the (`cache, scaledIndex`) combo here at all, but just the `childCache`. If you're going to use this same function for covering the root cache also, you might want to rename the function (and the `childCache` parameter). Maybe `clearCacheRange(cache, startIndex, endIndex)`

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -731,8 +731,24 @@ import { ItemCache } from '@vaadin/vaadin-grid/src/vaadin-grid-data-provider-mix
           }
           updateGridItemsInDomBasedOnCache(items);
         }
+        const cacheAndIndex = grid._cache.getCacheAndIndexByKey(pkey);
+        clearSubcache(cacheAndIndex.cache, cacheAndIndex.scaledIndex, index, length);
```
tomivirkki, line 735: The logic above actually seems to operate on full pages. So if `clear(...)` is called with index 0 and length 60, with pageSize of 50, if would actually delete the cache for pages 0 and 1, which means from index 0 to index _99_. The new logic should (?) align with that and not use the `length` per-se, but instead an `endIndex`, which is rounded upwards by grid.pageSize (`const endIndex = index + updatedPageCount * grid.pageSize;`).

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -731,8 +731,24 @@ import { ItemCache } from '@vaadin/vaadin-grid/src/vaadin-grid-data-provider-mix
...
+      /*
+       * Deletes items from grid cache, including the descendants of the affected
+       * items.
+       */
+      function clearSubcache(cache,scaledIndex, index, length) {
+        const end = length + index;
+        const childCache = cache.itemCaches[scaledIndex];
+        for(let itemIndex = index; itemIndex < end; itemIndex++) {
+          delete childCache.items[itemIndex];
+          delete childCache.itemCaches[itemIndex];
```
tomivirkki, line 747: The connector seems to add a property called `itemkeyCaches` to all caches, which is a map of key-cache pairs. Should that also be covered in the clearing logic (it would include references to caches that are deleted here)?

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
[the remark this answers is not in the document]
tulioag: Done

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
tulioag: Removed the function.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
tulioag: Done

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
tulioag: Done

tomivirkki (approved):

PR #623, opened 2021-02-04
chore: fix the update platform script [skip ci]

ZheSun88:
filter out vaadin-core, as this will not be update anyway from here.


ZheSun88:
picked.

PR #640, opened 2021-02-10
Add MessageList items API

pekam:
Resolves https://github.com/vaadin/vaadin-messages/issues/10

Also adds TestBench APIs needed in ITs.

The testing is mostly done in integration tests, because of
the beforeClientResponse optimization that makes it hard to
test with unit tests (copied from AvatarGroup).

### DIFF
--- a/vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
+++ b/vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
@@ -12,27 +12,114 @@
  * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
  * License for the specific language governing permissions and limitations under
  * the License.
- *
  */
 package com.vaadin.flow.component.messages;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
 
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.HasSize;
 import com.vaadin.flow.component.HasStyle;
 import com.vaadin.flow.component.Tag;
-import com.vaadin.flow.component.Text;
+import com.vaadin.flow.component.dependency.JsModule;
+import com.vaadin.flow.component.dependency.NpmPackage;
+import com.vaadin.flow.internal.JsonUtils;
+
+import elemental.json.JsonArray;
 
 /**
- * Server-side component for the {@code vaadin-message-list} element.
+ * Server-side component for the {@code vaadin-message-list} element. The
+ * component displays a list of messages that can be configured with
+ * {@link #setItems(Collection)}.
  *
  * @author Vaadin Ltd.
  */
 @Tag("vaadin-message-list")
-// @JsModule("@vaadin/vaadin-messages/src/vaadin-message-list.js")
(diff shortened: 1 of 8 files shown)


vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
```diff
@@ -12,27 +12,114 @@
...
+        setItems(items);
+    }
+
+    /**
+     * Sets the items that will be rendered as messages in this message list.
+     *
+     * @param items
+     *            the items to set
+     */
+    public void setItems(Collection<MessageListItem> items) {
```
Peppe, line 86: Should non-null items parameter be asserted? new ArrayList<>(null); will throw an exception anyhow, but could be good to be more specific about it.

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
```diff
@@ -12,27 +12,114 @@
...
+    }
+
+    /**
+     * Sets the items that will be rendered as messages in this message list.
+     *
+     * @param items
+     *            the items to set
+     */
+    public void setItems(MessageListItem... items) {
+        setItems(Arrays.asList(items));
```
Peppe, line 101: Same as previous comment. Arrays will throw if items==null.

vaadin-messages-flow-parent/vaadin-messages-testbench/src/main/java/com/vaadin/flow/component/messages/testbench/MessageElement.java
```diff
@@ -0,0 +1,85 @@
...
+    public String getText() {
+        return this.getPropertyString("textContent");
+    }
+
+    /**
+     * Gets the {@code time} property of this element.
+     *
+     * @return the {@code time} property
+     */
+    public String getTime() {
```
Peppe, line 45: We force using an `Instant` through the Java API. Should we force the response to be an instant of this method as well? Or is it better to be more native and return the type that `vaadin-message` gives (String)? Dunno.

Peppe (changes requested):

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
pekam: Good point. Done.

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
pekam: Done. (Technically `items` can't be `null`, because it's the array of arguments that you pass in, but the individual items in that array can be)

vaadin-messages-flow-parent/vaadin-messages-testbench/src/main/java/com/vaadin/flow/component/messages/testbench/MessageElement.java
pekam: Usually the TestBench APIs are closer to the web component APIs, because TB operates in the browser.

Peppe (approved):

PR #646, opened 2021-02-15
Add connector for MessageList

heruan:
Add MessageList connector to format time in the client with the current locale.

Closes https://github.com/vaadin/vaadin-messages/issues/25

### DIFF
--- a/vaadin-messages-flow-parent/vaadin-messages-flow-integration-tests/src/test/java/com/vaadin/flow/component/messages/tests/MessageListIT.java
+++ b/vaadin-messages-flow-parent/vaadin-messages-flow-integration-tests/src/test/java/com/vaadin/flow/component/messages/tests/MessageListIT.java
@@ -44,8 +44,8 @@ public class MessageListIT extends AbstractComponentIT {
 
         MessageElement msg = messages.get(0);
         Assert.assertEquals("Unexpected text content", "foo", msg.getText());
-        Assert.assertEquals("Unexpected time prop", "1 Jan 2021, 01:00",
-                msg.getTime());
+        Assert.assertTrue("Unexpected time prop",
+                msg.getTime().matches("Jan 1, 2021, [0-9]+:[0-9]+ [A|P]M"));
         Assert.assertEquals("Unexpected userName prop", "sender",
                 msg.getUserName());
         Assert.assertEquals("Unexpected userImage prop", "/test.jpg",
@@ -72,8 +72,8 @@ public class MessageListIT extends AbstractComponentIT {
         Assert.assertEquals("Unexpected text content", "foo2", msg.getText());
 
         clickElementWithJs("setTime");
-        Assert.assertEquals("Unexpected time prop", "2 Feb 2000, 03:02",
-                msg.getTime());
+        Assert.assertTrue("Unexpected time prop",
+                msg.getTime().matches("Feb 2, 2000, [0-9]+:[0-9]+ [A|P]M"));
 
         clickElementWithJs("setUserName");
         Assert.assertEquals("Unexpected userName prop", "sender2",
@@ -107,4 +107,14 @@ public class MessageListIT extends AbstractComponentIT {
                 msg.getUserName());
     }
 
+    @Test
+    public void changeLocale_timeFormatted() {
+        clickElementWithJs("setLocale");
+
+        List<MessageElement> messages = messageList.getMessageElements();
+        MessageElement msg = messages.get(0);
+
+        Assert.assertTrue("Unexpected time prop",
+                msg.getTime().matches("1 gen 2021, [0-9]+:[0-9]+"));
+    }
 }
--- a/vaadin-messages-flow-parent/vaadin-messages-flow-integration-tests/src/test/java/com/vaadin/flow/component/messages/tests/MessageListIT.java
+++ b/vaadin-messages-flow-parent/vaadin-messages-flow-integration-tests/src/test/java/com/vaadin/flow/component/messages/tests/MessageListIT.java
@@ -44,7 +44,7 @@ public class MessageListIT extends AbstractComponentIT {
 
         MessageElement msg = messages.get(0);
         Assert.assertEquals("Unexpected text content", "foo", msg.getText());
-        Assert.assertEquals("Unexpected time prop", "2021-01-01T00:00:00Z",
+        Assert.assertEquals("Unexpected time prop", "1 Jan 2021, 01:00",
                 msg.getTime());
         Assert.assertEquals("Unexpected userName prop", "sender",
                 msg.getUserName());
@@ -72,7 +72,7 @@ public class MessageListIT extends AbstractComponentIT {
         Assert.assertEquals("Unexpected text content", "foo2", msg.getText());
 
         clickElementWithJs("setTime");
-        Assert.assertEquals("Unexpected time prop", "2000-02-02T02:02:02Z",
+        Assert.assertEquals("Unexpected time prop", "2 Feb 2000, 03:02",
                 msg.getTime());
 
         clickElementWithJs("setUserName");
--- a/vaadin-messages-flow-parent/vaadin-messages-flow-integration-tests/src/test/java/com/vaadin/flow/component/messages/tests/MessageListIT.java
+++ b/vaadin-messages-flow-parent/vaadin-messages-flow-integration-tests/src/test/java/com/vaadin/flow/component/messages/tests/MessageListIT.java
@@ -117,4 +117,14 @@ public class MessageListIT extends AbstractComponentIT {
         Assert.assertTrue("Unexpected time prop",
                 msg.getTime().matches("1 gen 2021, [0-9]+:[0-9]+"));
     }
+
+    @Test
+    public void reattachElement_messagesRendered() {
+        clickElementWithJs("detachList");
+        clickElementWithJs("attachList");
+
+        List<MessageElement> messages = $(MessageListElement.class).first()
+                .getMessageElements();
+        Assert.assertEquals("Unexpected items count", 2, messages.size());
+    }
 }
--- a/vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
+++ b/vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
@@ -45,7 +45,8 @@ import elemental.json.JsonArray;
 @JsModule("./messageListConnector.js")
 @JsModule("@vaadin/vaadin-messages/src/vaadin-message-list.js")
 @NpmPackage(value = "@vaadin/vaadin-messages", version = "1.0.0-alpha1")
-public class MessageList extends Component implements HasStyle, HasSize, LocaleChangeObserver {
+public class MessageList extends Component
+        implements HasStyle, HasSize, LocaleChangeObserver {
 
     private List<MessageListItem> items = Collections.emptyList();
     private boolean pendingUpdate = false;
--- a/vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
+++ b/vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
@@ -28,6 +28,8 @@ import com.vaadin.flow.component.HasStyle;
 import com.vaadin.flow.component.Tag;
 import com.vaadin.flow.component.dependency.JsModule;
 import com.vaadin.flow.component.dependency.NpmPackage;
+import com.vaadin.flow.i18n.LocaleChangeEvent;
+import com.vaadin.flow.i18n.LocaleChangeObserver;
 import com.vaadin.flow.internal.JsonUtils;
 
 import elemental.json.JsonArray;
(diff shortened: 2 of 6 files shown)


vaadin-messages-flow-parent/vaadin-messages-flow/src/main/resources/META-INF/resources/frontend/messageListConnector.js
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright 2000-2019 Vaadin Ltd.
```
pekam, line 2:
```suggestion
 * Copyright 2000-2021 Vaadin Ltd.
```

vaadin-messages-flow-parent/vaadin-messages-flow/src/test/java/com/vaadin/flow/component/messages/tests/MessageListTest.java
```diff
@@ -23,8 +23,15 @@
...
 
+import com.vaadin.flow.component.UI;
 import com.vaadin.flow.component.messages.MessageList;
 import com.vaadin.flow.component.messages.MessageListItem;
+import com.vaadin.flow.dom.Element;
+import com.vaadin.flow.i18n.LocaleChangeEvent;
+import com.vaadin.flow.server.VaadinSession;
+
+import elemental.json.JsonArray;
+import static org.mockito.Mockito.*;
```
pekam, line 34: We should use individual imports.

vaadin-messages-flow-parent/vaadin-messages-flow-integration-tests/src/test/java/com/vaadin/flow/component/messages/tests/MessageListIT.java
```diff
@@ -44,7 +44,7 @@ public void setInitialItems_messagesRendered() {
 
         MessageElement msg = messages.get(0);
         Assert.assertEquals("Unexpected text content", "foo", msg.getText());
-        Assert.assertEquals("Unexpected time prop", "2021-01-01T00:00:00Z",
+        Assert.assertEquals("Unexpected time prop", "1 Jan 2021, 01:00",
```
pekam, line 47: These depend now on the server's locale and the client's timezone, so we need to figure out how to test them consistently in different environments. I don't have a solution in mind right away.

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/resources/META-INF/resources/frontend/messageListConnector.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+(function () {
+    const tryCatchWrapper = function (callback) {
+        return window.Vaadin.Flow.tryCatchWrapper(callback, 'Vaadin Message List', 'vaadin-messages');
+    };
+
+    window.Vaadin.Flow.messageListConnector = {
+        setItems: tryCatchWrapper(function (list, items, locale) {
+            const formatter = new Intl.DateTimeFormat(locale, { dateStyle: 'medium', timeStyle: 'short'});
```
pekam, line 24: Have you tested what happens in IE11?

vaadin-messages-flow-parent/vaadin-messages-flow/src/test/java/com/vaadin/flow/component/messages/tests/MessageListTest.java
```diff
@@ -67,4 +74,31 @@ public void setItems_nullCollection_throws() {
...
+    public void localeChanged_schedulesItemsUpdate() {
+        SpyMessageList spyMessageList = new SpyMessageList();
+        UI ui = new UI();
+        ui.getInternals().setSession(mock(VaadinSession.class));
+        LocaleChangeEvent localeChangeEvent = new LocaleChangeEvent(ui, ui.getLocale());
+        spyMessageList.localeChange(localeChangeEvent);
+        ui.add(spyMessageList);
+        ui.getInternals().getStateTree().runExecutionsBeforeClientResponse();
+        Element spyElement = spyMessageList.getElement();
+        verify(spyElement).executeJs(eq("window.Vaadin.Flow.messageListConnector.setItems(this, $0, $1)"), any(JsonArray.class), eq(ui.getLocale().toLanguageTag()));
```
pekam, line 88: I'm not sure if this is better than integration tests after all, because this depends on the implementation details (calling the JS function), and it doesn't guarantee that such a JS function even exists (this test would pass if we rename `setItems` to `foo`). :thinking:

pekam (changes requested):

vaadin-messages-flow-parent/vaadin-messages-flow-integration-tests/src/test/java/com/vaadin/flow/component/messages/tests/MessageListIT.java
heruan: Right, I'm setting the locale directly on the UI now to be consistent.

heruan: Right! There is a new method to set a specific timezone in Selenium 4, but we're on 3. I can set times to 12:00 GMT then check the strings starts with the correct date string: should be safe for all timezones. WDYT?

vaadin-messages-flow-parent/vaadin-messages-flow/src/test/java/com/vaadin/flow/component/messages/tests/MessageListTest.java
heruan: Yeah, I'm now setting/changing the locale directly on the UI in integrations tests.

vaadin-messages-flow-parent/vaadin-messages-flow-integration-tests/src/test/java/com/vaadin/flow/component/messages/tests/MessageListIT.java
pekam: Sounds like a plan! :+1: If we want to be really safe, we should check also that the time-part is included e.g. with a regular expression, but I don't think it's a must-have.

pekam: Should we also use `UI::setLocale` in `MessageListPage` to make sure that the locale is what we expect? I'm not sure what actually defines the default locale.

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/resources/META-INF/resources/frontend/messageListConnector.js
heruan: Just did: both IE11 and Safari do not support `dateStyle` and `timeStyle`, but using
```js
new Intl.DateTimeFormat(locale, {
    year: 'numeric', month: 'short', day: 'numeric',
    hour: 'numeric', minute: 'numeric'
});
```
gives the same result in Chrome/Firefox/Safari while in IE11 it's slightly different but still localized (e.g. for Italian it formats `17/feb/2021` while others `17 feb 2021`).

pekam: Good. :+1: Support for IE11 doesn't mean it needs to be perfectly consistent with other browsers, so that's good enough.

vaadin-messages-flow-parent/vaadin-messages-flow-integration-tests/src/test/java/com/vaadin/flow/component/messages/tests/MessageListIT.java
heruan: Yeah, I'm using `UI::setLocale` to set a consistent locale and then also to change it and test time formats accordingly.

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
```diff
@@ -40,9 +42,10 @@
  * @author Vaadin Ltd.
  */
 @Tag("vaadin-message-list")
+@JsModule("./messageListConnector.js")
 @JsModule("@vaadin/vaadin-messages/src/vaadin-message-list.js")
 @NpmPackage(value = "@vaadin/vaadin-messages", version = "1.0.0-alpha1")
-public class MessageList extends Component implements HasStyle, HasSize {
+public class MessageList extends Component implements HasStyle, HasSize, LocaleChangeObserver {
```
pekam, line 48: This file is not formatted by the Vaadin code conventions. This line exceeds the max line length.

I recommend using the `Save Actions` plugin in IntelliJ and configure it to reformat file and optimize imports on save.

pekam (changes requested):

vaadin-messages-flow-parent/vaadin-messages-flow-integration-tests/src/test/java/com/vaadin/flow/component/messages/tests/MessageListIT.java
pekam: Looks good! :+1:

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
heruan: It was in `Edit > Code Style > Java > Ensure right margin is not exceeded`, not sure why it didn't wrap with `Hard wrap at: 80` already set 😅

pekam: So it's still a bit unclear to me: did you get the auto-formatting to work? With my settings, it also wraps the `executeJs` line.

heruan: Seems like it doesn't automatically wrap long strings (or I didn't configure it correctly). I did it manually for now, I'll look for a setting!

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
```diff
@@ -124,9 +128,17 @@ void scheduleItemsUpdate() {
             getElement().getNode().runWhenAttached(
                     ui -> ui.beforeClientResponse(this, ctx -> {
                         JsonArray itemsJson = JsonUtils.listToJson(items);
-                        getElement().setPropertyJson("items", itemsJson);
+                        getElement().executeJs(
+                                "window.Vaadin.Flow.messageListConnector"
+                                        + ".setItems(this, $0, $1)",
+                                itemsJson, ui.getLocale().toLanguageTag());
```
pekam, line 134: I remembered now that this probably introduces a bug that is very common when updating a Flow component with JavaScript: If you detach and re-attach the component, the items will probably disappear.

The reason is that Flow gets rid of the client-side element instance on detach, and creates a new one on attach. Flow synchronizes all the properties from server to client, but with this change, `items` is no longer stored in the server-side `Element`'s properties.

The solution is to add an attach listener which schedules an update for the items. But it should be first tested that my assumption is correct.

We can create another ticket for that or fix in this PR, up to you.

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
pekam: Yeah, it doesn't wrap strings and I think it's fine.

But it starts the string from a new line now, which was the important thing. :) Because now my auto-formatter doesn't make that change when I'm working on something un-related.

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
heruan: I tested it and items are still rendered correctly. I guess it's because of the `getElement().getNode().runWhenAttached(...)` which runs the connector when the element is attached. Do you feel we better have a test for this?

pekam: I had to test myself and indeed it does work. :O But it's not because of `runWhenAttached`; that runs only one time, either immediately if already attached or otherwise as soon as attaching the next time.

With a breakpoint I found out that it's actually because of `LocaleChangeObserver`. For some reason `localeChange` is triggered on attach, which in turn calls `scheduleItemsUpdate`. Well, I guess it makes sense, since on attach the component gets a new reference to `UI` which owns the locale.

Anyway, now it works kind of "by accident", so I would prefer having a test case for it.

For reference, here's my changes to the test view (I guess you had something similar):
```java
addButton("detach", () -> remove(messageList));
addButton("attach", () -> addComponentAsFirst(messageList));
```

heruan: I had exactly that 🙂

pekam (approved):

PR #658, opened 2021-02-18
Add MessageInput component

heruan:
Add MessageInput component.

Closes https://github.com/vaadin/vaadin-messages/issues/30


vaadin-messages-flow-parent/vaadin-messages-flow-integration-tests/src/test/java/com/vaadin/flow/component/messages/tests/MessageInputIT.java
```diff
@@ -0,0 +1,47 @@
...
+    @Before
+    public void init() {
+        open();
+        messageInput = $(MessageInputElement.class).first();
+    }
+
+    @Test
+    public void submitValue_eventHasCorrectValue() {
+        getInShadowRoot(messageInput, By.tagName("vaadin-text-area"))
+                .sendKeys("foo");
```
pekam, line 42: The TestBench API should have the required methods so that you can avoid these details in the test code (when it makes sense).

I recommend adding `setValue` to `MessageInputElement`. Also, it can be implemented simply as setting the `value` property.

vaadin-messages-flow-parent/vaadin-messages-flow-integration-tests/src/test/java/com/vaadin/flow/component/messages/tests/MessageInputIT.java
```diff
@@ -0,0 +1,47 @@
...
+        getInShadowRoot(messageInput, By.tagName("vaadin-button")).click();
```
pekam, line 43: I recommend adding `submit` method to `MessageInputElement`.

Also, `getInShadowRoot` is an old util method that is no longer needed. The recommended way is to use the TestBench query method `$`, which knows to find children also from the shadow root. So the following should work:
```java
messageInput.$("vaadin-button");
```

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageInput.java
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * @author Vaadin Ltd.
+ */
+@Tag("vaadin-message-input")
+@JsModule("@vaadin/vaadin-messages/src/vaadin-message-input.js")
+@NpmPackage(value = "@vaadin/vaadin-messages", version = "v2.0.0-alpha1")
+public class MessageInput extends Component {
+
+    /**
+     * The {@code submit} event which is fired by the component.
```
pekam, line 40:
```suggestion
     * The {@code submit} event which is fired by {@link MessageInput} component.
```

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageInput.java
```diff
@@ -0,0 +1,99 @@
...
+     */
+    @DomEvent("submit")
+    public static class SubmitEvent extends ComponentEvent<MessageInput> {
+
+        private final String value;
+
+        /**
+         * Creates the event.
+         *
+         * @param source     the source component
```
pekam, line 50: The formatting seems to be off in JavaDocs. Does your auto-formatting setup work?

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageInput.java
```diff
@@ -0,0 +1,99 @@
...
+        public SubmitEvent(MessageInput source, boolean fromClient,
+                @EventData("event.detail.value") String value) {
+            super(source, fromClient);
+            this.value = value;
+        }
+
+        /**
+         * Gets the value of the input.
+         *
+         * @return the value of the input
```
pekam, line 63: The input field is cleared on submit, so I'd say "submitted value" instead of "value of the input".

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageInput.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /**
+     * Gets the current content of the input field.
+     *
+     * @return the value of the input field
+     */
+    public String getValue() {
+        return getElement().getProperty("value");
+    }
```
pekam, line 77: This is never synchronized from client to server. It only reflects what might be set with the server-side API.

But the web component doesn't really have any events currently that we could use for synchronizing... :/

Maybe we could skip this method for now, although then we'd have a setter without a getter.

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageInput.java
```diff
@@ -0,0 +1,99 @@
...
+     *
+     * @param value the value to set on the input field
+     */
+    public void setValue(String value) {
+        getElement().setProperty("value", value);
+    }
+
+    /**
+     * Adds a listener that is called when the user submits the value of the
+     * input field, which can be obtained with {@link SubmitEvent#getValue()}.
```
pekam, line 90: Could explain that it's fired when clicking the send button or pressing the enter key

pekam (changes requested):

vaadin-messages-flow-parent/vaadin-messages-flow-integration-tests/src/test/java/com/vaadin/flow/component/messages/tests/MessageInputIT.java
heruan: What about just having a `void submitValue(String)` method instead of the getter/setter?

vaadin-messages-flow-parent/vaadin-messages-flow-integration-tests/src/test/java/com/vaadin/flow/component/messages/tests/MessageInputIT.java
heruan: Oh, I've just had the same idea in the above comment 😄 What about a `submit` method with a string parameter, then?

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageInput.java
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Server-side component for the {@code vaadin-message-input} element. The
+ * component displays a text-area to input a message and a button to send it.
+ *
+ * @author Vaadin Ltd.
+ */
+@Tag("vaadin-message-input")
+@JsModule("@vaadin/vaadin-messages/src/vaadin-message-input.js")
+@NpmPackage(value = "@vaadin/vaadin-messages", version = "v2.0.0-alpha1")
+public class MessageInput extends Component {
```
pekam, line 37: For most components, it makes sense to implement `HasSize` and `HasStyle`, and I think that `MessageInput` is not an exception.

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageInput.java
pekam: The discussion in Slack is leaning towards the option of not having `getValue`/`setValue` at all at this point.

manolo:
Please follow the conventional commits convention in PR and commit messages https://docs.google.com/document/d/1yfH7DTF_MnXwI40qPsDRv2KhYTmnB9OkVkFrQXM4MkA/edit#heading=h.cvrrd6bxknne

This is a new feature, so it should have a good comment as well explaining what is this new component for

pekam (approved):

PR #669, opened 2021-02-19
chore:update script for V20.0 (#661) (CP:14.4)

ZheSun88:


scripts/prepareDeploy.sh
```diff
@@ -50,6 +50,7 @@ pomBase=`getBaseVersion $pomVersion`
 
 ### Load versions file for this platform release
 branch=$versionBase
+[ $branch = "20.0" ] && branch=master
```
manolo, line 53: I think the best option is to fix the bug in the following line and not hardcode the latest version in this file, so as it will work when next majors got to master. Replace `curl -s "https ...` with `curl -f -s "https ...`  in the following lines

manolo (changes requested):

scripts/prepareDeploy.sh
manolo: Well actually this is a CP, so my fault because I should realise the problem in the master commit

manolo (approved):

PR #724, opened 2021-03-11
chore:  exit build properly when error found in annotation [skip-ci]

ZheSun88:


scripts/lib/versions.js
```diff
@@ -86,17 +86,24 @@ async function getAnnotations(){
...
-      name: r[2],
-      annotation: r[3],
-      package: r[4],
-      version: r[5],
-      major: r[6],
-      minor: r[7],
-      updatedVersion: ''
-    };
+    try {
+      const r = /(.*(vaadin-.*)-parent.*):(.*value *= *"([^"]+).*version *= *"((\d+)\.(\d+)[^"]*).*)/.exec(line);
```
manolo, line 90: I think you can just check whether r is null which means that regex was not met, then exit 1
try/catch here seems overwhelming

manolo (changes requested):

manolo (approved):

PR #734, opened 2021-03-15
Fix: Call updateValue after server-side setHtml

tulioag:
Fixes: vaadin/vaadin-rich-text-editor#175

### DIFF
--- a/vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java
+++ b/vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java
@@ -1,5 +1,23 @@
 package com.vaadin.flow.component.richtexteditor;
 
+import java.io.Serializable;
+import java.util.Objects;
+
+/*
+ * #%L
+ * Vaadin Rich Text Editor for Vaadin 10
+ * %%
+ * Copyright (C) 2017 - 2020 Vaadin Ltd
+ * %%
+ * This program is available under Commercial Vaadin Developer License
+ * 4.0 (CVDLv4).
+ *
+ * See the file license.html distributed with this software for more
+ * information about licensing.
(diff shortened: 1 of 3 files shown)


vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java
```diff
@@ -268,6 +249,34 @@ String sanitize(String html) {
...
+                // Add a one-time listener if we are not in eager mode.
+                final String JS = "var listener = e => {"
+                    + "  this.$server.updateValue(e.detail.value);"
+                    + "  this.removeEventListener('value-changed', listener);"
+                    + "  listener = null; };"
+                    + "this.addEventListener('value-changed', listener);";
+                getElement().executeJs(JS);
+            }
+            getElement().callJsFunction("dangerouslySetHtmlValue", this.html);
+            requested = false;
```
sosa-vaadin, line 276:
```suggestion
        private boolean pending;

        void requestUpdate(String htmlValueString) {
            this.html =
                htmlValueString != null ? sanitize(htmlValueString) : null;
            if (!pending) {
                runBeforeClientResponse(ui -> this.execute());
                pending = true;
            }
        }

        void execute() {
            if (getValueChangeMode() != ValueChangeMode.EAGER) {
                // Add a one-time listener if we are not in eager mode.
                final String JS = "var listener = e => {"
                    + "  this.$server.updateValue(e.detail.value);"
                    + "  this.removeEventListener('value-changed', listener);"
                    + "  listener = null; };"
                    + "this.addEventListener('value-changed', listener);";
                getElement().executeJs(JS);
            }
            getElement().callJsFunction("dangerouslySetHtmlValue", this.html);
            pending = false;
```

sosa-vaadin (approved):
I believe a more appropriate name for the `requested` variable in the `HtmlSetRequest` class would be `pending`. Other than that, looks good to me.

sosa-vaadin (approved):

PR #735, opened 2021-03-16
feat: add MessageInput i18n API

pekam:
resolves https://github.com/vaadin/vaadin-messages/issues/37

The API details are consistent with components such as `DatePicker` and `Upload`:
- the i18n property setters return `this` for method chaining
- updating the properties on previously set i18n doesn't update the component automatically.


vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageInput.java
```diff
@@ -90,4 +95,35 @@ public Registration addSubmitListener(
...
+     * <p>
+     * Note: updating the object properties after setting the i18n will not
+     * update the component. To make the changes effective, you need to set the
+     * updated object again.
+     *
+     * @param i18n
+     *            the i18n object, not {@code null}
+     */
+    public void setI18n(MessageInputI18n i18n) {
+        Objects.requireNonNull(i18n, "The i18n object should not be null");
```
Peppe, line 125: null is the default value for `this.i18n`. Wouldn't it be okay then to reset the situation with `setI18n(null)`?

Peppe (changes requested):

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageInput.java
pekam: I agree that it would make sense, but
- The implementation would be a bit ugly. Although the value is `null` at server-side, the web component has a default value. So we should duplicate the default value in Java code to be able to reset the web component, and then we are in trouble if those get out of sync.
- This is consistent with other Java component i18n APIs.
- Adding support for setting `null` here is a non-breaking change, so adding it later is no issue.

So because the implementation is not trivial, I would prefer keeping all the i18n APIs consistent and change all of those together if/when someone asks for it and it becomes a priority.

Peppe (approved):

PR #741, opened 2021-03-18
feat: add EditOnClick feature (#480) (CP: 14.6)

DiegoCardoso:


sosa-vaadin (changes requested):
Needs one IT to check that `editMode` is entered on a single click.

DiegoCardoso:
> Needs one IT to check that `editMode` is entered on a single click.

Good point. But, since this is a cherry-pick, I believe that this should be added to master first and backported to this version.

sosa-vaadin (approved):

sosa-vaadin:
> > Needs one IT to check that `editMode` is entered on a single click.
> 
> Good point. But, since this is a cherry-pick, I believe that this should be added to master first and backported to this version.

You're righta. Go on then.

Issue #742, opened 2021-03-19
`@Generated` annotation in Generated elements is not available in modern JDKs


manolo:
We could get rid of that annotation [1] in favor of a jdoc comment, or a custom annotation. Otherwise we need to keep a hack [2] profile for compiling the components in modern JDKs

[1] https://github.com/vaadin/vaadin-flow-components/blob/master/vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/GeneratedVaadinButton.java#L108
[2] https://github.com/vaadin/vaadin-flow-components/blob/master/vaadin-button-flow-parent/vaadin-button-flow/pom.xml#L93

manolo:
Related discussion: https://vaadin.slack.com/archives/G9EU978QM/p1616406459029500

PR #748, opened 2021-03-22
chore: remove `@Generated` annotation and `javax.annotation-api` dep

manolo:
`@Generated` annotation is not needed since
- there are no generators for flow components anymore.
- It also depends on a package not available in modern JDKs
- It has source retention, thus no API changes with this change

Fixes: #742

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -47,7 +47,7 @@ tcLog() {
 # log in TC
 tcStatus() {
   [ "$1" = "0" ] && status=SUCCESS || status=FAILURE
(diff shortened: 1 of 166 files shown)


ZheSun88 (approved):
should we leave those content in the annotation, maybe as a comment? 
but not blocking

manolo:
@ZheSun88 those comments are not valid, because they say:
 - class generated, not modify, isn't valid anymore they are now maintained by hand
 - reported issue addresses are not valid either, since the links are incorrect.
 
So I'd rather not having them since do not clarify anything

PR #774, opened 2021-03-26
fix: change enabled state of radio-buttons when enabled state of radio-button-group changes

sosa-vaadin:
The `enabled` state of `radio-button`s inside a `radio-button-group` were not being changed in response to changes in the `enabled` state of the `radio-button-group`. Only the `disabled` property was being changed.

This has now being fixed.

Fixes https://github.com/vaadin/vaadin-radio-button/issues/190

### DIFF
--- a/vaadin-radio-button-flow-parent/vaadin-radio-button-flow-integration-tests/src/main/java/com/vaadin/flow/component/radiobutton/tests/RadioButtonGroupDemoPage.java
+++ b/vaadin-radio-button-flow-parent/vaadin-radio-button-flow-integration-tests/src/main/java/com/vaadin/flow/component/radiobutton/tests/RadioButtonGroupDemoPage.java
@@ -17,7 +17,6 @@
 
 package com.vaadin.flow.component.radiobutton.tests;
 
-import com.vaadin.flow.component.button.Button;
 import com.vaadin.flow.component.html.Anchor;
 import com.vaadin.flow.component.html.Div;
 import com.vaadin.flow.component.html.Hr;
@@ -81,7 +80,6 @@ public class RadioButtonGroupDemoPage extends DemoView {
         addItemIconGenerator();
         addDisabled();
         addDisabledItems();
-        addDisabledGroupWithItemEnabledProvider();
         addComponentAfterItems();
         addReadOnlyGroup();
         insertComponentsBetweenItems();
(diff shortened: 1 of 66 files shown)


sosa-vaadin:
Since this fix is done on the flow implementation, I would like to check the web component itself to confirm it doesn't suffer from the same issue.

sosa-vaadin:
> Since this fix is done on the flow implementation, I would like to check the web component itself to confirm it doesn't suffer from the same issue.

Even though enabling a disabled `vaadin-radio-group` will enable all of its buttons, regardless of them being disabled individually or not, the web component version does not have an 'ItemEnableProvider' object. To implement a similar behavior, it must be added as a custom implementation.

vaadin-radio-button-flow-parent/vaadin-radio-button-flow-integration-tests/src/main/java/com/vaadin/flow/component/radiobutton/tests/RadioButtonGroupDemoPage.java
```diff
@@ -285,6 +287,26 @@ private void addDisabledItems() {
                 valueInfo);
     }
 
+    private void addDisabledGroupWithItemEnabledProvider() {
```
tulioag, line 290: This shouldn't be a demo. Please create a new view or reuse one of the other ones

vaadin-radio-button-flow-parent/vaadin-radio-button-flow/src/main/java/com/vaadin/flow/component/radiobutton/RadioButtonGroup.java
```diff
@@ -556,6 +556,7 @@ protected boolean valueEquals(T value1, T value2) {
     private void updateEnabled(RadioButton<T> button) {
         boolean disabled = isDisabledBoolean()
                 || !getItemEnabledProvider().test(button.getItem());
+        button.setEnabled(!disabled);
```
tulioag, line 559: This might have security implications. I think the button's enabled status should come from the group status, like so
```java
boolean disabled = !this.isEnabled() ||   !getItemEnabledProvider().test(button.getItem());
``` 
I also think the code below this line should be place on the method `onEnabledStateChanged` in the `RadioButton` class.

tulioag (changes requested):

vaadin-radio-button-flow-parent/vaadin-radio-button-flow-integration-tests/src/main/java/com/vaadin/flow/component/radiobutton/tests/RadioButtonGroupDemoPage.java
sosa-vaadin: I moved the test to another class and modified that class so the component tested has an `ItemEnableProvider` object. It should not affect the other single test that it had and the class now covers the same case this one covered.

vaadin-radio-button-flow-parent/vaadin-radio-button-flow/src/test/java/com/vaadin/flow/component/radiobutton/RadioButtonGroupTest.java
```diff
@@ -156,6 +156,28 @@ public void selectDisabledItem_noRedundantEvent() {
...
+        List<RadioButton<String>> children = group.getChildren()
+                .map(child -> (RadioButton<String>) child)
+                .collect(Collectors.toList());
+
+        Assert.assertTrue(children.get(0).isEnabled());
+        Assert.assertFalse(children.get(1).isEnabled());
+
+        group.setEnabled(false);
+        Assert.assertFalse(children.get(0).isEnabled());
+        Assert.assertFalse(children.get(1).isEnabled());
```
tulioag, line 174: Can you add this block of test again at the end of this method? Just to make sure that if it was disabling again is working.

sosa-vaadin: Done.

tulioag (approved):

PR #865, opened 2021-04-30
fix: Ensure editor closing when page refreshed while using PreserveOnRefresh

TatuLund:
If Editor is not closed on detach and preserve on refresh is used, the ghost editor will cause an exception.

fixes: https://github.com/vaadin/vaadin-grid/issues/2176

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/PreserveOnRefreshPage.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/PreserveOnRefreshPage.java
@@ -31,6 +31,8 @@ import com.vaadin.flow.router.Route;
 @PreserveOnRefresh
 public class PreserveOnRefreshPage extends Div {
 
+    int count = 0;
+
     public PreserveOnRefreshPage() {
         Grid<Person> grid = new Grid<>();
         Person foo = new Person("foo", 20);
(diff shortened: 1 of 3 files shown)


DiegoCardoso:
While reviewing this, I noticed on misbehavior: 

After the editor is opened and the page is refreshed, clicking on the "Edit" button produces no effect on the Grid.

You can see this on the video below:

https://user-images.githubusercontent.com/262432/116712810-9ab4b280-a9aa-11eb-851d-b1216e1cb527.mp4

I changed the view and added a second button, which is defined as
```java
Button closeButton = new Button("close", event -> grid.getEditor().cancel());
```

As you can see, clicking on "Edit" and "Close" buttons work as expected. But, at the moment I refresh the page, it won't show the `TextField` anymore. Also notice that the "Closed" logs are being added when I click on the "Close" button.

The button is calling exactly what is being called on the `onDetach` method, so there's something odd happening about the time when this method is called. Another thing is that if I close the `Editor` before refreshing the page, then it works as expected.

TatuLund:
> After the editor is opened and the page is refreshed, clicking on the "Edit" button produces no effect on the Grid.

@DiegoCardoso yes, something goes wrong in closing the editor, something in state is not consistent ... I did some debugging, but I do not yet have idea what is wrong. This is also a bug, but different one. So this is becoming cluster of two bugs.

TatuLund:
@DiegoCardoso I think I found the solution.

DiegoCardoso (changes requested):
This change seems to fix the issue I pointed out. I tried it locally and couldn't see the same behavior again. I would suggest change the test to cover these steps as well to make sure it works properly.

One more thing: taking a closer look to the issue this PR is willing to fix, I noticed that these changes cover the bug described on this comment (https://github.com/vaadin/vaadin-grid/issues/1912#issuecomment-829432262). The issue described by the OP seems to be not covered by this PR.

TatuLund:
> I would suggest change the test to cover these steps as well to make sure it works properly.

@DiegoCardoso , done

DiegoCardoso (commented):
The change looks good to me. 

There's only need to fix the formatting issue running `mvn formatter:format`. (update branch first so it will have the change to make the formatter ignore HTML files)

TatuLund:
> There's only need to fix the formatting issue running

Done

DiegoCardoso (approved):

PR #876, opened 2021-04-30
fix: restore sorting indicators after column order change

TatuLund:
Column reordering messes sort indicators, this should be the fix, testing for regressions, do not merge

fixes: https://github.com/vaadin/vaadin-grid/issues/2175

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/SortingPage.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/SortingPage.java
@@ -94,9 +94,9 @@ public class SortingPage extends Div {
         NativeButton reOrder new NativeButton("Re-order", e-> {
             grid.setColumnOrder(ageColumn,nameColumn);
         });
-        reOrder.setId("reorder-button);
+        reOrder.setId("reorder-button");
 
-        add(button,reOrder);
+        add(button, reOrder);
 
         return grid;
     }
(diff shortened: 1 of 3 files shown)


sissbruecker:
@TatuLund 
> testing for regressions, do not merge

Can we mark this as a draft in the meantime?

TatuLund:
> Can we mark this as a draft in the meantime?

@sissbruecker The PR has been already updated and is ready for review

sissbruecker:
Considering that reordering the columns calls:
```
        grid.getElement()
                .executeJs("this._updateOrders(this._columnTree, null)");
```
on the client side, and that `setColumnOrder` does not affect the sort order state, it seems likely that the bug is in the web component and should probably be fixed there. Will have to dig some more to figure out what's happening in the web component.

TatuLund:
>  it seems likely that the bug is in the web component and should probably be fixed there.

Maybe, but not so obvious. The sorter directions seems to be also server state variable, and there is specific setSorterDirections function in gridConnector.js to sync them from server to client. 

Furthermore, it looks to me, that Java API for colum re-orderign relies on GridColumnOrderHelper, where the setColumnOrder method seems to rebuild the column layers, which in turn re sets the header components. This is I think what fundamentally overrides the sorting indicators. 

So as summary it is the server side logic that overrides the sorting indicators, and thus it is the server side responsibility to put them back.

Verify the above with the architect, and if I have not understood the logic right, please let me know. The logic is indeed convoluted.

sissbruecker:
@TatuLund You are right, I went over the GridColumnOrderHelper too quickly. The Flow component basically detaches all columns and then reattaches them in the desired order. The web component stores the state in the sorter element itself, which is recreated when reattaching a column, so the state is gone at that point.

Anyway, I opened a discussion in #ds-team if we want to enhance the web-component to store the sort state even if columns are detached, or if we just go with your solution of restoring the state from the Flow-side.

sissbruecker (approved):
We decided to not change the web components for now, so this change would be the most appropriate solution ATM. Approved.

PR #879, opened 2021-05-03
chore: apply formatter to files (14.5)

ZheSun88:
applying formatter to all files in branch 14.5, same action has been applied with other branches.


sissbruecker:
@ZheSun88 Can you please provide some background here - is this just about back-porting the `mvn formatter:format` command and formatting all related files or are there any other changes to look out for?

ZheSun88:
Hi @sissbruecker , yes, this PR is only about applying the formatter to this branch (as it was missed from previous round. )
I will update the commit message. Thanks.

manolo (approved):

PR #780, opened 2021-04-05
Fix OSGi related issues

stbischof:
**Fix: bnd export should only export project package** 493a714

PR #708 did not adress the fact that
bnd exports all classes from the classpath.
felix only uses the project sources
this could be fixed by explicit name the the exported package

**optimize manifest variables calculation** dd28c0d

bnd caclulates variables like the exported packagename
remove build-helper-maven-plugin and maven project.properties

**demo project - packageName ans osgify** c62abd6

rename packages that all demo projects uses the package pattern `.demo`
bnd-plugin exists in demo project but the generates MANIFEST.MF was not used
configure maven-jar-plugin that the bnd-generated MANIFEST.MF is used
configure bnd to calculate the package and bsn for demo-projects

**bnd plugin should run with default build** beff79b

bundles were only created in profile `attach-docs`
move bnd-maven-plugin and maven-jar-plugin-config
to default build section of component and demo projects

**use default Built-By behavior** a48a9f4

Manifest-Header Build-By had been overridden
this could also be done in bnd using:
`Build-By: vaadin`
the flow-server does not override this header
so we can remove this configuration of the maven-jar-plugin

**Organize Imports** 59facba

remove the unused imports detected by sonarcube
order all imports

**remove tailing whitespaced** eb7c1ac

**update maven-pom templates** b210d31

changes on flow-component and demo:
use bnd-maven-plugin not `felix:maven-bundle-plugin`
run bnd-maven-plugin with default build profile

**war profile fixes** f9e9879

fix readme - use port 8080
fix/ignore bnd issue when generating war

@denis-anisimov  could you have a look on this? at the moment an issue in #708 breakes osgi support - sorry for that. 493a714 fixes this concrete issue.

PR #708, opened 2021-03-08
use bnd maven plugin for building osgi bundles

stbischof:
This pr switches from felix plugin to newer bnd plugin.

It is a bit hard to build with the build scripts. It would help a lot to understand why you have the different maven profiles. Think there is a lot of optimization possible.

For the momenti tried this the automatic build of PR. 

If i login as guest to the the result i got:

```
404
You do not have enough permissions to access project with internal id: project28

Check that the URL is correct.

Previous page
Overview page
```

So i cant see if the manifest and the resource working properly. I will squash the commits if somebody tells me that is work's or i can see it in the ci

Issue #193, opened 2020-10-07
Remove demo module from master in Flow components


web-padawan:
Starting from V15 java components can't be demoed in vaadin.com making the demo artifact useless.
Verify the IT module has all the needed views for tests, if needed move the demo UI to IT module

---

Originally reported at https://github.com/vaadin/components-team-tasks/issues/508 but I could not transfer the issue from a private repo.

tomivirkki:
This is now finished. Closing.

PR #896, opened 2021-05-03
feat: use bnd-maven-plugin for building OSGi (in behalf of @stbischof)

manolo:
Continuation of  reverted [PR](https://github.com/vaadin/flow-components/pull/708) switching from felix to bnd plugin and other fixes proposed in [PR](https://github.com/vaadin/flow-components/pull/780) by @stbischof 
This change also renames packages in demo projects to match pattern `.demo`

There is a bunch of open questions from the author that we should answer before continuing

1. Should all demo projects use the *.demo package (some do, some not)?
2. Should demo jars be osgi-bundles?
3. Why all demo Projects waste time in calculating a OSGi-Manifest IF it is may not needed (?) AND the manifest file not picked up.
4. Should be, building an OSGi-Bundle, only a part of the java-doc profile attach-docs


manolo:
There are some classes that do not compile and chart tests broken after the refactoring, I'm fixing those to make the PR validation pass

manolo:
Build is green, after couple of fixes.

My opinion about your open questions?

> Should all demo projects use the *.demo package (some do, some not)?

Yes, we should align all demos

> Should demo jars be osgi-bundles?

It does not hurt, we do not have to maintain all pom.xml files in demos since we already have a template for generating them

> Why all demo Projects waste time in calculating a OSGi-Manifest IF it is may not needed (?) AND the manifest file not picked up.

It's not a big deal, it just some milliseconds per component, just in build time.

> Should be, building an OSGi-Bundle, only a part of the java-doc profile attach-docs

makes, sense, but still not a big deal

manolo:
In this ticket, design-system team plan to remove demos, I think we can still accept changes in this PR until demos are removed.
see https://github.com/vaadin/flow-components/issues/193

web-padawan:
I'm ok to accept changes to demos even though they are not published anywhere.

sissbruecker (approved):
Gave this a check. Apart from the OSGI related changes, which I'm not qualified to comment on and as such accept in good faith considering that nothing breaks, the remaining changes to demos look OK to me. Approved.

sissbruecker:
@manolo Do you see any other effort here, or can we merge this?

manolo:
@sissbruecker  LGTM, go ahead with merge.

PR #919, opened 2021-05-05
Fix: Prevent NPE with autoUpload=false and multiple files are being uploaded

tulioag:
Fixes: vaadin/vaadin-upload#358

### DIFF
--- a/vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/test/java/com/vaadin/flow/component/upload/tests/NonImmediateUploadIT.java
+++ b/vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/test/java/com/vaadin/flow/component/upload/tests/NonImmediateUploadIT.java
@@ -33,13 +33,14 @@ public class NonImmediateUploadIT extends AbstractUploadIT {
         UploadElement upload = $(UploadElement.class).waitForFirst();
         WebElement input = getInput(upload);
         fillPathToUploadInput(input, file1.getPath(), file2.getPath());
-        WebElement button = getButton(upload, buttonType);
+        WebElement button = findButtonInVaadinUploadFile(upload, buttonType);
         button.click();
         TestBenchElement element = $("span").id("error-handler-message");
         Assert.assertEquals("No errors", element.getText());
     }
 
-    private WebElement getButton(UploadElement upload, String buttonType) {
+    private WebElement findButtonInVaadinUploadFile(UploadElement upload,
+            String buttonType) {
         final String QUERY = String.format(
                 "return arguments[0]"
                         + ".shadowRoot.querySelector('vaadin-upload-file')"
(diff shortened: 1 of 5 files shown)


vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -88,8 +88,12 @@ public Upload() {
         DomEventListener allFinishedListener = e -> {
             JsonArray files = e.getEventData().getArray(elementFiles);
 
-            boolean isUploading = IntStream.range(0, files.length()).anyMatch(
-                    index -> files.getObject(index).getBoolean("uploading"));
+            boolean isUploading = IntStream.range(0, files.length())
+                    .anyMatch(index -> {
+                        final String KEY = "uploading";
+                        JsonObject object = files.getObject(index);
+                        return object.hasKey(KEY) && object.getBoolean(KEY);
```
sissbruecker, line 95: The construct here is a bit unfortunate, but it seems that there is no way to ask Flow to return an optional or a default value? If we need to do this more often, we might want to introduce a helper for that.

vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/test/java/com/vaadin/flow/component/upload/tests/NonImmediateUploadIT.java
```diff
@@ -0,0 +1,50 @@
...
+        UploadElement upload = $(UploadElement.class).waitForFirst();
+        WebElement input = getInput(upload);
+        fillPathToUploadInput(input, file1.getPath(), file2.getPath());
+        WebElement button = getButton(upload, buttonType);
+        button.click();
+        TestBenchElement element = $("span").id("error-handler-message");
+        Assert.assertEquals("No errors", element.getText());
+    }
+
+    private WebElement getButton(UploadElement upload, String buttonType) {
```
sissbruecker, line 42: Had to see this visually to understand that it queries a button from a nested `vaadin-upload-file` element. Maybe the method name should reflect that it queries a button within an upload-file element, e.g. `getUploadFileElementButton`.

sissbruecker (approved):
Looks good, just added some random thoughts.

vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/test/java/com/vaadin/flow/component/upload/tests/NonImmediateUploadIT.java
tulioag: Method was renamed

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
tulioag: It is a valid concern, but `JsonObject` is not part of Flow and it does not provide an API using Optional.

PR #927, opened 2021-05-06
fix: change how setSeries converts array to ArrayList

DiegoCardoso:
Because `setSeries(...series)` uses `Arrays.asList` to convert from array to `List`, one cannot call `addSeries` after calling `setSeries(...series)` since it throws a `UnsupportedOperationException`.

Fix vaadin/vaadin-charts#515

### DIFF
--- a/vaadin-charts-flow-parent/vaadin-charts-flow/src/test/java/com/vaadin/flow/component/charts/ConfigurationTest.java
+++ b/vaadin-charts-flow-parent/vaadin-charts-flow/src/test/java/com/vaadin/flow/component/charts/ConfigurationTest.java
@@ -24,7 +24,7 @@ public class ConfigurationTest {
 
         conf.setSeries(Arrays.asList(new ListSeries()));
         conf.addSeries(new ListSeries());
-        
+
         assertEquals(conf.getSeries().size(), 2);
     }
 
@@ -34,7 +34,7 @@ public class ConfigurationTest {
 
         List<Series> series = new ArrayList<>();
         conf.setSeries(series);
-        
+
         series.add(new ListSeries());
 
         assertEquals(conf.getSeries().size(), 0);
--- a/vaadin-charts-flow-parent/vaadin-charts-flow/src/test/java/com/vaadin/flow/component/charts/ConfigurationTest.java
+++ b/vaadin-charts-flow-parent/vaadin-charts-flow/src/test/java/com/vaadin/flow/component/charts/ConfigurationTest.java
@@ -1,0 +1,40 @@
+package com.vaadin.flow.component.charts;
+
+import static org.junit.Assert.assertEquals;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import org.junit.Test;
+
+import com.vaadin.flow.component.charts.model.Configuration;
+import com.vaadin.flow.component.charts.model.ListSeries;
+import com.vaadin.flow.component.charts.model.Series;
+
+/**
+ * Tests for the {@link Configuration}
+ *
+ */
+public class ConfigurationTest {
+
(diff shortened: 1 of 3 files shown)


sissbruecker (approved):

tulioag (changes requested):
Does it still throw an error if I call setSeries with an Unmodifiable collections?
```java
        conf.setSeries(Arrays.asList(new ListSeries(), new ListSeries()));
        conf.addSeries(new ListSeries());
```
Second, what happens if I add an element to the collection?
```java
       var list = new ArrayList<Series>(Arrays.asList(new ListSeries()));
       conf.setSeries(list);
       list.add(new ListSeries());
```

DiegoCardoso:
> Does it still throw an error if I call setSeries with an Unmodifiable collections?
> 
> ```java
>         conf.setSeries(Arrays.asList(new ListSeries(), new ListSeries()));
>         conf.addSeries(new ListSeries());
> ```

Yes, since the `setSeries(List)` will set whatever was passed to its local property and the next line will try to add a new element to a `Arrays#ArrayList` object.

> Second, what happens if I add an element to the collection?
> 
> ```java
>        var list = new ArrayList<Series>(Arrays.asList(new ListSeries()));
>        conf.setSeries(list);
>        list.add(new ListSeries());
> ```

For the same reason as explained above, the `conf.series` object is the same instance of `list`, so modifying one will modify the other.

sissbruecker:
So how about creating a shallow clone of the collection in `setSeries(list)`? That would solve both points (I would not expect Tulios second example to work).

tulioag:
My suggestion would be the same as @sissbruecker's to fix the first case, but I just want to be cautious not to break anything. Is it expected to be able to add an element to the collection, like in the second case, and have it be shown in the chart?

DiegoCardoso:
I guess it would make sense to change `setSeries(List)` to make a shallow clone of the instance instead of using it to set its internal property.

However, my feeling is that this is outside of the scope of the ticket this PR is trying to fix. Would it be better to create another ticket from @tulioag's comment?

BTW, this is the same approach taken on a similar bug fixed on Chart 4 (https://github.com/vaadin/charts/pull/576/).

tulioag:
I don't think it is outside of the scope of this ticket, since it would replace the current change and fix the current issue. It would be a more complete fix.

sissbruecker:
I agree with Tulio. I would not worry too much about the case where you modify the list after setting it in `setSeries`:
- the behaviour is not documented 
- there are clear APIs for replacing and adding series
- if someone runs into trouble, it should be obvious to fix
- in general I would say it's an edge-case, so it's less likely to be a problem in general

DiegoCardoso:
Apply the changes.

vaadin-charts-flow-parent/vaadin-charts-flow/src/test/java/com/vaadin/flow/component/charts/ConfigurationTest.java
```diff
@@ -0,0 +1,40 @@
...
+ *
+ */
+public class ConfigurationTest {
+
+    @Test(expected = Test.None.class)
+    public void configurationSetSeriesWithArraysAsListTest() {
+        Configuration conf = new Configuration();
+
+        conf.setSeries(Arrays.asList(new ListSeries()));
+        conf.addSeries(new ListSeries());
```
knoobie, line 26:
```suggestion
        conf.addSeries(new ListSeries());
        
        assertEquals(conf.getSeries().size(), 2);
```

knoobie:
Coming from VS-3804 - I would prefer the API to be consistent and self-explaining as @sissbruecker said. Currently it feels like a surprise bag

tulioag (approved):

PR #959, opened 2021-05-10
chore: fix package-name calculation by removing extra quotes

### LINKED ISSUE #957: [20.0] OSGi Export-Package declaration in some artifacts
Due to the general bnd configuration in vaadin-flow-componentes/pom.xml. There is no correct Export-Package declaration in some artifacts like:
* TextField (artifact vaadin-text-field-flow)
* AppLayout (artifact vaadin-app-layout-flow)
* OrderedLayout (artifact vaadin-ordered-layout-flow)

Instead, there are Private-Package declarations for packages to be exported:
`Private-Package: com.vaadin.flow.component.orderedlayout`

In the first sight, this seems to be a wrong package name replacement. Everything is fine for e.g. Button (artifact: vaadin-button-flow) or Grid (artifact: vaadin-grid-flow)

Obviously all artifacts with a dash in their names e.g. text-field, app-layout are not replaces in a correct manner

stbischof:
fixes #957

should also be added to v20

Alternatives:
- package-info.java with `@org.osgi.annotation.bundle.Export` Annotation in every bundle that should be exported.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -171,9 +171,10 @@
                     <configuration>
                         <bnd><![CDATA[
                           regexArtifactId: vaadin-(.*)-flow(-.*)?
+                          dash: -
                           replacementPattern: $1
                           packageNameDash: ${replacestring;${project.artifactId};${regexArtifactId};${replacementPattern}}
-                          packageNameTmp: ${replacestring;${packageNameDash};'-'}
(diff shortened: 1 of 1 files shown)


manolo (approved):

stbischof:
@manolo Thank you for the fast support!!!! V20 would be nice

PR #973, opened 2021-05-10
fix: disable client side validation

sissbruecker:
## Description

The original issue occurs because the select web component triggers its own client-side validation in several places (specifically to the issue, when the dropdown is closed), which then overwrites the validation state set by the server. Investigating this also lead to another issue that changing the validation state on the client through JS / dev tools also changed the validation state on the server, which should not be allowed.

This change fixes that by:
- overwriting the client-side validation method with a no-op, as is done for other field components
- disabling synchronization of the client side validation state to the server
- semi-related change: fixes the initial value of the client side component to be an empty string instead of the string `null` which initially lead the client side component to pass validation, because the value was not empty

Fixes https://github.com/vaadin/vaadin-select/issues/265

## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-select-flow-parent/vaadin-select-flow-integration-tests/pom.xml
+++ b/vaadin-select-flow-parent/vaadin-select-flow-integration-tests/pom.xml
@@ -94,6 +94,17 @@
             <groupId>com.vaadin</groupId>
             <artifactId>flow-polymer-template</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-grid-flow</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-grid-testbench</artifactId>
+            <version>21.0-SNAPSHOT</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
--- a/vaadin-select-flow-parent/vaadin-select-flow-integration-tests/pom.xml
+++ b/vaadin-select-flow-parent/vaadin-select-flow-integration-tests/pom.xml
@@ -102,7 +102,7 @@
         <dependency>
             <groupId>com.vaadin</groupId>
             <artifactId>vaadin-grid-testbench</artifactId>
-            <version>21.0-SNAPSHOT</version>
+            <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/vaadin-select-flow-parent/vaadin-select-flow/src/main/java/com/vaadin/flow/component/select/FieldValidationUtil.java
+++ b/vaadin-select-flow-parent/vaadin-select-flow/src/main/java/com/vaadin/flow/component/select/FieldValidationUtil.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2000-2019 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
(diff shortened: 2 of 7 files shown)


vaadin-select-flow-parent/vaadin-select-flow/src/main/java/com/vaadin/flow/component/select/FieldValidationUtil.java
```diff
@@ -0,0 +1,69 @@
...
+
+import com.vaadin.flow.internal.StateNode;
+
+/**
+ * Utility class for select Flow component to disable client side
+ * validation.
+ *
+ * @author Vaadin Ltd
+ */
+final class FieldValidationUtil {
```
sissbruecker, line 26: Decided to duplicate this for now, as is done for multiple other components. Besides for the Select we need a different `validate` implementation than for the text fields. The text fields override the `validate` implementation using [`checkValidity`](https://github.com/vaadin/flow-components/blob/master/vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/FieldValidationUtil.java#L58), which does not exist on the Select component.

vaadin-select-flow-parent/vaadin-select-flow/src/main/java/com/vaadin/flow/component/select/generated/GeneratedVaadinSelect.java
```diff
@@ -392,13 +392,12 @@ protected void setRequired(boolean required) {
...
      * <p>
-     * This property is synchronized automatically from client side when a
-     * 'invalid-changed' event happens.
+     * This property is not synchronized automatically from the client side, so
+     * the returned value may not be the same as in client side.
      * </p>
      *
      * @return the {@code invalid} property from the webcomponent
      */
-    @Synchronize(property = "invalid", value = "invalid-changed")
```
sissbruecker, line 401: This effectively disabled the synchronization of the prop to the server, checked with @yuriy-fix that modifying the generated classes should be fine.

vaadin-select-flow-parent/vaadin-select-flow/src/main/java/com/vaadin/flow/component/select/Select.java
```diff
@@ -123,6 +123,10 @@ public Select() {
 
         getElement().setProperty("invalid", false);
         getElement().setProperty("opened", false);
+        // Trigger model-to-presentation conversion in constructor, so that
+        // the client side component has a correct initial value of an empty
+        // string
+        setPresentationValue(null);
```
tulioag, line 129: This seems to conflict with the constructor in `GeneratedVaadinSelect` called in line `121`. There you have also a call to setPresentationValue depending on some conditions.
```java
        if ((getElement().getProperty("value") == null
                || !isInitialValueOptional) && initialValue != null) {
            setPresentationValue(initialValue);
        }
```
 Would it be possible to change or remove the conditions in  `GeneratedVaadinSelect` instead of adding this call here?

vaadin-select-flow-parent/vaadin-select-flow-integration-tests/pom.xml
```diff
@@ -94,6 +94,17 @@
...
         </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-grid-flow</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-grid-testbench</artifactId>
+            <version>21.0-SNAPSHOT</version>
```
tulioag, line 105:
```suggestion
            <version>${project.version}</version>
```

tulioag (changes requested):

tulioag (approved):

PR #978, opened 2021-05-11
feat: use web components 21.0.0-alpha2, fix tests

web-padawan:
## Description

1. Updated the web components to new major
2. Added the `@vaadin/vaadin-template-renderer` to bring back removed Polymer template APIs.

PR #979, opened 2021-05-12
feat: add theme API to MessageListItem

pekam:
**NOTE**: This depends on https://github.com/vaadin/flow-components/pull/978 and won't pass in CI before that is merged.

## Description

Feature: add and remove `theme` attribute values to messages in a `MessageList`. Only with strings at this point, since no theme variants have been implemented so far. The API is copied from `HasTheme` interface, which can be directly implemented only by component classes.

This enables developers to apply custom theming per each message in a MessageList.

Related issue: https://github.com/vaadin/collaboration-engine/issues/30

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [x] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/

### DIFF
--- a/vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageListItem.java
+++ b/vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageListItem.java
@@ -18,9 +18,14 @@ package com.vaadin.flow.component.messages;
 import java.io.Serializable;
 import java.net.URI;
 import java.time.Instant;
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.LinkedHashSet;
 import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
 
+import com.fasterxml.jackson.annotation.JsonGetter;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
@@ -57,6 +62,8 @@ public class MessageListItem implements Serializable {
     private Registration pendingRegistration;
     private Command pendingHandle;
     private AbstractStreamResource imageResource;
+
+    private Set<String> themeNames = new LinkedHashSet<>();
 
     /**
      * Creates an empty message list item. Use the setter methods to configure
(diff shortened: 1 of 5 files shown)


heruan (approved):

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageListItem.java
```diff
@@ -282,6 +289,52 @@ public void setUserColorIndex(Integer userColorIndex) {
...
     }
 
+    /**
+     * Adds one or more theme names to this message. Multiple theme names can be
+     * specified by using multiple parameters.
+     *
+     * @param themeNames
+     *            the theme name or theme names to be added to the message
+     */
+    public void addThemeNames(String... themeNames) {
```
sissbruecker, line 299: I don't think it's a good idea to just reimplement the `HasTheme` interface here. Maybe we should extract an interface that works without Flow specifics. Will open discussion in team.

sissbruecker (changes requested):

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageListItem.java
pekam: It's indeed not a good idea, but it's the best one we have. :)

Redesigning the `HasTheme` interface would either require a breaking change or result in weird API for an interface that _has theme but no element_. Also, if we eventually want go that way, this PR doesn't prevent doing that afterwards and then refactoring `MessageListItem` to implement the interface once it is available, without blocking this PR.

Another option would be to change the `<vaadin-message-list>` API so that you render `<vaadin-message>` elements inside it instead of passing data with the `items` property. Then Flow users would create `Message` components instead of `MessageListItems`, and `Message` component could implement `HasTheme`. But that would be a big change to do at this point, and that API design has its downsides as well. I was personally strongly in favor of the `items` API when we created these components, because it's a data-heavy component, and I want to just pass in JSON data that I've received over the network and mapped to compatible format, instead of rendering elements. Since then, I've been slowly changing my mind, because it sucks to add new API in the item objects for every new property/attribute of the message element.

Anyway, we decided together with the CE team that this shortcut with partial copy of the `HasTheme` API is the best trade-off.

pekam: > Another option would be to change the <vaadin-message-list> API so that you render <vaadin-message> elements inside it instead of passing data with the items property. 

Oh, and btw, one important reason to have the JSON object API instead of rendering elements: This enables us to implement lazy loading (`dataProvider` API) in the future.

sissbruecker (approved):

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageListItem.java
sissbruecker: Reiterating from internal discussion: I'm general fine with the changes, understanding the problem and seeing that this is currently the only use case for reusing the theming API. In the future, if we want to apply the same theming mechanism to other classes, we should consider extracting an interface to ensure those classes use the same theming API.

Issue #993, opened 2021-05-19
When using CheckBoxGroup#setItemEnabledProvider disabled checkboxes do not show its value


TatuLund:
When using CheckBoxGroup#setItemEnabledProvider disabled checkboxes do not show its value.

This is because the enabled provider and setting value are done in wrong order when checkboxes are updated, hence the later does not apply.

Tested Vaadin 15.5.5

PR #994, opened 2021-05-19
fix: Disable checkbox after applying value

TatuLund:
Fixes: https://github.com/vaadin/flow-components/issues/993


TatuLund:
Just observation: The existing test for use case asserts value on CheckboxGroup level, but not by individual Checkbox and thus the test did not fail:

https://github.com/vaadin/flow-components/blob/master/vaadin-checkbox-flow-parent/vaadin-checkbox-flow-integration-tests/src/test/java/com/vaadin/flow/component/checkbox/tests/CheckboxGroupIT.java#L89

tulioag (changes requested):
We need to add a test as well.

TatuLund:
> We need to add a test as well.

@tulioag I am investigating this at the moment. As seen in my previous comment, there was kind of test already, but it was not sufficient.

tulioag (approved):

Issue #1411, opened 2020-08-19
Changing i18n provider multiple times is not reflected in the front end


sandronm:
This issue is a copy of the following one on vaadin-date-picker-flow:
https://github.com/vaadin/vaadin-date-picker-flow/issues/233

I checked the code made on the flow part and everything seems ok.

```   private void setI18nWithJS() {
        runBeforeClientResponse(ui -> {
            JsonObject i18nObject = (JsonObject) JsonSerializer.toJson(i18n);
            for (String key : i18nObject.keys()) {
                getElement().executeJs("this.set('i18n." + key + "', $0)",
                        i18nObject.get(key));
            }
        });
    }
``` 

So when I debug the Flow code I can see there is an execution of the following js code:

``` this.set('i18n.weekdaysShort')```  and it seems to be ok but apparently the change on the weekdaysShort is not reflected.

Could you fix this one?

PR #1742, opened 2021-06-01
fix: update date picker overlay after setting I18N settings 

sissbruecker:
## Description

The data binding in the date picker web component is set up so that it only detects changes to the `i18n` property itself, but not to individual sub-properties like `i18n.monthNames`. This can lead to the overlay not updating if the overlay was already opened once, which means that the initial data bindings have been set up. This was noticable in the Flow component, which does not replace the I18N object but changes each sub-property individually.

This change fixes that by making the Flow component replace the whole I18N object in the WC instead of changing individual sub-properties. It also fixes a bug where the Flow component would set null values into the I18N object of the WC which could lead to errors.

I also investigated changing the data binding behaviour of the WC to detect changes to sub-properties, but that was not easily solvable. Plus the API for setting sub-properties will likely disappear with removing Polymer.

Fixes https://github.com/vaadin/flow-components/issues/1411

## Type of change

- [x] Bugfix


vaadin-date-picker-flow-parent/vaadin-date-picker-testbench/src/main/java/com/vaadin/flow/component/datepicker/testbench/DatePickerElement.java
```diff
@@ -30,6 +37,94 @@
...
+                            this.getCommandExecutor()))
+                    .collect(Collectors.toList());
+        }
+
+        /**
+         * Gets the today button from the overlays toolbar
+         * 
+         * @return
+         */
+        public ButtonElement getTodayButton() {
```
yuriy-fix, line 67: Nit: we can use `WebElement` over here without adding a dependency on `vaadin-button-testbench`.

yuriy-fix (changes requested):
It seems that there is no need in `ShadowDomHelper`. 
I have tried to update the logic and it seems that you can select all needed elements in shadow root with `ElementQuery`. (using `.id()` / `.attribute()`)
For example, in `getOverlayContent`:
`overlay.$("div").id("content")` for selecting the `content`

vaadin-date-picker-flow-parent/vaadin-date-picker-testbench/src/main/java/com/vaadin/flow/component/datepicker/testbench/DatePickerElement.java
sissbruecker: Would like to keep this more explicit, might also be useful for future use.

sissbruecker:
@yuriy-fix Thanks for the suggestion, I made the changes and the code is now much easier to process.

One issue came up with testing month names, where I now needed to introduce a wait until the month calendar element rerendered. See [change here](https://github.com/vaadin/flow-components/pull/1742/commits/895e82d8a7ee7d5af647e953b746286d1e9649ae#diff-1cbcdf70c757429d8c33a2774813b1abff988d974665f5c89bbabf2640ee45c6R82-R91). I removed the assertion there, as the `waitUntil` implies that the test would fail if it doesn't resolve. Wasn't sure though if that is the correct pattern.

Also while cleaning up the `DatePickerElement`, I ended up with an [empty class](https://github.com/vaadin/flow-components/pull/1742/commits/895e82d8a7ee7d5af647e953b746286d1e9649ae#diff-20bad5e1573bd0eee6393f5cc4f429a14bfb3cc076a84373fc8275b386e2369aR92-R93). I would still like to keep it for consistency.

yuriy-fix (approved):

PR #1780, opened 2021-06-09
chore: use english locale for tests

sissbruecker:
## Description

Currently tests are run with whatever locale is set as system locale. However there are some test that contain assertions with date / number formats specific to english, which makes tests fail if you use a system locale that uses different formats (e.g. I'm using German locale and these tests fail for me).

To fix this I added a Maven config file that sets english as the default locale for the JVM that Maven starts. That way tests should run deterministically regardless of the system locale.


vaadin-text-field-flow-parent/vaadin-text-field-flow/src/test/java/com/vaadin/flow/component/textfield/tests/BigDecimalFieldTest.java
```diff
@@ -33,6 +34,7 @@
     @Before
     public void init() {
         field = new BigDecimalField();
+        field.setLocale(Locale.US);
```
sissbruecker, line 37: For some reason one test in this suite was not affected by the global setting. Configuring the field explicitly works though. Since it's only this one occurrence it should be OK.

ZheSun88 (approved):

PR #1814, opened 2021-06-17
perf: remove the unnecessary span wrapper from virtual list items

tomivirkki:
While building `VirtualList` on top of a copy of `IronList` an extra `<span>` wrapper inside the item template should've been removed but wasn't. This PR removes the unnecessary wrapper element.

Before:
Screenshot 2021-06-17 at 12 03 18

After:
Screenshot 2021-06-17 at 12 02 01

### DIFF
--- a/vaadin-virtual-list-flow-parent/vaadin-virtual-list-flow-integration-tests/src/test/java/com/vaadin/flow/component/virtuallist/tests/VirtualListIT.java
+++ b/vaadin-virtual-list-flow-parent/vaadin-virtual-list-flow-integration-tests/src/test/java/com/vaadin/flow/component/virtuallist/tests/VirtualListIT.java
@@ -380,7 +380,7 @@ public class VirtualListIT extends AbstractComponentIT {
     @Test
     public void numberRenderer() {
         List<TestBenchElement> items = $("vaadin-virtual-list")
-                .id("list-with-numbers").$("span").all();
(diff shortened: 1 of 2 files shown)


vaadin-virtual-list-flow-parent/vaadin-virtual-list-flow-integration-tests/src/test/java/com/vaadin/flow/component/virtuallist/tests/VirtualListIT.java
```diff
@@ -380,7 +380,8 @@ public void nativeButtonRenderer() {
     @Test
     public void numberRenderer() {
         List<TestBenchElement> items = $("vaadin-virtual-list")
-                .id("list-with-numbers").$("span").all();
+                .id("list-with-numbers")
+                .$("div[style*=\"position: absolute;\"]").all();
```
vursen, line 384: Would `:scope > div` work here? 

Seems that the `:scope` selector is well-supported according to [caniuse](https://caniuse.com/?search=%3Ascope).

tomivirkki: Would be cleaner but for some reason with that selector the tests fail:

Screenshot 2021-06-17 at 12 20 00

vursen (approved):

Issue #1002, opened 2021-05-17
setSplitterPosition doesn't change split position inside a `display:block` container


OlliTietavainenVaadin:
### Description
In some conditions, calling `SplitLayout::setSplitterPosition` doesn't do anything. This seems to occur when you have a parent container that has the `display` CSS property set to `block` (which Vaadin Start Views have by default).

### Expected outcome
Calling `setSplitterPosition` should react to the input value.

### Actual outcome
`setSplitterPosition` doesn't do anything.

### Live Demo
Example view: 
```
public class HelloWorldView extends HorizontalLayout {

    public HelloWorldView() {
        getElement().getStyle().set("display", "block");
        SplitLayout sl = new SplitLayout(new Label("First"), new Label("Second"));
        sl.setOrientation(SplitLayout.Orientation.VERTICAL);
        Button hideSecond = new Button("Hide Second", e -> sl.setSplitterPosition(100));
        Button hideFirst = new Button("Hide First", e -> sl.setSplitterPosition(0));
        add(sl, hideSecond, hideFirst);
    }

}
```

### Steps to reproduce
1. Run the app
2. Click either of the buttons
3. Nothing happens
4. Comment out the row that sets the `display: block` style
5. Try again

### Browsers Affected

- [x ] Chrome (only tested browser)

Tested with Vaadin 14.5.5 and 19.0.7, reproduces in both

sissbruecker:
The issue here is not related to the parent being a block container, but comes from the fact that the SplitLayout does not have an explicit height. When using `setSplitterPosition`, it sets a percentage height for the children. However the SplitLayout's height is not explicit, and because of that is calculated based on the contents of the children themselves. This creates a kind of loop where the child wants to be some percentage of the height of the parent, but the parent needs the childrens height to calculate its own height. This SO answer explains it pretty well I think: https://stackoverflow.com/a/1622097/11368281

We discussed changing the logic to calculate the child sizes based on the effective height of the combined childrens contents, but that would lead to even more edge-cases and non-deterministic behaviour.

As such we decided to not support this use-case, and instead add a note to the docs that a vertical SplitLayout needs a well-defined height, either as an absolute value or a percentage. Note that when using a percentage height on the SplitLayout, it's ancestor(s) needs to have an explicitely defined height too.

@OlliTietavainenVaadin Regarding the start.vaadin.com templates, I can offer to take a look at what the problem is there and how we can change them to provide a well-defined height for the SplitLayout.

[...]

PR #1819, opened 2021-06-18
refactor: use flex-basis instead of width/height

sissbruecker:
## Description

Currently the SplitLayout flow component uses the width/height styles when programmatically setting the splitter position. This does not align with the web component, which uses flex-basis when moving the splitter, and also does not behave correctly when changing the splitter orientation.

This change addresses that by changing the `setSplitterPosition` logic to use flex-basis instead of width/height. This aligns it with the web component and fixes the bug when changing the orientation. I added a test for changing the orientation and improved the other integration tests to check the actual width / height of the children instead of testing implementaiton details.

This came up as a possible improvement when working on #1002 , but there is no related ticket.


vaadin-split-layout-flow-parent/vaadin-split-layout-flow-integration-tests/src/test/java/com/vaadin/flow/component/splitlayout/tests/SplitLayoutAssertions.java
```diff
@@ -0,0 +1,64 @@
...
+
+class SplitLayoutAssertions {
+
+    // Margin of error to account for rounding errors
+    private static final int MARGIN = 2;
+
+    public static void assertChildWidthInPercentage(
+            SplitLayoutElement layoutElement, TestBenchElement child,
+            double percentage) {
+        int layoutWidth = layoutElement.getWrappedElement().getSize()
```
yuriy-fix, line 36: nit: we can access `getSize` directly from elements. (i.e. `layoutElement.getSize`)

yuriy-fix (approved):
One small suggestion, otherwise :lgtm:

Issue #1839, opened 2021-06-23
[Grid] SortOrders get reset on changing the header of a column


wofl:
### Description

The sort orders of a grid reset when there is a new header set on the server side. It seems like this reset is only visually but not respresented on the data side. Both the grid webcomponent and the server side still keep the "right" sort orders (e.g. take a look in $grid._sorters through the dev tools in your browser of choice). Even the gridConnector#sorterChangeListener is called from the client side because it seems like it was notified that something changed ... but it just sends the "right" sort order that is already set on the server side and thus changing nothing.

### Expected outcome

The sort orders should **not** reset when changing the header or anything other regarding the column.

### Actual outcome

The sort orders get reset and not even synced to the server thus preventing any workaround with setting the sort orders again from the server side.

### Live Demo (optional)

![GridSortOrderReset](https://user-images.githubusercontent.com/3864493/123113799-209b2a00-d43f-11eb-896c-436e058f1b02.gif)

### Minimal reproducible example

```
@Route("gridSortOrderReset")
public class GridSortOrderReset extends VerticalLayout {
    public GridSortOrderReset() {
        final Grid<Integer> grid = new Grid<>();
        grid.setItems(0, 1, 2, 3, 4);
        final Span headerText = new Span("header");
        final Grid.Column<Integer> col = grid.addColumn(Object::toString)
                .setHeader(headerText)
                .setSortable(true);
        add(grid);
        final Button change = new Button("Change header");
        change.addClickListener(event -> col.setHeader(headerText));
        add(change);
    }
}
```

### Steps to reproduce

1. Click on the header of the column
2. Click on the "Change header" button

### Environment

Vaadin 14.6.3

### Browsers Affected
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE 11
- [ ] iOS Safari
- [ ] Android Chrome
(others not tested but should apply to all)

PR #1841, opened 2021-06-23
fix: Update client side sort indicators after header is updated

TatuLund:
Fixes: #1839


sissbruecker:
Added integration tests for the scenarios. Note that there is another bug (probably in the WC?) that leads to the column showing a numbered sort direction after changing the header:
Bildschirmfoto 2021-06-24 um 08 10 42

sosa-vaadin (approved):

PR #1875, opened 2021-07-06
fix: avoid adding a second grid to a crud when using a constructor without the grid argument

### LINKED ISSUE #1005: Flow: CRUD constructor with grid vs setGrid
Shouldn't these produce the same end result?

**Constructor with grid**
```
Crud<Person> crud = new Crud<>(
  Person.class,
  createGrid(),
  createEditor()
);
```

**Result:**

Screenshot 2021-04-22 at 13 56 05

**setGrid**
```
crud = new Crud<>(
  Person.class,
  createEditor()
);
crud.setGrid(createGrid());
```

**Result:**

Screenshot 2021-04-22 at 13 57 25

sosa-vaadin:
## Description

When creating a Crud using the following constructor and using the `setGrid()` method: 

```java
crud = new Crud<>(
  Person.class,
  createEditor()
);
crud.setGrid(createGrid());
```

Two grids would show, instead of just one. The problem is that the first grid is added by default when the constructor runs but it is **not** removed when the `setGrid()` method is called. The crud fails to remove the grid because of an equality check (using the equality operator `==`) instead of the `equals()` method, when checking if the Grid's parent is the Crud itself.

Fixes #1005

## Type of change

- [x] Bugfix

## Checklist

### DIFF
--- a/vaadin-crud-flow-parent/vaadin-crud-flow-integration-tests/src/test/java/com/vaadin/flow/component/crud/test/BasicUseIT.java
+++ b/vaadin-crud-flow-parent/vaadin-crud-flow-integration-tests/src/test/java/com/vaadin/flow/component/crud/test/BasicUseIT.java
@@ -5,13 +5,9 @@ import java.util.List;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
-import org.openqa.selenium.By;
-import org.openqa.selenium.WebElement;
 
 import com.vaadin.flow.component.button.testbench.ButtonElement;
-import com.vaadin.flow.component.crud.examples.Person;
 import com.vaadin.flow.component.crud.testbench.CrudElement;
-import com.vaadin.flow.component.grid.Grid;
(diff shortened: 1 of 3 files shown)


vaadin-crud-flow-parent/vaadin-crud-flow-integration-tests/src/test/java/com/vaadin/flow/component/crud/test/BasicUseIT.java
```diff
@@ -27,6 +31,14 @@ public void crudIsPresent() {
         Assert.assertNotNull($(CrudElement.class).first().getGrid());
     }
 
+    @Test
+    public void crudReplacesGrid() {
+        CrudElement crud = $(CrudElement.class).waitForFirst();
```
DiegoCardoso, line 36: NIT: the `crud` variable is never used, so it doesn't need to be declared.

Also, there are some unused imports.

DiegoCardoso (approved):
The changes look good to me, only one small suggestion left.

Issue #856, opened 2021-04-29
Create a lit-based renderer


tomivirkki:
The Lit-based renderer should work similarly to the current `TemplateRenderer.of()`  but with a different binding syntax.

The API design (wip):

```java
grid.addColumn(LitRenderer.<Person>of("Hello ${item.name}")
    .withProperty("name", (i) -> i.getFirstName()))
    .setHeader("Name");
```

# Implementation notes
- Prototype [here](https://vaadin.slack.com/archives/C3TGRP4HY/p1583914760237500) and [here](https://vaadin.slack.com/archives/C6X43FE8M/p1607938970141700?thread_ts=1607937403.141000&cid=C6X43FE8M)

The code for the `LitRenderer` class should be in the flow-components repo (not in the flow repo, where `PolymerTemplate` is currently).

Part of: https://github.com/vaadin/flow-components/issues/916

PR #1881, opened 2021-07-08
feat!: do not use the `Renderer` base class for `LitRenderer`

vlukashov:
## Description

The `Renderer` class has the use of Polymer and `<template>` tags as a part of it's API contract. It's visible in
 - template element as a the constructor argument
 - reference to Polymer in the `withEventHandler` method name
 - `getTemplateElement` method in the `Rendering` interface

The `LitRender` class has a different set of assumptions in its API contract:
 - use of JS renderer functions instead of `<template>` tags
 - use of Lit template syntax instead of the Polymer template syntax

Since these 2 sets are sufficiently different, there is no point in making `LitRenderer` extend the `Renderer` class. And since the current `Renderer` class is planned to be deprecated and removed, it's better not to add more usage of it.

Fixes https://github.com/vaadin/flow-components/issues/856

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [ ] I have read the contribution guide: https://vaadin.com/docs-beta/latest/guide/contributing/overview/


vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/java/com/vaadin/flow/renderer/LitRenderer.java
```diff
@@ -62,25 +64,23 @@
...
 
-        // Generate a unique (in scope of the UI) namespace for the renderer
-        // properties.
-        int litRendererCount = UI.getCurrent().getElement()
-                .getProperty("__litRendererCount", 0);
-        UI.getCurrent().getElement().setProperty("__litRendererCount",
-                litRendererCount + 1);
+        // Generate a unique (in scope of the JVM instance) namespace for
+        // the renderer properties.
+        int litRendererCount = instanceCounter.incrementAndGet();
```
tomivirkki, line 83: Using a static counter for the namespace may be problematic since the accumulated number (available for everyone) could be used to get an approximation of how many users an application has.

Issue #1016, opened 2021-03-23
TreeGrid scrollToIndex problem on large tree with expanded nodes


tarekoraby:
### Description
TreeGrid scrollToIndex is not working properly on a large tree with expanded nodes. 

### Steps to reproduce
In a 14.4.10 project, add the following to a view and click on the button

```
TreeGrid<String> grid = new TreeGrid<>();
grid.addHierarchyColumn(String::toString).setHeader("Name");
TreeData<String> td = grid.getTreeData();
td.addItem(null, "root");
for (int i = 0; i < 100; i++) {
    td.addItem("root", "child " + i);
    for (int k = 0; k < 100; k++) {
        td.addItem("child " + i, "grand-child " + i + "-" + k);
    }
}
Button button = new Button("Expand and scroll to index 500", e -> {
    grid.expandRecursively(Stream.of("root"), 2);
    grid.scrollToIndex(500);
    // UI.getCurrent().beforeClientResponse(grid, ctx -> grid.scrollToIndex(500));
});
add(grid, button);
```

### Expected outcome
The tree expands and the scrolls to index 500

### Actual outcome
The tree expands, but no scrolling occurs.

A second button click is required to scroll to the index. Notice, also that workaround of calling `UI.getCurrent().beforeClientResponse(grid, ctx -> grid.scrollToIndex(500));` (which works in the case of shallow trees) doesn't seem to work with large trees with expanded nodes.

sosa-vaadin:
Due to the difficulties encountered when using the `scrollToIndex(int)` method with `TreeGrid`, we're discouraging its use with `TreeGrid`. The official documentation for `TreeGrid` and the JavaDoc for the method were updated to warn users about these issues.

A [new issue](https://github.com/vaadin/web-components/issues/2228) was opened to eventually provide an implementation with predictable behavior.

TatuLund:
Just pointing earlier comment https://github.com/vaadin/flow-components/issues/1016#issuecomment-829952022

> Introduce eager loading to the tree grid, so that children of an expanded node are always loaded eagerly [Github link](https://github.com/vaadin/vaadin-grid/issues/1914#issuecomment-683806711). That would allow gradually expanding nodes until the specific item is loaded. At that point the grid could identify the scroll position for that item.

This has been effectively done now: https://github.com/vaadin/flow-components/issues/2648

PR #1882, opened 2021-07-08
docs: add warning about unreliable behaviour of scrollToIndex(int) in TreeGrid

sosa-vaadin:
## Description

The `scrollToIndex(int)` method in `TreeGrid` is directly inherited from `Grid` but has some issues. The grid needs to know the entire tree structure to accurately calculate a node's index (in a pre-order traversal method). Since it uses lazy loading for performance reasons, there is no reliable way to consistently scroll to a node's position.

Let's document this behavior in the `TreeGrid` class version of the method and discourage the use of it until we find a proper solution.

Related: #1016, https://github.com/vaadin/web-components/issues/2228


DiegoCardoso (approved):

manolo:
Shouldn't this be cherry-picked to 20.0 ?

PR #1903, opened 2021-07-14
fix: disabled items can now be part of a CheckboxGroup's valid value

### LINKED ISSUE #1185: CheckBoxGroup: Selection issue when using ItemEnabledProvider
### Description
CheckBoxes can't be selected, when applying a value before adding a ItemEnabledProvider.
The CheckBox is not preselected if the value is applied before the ItemEnabledProvider is set.

checkboxgroup

### Example 
```
@Route("test")
public class Test extends VerticalLayout {

    private static final String A = "A";
    private static final String B = "B";
    private static final String C = "C";

    public Test() {

        CheckboxGroup<String> checkboxGroup1 = new CheckboxGroup<String>();
        checkboxGroup1.setLabel("without preselected value");
        checkboxGroup1.setItems(A, B, C);
        checkboxGroup1.setItemEnabledProvider(value -> !value.equals(B));

        CheckboxGroup<String> checkboxGroup2 = new CheckboxGroup<String>();
        checkboxGroup2.setLabel("with preselected value");
        checkboxGroup2.setItems(A, B, C);
        checkboxGroup2.setValue(Set.of(B));
        checkboxGroup2.setItemEnabledProvider(value -> !value.equals(B));

        CheckboxGroup<String> checkboxGroup3 = new CheckboxGroup<String>();
        checkboxGroup3.setLabel("with preselected value and values set after ItemEnabledProvider");
        checkboxGroup3.setItems(A, B, C);
        checkboxGroup3.setItemEnabledProvider(value -> !value.equals(B));
        checkboxGroup3.setValue(Set.of(B));

        add(checkboxGroup1, checkboxGroup2, checkboxGroup3);

    }

}

```

### Versions
- Vaadin / Flow version 14.1.3
- Java version jdk-11.0.2
- OS version Windows 10
- Browser version Google Chrome 77.0.3865.75  (64-Bit) / Firefox 72.0.2 (64-Bit)

sosa-vaadin:
## Description

EDIT: the "valid" logic was replaced for one where the values of disabled items are checked to see if they changed. If they did, then it's an invalid change and will be rolled back.

The value of a CheckboxGroup is a Set of all selected (checked) items. Items in a CheckboxGroup can have their enabled state managed with an ItemEnabledProvider. This ItemEnabledProvider is also used to check whether the value itself of the CheckboxGroup is valid. If any selected (checked) item is **disabled** (when checked against the ItemEnabledProvider), then the CheckboxGroup is said to have an invalid value. This check happens every time the value changes.

If the new value is invalid, it immediately rolls back to its previous value. This explains the problems encountered in the issue this PR is trying to fix. By removing this "valid" value logic, any item, whether enabled or disabled, will not invalid the CheckboxGroup's value. The user will still be unable to select (by clicking or focusing) a disabled checkbox item, but its selection can still be set programmatically. Any combination of selected items should be a valid value for the group, regardless of those items being enabled or disabled.

Fixes #1185 

## Type of change

- [x] Bugfix

## Checklist


sissbruecker (commented):
Preliminary review. 

I have some trouble understanding what the previous validation logic was about, and therefore am not sure about it's complete removal. I'm wondering if there is anything left that should still be validated (can't think of anything though)?. Need to clarify that.

However I would suggest to add regression ITs for the scenarios that were discussed in the issue:
- set item enabled provider to disable checkboxes, verify that a disabled checkbox can be changed programmatically
- preselect value in a checkbox, set item enabled provider to disable some checkboxes, verify that non-disabled checkboxes can still be changed manually / from the browser

sissbruecker:
Added regression tests

sissbruecker (dismissed):

alvarezguille (approved):
LGTM

PR #1906, opened 2021-07-16
fix: allow partial I18N instances for Upload

sissbruecker:
## Description

Currently the I18N feature of the Upload component only works when setting a fully specified `UploadI18N` instance. Just overriding a single label like this:
```
upload.setI18N(new UploadI18N().setAddFiles(new UploadI18N.AddFiles().setOne("Upload Report...")));
```

does not work, because it results in null values being set in the web components I18N config, which in turn results in null reference errors.

This change fixes that by:
- removing all null values from the UploadI18N instance before setting it in the web component
- merging the config with the existing default config from the web component, so that as a fallback the default translations from the web component are used

Fixes: No related issue

## Type of change

- [x] Bugfix

## Checklist

### DIFF
--- a/vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
+++ b/vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
@@ -41,6 +41,7 @@ import com.vaadin.flow.shared.Registration;
 import elemental.json.JsonArray;
 import elemental.json.JsonNull;
 import elemental.json.JsonObject;
+import elemental.json.JsonType;
 
 /**
  * Server-side component for the {@code vaadin-upload} element.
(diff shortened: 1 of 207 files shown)


vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -560,17 +561,48 @@ public void setI18n(UploadI18N i18n) {
...
+                                + "  status: uploadingStatus,"
+                                + "  remainingTime: uploadingRemainingTime,"
+                                + "  error: uploadingError," + "};"
+                                + "const units = $0.units || this.i18n.units;"
+                                + "this.i18n = Object.assign({}, this.i18n, $0, {"
+                                + "  addFiles: addFiles,"
+                                + "  dropFiles: dropFiles,"
+                                + "  uploading: uploading," + "  units: units,"
+                                + "});",
+                        i18nJson);
```
sissbruecker, line 591: Couldn't identify an easier / safer way to do a deep merge. Considered:
- importing a merge function from a library like lodash - that might have too big of an impact for this change alone
- copying a merge function from Github / SO - might have bugs or not work as expected in this use case (e.g. merging `units` array). Also would be terrible to put such a complex function inline into the Java code

Still, I'm open for other ideas on improving this. I tried to make the tests rather extensive to make up for this.

Note that the `units` array is not merged. If the UploadI18N config specifies one, it will always replace the one from the web component, even if the number of elements don't match. However that is the same behaviour as we have today.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -560,17 +561,48 @@ public void setI18n(UploadI18N i18n) {
...
+
+        // Remove null values so that we don't overwrite existing WC
+        // translations with empty ones
+        deeplyRemoveNullValuesFromJsonObject(i18nJson);
+
+        // Assign new I18N object to WC, by deeply merging the existing
+        // WC I18N, and the values from the new UploadI18N instance,
+        // into an empty object
+        getElement().executeJs(
+                "const dropFiles = Object.assign({}, this.i18n.dropFiles, $0.dropFiles);"
```
sosa-vaadin, line 580: Would it make more sense to use a StringBuilder since a lot of concatenation is happening here?

sosa-vaadin (approved):
I can only consider using StringBuilder where there's a ton of concatenation. I don't see any major problems with this though.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
sissbruecker: Looks like the compiler will [use StringBuilder under the hood](https://stackoverflow.com/a/1532499/11368281) if it thinks that it would optimize the performance. In terms of readability I prefer the String concatenation to chained `.append`  calls

PR #1909, opened 2021-07-19
fix: emptyItem now always uses emptySelectionCaption value for its label and text (value)

### LINKED ISSUE #1085: [BUG] NPE when Select is configured for EmptySelectionAllowed
**Describe the bug**
Configuring a Select component to allow/use an empty selection causes an NPE. 

**To Reproduce**
Set the flag to true on a select component -> select.setEmptySelectionAllowed( true );
Set the label generator (required for non String(y) objects) -> select.setItemLabelGenerator( <some method> );

Once the view/layout is constructed, the Select code tries to use the label generator on the empty selection item, which was constructed/set earlier in the Select code.. and contains a null object

Fix seems pretty easy -> Select.class in private void updateItem(VaadinItem<T> vaadinItem)

`
        if (this.getItemLabelGenerator() != null) {
            vaadinItem.getElement().setAttribute("label", this.getItemLabelGenerator().apply(item));
        } else if (item == this.emptySelectionItem) {
            vaadinItem.getElement().setAttribute("label", "");
        } else {
            vaadinItem.getElement().removeAttribute("label");
        }
`

Just need to put the check for the emptySelectionItem first, and then check for the label generator.

**Expected behavior**
This shouldn't NPE when trying to use an empty selection.

**Screenshots**

**Desktop (please complete the following information):**
This is going to happen no matter where it's executed.

**Additional context**

sosa-vaadin:
## Description

Using an `ItemLabelGenerator` with empty item selection allowed (`setEmptySelectionAllowed(true)`), the label generator must do null checks to avoid throwing a `NullPointerException`. The `emptySelectionCaption` is completely ignored in this scenario. Users should be able to use `emptySelectionCaption` to customize the label for the "empty" selection, instead of the label generator.

Fixes #1085 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-select-flow-parent/vaadin-select-flow-integration-tests/src/test/java/com/vaadin/flow/component/select/test/EmptySelectionItemIT.java
+++ b/vaadin-select-flow-parent/vaadin-select-flow-integration-tests/src/test/java/com/vaadin/flow/component/select/test/EmptySelectionItemIT.java
@@ -90,9 +90,8 @@ public class EmptySelectionItemIT extends AbstractSelectIT {
         verify.selectedItem("caption", "caption");
 
         page.toggleItemLabelGenerator(true);
-        // the selected label should be updated accordingly to replace the
-        // caption
-        verify.selectedItem("caption", "null-LABEL");
+        // the selected label should stay the same
+        verify.selectedItem("caption", "caption");
 
         page.toggleItemLabelGenerator(false);
         verify.selectedItem("caption", "caption");
--- a/vaadin-select-flow-parent/vaadin-select-flow/src/main/java/com/vaadin/flow/component/select/Select.java
+++ b/vaadin-select-flow-parent/vaadin-select-flow/src/main/java/com/vaadin/flow/component/select/Select.java
@@ -804,6 +804,7 @@ public class Select<T> extends GeneratedVaadinSelect<Select<T>, T>
         vaadinItem.removeAll();
         T item = vaadinItem.getItem();
 
+        // handle the item's text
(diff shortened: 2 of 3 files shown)


sissbruecker (changes requested):
I think this would be a breaking change for multiple reasons:
- there might be users who rely on the item generator to generate a specific label for the `null` item
- the API docs for `setEmptySelectionCaption` currently state that it is **not used** for the label, but that the placeholder will be used if no item label generator is defined

Also the API docs for `setEmptySelectionCaption` specify exactly the behaviour as it was implemented:
- emptySelectionCaption is only used for the text of the empty item in the dropdown
- the label for when the empty item is selected is either:
  - the result of the item generator, when one is provided
  - or the placeholder text
  - or empty string if there is no placeholder

In my opinion this isn't a bug at all, it's the expected and documented behaviour.

Options I see here:
- We can choose to do nothing
- We can do a fix by wrapping the item label generator call for the empty item into a `try/catch`, and use the placeholder, or an empty string, as fallback if an exception occurs
- We can choose to change that behaviour, but then as a something feature-like in a new major
  - personally I would favor keeping the current behaviour as it gives you more flexibility in that you can specify different values for the empty item in the dropdown, and for the label when it is selected
- We can choose to make it more obvious that the item label generator might receive a null value, e.g. by using `Optional`, again a breaking change in a new major

sosa-vaadin:
Closing this as this behavior is what was intended and the responsibility of handling null values in the item label generator falls under the developer.

PR #1992, opened 2021-08-10
refactor: replace NpmPackage annotation with custom date-fns bundle

sissbruecker:
## Description

Changes the date picker to include a custom bundle of the date-fns library to allow back-porting the custom date format feature to Vaadin 14. The reason for the change is that the current mechanism of including the date-fns dependency though `@NpmPackage` is not compatible with Vaadin 14 compatibility mode. Instead of only applying this change to the version 14 branch, I would suggest to also use this solution in the latest version to make future maintenance / back-porting easier. This solution can be removed / replaced with `@NpmPackage` at some point when 14 is not maintained anymore.

General solution outline:
- adds a `date-fns` folder to the Flow date picker Maven module
- that folder contains a node.js / rollup build that bundles only the required date-fns function into a single JS file
- the rollup build is automatically executed during the `generate-resources` Maven phase, using the `maven-frontend-plugin`
- the resulting bundle is copied to the target output, so that it will be packaged into the date picker JAR file, together with the date picker connector
- the bundle is imported using `@JsModule`, same as the date picker connector
- the bundle exposes the date-fns functions as a global under `window.Vaadin.Flow.datepickerDateFns`
- the date picker connector uses this global to access these functions

TODO:
- [ ] Check browser support for Vaadin 14 and if we need polyfills for the date-fns code
- [ ] Check with platform team regarding:
  - changes to the Maven build process
  - OSGI support

### DIFF
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/date-fns/.gitignore
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/date-fns/.gitignore
@@ -1,0 +1,7 @@
+node
+output
+node_modules
+
+# Include package.json, which is excluded by outer .gitignore
+!package.json
+!package-lock.json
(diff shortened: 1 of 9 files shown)


sissbruecker:
Closing in favor of #2002 as we decided to apply the custom date-fns bundle workaround only to Vaadin 14.

Issue #1914, opened 2021-07-20
Memory leak when repeatedly setting DataProvider


edler-san:
Minimum example:
```
@Route("")
public class MainView extends VerticalLayout {

public MainView() {
    Grid<Object> grid = new Grid<>();
    ListDataProvider<Object> provider = new ListDataProvider<>(new ArrayList<>());
    grid.setDataProvider(provider);

    add(grid);
  
    Button refreshButton = new Button("Refresh");
    refreshButton.addClickListener(e -> provider.refreshAll());
    add(refreshButton);
    }
}
```
When the setDataProvider method is called, it internally will call "handleAttach" on the DataCommunicator. This will register a DataProviderListener and return a DataProviderUpdateRegistration. However, when adding the Grid to a layout, handleAttach will be called again which will register another DataProviderListener, because the handleAttach method of the DataCommunicator does not check if the DataProviderUpdateRegistration already has a value.

Now there are two identical DataProviderListeners, with one having no registration to remove it.
Once you update/refresh the grid, BOTH of these Listeners will be called in the fireEvent method of the AbstractDataProvider, which causes unnecessary calls to the reset() method. When you then click the Refresh button below, you will see that the fireEvent method will have three listeners, where two of those will point to the same DataCommunicator, and so on.

Tested with Vaadin 14.6.6.

TatuLund:
Note, this can happen with ComboBox too.

mshabarov:
Reopened, because the proposed change breaks `ComboBox` filtering tests and lead to regression. Need to consider a fix on `ComboBox` side.

mshabarov:
The fix on `ComboBox` side is proposed in the corresponding PR.

Here are some clarifications why the tests in CB failed:

How it worked before both fixes (in Flow and ComboBox):
1. DataCommunicator adds a data provider listener upon its `setDataProvider` method.
2. ComboBox adds a data provider listener upon its `setDataProvider` method.
3. DataCommunicator adds one more data provider listener upon attach event.
4. ComboBox adds one more data provider listener upon attach event.

So finally DC listener comes before CB listener and this works fine.

How it worked after the fix in Flow:
1. DataCommunicator adds a data provider listener upon its `setDataProvider` method.
2. ComboBox adds a data provider listener upon its `setDataProvider` method.
3. DataCommunicator REMOVES the old data provider listener upon attach event and adds a new one.
4. ComboBox adds one more data provider listener upon attach event.

[...]

PR #2005, opened 2021-08-12
fix: Remove old data provider listener on component attach

mshabarov:
## Description

This change is a fix for ComboBox and related to the fix made on the flow side to get rid of memory leaks caused by old listeners being hanging in the list - https://github.com/vaadin/flow/pull/11471.

Null-value checking of `dataProviderListener` is not needed in attach handler because:
1. Calling to a `remove()` is more robust and this is a normal way of handling listeners registrations in flow. Additionally, if, for some reason, a new data provider will be set up to the component, it's more correct to remove the old listener sitting in the old data provider.
2. Not calling of `remove()` method gives an inconsistency with the flow, which now (with the above fix) removes the listener on attach event.

The inconsistency described in 2. causes the ComboBox's and DataCommunicator data provider listeners order violation.
This causes a side-effects in client-side filtering, it's seen from the failed tests.
I didn't dig deeply into the details what a the side effect in deep, but it's clear that it is somehow overrides/cleans filter/items/size variables in data communicator -> break it's internal state.

How it worked before both fixes:
1. DataCommunicator adds a data provider listener upon its `setDataProvider` method.
2. ComboBox adds a data provider listener upon its `setDataProvider` method.
3. DataCommunicator adds one more data provider listener upon attach event.
4. ComboBox adds one more data provider listener upon attach event.

So finally DC listener comes before CB listener and this work fine.

How it worked after the fix in Flow:
1. DataCommunicator adds a data provider listener upon its `setDataProvider` method.
2. ComboBox adds a data provider listener upon its `setDataProvider` method.
3. DataCommunicator REMOVES the old data provider listener upon attach event and adds a new one.
4. ComboBox adds one more data provider listener upon attach event.

So now the CB listener comes first (because the first listener in the queue has been removed), and this violates the order of listener execution. And this causes a side-effects and test failures.

How it works after the fix in Flow and ComboBox:
1. DataCommunicator adds a data provider listener upon its `setDataProvider` method.
2. ComboBox adds a data provider listener upon its `setDataProvider` method.
3. DataCommunicator REMOVES the old data provider listener upon attach event and adds a new one.
4. ComboBox REMOVES the old data provider listener upon attach event and adds a new one.

Thus, there are only two listeners in the end and they are in the right order -> works fine.

No tests added, because this change is covered by tests in `FilteringIT`.

Related-to https://github.com/vaadin/flow-components/issues/1914

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
@@ -942,7 +942,7 @@ public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
     protected void onAttach(AttachEvent attachEvent) {
         super.onAttach(attachEvent);
         DataProvider<T, ?> dataProvider = getDataProvider();
-        if (dataProvider != null && dataProviderListener == null) {
+        if (dataProvider != null) {
             setupDataProviderListener(dataProvider);
         }
 
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridTest.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridTest.java
@@ -17,9 +17,6 @@
 package com.vaadin.flow.component.grid;
 
 import java.util.Arrays;
-import java.util.Collections;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Stream;
 
 import org.junit.Assert;
@@ -27,13 +24,10 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 
-import com.vaadin.flow.component.UI;
 import com.vaadin.flow.component.grid.dataview.GridListDataView;
 import com.vaadin.flow.data.provider.DataCommunicatorTest;
 import com.vaadin.flow.data.provider.DataProvider;
-import com.vaadin.flow.data.provider.DataProviderListener;
-import com.vaadin.flow.data.provider.ListDataProvider;
-import com.vaadin.flow.shared.Registration;
+import com.vaadin.tests.DataProviderListenersTest;
 
 public class GridTest {
 
@@ -111,64 +105,10 @@ public class GridTest {
     }
 
     @Test
-    public void dataProviderListeners_gridAttached_oldDataProviderListenerRemoved() {
-        AtomicInteger listenersCount = new AtomicInteger(0);
-        AtomicBoolean oldListenerRemoved = new AtomicBoolean();
-
-        // given
-        Grid<String> grid = new Grid<>();
-        ListDataProvider<String> dataProvider = new ListDataProvider<String>(
-                Collections.emptyList()) {
-            @Override
-            public Registration addDataProviderListener(
-                    DataProviderListener<String> listener) {
-                listenersCount.incrementAndGet();
-                Registration registration = super.addDataProviderListener(
-                        listener);
-                // the first listener is added by Grid in 'setDataProvider'
-                // and it is the one we want to be removed.
-                return listenersCount.get() == 1
-                        ? Registration.combine(registration,
-                                () -> oldListenerRemoved.set(true))
-                        : registration;
-            }
-        };
-
-        // when
-        grid.setDataProvider(dataProvider);
-
-        // then
-        Assert.assertEquals(
-                "Expected exactly two data provider listeners added by "
-                        + "DataCommunicator's and Grid's 'setDataProvider' methods",
-                2, listenersCount.get());
-        Assert.assertFalse(
-                "Expected data provider listener added by Grid not to be "
-                        + "removed before Grid being attached",
-                oldListenerRemoved.get());
-
-        // given
-        listenersCount.set(0);
-
-        // when
-        DataCommunicatorTest.MockUI mockUI = new DataCommunicatorTest.MockUI();
-        mockUI.add(grid);
-        fakeClientCommunication(mockUI);
-
-        // then
-        Assert.assertEquals(
-                "Expected exactly two data provider listeners added by "
-                        + "DataCommunicator's and Grid's attach handle methods",
-                2, listenersCount.get());
-        Assert.assertTrue(
-                "Expected old data provider listener to be removed after "
-                        + "grid being attached",
-                oldListenerRemoved.get());
-    }
-
-    private void fakeClientCommunication(UI ui) {
-        ui.getInternals().getStateTree().runExecutionsBeforeClientResponse();
-        ui.getInternals().getStateTree().collectChanges(ignore -> {
-        });
+    public void dataProviderListeners_gridAttachedAndDetached_oldDataProviderListenerRemoved() {
+        DataProviderListenersTest
+                .checkOldListenersRemovedOnComponentAttachAndDetach(
+                        new Grid<>(), 2, 2, new int[] { 0, 2 },
+                        new DataCommunicatorTest.MockUI());
     }
 }
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridTest.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridTest.java
@@ -17,6 +17,9 @@
 package com.vaadin.flow.component.grid;
 
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Stream;
 
 import org.junit.Assert;
@@ -24,8 +27,13 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 
+import com.vaadin.flow.component.UI;
 import com.vaadin.flow.component.grid.dataview.GridListDataView;
+import com.vaadin.flow.data.provider.DataCommunicatorTest;
 import com.vaadin.flow.data.provider.DataProvider;
+import com.vaadin.flow.data.provider.DataProviderListener;
+import com.vaadin.flow.data.provider.ListDataProvider;
+import com.vaadin.flow.shared.Registration;
 
 public class GridTest {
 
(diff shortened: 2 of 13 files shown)


vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -942,7 +942,7 @@ private void executeRegistration(PropertyChangeEvent event) {
     protected void onAttach(AttachEvent attachEvent) {
         super.onAttach(attachEvent);
         DataProvider<T, ?> dataProvider = getDataProvider();
-        if (dataProvider != null && dataProviderListener == null) {
+        if (dataProvider != null) {
```
pleku, line 945: Unit test please ?

pleku (changes requested):

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
mshabarov: Initially, I was thinking that this change is covered by `FilteringIT`. But this filtering test is a separate thing, and it's quite hard to make connections between filtering and DP listeners. So, it's a good idea to add a specific test covering this particular case with DP listeners.

mshabarov: The `Grid` has the same code (not leading to a test failures though), so I'll fix it as well.

mshabarov: And some other component which use data provider...:)

mshabarov: Done

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridTest.java
```diff
@@ -101,4 +109,66 @@ public void setAllRowsVisibleProperty_isHeightByRowsAndIsAllRowsVisibleWork() {
...
+                Collections.emptyList()) {
+            @Override
+            public Registration addDataProviderListener(
+                    DataProviderListener<String> listener) {
+                listenersCount.incrementAndGet();
+                Registration registration = super.addDataProviderListener(
+                        listener);
+                // the first listener is added by Grid in 'setDataProvider'
+                // and it is the one we want to be removed.
+                return listenersCount.get() == 1
```
pleku, line 130: I'm trying to wrap my head around why is this needed for the test, but cannot, so please elaborate ? I understand collecting the number of listeners present, but don't understand why this changes the logic of the method directly. Seems odd.

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridTest.java
```diff
@@ -101,4 +109,66 @@ public void setAllRowsVisibleProperty_isHeightByRowsAndIsAllRowsVisibleWork() {
...
+                "Expected data provider listener added by Grid not to be "
+                        + "removed before Grid being attached",
+                oldListenerRemoved.get());
+
+        // given
+        listenersCount.set(0);
+
+        // when
+        DataCommunicatorTest.MockUI mockUI = new DataCommunicatorTest.MockUI();
+        mockUI.add(grid);
```
pleku, line 155: As the detach is also removing the listener(s) and it is a critical part of keeping things working, is there another unit test that verifies that detach does it or could you possibly also include verifying that detach cleans up the listeners, so that we can guarantee that the next attach will work ?

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridTest.java
```diff
@@ -101,4 +109,66 @@ public void setAllRowsVisibleProperty_isHeightByRowsAndIsAllRowsVisibleWork() {
...
+                // and it is the one we want to be removed.
+                return listenersCount.get() == 1
+                        ? Registration.combine(registration,
+                                () -> oldListenerRemoved.set(true))
+                        : registration;
+            }
+        };
+
+        // when
+        grid.setDataProvider(dataProvider);
```
pleku, line 138: The logic feels off to me that the listeners are added
a) when the component is attached - OK
b) when the data provider is added - I would expect that here it is only added if the component was attached to begin with

vaadin-radio-button-flow-parent/vaadin-radio-button-flow/src/test/java/com/vaadin/flow/component/radiobutton/RadioButtonGroupTest.java
```diff
@@ -417,4 +424,63 @@ public void addNullOption_setValue() {
...
+        // when
+        DataCommunicatorTest.MockUI mockUI = new DataCommunicatorTest.MockUI();
+        mockUI.add(rbg);
+        fakeClientCommunication(mockUI);
+
+        // then
+        Assert.assertEquals(
+                "Expected exactly one data provider listener added by "
+                        + "RBG's attach handle method",
+                2, listenersCount.get());
```
pleku, line 474: Again, it verifies for against two but there should be only one ?

pleku (changes requested):
There is a lot of repetition in the unit tests, which probably could be refactored into something reusable so that maintenance could be easier ?

vaadin-list-box-flow-parent/vaadin-list-box-flow/src/test/java/com/vaadin/flow/component/listbox/test/ListBoxUnitTest.java
```diff
@@ -163,6 +171,65 @@ public void setIdentifierProviderOnId_setItemWithNullId_shouldThrowException() {
...
+        // when
+        DataCommunicatorTest.MockUI mockUI = new DataCommunicatorTest.MockUI();
+        mockUI.add(listBox);
+        fakeClientCommunication(mockUI);
+
+        // then
+        Assert.assertEquals(
+                "Expected exactly one data provider listener added by "
+                        + "ListBox's attach handle method",
+                2, listenersCount.get());
```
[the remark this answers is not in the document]
mshabarov: Expected count = 2 here is the accumulated value. I will reset the counter before component attach in order to improve the readability.

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridTest.java
mshabarov: The goal of doing that is no only check that the given listener has been removed. Since there are several listeners added, including ones added by DataCommunicator, I use the known order of this listener. In case of Grid, the listener added in `setDataProvider` method by Grid is the first one in the list and it is supposed to be removed upon  `onAttach` method.

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridTest.java
mshabarov: Yes, good suggestion, I'll check and add if needed.

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridTest.java
mshabarov: Basically, I agree that in most cases it's a non sense to call `refreshAll` or `refreshItem` on detached component, because the updated won't be seen anyway. But there are a couple of concerns to do that:
1. Some 'exotic' cases when `refresh` methods are called even before component attach and some logic relies on that.
2. Should we distinguish between 'initial' attach and further re-attach events? I mean, for the initial attach we don't need to add the listener, but should we add the listener upon `setDataProvider` between further reattachments? The logic goes more complex and this is a door for regressions.

vaadin-radio-button-flow-parent/vaadin-radio-button-flow/src/test/java/com/vaadin/flow/component/radiobutton/RadioButtonGroupTest.java
mshabarov: `listenersCount` will be re-set so as to validate agains expected '1' value, not '2'.

vaadin-list-box-flow-parent/vaadin-list-box-flow/src/main/java/com/vaadin/flow/component/listbox/ListBoxBase.java
```diff
@@ -102,6 +102,7 @@ public void setDataProvider(DataProvider<ITEM, ?> dataProvider) {
         this.dataProvider.set(Objects.requireNonNull(dataProvider));
         DataViewUtils.removeComponentFilterAndSortComparator(this);
         clear();
+        rebuild();
```
pleku, line 105: Did you just fix another regression or what is this change about ? 😄 

Wondering if it should be highlighted as an additional fix. Was there a test failure that occurred that caught it ? IT? Or unit test ?

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridTest.java
pleku: I think it should work in simply following heurestic:
When setting a data provider
a) if component is attached, initialize everything (most things only take effect on before-client-response phase)
b) if component is not attached, wait for it to be attached before doing everything
On detach things are cleaned up as much as possible, but it may wary (currently) between components whether the data is reused or not (probably not). 
On setting a new data provider, a) and b) rules should always apply.

So I don't at the moment see the reason for the more complex logic between initial and reattach (when it comes to data provider things, browser related things are different).

vaadin-list-box-flow-parent/vaadin-list-box-flow/src/main/java/com/vaadin/flow/component/listbox/ListBoxBase.java
mshabarov: Good question!
This was another regression, which was revealed when I deleted registration `null` check in `ListBox`.
It fails `ListBoxIT.componentsBetweenItems` tests and the reason was that `rebuild()` method was in the wrong place and it was called mistakenly in `setupDataProviderListener`.
I placed it next to `clear()` method, as it done in the data provider listener.

I can place it into additional fix, but then I'll need to revert all changes in ListBox in this PR.

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridTest.java
mshabarov: Added

vaadin-list-box-flow-parent/vaadin-list-box-flow/src/test/java/com/vaadin/flow/component/listbox/test/ListBoxUnitTest.java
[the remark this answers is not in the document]
mshabarov: Reworked.

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridTest.java
mshabarov: Thanks for clarifications! I'll try to implement this logic.

vaadin-list-box-flow-parent/vaadin-list-box-flow/src/main/java/com/vaadin/flow/component/listbox/ListBoxBase.java
pleku: It is fine I think (like this), but please just highlight this in the commit description when merging.

pleku (approved):

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridTest.java
mshabarov: @pleku you approved this PR, so is it fine if I implement this suggestion in the separate PR?

pleku: Yes I think it should be a separate PR as it is additional and between refactor/bugfix. And another issue should be created as this does handle the original one.

pleku: And the point about calling `refresh` or `refreshAll` doesn't make sense in my opinion if the component is not attached - since there should not be anything cached (or is there and then why?) then the _refresh_ is a NOOP as all the items will be anyway updated/refetched once the component is attached again.

In case some component retains the fetched items over detach+reattach, that sounds like premature optimization to me.

PR #2006, opened 2021-08-14
feat: add enum for lumo icons

### LINKED ISSUE #1878: Java API for lumo icons is missing
Reporting here since 
* `vaadin-lumo-styles` webjar is used by `vaadin-lumo-theme` module
https://github.com/vaadin/flow/blob/master/flow-theme-integrations/lumo/pom.xml

* `lumo/icons.html` HtmlImport is done in `Lumo`
https://github.com/vaadin/flow/blob/master/flow-theme-integrations/lumo/src/main/java/com/vaadin/flow/theme/lumo/Lumo.java#L36

maltindal:
The java API for the lumo icons is missing.

This change adds a dedicated enum similar to the VaadinIcon enum.

Also adds demo for creating all lumo icons. For this a method has been extracted for creating one layout containing the icon with the name.

Fixes #1878.

Usage of `LumoIcon` will require to have `@JsModule("@vaadin/vaadin-lumo-styles/vaadin-iconset.js")` set. The demo already had this included already. Should we give a hint about this the javadoc of `LumoIcon`?

### DIFF
--- a/vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/LumoIcon.java
+++ b/vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/LumoIcon.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2000-2021 Vaadin Ltd.
+ * Copyright 2000-2022 Vaadin Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -13,7 +13,10 @@
  * License for the specific language governing permissions and limitations under
  * the License.
  */
-package com.vaadin.flow.component.icon;
+package com.vaadin.flow.theme.lumo;
+
+import com.vaadin.flow.component.icon.Icon;
+import com.vaadin.flow.component.icon.IconFactory;
 
 import java.util.Locale;
 
--- a/vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/LumoIcon.java
+++ b/vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/LumoIcon.java
@@ -27,6 +27,9 @@ import java.util.Locale;
  * <p>
  * These instances can be used to create {@link Icon} components by using their
  * {@link #create()} method.
+ * <p>
+ * NOTE: Using this enum will also include the Vaadin icon set in the frontend
+ * bundle.
  *
  * @author Vaadin Ltd
  */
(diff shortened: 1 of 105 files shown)


sissbruecker:
Thanks for the contribution, sorry it took so long to process this. I've moved the implementation into the `vaadin-lumo-theme-flow` module, which seems more appropriate.

The `@vaadin/vaadin-lumo-styles/vaadin-iconset.js` module gets automatically imported from the `Lumo` theme class, so as long as that is used, the icons will be loaded. Seems like a reasonable assumption that the `Lumo` class must be used in order for the icon to be available.

One other concern is that the implementation is based on the `Icon` class, which will also load the Vaadin icon set, even if it's not used by the application. Ideally we would have an icon base class and then specialized icon classes for Vaadin and Lumo icon sets, which then have annotations to load the respective icon set. However I can not see how we could accomplish this without a breaking change, given that the `IconFactory` expects to return an `Icon` (the Vaadin icon set specific one).
We'll check if we are OK with the Lumo enum pulling in both icon sets, otherwise we would have to put this on hold until we can introduce a breaking change in a future major.

sissbruecker:
We decided to reuse the existing `Icon` class, which leads to the `LumoIcon` enum also including the Vaadin icon set. Added a note about this to the JavaDoc.

vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/LumoIcon.java
```diff
@@ -0,0 +1,50 @@
...
+package com.vaadin.flow.theme.lumo;
+
+import com.vaadin.flow.component.icon.Icon;
+import com.vaadin.flow.component.icon.IconFactory;
+
+import java.util.Locale;
+
+/**
+ * Enumeration of all icons in the
+ * <a href= "https://vaadin.com/docs/latest/ds/foundation/icons">Lumo Icons</a>
```
web-padawan, line 25: Let's use the new docs link:

```suggestion
 * <a href= "https://vaadin.com/docs/latest/components/icons">Lumo Icons</a>
```

web-padawan (commented):
Let's add the correct `vaadin-iconset.js` to the `Lumo` class, as it still imports the wrong one:

https://github.com/vaadin/flow-components/blob/6c87286269073c5e023269cd0c42c7609360e3bf/vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/Lumo.java#L41

It should be as follows

```java
@JsModule("@vaadin/vaadin-lumo-styles/vaadin-iconset.js")
```

vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/LumoIcon.java
sissbruecker: Done

sissbruecker:
Updated the JS import on the `Lumo` class to use `@vaadin/vaadin-lumo-styles/vaadin-iconset.js`

Issue #2018, opened 2021-08-19
Setting GridVariant.LUMO_COMPACT when Grid is attached does not work


TatuLund:
In Vaadin 14 Flow I have the following code:

```
Grid grid = new Grid();
setupGrid(grid);

Button compactButton = new Button("Compact", 
    click -> grid.addThemeVariants(GridVariant.LUMO_COMPACT));
Button normalButton = new Button("Normal", 
    click -> grid.removeThemeVariants(GridVariant.LUMO_COMPACT));

```

The issue is when I click on the above buttons only the header of the grid seems to be redrawn when the buttons are clicked, the rows below the header (all the rows of the table) do not appear to be affected. They seem to stay at whatever variant they were initially set at when the screen was initially drawn (other than the header).

TatuLund:
The workaround is to force Grid to redraw upon change

```
        Button compactButton = new Button("Compact", 
        	    click -> {
        	    	grid.addThemeVariants(GridVariant.LUMO_COMPACT);
        	    	grid.getElement().executeJs("this.notifyResize()");
        	    });
        Button normalButton = new Button("Normal", 
        	    click -> {
        	    	grid.removeThemeVariants(GridVariant.LUMO_COMPACT);
        	    	grid.getElement().executeJs("this.notifyResize()");
        	    });
```

TatuLund:
This may be related to https://github.com/vaadin/flow-components/issues/1381

web-padawan:
Could not reproduce using the latest Vaadin version. Closing as stale.

PR #2019, opened 2021-08-19
fix: Force grid to redraw after GridVariant.LUMO_COMPACT is applied or removed

TatuLund:
Fixes: https://github.com/vaadin/flow-components/issues/2018

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
@@ -2204,6 +2204,9 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
     public void addThemeVariants(GridVariant... variants) {
         getThemeNames().addAll(Stream.of(variants)
                 .map(GridVariant::getVariantName).collect(Collectors.toList()));
+        if (Stream.of(variants).anyMatch(variant -> variant == GridVariant.LUMO_COMPACT)) {
+            getElement().executeJs("this.notifyResize()");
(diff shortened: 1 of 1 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -2204,6 +2204,9 @@ public FooterRow appendFooterRow() {
     public void addThemeVariants(GridVariant... variants) {
         getThemeNames().addAll(Stream.of(variants)
                 .map(GridVariant::getVariantName).collect(Collectors.toList()));
+        if (Stream.of(variants).anyMatch(variant -> variant == GridVariant.LUMO_COMPACT)) {
```
tomivirkki, line 2207: This will only work if a developer uses `addThemeVariants` API for the purpose. If you use `grid.getThemeNames().add(GridVariant.LUMO_COMPACT.getVariantName());`, the issue is still there => should be fixed on the Web Component (notifyResize on `theme` -attribute change)

tomivirkki (changes requested):
The fix shouldn't target `master` since the vaadin-grid in V22 uses ResizeObservers internally and doesn't need `notifyResize` invoked. Only needs to be fixed for V14.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
TatuLund: Good point, I will close this in favor of that.

PR #2026, opened 2021-08-19
fix: disable on click behavior after detach and reattach

### LINKED ISSUE #1401: `disableOnClick` does not work after a Button is detached and re-attached
### Description of the bug
`disableOnClick` does not work after detach and re-attach of the `Button`.
This can happen for example when a `Dialog` containing this button is closed and then reopened.

- Create a `Dialog` subclass that has a button that shows a `Notification` and then closes the dialog when clicked.
- Call 'setDisableOnClick(true)` when the button is created.
- Create a UI with 1 instance of this dialog and a `Button` to open it.
- Run the application.
- Press the button to open the dialog.
- Double-click the button to close the dialog.  The notification will only show once.
- Press the button to open the dialog again.
- Double-click the button to close the dialog again.

### Expected behavior
The notification should again only show once.
### Actual behavior
The notification is shown twice.

This is because a hack is used to add a client-side event handler to disable on click:

```
/**
 * Initialize client side disabling so disabled if immediate on click even
 * if server-side handling takes some time.
 */
private void initDisableOnClick() {
    if (!disableOnClickConfigured) {
        getElement().executeJs("var disableEvent = function () {"
                + "if($0.getAttribute('disableOnClick')){"
                + " $0.setAttribute('disabled', 'true');" + "}" + "};"
                + "$0.addEventListener('click', disableEvent)");
        disableOnClickConfigured = true;
    }
}
```

This is gone after the button has been detached.
And it doesn't help to call `disableOnClick(true)` again because `disableOnClickConfigured` will still be `true` so the `executeJs` part is skipped.

I guess it would be best that the client-side `vaadin-button` component supports the `disableOnClickConfigured` property itself.

Also, I'm surprised that the second click event is handled because the first click already closes the `Dialog` so should detach the button.  Are click events for detached buttons still handled on the server side?  I never experienced this in Vaadin 8.

See https://stackoverflow.com/questions/63634513/reliability-of-button-disableonclick-in-vaadin-14/63635292?noredirect=1#comment112614454_63635292.

### Versions:
    - Vaadin / Flow version: 14.3.4

paodb:
## Description

When enabling disabledOnClick, a javascript function is configured so the button gets immediately disabled (on client side) on click.
When detaching and re attaching the button, this JavaScript function is not working anymore, so then the button is disabled by doDisableOnClick(), but after a server side roundrip. 
So in this case if a click listener takes too long to be processed, the button will behave differently after detaching and reattaching again.
With the proposed change, the JavaScript function will be configured again by initDisableOnClick() method and button will be disable inmediately on click.

An IT is provided that will fail if the button is not immediately disabled on click after the detach-reattach cycle, but works fine with the fix.

Fixes #1401

## Type of change

- [X] Bugfix
- [ ] Feature

## Checklist

- [X] I have read the contribution guide: https://vaadin.com/docs-beta/latest/guide/contributing/overview/
- [X] I have added a description following the guideline.
- [X] The issue is created in the corresponding repository and I have referenced it.
- [X] I have added tests to ensure my change is effective and works as intended.
- [X] New and existing tests are passing locally with my change.
- [X] I have performed self-review and corrected misspellings.

### DIFF
--- a/vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/test/java/com/vaadin/flow/component/button/tests/DetachReattachDisableOnClickButtonIT.java
+++ b/vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/test/java/com/vaadin/flow/component/button/tests/DetachReattachDisableOnClickButtonIT.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2000-2021 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component.button.tests;
+
+import com.vaadin.flow.component.button.testbench.ButtonElement;
+import com.vaadin.flow.testutil.TestPath;
+import com.vaadin.tests.AbstractComponentIT;
+import org.junit.Assert;
+import org.junit.Test;
+import org.openqa.selenium.By;
+import org.openqa.selenium.support.ui.ExpectedConditions;
+
+@TestPath("vaadin-button/detach-reattach-disable-on-click-button")
+public class DetachReattachDisableOnClickButtonIT extends AbstractComponentIT {
+
+  private ButtonElement getDisableOnClickButton() {
+    return $(ButtonElement.class).id("disable-on-click");
+  }
+
+  private void assertDisableOnClickButtonEnabled(ButtonElement disableOnClickButton) {
+    Assert.assertTrue(
+        "'Disable on click' button should be enabled", disableOnClickButton.isEnabled());
+  }
+
+  private void assertDisableOnClickButtonDisabled(ButtonElement disableOnClickButton) {
+    Assert.assertFalse(
+        "'Disable on click' button should be disabled", disableOnClickButton.isEnabled());
+  }
+
+  private void clickDisableOnClickButton(ButtonElement disableOnClickButton) {
+    getCommandExecutor().disableWaitForVaadin();
+
+    // Click 'Disable on click' button
+    disableOnClickButton.click();
+
+    // Check 'Disable on click' button is disabled
+    assertDisableOnClickButtonDisabled(disableOnClickButton);
+
+    waitUntil(
+        ExpectedConditions.elementToBeClickable($(ButtonElement.class).id("disable-on-click")),
+        2000);
+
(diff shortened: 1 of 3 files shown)


vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java
```diff
@@ -421,4 +423,14 @@ private void doDisableOnClick() {
...
     }
+
+    @Override
+    protected void onAttach(AttachEvent attachEvent) {
+        initDisableOnClick();
+    }
+
+    @Override
+    protected void onDetach(DetachEvent detachEvent) {
+        disableOnClickConfigured = false;
```
sissbruecker, line 434: I think the `disableOnClickConfigured` property can be removed. It seems the whole purpose of the flag was to check that the listener is not configured multiple times, when calling `setDisabledOnClick` multiple times in a row. Now that we always execute `initDisableOnClick` on attach, it will always be true after attaching the component, plus there is no need to run `initDisableOnClick` again in `setDisabledOnClick`.

So in summary we can:
- remove `disableOnClickConfigured`
- remove the `initDisableOnClick` call here: https://github.com/vaadin/flow-components/blob/master/vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java#L323
- remove the onDetach listener

vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/test/java/com/vaadin/flow/component/button/tests/DetachReattachDisableOnClickButtonIT.java
```diff
@@ -0,0 +1,96 @@
...
+                $(ButtonElement.class).id("disable-on-click")), 2000);
+
+        // Check 'Disable on click' button is enabled again
+        assertDisableOnClickButtonEnabled(disableOnClickButton);
+
+        getCommandExecutor().enableWaitForVaadin();
+    }
+
+    @Test
+    public void detachReattachDisableOnClickButtonTest() {
```
sissbruecker, line 64: Naming: It's not clear what this actually tests. The name should indicate that we want to assert that the button still gets disabled after detaching and reattaching. For example `testShouldBeDisabledOnClickAfterDetachingAndReattaching`

vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java
paodb: Didn't realized ```disabledOnClickConfigured``` flag was not necessary. I updated the branch with the proposed changes.

vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/test/java/com/vaadin/flow/component/button/tests/DetachReattachDisableOnClickButtonIT.java
paodb: Test was renamed.

sissbruecker (approved):
Thanks Paola, this is overall a very wholesome first contribution

Issue #1562, opened 2018-11-27
DatePickerI18n: Invaild Parameter for setFirstDayOfWeek causes browser tab to hang


AB-xdev:
### Description
A malformed parameter for `DatePickerI18n.setFirstDayOfWeek(int firstDayOfWeek)` e.g. -1 causes a browser tab to hang when you open the connected datepicker

### How to reproduce

Make a simple Vaadin Flow Project, create a class and change the content as follow:
```
@Route("")
public class DemoUI extends Div
{
	public DemoUI()
	{
		final DatePicker dp = new DatePicker(LocalDate.now());
		
		final List<String> weekdays = new ArrayList<>();
		final List<String> weekdaysshort = new ArrayList<>();
		
		for(int i = 1; i <= 7; i++)
		{
			final DayOfWeek dof = DayOfWeek.of(i);
			
			weekdays.add(dof.getDisplayName(TextStyle.FULL, Locale.getDefault()));
			weekdaysshort.add(dof.getDisplayName(TextStyle.SHORT, Locale.getDefault()));
		}

		final List<String> months = new ArrayList<>();
		for(int i = 1; i <= 12; i++)
		{
			final Month m = Month.of(i);
			months.add(m.getDisplayName(TextStyle.FULL, Locale.getDefault()));
		}

		final DatePickerI18n dpi18n =
			new DatePickerI18n()
			.setWeek("Woche")
			.setCalendar("Kalender")
			.setClear("Leeren")
			.setToday("Heute")
			.setCancel("Abbrechen")
			.setMonthNames(months)
			.setWeekdays(weekdays)
			.setWeekdaysShort(weekdaysshort)
			.setFirstDayOfWeek(-1);
		
		dp.setI18n(dpi18n);
		
		this.add(dp);
	}
}
```
Open the page in a browser and open the DatePicker. 

**Result:** Your tab hangs. Firefox asks if you want to stop the website.

### Affected Browsers
- [x] Firefox v63.0.3
- [x] Chrome v70.0.3538.110
- [x] Edge v42.17134.1.0

paodb:
Still an issue on the latest LTS version, 14.6.8. 

Example to reproduce: 

```java
DatePicker germanDatePicker = new DatePicker();
germanDatePicker.setLabel("German");
germanDatePicker.setValue(LocalDate.now());

DatePickerI18n datePickerI18n = new DatePickerI18n();
datePickerI18n.setWeek("Woche");
[...]

PR #2046, opened 2021-08-25
fix: Validate first day of the week

TatuLund:
If not valid throw exception

Fixes: #1562


sosa-vaadin (approved):
I'm not 100%, but maybe the test could be a lot shorter by omitting most of the setup and just calling `datePickerI18n.setFirstDayOfWeek(7)`

PR #2066, opened 2021-08-30
feat: allow theming individual menu items

eriklumme:
This change enables theming individual menu bar items, for example to display one item as the primary item. As the menu bar uses the context menu to display submenus, most of the changes are in there.

Depends on https://github.com/vaadin/web-components/pull/2401

Part of vaadin/flow-components#880

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/java/com/vaadin/flow/component/contextmenu/MenuItemBase.java
+++ b/vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/java/com/vaadin/flow/component/contextmenu/MenuItemBase.java
@@ -15,6 +15,9 @@
  */
 package com.vaadin.flow.component.contextmenu;
 
+import java.io.Serializable;
+import java.util.Objects;
+
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.HasComponents;
 import com.vaadin.flow.component.HasEnabled;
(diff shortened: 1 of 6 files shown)


eriklumme:
Some concerns:

* I don't use the `HasTheme` interface, because it extends `HasElement`. If I extended that, I'd have to most, if not all, of the 9 methods it defines. Alternatively I could create my custom `ThemeList`, but `ThemeListImpl` has no default constructor, and I'd also have to override `setThemeName` and `getThemeName` then. Does it have to implement `HasTheme`, and if so, which approach is preferable?
* I added a simple demo to the menu bar, as the existing `MenuBarVariant`s seem to work on the main menu buttons. They don't work for the context menu items, though, so I added no example for that. I figured if and when variants are added for submenu, then examples could be added to `MenuBar` and/or `ContextMenu`.

eriklumme:
Note that this PR depends on https://github.com/vaadin/web-components/pull/2401

sissbruecker (commented):
@yuriy-fix My concern here is that this custom implementation of the theming API differs from the custom one we already have in `MessageListItem` ([see here](https://github.com/vaadin/flow-components/blob/master/vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageListItem.java#L306-L333)):
- here we implement the `setThemeName` and `getThemeName` from `HasTheme`
- in MessageListItem we implement `addThemeNames`, `removeThemeNames`, and `hasThemeName` from `HasTheme`

Personally I think we should align this:
- we already have quite a few DX issues with differing APIs in Flow components
- here we have a chance to improve on that with just a little bit more effort, and taking a little bit more time
- once we introduce an API it becomes a pain to change it, customers will be angry

Since MessageListItem already has a stable API, it would be much easier to change the API in this PR, so I would suggest to do that.

sissbruecker (approved):
Looks good. 

Just one thing to verify: I assume that the menu bar theme variants can only be applied to root items / buttons, and not to sub items showing up in the context menu - at least the latter did not work from my testing.

PR #2096, opened 2021-09-06
Override __isSaveBtnDisabled to preserve server side enabled state

yuriy-fix:
Doesn't allow client to override save button disabled state when form is dirty

### DIFF
--- a/vaadin-crud-flow-parent/vaadin-crud-flow/src/main/java/com/vaadin/flow/component/crud/Crud.java
+++ b/vaadin-crud-flow-parent/vaadin-crud-flow/src/main/java/com/vaadin/flow/component/crud/Crud.java
@@ -89,6 +89,7 @@ public class Crud<E> extends Component implements HasSize, HasTheme {
     private Grid<E> grid;
     private CrudEditor<E> editor;
     private E gridActiveItem;
+    private boolean toolbarVisible = true;
     private boolean saveBtnDisabledOverridden;
 
     final private Button saveButton;
@@ -583,7 +584,7 @@ public class Crud<E> extends Component implements HasSize, HasTheme {
 
     /**
      * Controls visiblity of toolbar
-     * 
+     *
      * @param value
      */
     public void setToolbarVisible(boolean value) {
@@ -597,7 +598,7 @@ public class Crud<E> extends Component implements HasSize, HasTheme {
 
     /**
      * Gets visiblity state of toolbar
-     * 
+     *
      * @param
      * @return true if toolbar is visible false otherwise
      */
@@ -1001,7 +1002,7 @@ public class Crud<E> extends Component implements HasSize, HasTheme {
 
         /**
          * Gets new item being created
-         * 
+         *
          * @return a new instance of bean type
          */
         @Override
--- a/vaadin-crud-flow-parent/vaadin-crud-flow/src/main/java/com/vaadin/flow/component/crud/Crud.java
+++ b/vaadin-crud-flow-parent/vaadin-crud-flow/src/main/java/com/vaadin/flow/component/crud/Crud.java
@@ -16,6 +16,7 @@ package com.vaadin.flow.component.crud;
  * #L%
  */
 
+import com.vaadin.flow.component.AttachEvent;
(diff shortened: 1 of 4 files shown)


vaadin-crud-flow-parent/vaadin-crud-flow/src/main/java/com/vaadin/flow/component/crud/Crud.java
```diff
@@ -190,25 +190,20 @@ public Crud() {
         deleteButton.getElement().setAttribute("slot", "delete-button");
         deleteButton.addThemeNames("tertiary", "error");
         getElement().appendChild(deleteButton.getElement());
-
-        addAttachListener(e -> {
-            getElement()
-                    .executeJs("this.__validate = function () {return true;}");
```
DiegoCardoso, line 196: This call was here before the changes introduced by the feature branch. Should we keep it there?

vaadin-crud-flow-parent/vaadin-crud-flow/src/main/java/com/vaadin/flow/component/crud/Crud.java
```diff
@@ -190,25 +190,20 @@ public Crud() {
...
-            this.setDirty(true);
+    private class SaveButton extends Button {
+        @Override
+        public void onEnabledStateChanged(boolean enabled) {
+            super.onEnabledStateChanged(enabled);
+            overrideSaveDisabled(enabled);
         }
     }
 
+    private void overrideSaveDisabled(boolean enabled) {
```
sissbruecker, line 203: Consider moving this into the SaveButton class

vaadin-crud-flow-parent/vaadin-crud-flow/src/main/java/com/vaadin/flow/component/crud/Crud.java
```diff
@@ -190,25 +190,20 @@ public Crud() {
...
+        getElement().executeJs("this.__isSaveBtnDisabled = () => {return $0;}", !enabled);
```
sissbruecker, line 204: This change will be undone when detaching + reattaching the component. I think it would need to be triggered again in onAttach.

hdamr: Isn't the only case that this part gets executed is when `onEnabledStateChanged` is called? I think it is independent of attach/detaching crud element to the DOM.

sissbruecker: `executeJs` calls do not persist over detaching and reattaching, they have to be called again after reattaching the component.

So for the following case:
- set button to disabled
- detach Crud
- reattach Crud

the `__isSaveBtnDisabled` override will not be in place anymore.

vaadin-crud-flow-parent/vaadin-crud-flow/src/main/java/com/vaadin/flow/component/crud/Crud.java
yuriy-fix: True that. Restored.

vaadin-crud-flow-parent/vaadin-crud-flow/src/main/java/com/vaadin/flow/component/crud/Crud.java
yuriy-fix: I have considered it initially, but it's easier to override the `__isSaveBtnDisabled` from the Crud scope without introducing new fields/getters 🤔

vaadin-crud-flow-parent/vaadin-crud-flow/src/main/java/com/vaadin/flow/component/crud/Crud.java
yuriy-fix: Good point! The logic is now in place.

sissbruecker (approved):

PR #2098, opened 2021-09-07
fix: duplicate value changed event

### LINKED ISSUE #1172: Truncate string value based on "step" when sending value to client to avoid unnecessary value changes
I think it's this part of the code which handles converting the LocalTime to a string when the value is sent from server to client (when setting the value on server):
https://github.com/vaadin/vaadin-time-picker-flow/blob/master/vaadin-time-picker-flow/src/main/java/com/vaadin/flow/component/timepicker/TimePicker.java#L53-L55

This FORMATTER should probably be modified to include simple logic like here:
https://github.com/vaadin/vaadin-time-picker/blob/master/src/vaadin-time-picker.html#L578-L602

so that the milliseconds or seconds and milliseconds are trimmed out of the string value in case `step` is divisible by 1 or less than 1.

This should also avoid extra value-changed round trips when setting the value from the server (in case the value is set to read-only) since the client side will first adjust the value formatting automatically, but the second time won't (another issue).

### LINKED ISSUE #1986: DateTimePicker and DatePicker value change listener get fired twice when setting value with setValue() method.
### Description
When setting a value for DateTimePicker or DatePicker using setValue() method, the value change listener gets fire twice.
When loading  the view the first time (or refreshing the page) one event is reported to come from the server, and one from the client, when there has been no client interactions to the field, just a setValue() on the server. 
When calling setValue(), the value change listener fires two times, but at least both of them are reported as coming from the server.

### Expected outcome

Value change listener to fire once and with event from the right source.

### Actual outcome

Value change listener is being fired twice and events are not always from server.

```  
 public MainView() {
        DateTimePicker dateTimePicker = new DateTimePicker();
        dateTimePicker.addValueChangeListener(event -> {
          if (event.isFromClient()) {
              System.out.println("DateTimePicker event is from client!");
          } else {
              System.out.println("DateTimePicker event is from server!");
          }        
        });
        add(dateTimePicker);
        dateTimePicker.setValue(LocalDateTime.now());
        Button setDateTime = new Button("Set date time");
        setDateTime.addClickListener(event -> {
          System.out.println("Button clicked");
          dateTimePicker.setValue(LocalDateTime.now().minusDays(5));
        });        
        add(setDateTime);
    }
```

### Steps to reproduce

1. Add a DateTimePicker component to a page with no initial value.
2. Add a ValueChangeListener to the DateTimePicker component to check if fired event is from client or from server.
3. Add a button to the same page, to set a value to the DateTimePicker component on button click.
4. On page initializatin (or by refreshing) check console to see fired events and where the come from.
5. Click button and check console to see fired events.

### Environment

Vaadin platform 14.6.7.

### Browsers Affected
Tested on Chrome.

sissbruecker:
## Description

Currently the DateTimePicker can trigger an additional value changed event, when the value is updated on the server side. There seem to be two different issues:

1. Using `TimePicker.setValue` with a time value that has sub-millisecond precision (e.g. `LocalTime.now()`), the value is truncated to millisecond precision, which means the TimePicker will store a different time as field value than the DateTimePicker. After setting the value in the web component, that will trigger a value changed event from the client. The DateTimePicker now reads back the values from the DatePicker and TimePicker, and calls `setModelValue` with the new values and an indicator that the value is from the client-side. This triggers the duplicate value changed event from the client, as the time values differ.
  This also results in a different issue that `TimePicker.getValue` might return a different value than what was set by the user.
2. The TimePicker web component truncates time values based on the `steps` setting, resulting in a lower precision value on the client-side. By default, the web component uses a minute precision, by changing the step setting the web component can also support second and millisecond precision values. However neither the TimePicker, nor the DateTimePicker Flow components mirror that logic. There is an [existing workaround](https://github.com/vaadin/flow-components/blob/master/vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow/src/main/java/com/vaadin/flow/component/datetimepicker/DateTimePicker.java#L241-L250) for a separate issue in the DateTimePicker which seems to deal with this inconsistency. The problem with the workaround is that it can lead to a duplicate value changed event from the server-side, as it's effectively a second server-side setValue call.

This change addresses these issues by:
- changing the behaviour of `TimePicker.setValue` to not truncate the value, but always keep the value that was passed in. This is a small behaviour change, but seems like a general improvement, as the TimePicker now keeps the value that was passed in as is.
- adding custom presentation<->model converters to the time picker, to deal with the fact that the server-side value might have a higher precision than the client-side value:
  - when converting to the presentation value, the Flow component now respects the steps setting, and produces a time value string that is truncated to that precision. I'm not sure if that is necessary for this fix, but it also can't hurt.
  - when converting to the model value, the Flow component checks if the new value is equal to the current value, if the current value would be truncated to the currently configured precision. If the values are equal, then it returns the old value (with potentially higher precision).
    - while this is not the intended purpose of the converter, I could not find an alternative way to prevent the Flow `AbstractField` from changing the internal value. I checked `valueEquals`, however that is not applicable is it isn't used to check if the incoming client-side value is different from the new value that was just set.
- removing the workaround mentioned in (2). This doesn't seem necessary anymore now that the TimePicker always keeps the value that was set, disregarding any changes if the changed value matches the old value when respecting the configured precision.

Fixes #1986
Fixes #1172

## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-time-picker-flow-parent/vaadin-time-picker-flow-demo/src/main/java/com/vaadin/flow/component/timepicker/demo/TimePickerView.java
+++ b/vaadin-time-picker-flow-parent/vaadin-time-picker-flow-demo/src/main/java/com/vaadin/flow/component/timepicker/demo/TimePickerView.java
@@ -291,6 +291,7 @@ public class TimePickerView extends DemoView {
 
         public void setStep(Duration step) {
             this.step = step;
+            updateValue();
         }
 
         private void updateValue() {
--- a/vaadin-time-picker-flow-parent/vaadin-time-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/timepicker/tests/TimePickerLocalizationIT.java
+++ b/vaadin-time-picker-flow-parent/vaadin-time-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/timepicker/tests/TimePickerLocalizationIT.java
@@ -156,13 +156,13 @@ public class TimePickerLocalizationIT extends AbstractComponentIT {
 
         errors.add(verifyFormat());
         errors.add(verifyValueProperty("17:22:33"));
-        errors.add(verifyServerValue("17:22:33.0"));
+        errors.add(verifyServerValue("17:22:33.123"));
 
         selectStep("1m");
 
         errors.add(verifyFormat());
         errors.add(verifyValueProperty("17:22"));
-        errors.add(verifyServerValue("17:22:0.0"));
+        errors.add(verifyServerValue("17:22:33.123"));
 
         selectStep("1h");
 
@@ -170,7 +170,7 @@ public class TimePickerLocalizationIT extends AbstractComponentIT {
         // when changing granularity from 1min -> 60min
         errors.add(verifyFormat());
         errors.add(verifyValueProperty("17:22"));
-        errors.add(verifyServerValue("17:22:0.0"));
+        errors.add(verifyServerValue("17:22:33.123"));
 
         errors.removeIf(item -> item == null);
 
--- a/vaadin-time-picker-flow-parent/vaadin-time-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/timepicker/tests/TimePickerValueChangeIT.java
+++ b/vaadin-time-picker-flow-parent/vaadin-time-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/timepicker/tests/TimePickerValueChangeIT.java
@@ -39,39 +39,47 @@ public class TimePickerValueChangeIT extends AbstractComponentIT {
 
     @Test
     public void setServerSideValueShouldTriggerSingleServerSideChangeEvent() {
-        TestBenchElement setCurrentTimeButton = $("button").id("set-current-time");
+        TestBenchElement setCurrentTimeButton = $("button")
+                .id("set-current-time");
 
         setCurrentTimeButton.click();
 
         String[] entries = getChangeLogEntries();
         Assert.assertEquals(1, entries.length);
-        Assert.assertTrue("Event should originate from server-side", entries[0].startsWith("source: server"));
+        Assert.assertTrue("Event should originate from server-side",
+                entries[0].startsWith("source: server"));
     }
 
     @Test
     public void setServerSideValueWithSecondsPrecisionShouldTriggerSingleServerSideChangeEvent() {
-        TestBenchElement setSecondsPrecisionButton = $("button").id("set-seconds-precision");
-        TestBenchElement setCurrentTimeButton = $("button").id("set-current-time");
+        TestBenchElement setSecondsPrecisionButton = $("button")
+                .id("set-seconds-precision");
+        TestBenchElement setCurrentTimeButton = $("button")
+                .id("set-current-time");
 
         setSecondsPrecisionButton.click();
         setCurrentTimeButton.click();
 
         String[] entries = getChangeLogEntries();
         Assert.assertEquals(1, entries.length);
-        Assert.assertTrue("Event should originate from server-side", entries[0].startsWith("source: server"));
+        Assert.assertTrue("Event should originate from server-side",
+                entries[0].startsWith("source: server"));
     }
 
     @Test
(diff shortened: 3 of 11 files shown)


vaadin-time-picker-flow-parent/vaadin-time-picker-flow-demo/src/main/java/com/vaadin/flow/component/timepicker/demo/TimePickerView.java
```diff
@@ -291,6 +291,7 @@ public void setLocalTime(LocalTime localTime) {
 
         public void setStep(Duration step) {
             this.step = step;
+            updateValue();
```
sissbruecker, line 294: This is part of a component that is used in test setups to verify the time formats for different locales. The component was relying on the fact that changing the step property would trigger a value change in the TimePicker Flow component, which in turn lead to the `updateValue` method being called.

Now changing the step will keep the old value in the component, so it was necessary to trigger `updateValue` additionally here.

vaadin-time-picker-flow-parent/vaadin-time-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/timepicker/tests/TimePickerLocalizationIT.java
```diff
@@ -156,21 +156,21 @@ private void runReduceStepTest(Locale locale, String initialValue4PM,
 
         errors.add(verifyFormat());
         errors.add(verifyValueProperty("17:22:33"));
-        errors.add(verifyServerValue("17:22:33.0"));
+        errors.add(verifyServerValue("17:22:33.123"));
```
sissbruecker, line 159: Server side value should not change when just changing the steps / precision.

vaadin-time-picker-flow-parent/vaadin-time-picker-flow/src/main/java/com/vaadin/flow/component/timepicker/GeneratedVaadinTimePicker.java
```diff
@@ -937,10 +937,12 @@ public GeneratedVaadinTimePicker(T initialValue, T defaultValue,
...
             Class<P> elementPropertyType,
             SerializableBiFunction<R, P, T> presentationToModel,
-            SerializableBiFunction<R, T, P> modelToPresentation) {
+            SerializableBiFunction<R, T, P> modelToPresentation,
+            boolean isInitialValueOptional) {
         super("value", defaultValue, elementPropertyType, presentationToModel,
                 modelToPresentation);
-        if (initialValue != null) {
+        if ((getElement().getProperty("value") == null
+                || !isInitialValueOptional) && initialValue != null) {
```
sissbruecker, line 945: Needed to change this constructor to support the `isInitialValueOptional` option. The implementation is copied from the other constructor that was previously used by `TimePicker`.

vaadin-time-picker-flow-parent/vaadin-time-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/timepicker/tests/TimePickerValueChangeIT.java
```diff
@@ -0,0 +1,98 @@
...
+import com.vaadin.flow.component.timepicker.testbench.TimePickerElement;
+import com.vaadin.flow.testutil.TestPath;
+import com.vaadin.testbench.TestBenchElement;
+import com.vaadin.tests.AbstractComponentIT;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+@TestPath("vaadin-time-picker/time-picker-value-change")
+public class TimePickerValueChangeIT extends AbstractComponentIT {
```
sissbruecker, line 27: The behaviour tested in this IT was actually working correctly before the change. Since we are adding custom logic to the value handling, I thought it couldn't hurt to add those here as well.

sissbruecker:
Changed this back to a draft as I came up with an alternative solution that might be preferrable to this one: #2099

sissbruecker:
Closing this for now in favor of #2099

PR #2099, opened 2021-09-07
fix: prevent duplicate value changed events from DateTimePicker.setValue

### LINKED ISSUE #1986: DateTimePicker and DatePicker value change listener get fired twice when setting value with setValue() method.

### LINKED ISSUE #2101: Setting steps after initializing DateTimePicker truncates time value
### Description

Changing the steps setting some time after the DateTimePicker was initialized, leads to seconds and milliseconds being removed from the value, or just not being displayed correctly, when setting a value from the server-side.

### Expected outcome

When setting a value from the server-side, the DateTimePicker displays the correct values for seconds and the milliseconds.

### Actual outcome

The DateTimePicker displays `00` for seconds, or `000` for milliseconds.

There is a small ficker in the current version, where the correct values are displayed, before being changes to zeros.

### Live Demo (optional)

https://user-images.githubusercontent.com/357820/132477730-5162ed9f-324a-4456-92df-d48873806212.mp4

```java

    public HelloWorldView() {
        DateTimePicker dateTimePicker = new DateTimePicker();

        Button setStepsButton = new Button("Set second precision", e -> {
            dateTimePicker.setStep(Duration.of(30, ChronoUnit.SECONDS));
        });

        Button setValueButton = new Button("Set current date time", e -> {
            dateTimePicker.setValue(LocalDateTime.now());
        });

        add(dateTimePicker, setStepsButton, setValueButton);
    }
}
```

### Steps to reproduce

1. Add a DateTimePicker to a page.
2. Add button that changes the steps to 30 seconds
3. Add button that sets the current date and time
4. Run, open in browser
5. Click button to change steps
6. Click button to set current date and time

### Additional notes

The issue does not occur when setting the steps immediately after constructing the DateTimePicker:
```
DateTimePicker dateTimePicker = new DateTimePicker();
dateTimePicker.setStep(Duration.of(30, ChronoUnit.SECONDS));
```

The web component doesn't seem to be affected by this, and changing the value on the client side works as well.

### Environment

- Vaadin 14.6.8
- MacOS

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Safari

sissbruecker:
## Description

Alternative solution to #2098 

This approach refactors the DateTimePicker to remove the quirky logic that uses the individual date and time picker components for presentation values. Instead the component now extends `AbstractSinglePropertyField`, and manipulates the `value` property of the DateTimePicker web component directly, similar to other input components. Additionally it synchronizes the presentation value of the DatePicker and TimePicker components to match the field value of the parent DateTimePicker.

Fixes #1986
Fixes #2101  (fixing this as part of the refactoring was necessary to prevent additional client side value change events)

## Type of change

- [x] Bugfix


sissbruecker:
Feedback from testing by Guillermo:

`DateTimePickerValueChangeIT.setClientSideValueShouldTriggerSingleClientSideChangeEvent` seem to fail every couple of re-runs, I restarted the CI to see if it’s an issue locally.
Update: can’t reproduce locally

I’m getting this quite often in the demo/IT view (/vaadin-date-time-picker)
```
Uncaught RangeError: Maximum call stack size exceeded
```
seems related to this comment that was removed
```
        // Get the value from client after possible adjustment.
        // This is a workaround for infinite value change loop which can be
        // triggered when the field is set as read-only.
```

I get some strange behavior as well in the value change view (/vaadin-date-time-picker/date-time-picker-value-change)
If I set value and then set precision to millis then I can’t open the time picker overlay and after clicking the time picker twice I get a value change from client

sissbruecker:
Addressed review comments:
> DateTimePickerValueChangeIT.setClientSideValueShouldTriggerSingleClientSideChangeEvent seem to fail every couple of re-runs, I restarted the CI to see if it’s an issue locally.
Update: can’t reproduce locally

This was a bug in the date time picker test bench element setting invalid values into the web component, which resulted in the missing client-side change event

>I’m getting this quite often in the demo/IT view (/vaadin-date-time-picker)
> Uncaught RangeError: Maximum call stack size exceeded
> seems related to this comment that was removed
>        // Get the value from client after possible adjustment.
>        // This is a workaround for infinite value change loop which can be
>        // triggered when the field is set as read-only.

Fixed by truncating the LocalDateTime instance in the DateTimePicker to milliseconds, same as the TimePicker does. Otherwise it was possible to set an invalid value into the web component, as that only supports a max precision of milliseconds.

>I get some strange behavior as well in the value change view (/vaadin-date-time-picker/date-time-picker-value-change)
If I set value and then set precision to millis then I can’t open the time picker overlay and after clicking the time picker twice I get a value change from client

The overlay not opening in this case is expected, the overlay stays closed if the steps interval becomes too small, as there would be too many values in the drop down to be practical.

The client change event when clicking on the overlay toggle button is due to the blur event causing a value change. It seems the DateTimePicker has a truncated time value when compared to the internal time picker. Still looking into that.

vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow/src/main/java/com/vaadin/flow/component/datetimepicker/DateTimePicker.java
```diff
@@ -374,7 +385,10 @@ public String getTimePlaceholder() {
      *            an hour evenly
      */
     public void setStep(Duration step) {
-        timePicker.setStep(step);
+        Objects.requireNonNull(step, "Step cannot be null");
+
+        getElement().setProperty("step",
+                StepsUtil.convertDurationToStepsValue(step));
```
sissbruecker, line 391: Adopted the same logic for setting the step from the TimePicker Flow component. The previous solution of setting the step on the time picker itself was flawed, as the date time picker web component has its own step property and related logic. This lead to a different bug: https://github.com/vaadin/flow-components/issues/2101

Setting the step on the date time picker component itself fixes the issue.

vaadin-time-picker-flow-parent/vaadin-time-picker-flow/src/main/java/com/vaadin/flow/component/timepicker/StepsUtil.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.time.Duration;
+
+/**
+ * Utility class around the time picker steps functionality. The logic in here is extracted for reuse by the DateTimePicker component.
+ *
+ * <p>
+ *     NOTE: This class is not part of the public API surface and for internal use only
+ */
+public class StepsUtil {
```
sissbruecker, line 28: Extracted this logic, so that it can be reused by the DateTimePicker

hdamr (approved):
LGTM

PR #2140, opened 2021-09-20
chore: improve flaky ServerSideEventsIT

sissbruecker:
## Description

Improve the `ServerSideEventsIT` to remove some of the flaky behavior related to event order and unexpected axis extremes changed events.

For the tests where I removed assertions related to the event order, I could not see why we would want to test the order at all. My assumption is that the previous setup was not intentional and was just based on the observed event order at the time the tests were written.


hdamr (approved):

vaadin-charts-flow-parent/vaadin-charts-flow-demo/src/main/java/com/vaadin/flow/component/charts/examples/dynamic/ServerSideEvents.java
```diff
@@ -29,7 +29,7 @@
 import com.vaadin.flow.component.charts.model.XAxis;
 import com.vaadin.flow.component.charts.model.YAxis;
 import com.vaadin.flow.component.checkbox.Checkbox;
-import com.vaadin.flow.component.html.Label;
+import com.vaadin.flow.component.html.*;
```
ZheSun88, line 32: by the java code convention in Vaadin, we use direct import only.

sissbruecker: Done

PR #2156, opened 2021-09-27
feat: Add support for theme variants for DatePicker

TatuLund:
fixes: #1604

### DIFF
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePickerVariant.java
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePickerVariant.java
@@ -15,9 +15,9 @@
  */
 package com.vaadin.flow.component.datepicker;
 
- /**
-  * Set of theme variants applicable for {@code vaadin-date-picker} component.
-  */
+/**
+ * Set of theme variants applicable for {@code vaadin-date-picker} component.
+ */
 public enum DatePickerVariant {
     LUMO_SMALL("small"), LUMO_ALIGN_CENTER("align-center"), LUMO_ALIGN_RIGHT(
             "align-right"), LUMO_HELPER_ABOVE_FIELD("helper-above-field");
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePickerVariant.java
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePickerVariant.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2000-2018 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component.datepicker;
+
+ /**
+  * Set of theme variants applicable for {@code vaadin-date-picker} component.
+  */
+public enum DatePickerVariant {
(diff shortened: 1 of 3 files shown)


vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePickerVariant.java
```diff
@@ -0,0 +1,39 @@
...
+ * the License.
+ */
+package com.vaadin.flow.component.datepicker;
+
+/**
+ * Set of theme variants applicable for {@code vaadin-date-picker} component.
+ */
+public enum DatePickerVariant {
+    LUMO_SMALL("small"), LUMO_ALIGN_CENTER("align-center"), LUMO_ALIGN_RIGHT(
+            "align-right"), LUMO_HELPER_ABOVE_FIELD("helper-above-field");
```
sissbruecker, line 23: Is there any reference for a comprehensive list of applicable themes? To check this I cross-referenced several style modules in the web-components repo, but I'm not sure if I found all of them.

One theme that is missing is `align-left`. I assume that it was left out intentionally since it seems to be the default. We should still add it to have the complete set. It might also be required when using RTL, and you want to align he text left for some reason.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/test/java/com/vaadin/flow/component/datepicker/DatePickerVariantTest.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+package com.vaadin.flow.component.datepicker;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class DatePickerVariantTest {
+
+    private final DatePicker datePicker = new DatePicker();
```
sissbruecker, line 24: Please initialize the instance in a setup method, reusing the same instance between tests is not a good practice and can lead to issues that are hard to debug.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/test/java/com/vaadin/flow/component/datepicker/DatePickerVariantTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Test
+    public void addAndRemoveLumoAlignCenterVariant_themeAttributeUpdated() {
+        assertThemeAttribute(null);
+        datePicker.addThemeVariants(DatePickerVariant.LUMO_ALIGN_CENTER);
+        assertThemeAttribute("align-center");
+        datePicker.removeThemeVariants(DatePickerVariant.LUMO_ALIGN_CENTER);
```
sissbruecker, line 31: Similarly removing the theme variants in every test is not necessary then.

We should rather have these additional tests:
- verify that all theme variants can be removed
- add multiple variants, and verify that a single theme variant can be removed, while the others are kept in place

I would also suggest to add a test to verify that calling `addThemeVariants` multiple times does indeed concatenate them, instead of overwriting the results of previous calls.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePickerVariant.java
TatuLund: > One theme that is missing is align-left. I assume that it was left out intentionally since it seems to be the default. 

Yes, I just followed how it was done in TextField.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/test/java/com/vaadin/flow/component/datepicker/DatePickerVariantTest.java
TatuLund: It is cargo cult from https://github.com/vaadin/flow-components/blob/master/vaadin-dialog-flow-parent/vaadin-dialog-flow/src/test/java/com/vaadin/flow/component/dialog/DialogThemeVariantTest.java

sissbruecker: Similarly, existing tests not using best-practices is not a good reason to not apply best practices to new tests. Let's fix it.

TatuLund: As contributor I can rely only on what is already there ... but fixed as requested now.

sissbruecker (approved):

PR #2165, opened 2021-09-29
fix: make DateTimePicker apply initial steps

### LINKED ISSUE #2164: Setting steps in DateTimePicker before attaching it does not work
### Description

Setting the steps property of the DateTimePicker before it has been added to the component hierarchy does not work currently. The issue seems to be with [this piece of code](https://github.com/vaadin/web-components/blob/master/packages/date-time-picker/src/vaadin-date-time-picker.js#L547-L558) in the web component, which initializes the DateTimePicker's step from the TimePicker, if the TimePicker has been slotted (which the Flow DateTimePicker does). Which means the step value set on the DateTimePicker will be overwritten by the undefined value of the TimePicker.

### Expected outcome

Setting the steps property should apply it to the web component.

### Actual outcome

Setting the steps property does not apply it to the web component.

### Live Demo (optional)

n/a

```java
DateTimePicker picker = new DateTimePicker();
picker.setSteps(Duration.ofMillis(500));

add(picker);
```

### Steps to reproduce

Clear steps describing how to reproduce the issue.

1. Instantiate a DateTimePicker
2. Set the steps to some value that should force millisecond precision
3. Add the picker to the page
3. -> picker uses defautl steps of half an hour

### Environment

- Tested with Vaadin 22 and 14.7.1

### Browsers Affected

n/a

sissbruecker:
## Description

Fixes the DateTimePicker to apply steps if they were set before attaching the component. The issue is that the web component synchronizes the steps property from the TimePicker to the DateTimePicker, when using a slotted TimePicker. To solve this, I changed `setSteps` to additionally set the value on the TimePicker. I could not observe any issues with the solutions, it works when setting the steps before and after attaching the component.

Fixes #2164

## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow-integration-tests/src/main/java/com/vaadin/flow/component/datetimepicker/DateTimePickerStepPage.java
+++ b/vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow-integration-tests/src/main/java/com/vaadin/flow/component/datetimepicker/DateTimePickerStepPage.java
@@ -16,7 +16,7 @@
 package com.vaadin.flow.component.datetimepicker;
 
 import com.vaadin.flow.component.html.Div;
-import com.vaadin.flow.component.html.Hr;
+import com.vaadin.flow.component.html.H1;
 import com.vaadin.flow.component.html.NativeButton;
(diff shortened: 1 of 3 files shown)


alvarezguille (approved):
LGTM

PR #2242, opened 2021-10-18
feat: Make CellEditStartedEvent#getPath public

TatuLund:
Based on JavaDoc it looks like it was the original intention and it has been private just by accident

### DIFF
--- a/vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/java/com/vaadin/flow/component/gridpro/GridPro.java
+++ b/vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/java/com/vaadin/flow/component/gridpro/GridPro.java
@@ -514,7 +514,7 @@ public class GridPro<E> extends Grid<E> {
          *
          * @return the key of the column
          */
-        private String getPath() {
+        public String getPath() {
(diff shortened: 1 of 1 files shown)


sissbruecker (approved):

tomivirkki:
> Based on JavaDoc it looks like it was the original intention and it has been private just by accident

Doesn't seem accidental. The related [discussion can be found in the original PR](https://github.com/vaadin/vaadin-grid-pro-flow/pull/38/files#diff-691110fe53a87b2de3d95c03246989d3c568955a26aa490ca253ca0f049237e6R452). The change was made in [this commit](https://github.com/vaadin/vaadin-grid-pro-flow/pull/38/commits/403a429aaf7e484cf1d09560804dbba55a33fc80#diff-691110fe53a87b2de3d95c03246989d3c568955a26aa490ca253ca0f049237e6L522).

sissbruecker:
As Tomi pointed out, the property was made private intentionally, so we do not just want to expose it and introduce a new API in the process.

Please file an enhancement ticket instead describing the use case. We might be able to come up with an existing solution, or we can consider introducing a more appropriate API. For example by exposing the column instance itself rather than the path, which is how similar APIs on Grid events work.

PR #2298, opened 2021-11-01
fix: isMultiSort will now return true when appropriate

### LINKED ISSUE #2227: Grid.isMultiSort() always returns false
### Description

After enabling multisort on a grid using `grid.setMultiSort(true)`, `grid.isMultiSort()` still returns `false`.

### Expected outcome

After enabling multisort on a grid using `grid.setMultiSort(true)`, `grid.isMultiSort()` should return `true`.

### Actual outcome

... it returns `false`.

```
grid.setItems(list);

final Button btnRunTest = new Button("runTest", e -> {
    var protocol = "";
    protocol += "calling grid.setMultiSort(true)\n";
    grid.setMultiSort(true);
    protocol += "now grid.isMultiSort() returns " + grid.isMultiSort()+ "\n\n";

    protocol += "calling grid.setMultiSort(false)\n";
    grid.setMultiSort(false);
    protocol += "now grid.isMultiSort() returns " + grid.isMultiSort() + "\n";

     System.out.println(protocol);
});
```
Prints: 
```
calling grid.setMultiSort(true)
now grid.isMultiSort() returns false                           <<<<< which is not expected

calling grid.setMultiSort(false)
now grid.isMultiSort() returns false
```

### Steps to reproduce

Add the code above to a view and click on the button. Check the protocol.

### Environment

- Vaadin v20.0.3
- OSX 11.6

### Browsers Affected
- [x] Chrome
- [?] Firefox
- [x] Safari
- [?] Edge
- [?] IE 11
- [?] iOS Safari
- [?] Android Chrome

### Analysis

`Grid.java.setMultiSort` and `isMultiSort` are not compatible.

```
    public void setMultiSort(boolean multiSort) {
        this.getElement().setAttribute("multi-sort", multiSort);
    }
```
When called with `multiSort = true`, the value `""` (empty String) is set on the element (see `Element.setAttribute(String, boolean)`). 

But when `isMultiSort` is queried, it maps `null` to `false`, `"true"` (case insensitive) to `true` and everything else to `false`. It does not deal with `""` (empty String).
```
    public boolean isMultiSort() {
        String multiSort = this.getElement().getAttribute("multi-sort");
        return multiSort == null ? false : Boolean.valueOf(multiSort);
    }
```

### Primitive Fix
```
    public boolean isMultiSort() {
        return this.getElement().getAttribute("multi-sort") != null;
    }
```

### Better fix
provide `Element.getBooleanAttribute` and use this in grid (and on other occasions where booleans are set/read).

sosa-vaadin:
## Description

The `Grid.isMultiSort()` method checks for the value of the `multi-sort` attribute. The attribute is just added with an empty value, and since the value is empty, this method returns `false`. Adding `multi-sort` attribute to a grid is considered as setting it to `true`.

Fixes #2227 

## Type of change

- [x] Bugfix


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -3092,8 +3092,9 @@ public void setMultiSort(boolean multiSort) {
      *         {@code false} otherwise
      */
     public boolean isMultiSort() {
-        String multiSort = getElement().getAttribute("multi-sort");
-        return multiSort == null ? false : Boolean.valueOf(multiSort);
+        return Optional.ofNullable(getElement().getAttribute("multi-sort"))
```
hdamr, line 3095: I think usage of stream API and Optional is excess work here, because we are not dealing with any stream and the only values that we deal with are `null` and `""`, so I think it is better to change it with basic `if ... else` statements.

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridSortingTest.java
```diff
@@ -263,6 +263,14 @@ public void template_renderer_non_comparable_property() {
                 grid.getDataCommunicator().getBackEndSorting());
     }
 
+    @Test
+    public void checkMultiSort() {
```
hdamr, line 267: I suggest to separate two cases, one correctly reading the initial value, the other reading the correct value after we change it via code.

hdamr (changes requested):

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
sosa-vaadin: I'm not using the Stream API here. I'm using `Optional` to make the code `null` safe. Also to make an empty value of the `multi-sort` attribute the same as `true`.

The current change is equivalent to:
```Java
String multiSort = getElement().getAttribute("multi-sort");
if (multiSort != null && multiSort.length() == 0) {
    multiSort = "true";
}
return multiSort == null ? false : Boolean.valueOf(multiSort);
```

hdamr: IMO if we reduce it only to this code it works, unless I am missing something, no need to use `valueOf`:

String multiSort = getElement().getAttribute("multi-sort");
 return multiSort == null ? false : true;

sosa-vaadin: It does given the current implementation. However, it is assuming we will never change it to `multi-sort="false"` one day. I was just being extra cautious.

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridSortingTest.java
sosa-vaadin: Done.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
sosa-vaadin: I changed the implementation to not use `Optional` so it is a bit easier to read.

hdamr (approved):

PR #2340, opened 2021-11-15
fix: clear input element value when clearing custom value in ComboBox

### LINKED ISSUE #1090: ComboBox.clear() has no effect when custom values are enabled and a custom value is entered
@epricer commented on [Fri Oct 09 2020](https://github.com/vaadin/vaadin-combo-box-flow/issues/401)

When

1. enabling custom values for a ComboBox 
2. typing a custom value
3. invoking the clear() method

The value displayed in the combobox is not cleared.

A sample test harness - reproducible in 14.4 and 14.3.7

```java
package tedpricer.vaadincomboboxbug;

import com.vaadin.flow.component.combobox.ComboBox;
import com.vaadin.flow.component.html.Div;

import java.util.List;

@Route("")
public class TheView extends Div {

    public TheView() {
        comboBox.setItems(List.of("One", "Two"));
        comboBox.addCustomValueSetListener(event -> {
            System.out.println("Custom Value Entered: " + event.getDetail());
        });
        Button clearButton = new Button("Click me to clear", buttonClickEvent -> {
            comboBox.clear();
        });
        add(comboBox, clearButton);
    }
}
```

a very ugly workaround:
```java
comboBox.getElement().executeJs("this.shadowRoot.querySelector('vaadin-text-field').value = null;");
```

sissbruecker:
## Description

Currently clearing the ComboBox using `ComboBox.clear()`, or using `ComboBox.setValue(null)`, does not clear the internal input element's value after entering a custom value. One cause here is that the [connector](https://github.com/vaadin/flow-components/blob/master/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js#L337) prevents propagating the value change event, probably because Flow uses item keys as values, and setting a custom value could interfere with that. That also means that the state of the value property is never updated on the server-side, which means setting the value to null/empty does not trigger a client-side update, since the values are the same.

To fix this I used a different property `_inputElementValue`, which is already being used by the [`refreshValue`](https://github.com/vaadin/flow-components/blob/master/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L398) method to clear the value of the internal input element. There is a similar issue here, in that the property is not observable, so Flow never gets notified about changes to it from the client-side. To fix this, I added an event listener for the `custom-value-set` event, and synchronize the `_inputElementValue` property there manually. That means that when clearing the value / setting it to null, the code linked above will trigger a client update, because now there is an actual value change.

Fixes #1090 

## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/ClearValueIT.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/ClearValueIT.java
@@ -24,6 +24,7 @@ import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.openqa.selenium.By;
+import org.openqa.selenium.Keys;
 
 @TestPath("vaadin-combo-box/clear-value")
 public class ClearValueIT extends AbstractComponentIT {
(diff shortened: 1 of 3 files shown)


hdamr (commented):
Couldn't we override `clear` method inside combo-box and call `refreshValue` there and have the same effect without an extra event listener?

sissbruecker:
> Couldn't we override clear method inside combo-box and call refreshValue there and have the same effect without an extra event listener?

@hdamr `clear` already triggers `refreshValue`, through `setValue`. The problem is that [this line](https://github.com/vaadin/flow-components/blob/master/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L398) in `refreshValue` currently has no effect, because the property is never updated from the client-side. So regardless what custom value the user has entered on the client, Flow still thinks that the property contains an empty string. Setting the property to an empty string again, Flow compares the new value with the previous value and determines that they are the same, and that no client update is needed. Thus the value is never cleared on the client. This fix introduces the event listener for custom values, so that the property is changed to something else than an empty string, which now causes the line above to actually trigger a client update, because old and new value are not the same.

hdamr (approved):

simasch:
@sissbruecker This is not fixed

PR #2357, opened 2021-11-17
fix: fix selecting an item in a single select grid

### LINKED ISSUE #2354: Grid with CallbackDataProvider selection didn't work on Dialog open
Vaadin Flow 14. 
I have a Button, on click it will:
- open Dialog (if not opened) with Grid inside
- in new Thread with UI.GetCurrent().access():
     - scroll to needed item in the Grid
     - select same needed item in the Grid

By button click:
- if Dialog closed, then it opens, showing Grid, scrolling, but no select;
- if Dialog already opened, then scrolling to item and select this item;

I tryed to use addAttachListener() and addOpenedChangeListener(), and make operations on Grid on that events, but no luck.

Where am i wrong?

pleku:
Fixes selecting an item when an identifier provider is used through the
data provider or the data view API.

Fixes #2354

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/AbstractGridSingleSelectionModelTest.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/AbstractGridSingleSelectionModelTest.java
@@ -1,0 +1,77 @@
+package com.vaadin.flow.component.grid;
+
+import java.util.stream.Stream;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import com.vaadin.flow.data.provider.CallbackDataProvider;
+
+public class AbstractGridSingleSelectionModelTest {
+
+    private static final Person PERSON_C = new Person("c", 3);
+    private static final Person PERSON_B = new Person("b", 2);
+    private static final Person PERSON_A = new Person("a", 1);
+
+    private Grid<Person> grid;
+    private AbstractGridSingleSelectionModel<Person> selectionModel;
+    private CallbackDataProvider<Person, Void> spy;
+
+    @Before
+    public void init() {
+        grid = new Grid<>();
+        selectionModel = (AbstractGridSingleSelectionModel<Person>) grid
+                .getSelectionModel();
+        final CallbackDataProvider<Person, Void> dataProvider = new CallbackDataProvider<>(
+                query -> Stream.of(PERSON_A, PERSON_B, PERSON_C), query -> 3,
+                Person::getName);
+        spy = Mockito.spy(dataProvider);
+        grid.setItems(spy);
+    }
+
+    @Test // #2354
+    public void selectItem_dataProviderWithIdentifierProvider_identityUsedForEqualsComparison() {
+        grid.select(PERSON_A);
+
+        // called 3 times - once by selection model and twice by KeyMapper
+        Mockito.verify(spy, Mockito.times(3)).getId(PERSON_A);
+        Mockito.verify(spy, Mockito.never()).getId(PERSON_B);
+
+        Mockito.reset(spy);
+        grid.select(PERSON_B);
+
+        // called 3 times - once by selection model and twice by KeyMapper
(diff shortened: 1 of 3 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/AbstractGridSingleSelectionModelTest.java
```diff
@@ -0,0 +1,77 @@
...
+                .getSelectionModel();
+        final CallbackDataProvider<Person, Void> dataProvider = new CallbackDataProvider<>(
+                query -> Stream.of(PERSON_A, PERSON_B, PERSON_C), query -> 3,
+                Person::getName);
+        spy = Mockito.spy(dataProvider);
+        grid.setItems(spy);
+    }
+
+    @Test // #2354
+    public void selectItem_dataProviderWithIdentifierProvider_identityUsedForEqualsComparison() {
```
sissbruecker, line 34: It's very hard to figure out how these tests are related to the fix:
- you fixed the early return in `select`, and `selectFromClient`, which purpose seems to be to prevent execution selection logic, if the item is already selected
- so it seems that our expectations are:
  - that selection logic is executed when a different item is selected
  - that selection logic is not executed, when the same item is selected
- yet all the tests do is select a few different items, and then make some assertions about how often the data provider was called

I would suggest to restructure the tests to:
- add test cases for these four scenarios:
  - select should select item if it is not already selected
  - select should not select item if it is already selected
  - selectFromClient should select item if it is not already selected
  - selectFromClient should not select item if it is already selected
- the test names shouls accurately reflect these scenarios, rather than the internal details that are used to accomplish this (`identityUsedForEqualsComparison`)
- the tests should not rely on internal details of the class, such as how often the data provider is called when selecting an item. This is unreliable, and an unrelated change can quickly break this test, while a developer now has to figure out why they have to replace a `Mockito.times(3)` with a `Mockito.times(4)`
- instead the tests should rely on something that is observable from outside. For example, the selection model [fires an event](https://github.com/vaadin/flow-components/blob/master/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridSingleSelectionModel.java#L194) that can be observed to check if selection has changed

sissbruecker (changes requested):

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/AbstractGridSingleSelectionModelTest.java
pleku: Thanks for the prompt review!

I agree mostly with your comments, but I'd want to note that the existing code for `AbstractSingleSelectionModel` is not unit tested at least directly (nor probably indirectly other than in ITs). The test cases you mention _should already be there_ but are not.

For tests in this PR, my interest is only in making sure that **the data provider identity provider will be used**. The four test cases might each pass depending on the used bean type and its `equals` and `hashCode` implementation even if the code is broken and doesn't use the identifier provider (`dataProvider.getId(item)` ). It needs to be validated that the method is called from the data provider, and thus I will need to validate that.

I agree that checking how many times it has been invoked is fragile. Instead of making the tests more complex by mocking away some internal parts, I instead opted for commenting on the test _why it is called so many times_ so that if it changes, it will be easier to see it. 

> instead the tests should rely on something that is observable from outside. For example, the selection model fires an event that can be observed to check if selection has changed

I could try to use this but then I'd have to add a lot of more code because it requires a bean that has invalid equals&hashcode to make sure it won't pass when it should not. You can call me lazy, but I should be working on other things instead of this PR. I cannot promise that I'll have time to make these changes

sissbruecker: Updated the tests to use selection change listener.

sissbruecker (approved):

Issue #2379, opened 2021-11-25
getLocale on DatePicker overwritten


mheicke:
### Description

The default implementation of `getLocale` in `com.vaadin.flow.component.Component` does return the locale of the current UI so developers are used to be able of just using component.getLocale() to get the locale from a component. It's also used for other internal elements like translation etc.
`com.vaadin.flow.component.datepicker.DatePicker` overwrites `getLocale` with `return locale;` which initially is null and gets initiated with the locale of the UI on the onAttach event. This will mean, that getLocale() of a datePicker returns null for a while getLocale of every other component returns the currentUI locale. I would consider this as a bug in the logic and even if it is not, it might lead to potential NPE as it already did in our application. 
I recommand to overwrite getLocale in DatePicker in a way which returns the locale variable if it is not null and returns super.getLocale otherwise.
`DateTimePicker` does it a little bit differently here and fills `locale` with the currentUIs locale in the constructor which is different again to the DatePicker implementation. I would here - as well - remove the setting of locale in the constructor and overwrite the getLocale with the use of if (locale == null) return super.getLocale()

- Seems like the new getLocale implementation happened here: https://github.com/vaadin/vaadin-date-picker-flow/pull/54
- And this is the commit where the setLocale from the constructor was removed, leading to NPEs in our code: https://github.com/vaadin/flow-components/commit/863e2b0a1fa376f4d309468096403f279e504ae3 

### Expected outcome

getLocale() of DatePicker should return the currentUI locale as every other component if locale is not set

### Actual outcome

getLocale() of DatePicker returns null

### Minimal reproducible example

```
class PublicLocaleTextField extends TextField {
	@Override
	public Locale getLocale() {
		return super.getLocale();
	}
}

PublicLocaleTextField textField = new PublicLocaleTextField();
Objects.requireNonNull(textField.getLocale()); // no exception

DatePicker datePicker = new DatePicker();
Objects.requireNonNull(datePicker.getLocale());  // throws NullPointerException
```

### Steps to reproduce

1. Create a DatePicker
2. Call getLocale() before onAttach function is called
3. Create your own component which makes getLocale public or just uses it inside of the component
4. Call getLocale() on or in this component before onAttach function is called

### Environment

- Vaadin 14.7.5

### Browsers Affected
- [x ] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge
- [x] IE 11
- [x] iOS Safari
- [x] Android Chrome

dskbrd:
[...]

PR #2380, opened 2021-11-25
fix: Fallback to super.getLocale() if locale not yet set

TatuLund:
fixes: #2379


sissbruecker:
Closing in favor of #2416

PR #2390, opened 2021-11-30
fix: prevent HasItem.getItem throwing ClassCastException

### LINKED ISSUE #1040: DataSeries ClassCastException
PointClickEvent implements HasItem. HasItem casts to DataSeries. 

So if your series is anything else than DataSeries (or derived from) this will result in ClassCastException. 
Like when using Organization Chart and NodeSeries (derived from AbstractDataSeries, not DataSeries) resulting in exception.

```
chart.addPointClickListener(e -> {
  DataSeriesItem item = e.getItem(); // exception here when using NodeSeries
});
```

HasItem.java
```
default DataSeriesItem getItem() {
  return ((DataSeries) getSeries()).get(getItemIndex());
}
```

sissbruecker:
## Description

Context: Charts

Currently using `HasItem.getItem` (for example through `PointClickEvent.getItem`) with any series other than `DataSeries` results in a `ClassCastException`. The issue is that the method assumes that all series will be a `DataSeries`, and it can only return a `DataSeriesItem` as well. Additionally there are no obvious solutions for how to get the item from an event for other series, like `NodeSeries` or `ListSeries`.

As this is a common API we can not modify its signature or name to reflect that it only works with DataSeries. We also can not introduce major new APIs to handle other series types. Instead this change contains:
- throw a more appropriate `UnsupportedOperationException` with a relevant message that indicates that this method only works for `DataSeries`
- update the `HasItem.getItem` JavaDoc to note about the fact
- in order to support retrieving items from series other than `DataSeries`:
  - add a `HasItem.getItemId` method that returns the ID of the item, and implement it in all event classes implementing `HasItem`. The item ID be used to manually identify the item in the series data, given that the series items support setting an ID.
  - add JavaDoc to the method to note about its usage as an alternative to `HasItem.getItem` and add example code for `NodeSeries`
  - add JavaDoc to `HasItem.getItemIndex` to note about its usage as an alternative to `HasItem.getItem` and add example code for `ListSeries`

I did manual testing to verify that the item / point ID is provided by all events that implement `HasItem`.

Fixes #1040 

## Type of change

- [X] Bugfix

### DIFF
--- a/vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/events/HasItem.java
+++ b/vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/events/HasItem.java
@@ -14,8 +14,12 @@ package com.vaadin.flow.component.charts.events;
  */
 
 import com.vaadin.flow.component.charts.Chart;
+import com.vaadin.flow.component.charts.model.AbstractSeriesItem;
 import com.vaadin.flow.component.charts.model.DataSeries;
 import com.vaadin.flow.component.charts.model.DataSeriesItem;
+import com.vaadin.flow.component.charts.model.ListSeries;
+import com.vaadin.flow.component.charts.model.NodeSeries;
+import com.vaadin.flow.component.charts.model.NodeSeriesItem;
 import com.vaadin.flow.component.charts.model.Series;
 
 /**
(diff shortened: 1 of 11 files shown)


sissbruecker:
@hdamr Updated the JavaDocs for `HasItem.getItemId`

vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/events/HasItem.java
```diff
@@ -27,18 +32,75 @@
...
-     * Returns the item that was clicked
+     * The ID of the series item that is associated with the event. Can be used
+     * to identify the item within the series.
+     * <p>
+     * Example for {@link NodeSeries}:
+     *
+     * <pre>
+     * String id = this.getItemId();
+     * NodeSeries series = (NodeSeries) this.getSeries();
+     * Optional&lt;Node&gt; nodeForId = series.getNodes().stream()
```
hdamr, line 62: This line shows correctly in the IDE, but it is shown different than other places.

hdamr (changes requested):
Just a minor change required IMO

vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/events/HasItem.java
sissbruecker: Can you be more specific what you mean by different, and other places? JavaDoc is basically HTML, so if we want the `<Node>` char sequence to show up correctly, then the characters have to be escaped, otherwise they will be interpreted as HTML tag.

hdamr: Like the line above where we have `<pre>` tag.

hdamr (approved):
It looks good IMO

Issue #2389, opened 2021-11-29
Grid: rows are empty if setItems is called after scrolling to bottom of grid


drewharvey:
Vaadin 22 rc1

### Description

If you scroll to the bottom of the current grid rows, then call `grid.setItems` with a new data set, the new rows are displayed as empty.

### Use Case

Our project uses filters in header rows to fetch new data.  So if the user scrolls to the very bottom of a grid, then enters text into the filter header, the new data that is returns will show as empty rows.

### Live Demo (optional)

![grid-empty-rows](https://user-images.githubusercontent.com/2496955/143950478-8ab9b8a5-cde2-4819-847c-7f0b2a54676a.gif)

### Minimal reproducible example

```
private TextField filter;

public AAAGridProTestView() {

    // setup grid (or GridPro)
    Grid<String> grid = new Grid<>();
    grid.setWidth("500px");
    grid.setHeight("200px");

    grid.addColumn(item -> item)
            .setHeader("A");

    Grid.Column<String> colB =
            grid.addColumn(item -> item)
            .setHeader("B")
            .setKey("B");

    grid.addColumn(item -> item)
            .setHeader("C");

    // add text field filter to column B
    filter = new TextField();
    filter.setValueChangeMode(ValueChangeMode.EAGER);
    grid.appendHeaderRow().getCell(colB).setComponent(filter);

    // fetch different data set depending on if filter has value
    filter.addValueChangeListener(event -> {
        if (event.getValue() != null && !event.getValue().isEmpty()) {
            grid.setItems(generateFilteredData());
        } else {
            grid.setItems(generateFullData());
        }
    });

    // set default data
    grid.setItems(generateFullData());

    add(grid);

    // add buttons to show the bug can trigger simply by calling grid.setItems from anywhere
    add(new Button("Show filtered data", e -> grid.setItems(generateFilteredData())));
    add(new Button("Show unfiltered data", e -> grid.setItems(generateFullData())));
}

private List<String> generateFullData() {
    List<String> items = new ArrayList<>();
    for (int i = 0; i < 100; i++) {
        items.add("Unfiltered " + i);
    }
    return items;
}

private List<String> generateFilteredData() {
    return new ArrayList<String>() {{
        add("Filtered 1");
        add("Filtered 2");
    }};
}

```

PR #2392, opened 2021-12-01
fix: make grid connector request content update on missing data

tomivirkki:
Fixes https://github.com/vaadin/flow-components/issues/2389

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridViewIT.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridViewIT.java
@@ -694,7 +694,7 @@ public class GridViewIT extends GridViewBase {
     }
 
     @Test
-    public void gridWithFiltering_scrolledToEnd() {
+    public void scrollToEnd_filter_rowsUpdated() {
         // Open /vaadin-grid-it-demo/filtering
         openTabAndCheckForErrors("filtering");
         GridElement grid = $(GridElement.class).id("grid-with-filters");
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridViewIT.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridViewIT.java
@@ -705,7 +705,8 @@ public class GridViewIT extends GridViewBase {
         grid.findElement(By.tagName("vaadin-text-field")).sendKeys("100");
         waitUntil(driver -> grid.getRowCount() == 1);
 
-        // Expect the one remaining row's first cell to contain text "Person 100"
+        // Expect the one remaining row's first cell to contain
+        // text "Person 100"
         Assert.assertEquals("Person 100", grid.getCell(0, 0).getText());
     }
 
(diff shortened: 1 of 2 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridViewIT.java
```diff
@@ -693,6 +693,23 @@ public void gridWithFiltering() {
         waitUntil(driver -> grid.getCell(0, 0).getText().contains("Person 6"));
     }
 
+    @Test
+    public void gridWithFiltering_scrolledToEnd() {
```
sissbruecker, line 697: The test name could use an expectation, and filtering and scrolling run in different order. Maybe `scrollToEnd_filter_rowsUpdated`?

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridViewIT.java
```diff
@@ -693,6 +693,23 @@ public void gridWithFiltering() {
...
+
+        // Scroll to the end of the grid
+        grid.scrollToRow(grid.getRowCount() - 1);
+        // Filter "Name" column with "100"
+        grid.findElement(By.tagName("vaadin-text-field")).sendKeys("100");
+        waitUntil(driver -> grid.getRowCount() == 1);
+
+        // Expect the one remaining row's first cell to contain
+        // text "Person 100"
+        Assert.assertEquals("Person 100", grid.getCell(0, 0).getText());
```
sissbruecker, line 710: When removing the fix, I'm getting a bit of a different actual value here compared to what is described in the issue. In the issue, and when testing manually, the cell seems to be empty. However when running the IT, it contains a value:
```
 GridViewIT.gridWithFiltering_scrolledToEnd:710 expected:<Person [100]> but was:<Person [489]>
```
Apart from that I can confirm that the change fixes the issue, and that the test fails when removing the fix, so maybe this is OK.

sissbruecker (commented):
Looks good, verified that it works with manual testing. 

I see that this only targets master, 22, and 21. Did you check if the bug is also in 14?

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridViewIT.java
tomivirkki: scrollToEnd_filter_rowsUpdated ANY_Chrome_ (com vaadin flow component grid it GridViewIT)

In the error screenshot (as in the original issue), the row seems visually empty. The cells however do contain text as the test output shows, but the row is in `[part~='row'][loading]` state so the content has `opacity: 0`.

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridViewIT.java
tomivirkki: Done

tomivirkki:
> I see that this only targets master, 22, and 21. Did you check if the bug is also in 14?

The issue doesn't reproduce with V14. The `<vaadin-grid>` Web Component has diverged quite a lot between V14-V2*. The V14 grid connector most likely works in this case (by accident) because the older `<vaadin-grid>` makes excess update requests for each row. The new version [was optimized](https://github.com/vaadin/web-components/pull/2910) to make just one data request / index.

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridViewIT.java
sissbruecker: Got it, thanks for the explanation.

sissbruecker (approved):

PR #2408, opened 2021-12-06
fix: fix duplicate inputs when using DateTimePicker in a Lit template

sissbruecker:
## Description

When using a DateTimePicker in a Lit template, the Flow component seems to add it's custom date and time pickers at a later stage, at which the web component has already created its own slotted children. This results in duplicates.

This change modifies the initialization logic of the Flow component to first remove the default contents of the slots before adding its custom content.

Fixes https://github.com/vaadin/web-components/issues/3149

## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow/src/main/java/com/vaadin/flow/component/datetimepicker/DateTimePicker.java
+++ b/vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow/src/main/java/com/vaadin/flow/component/datetimepicker/DateTimePicker.java
@@ -145,8 +145,8 @@ public class DateTimePicker
             synchronizeChildComponentValues(initialDateTime);
         }
 
-        addToSlot(datePicker, "date-picker");
-        addToSlot(timePicker, "time-picker");
+        replaceDefaultSlotContent("date-picker", datePicker);
+        replaceDefaultSlotContent("time-picker", timePicker);
 
         setLocale(UI.getCurrent().getLocale());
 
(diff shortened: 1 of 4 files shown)


vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow/src/main/java/com/vaadin/flow/component/datetimepicker/DateTimePicker.java
```diff
@@ -292,11 +292,21 @@ private void synchronizeChildComponentValues(LocalDateTime value) {
...
+        component.getElement().setAttribute("is-flow-slotted-child", "");
+
+        // executeJs call to remove the default slot contents. executeJs
+        // actually runs after appendChild, so the removal logic would also
+        // remove the replacement component we add below. To fix that we check
+        // for a custom attribute and only remove children that do not have it.
+        getElement().executeJs(
+                "this.querySelectorAll(`[slot=\"${$0}\"]`).forEach(child => !child.hasAttribute('is-flow-slotted-child') && child.remove());",
+                slot);
         getElement().appendChild(component.getElement());
```
sissbruecker, line 310: This is pretty dumb, but I could not find a way to synchronize the executeJs and appendChild calls.

I was thinking about adding a mode to the web component to prevent it from creating its own default content by setting an attribute, but I think that doesn't work with templates, as the creation of the element is controlled by the user's template.

sissbruecker: Actually dedicated methods on the web component for replacing the date picker and time picker might work.

sissbruecker:
Closing in favor of https://github.com/vaadin/web-components/pull/3153

PR #2409, opened 2021-12-07
fix: register custom editor onAttach

### LINKED ISSUE #2466: Server-side date parsing
### Describe your motivation
Some applications use various specialized strings for quickly entering dates, such as `+1` or `tomorrow` for tomorrow. While we can extend the parsing formats supported by Date Picker via date-fns in the future, it is likely that there will always be custom, application-specific strings that even date-fns does not support. Providing a solution for server-side parsing of dates, that allow developers to provide their own parsing, would solve these use cases.

A way to provide server-side date parsing as a fallback (if the parsing formats provided are unable to parse the entered string). The API could be something along the lines of
```
DatePicker dp = new DatePicker();
dp.setFallbackParser(s -> {
  // custom logic for parsing the input string, e.g.
  if (s.equals("tomorrow")) {
    return LocalDate.now().plusDays(1);
  } else {
    return null;
}
```

TatuLund:
Fixes [#2466](https://github.com/vaadin/web-components/issues/2466)

Fixes https://github.com/vaadin/flow-components/issues/2319

### DIFF
--- a/vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/java/com/vaadin/flow/component/gridpro/EditColumnConfigurator.java
+++ b/vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/java/com/vaadin/flow/component/gridpro/EditColumnConfigurator.java
@@ -8,10 +8,10 @@ package com.vaadin.flow.component.gridpro;
  * %%
  * This program is available under Commercial Vaadin Developer License
  * 4.0 (CVDLv4).
- * 
+ *
  * See the file license.html distributed with this software for more
  * information about licensing.
- * 
+ *
  * For the full License, see <https://vaadin.com/license/cvdl-4.0>.
  * #L%
  */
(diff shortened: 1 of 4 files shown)


vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/java/com/vaadin/flow/component/gridpro/EditColumnConfigurator.java
```diff
@@ -106,16 +109,21 @@
     public <V> Column<T> custom(AbstractField<?, V> component,
             ItemUpdater<T, V> itemUpdater) {
         column.getElement().appendVirtualChild(component.getElement());
-        column.getElement().getNode()
-                .runWhenAttached(ui -> ui.beforeClientResponse(column,
-                        context -> setEditModeRenderer(component)));
+        if (attachRegistration != null) {
+            attachRegistration.remove();
+            attachRegistration = null;
+        }
```
sissbruecker, line 115: Safe-guard in case `custom` is called multiple times.

sissbruecker (approved):

ZheSun88:
looks like the commit has been picked.. 
i will close the auto created one..

Issue #2383, opened 2021-11-26
TreeGrid: refreshAll() creates empty handles with items that changed their hashCode although getId is overridden


werwolfberlin:
When calling refreshAll() on a dataprovider which has items that changed their hashCode (and equals) the TreeGrid shows only expand/collapse handles, but nothing bleow although getId(Item) is overwritten. 

![image](https://user-images.githubusercontent.com/19834316/143589911-3321ab19-0437-4fdd-834e-883b85921d82.png)

See the minimal example file and add it as a view do a V14-V22 application.
We suspect that something is caching the hierarchy based on the beans hashCode/equals instead of getId(item).
I added some breakpoints in hasChildren(), getChildCount(...) and fetchChildren(...). only hasChildren() was hit because it was used to determin wether to show the expand/collapse handle or not.

Last position where we lost the trace in debugging because of lambdas was in TreeGrid.java:

`
        @Override
        public TreeGridUpdateQueue startUpdate(int sizeChange) {
            return (TreeGridUpdateQueue) updateQueueFactory.apply(data,  sizeChange);
        }
`

Code example:
`
package com.example.application.views.helloworld;

import com.vaadin.flow.component.button.Button;
import com.vaadin.flow.component.notification.Notification;
import com.vaadin.flow.component.orderedlayout.HorizontalLayout;
import com.vaadin.flow.component.orderedlayout.VerticalLayout;
import com.vaadin.flow.component.textfield.TextField;
import com.vaadin.flow.component.treegrid.TreeGrid;
import com.vaadin.flow.data.provider.InMemoryDataProvider;
import com.vaadin.flow.data.provider.hierarchy.AbstractHierarchicalDataProvider;
import com.vaadin.flow.data.provider.hierarchy.HierarchicalQuery;
import com.vaadin.flow.function.SerializableComparator;
import com.vaadin.flow.function.SerializablePredicate;
import com.vaadin.flow.router.Route;
import com.vaadin.flow.router.PageTitle;
import com.example.application.views.MainLayout;
import com.vaadin.flow.router.RouteAlias;

import java.util.*;
import java.util.concurrent.atomic.AtomicInteger;
import java.util.stream.Collectors;
import java.util.stream.Stream;

@PageTitle("TreeGridIssue")
@Route(value = "tree-grid-issue", layout = MainLayout.class)
@RouteAlias(value = "", layout = MainLayout.class)
public class TreeGridIssue extends VerticalLayout {

    public TreeGridIssue() {

[...]

Issue #464, opened 2020-11-30
TreeGrid.defaultUniqueKeyProvider is in no way providing a unique key


Artur-:
The implementation here https://github.com/vaadin/vaadin-flow-components/blob/master/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java#L147 is
```
    private final ValueProvider<T, String> defaultUniqueKeyProvider = item -> String.valueOf(item.hashCode());
```

There is no way you can get a unique key by using `hashCode`

rolfsmeds:
Possibly related:
https://github.com/vaadin/vaadin-grid/issues/1925
https://github.com/vaadin/vaadin-grid/issues/1977

PR #2411, opened 2021-12-07
fix: Use getDataProvider().getId(item) instead of item.hashCode()

### LINKED ISSUE #1270: TreeGrid breaks when items have the same hashCode
Sample code:
```
@Route(value = "")
public class TreeTest extends VerticalLayout {
    private final TreeGrid<String> treeGrid = new TreeGrid<>();

    public TreeTest() {
        TreeData<String> stringTreeData = new TreeData<>();
        // "Aa".hashCode() == "BB".hashCode()
        stringTreeData.addItem(null, "Aa");
        stringTreeData.addItem("Aa", "BB");
        treeGrid.setDataProvider(new TreeDataProvider<>(stringTreeData));
        treeGrid.addHierarchyColumn(s -> s);
        treeGrid.expand("Aa");
        add(treeGrid);
    }
}
```

This fails on the client:
Selection_358

If one of the strings is changed, everything works fine.

In our application we have this problem with a TreeGrid that displays JPA entites. The user can create new entries in the tree (that do not have an identifier assigned by jpa yet), so they have the same hash code. We want to save those new entities only when the user clicks 'Save'.

It would also break with entities that follow the recommendation of the hibernate developer advocate for implementing hash code: https://vladmihalcea.com/how-to-implement-equals-and-hashcode-using-the-jpa-entity-identifier/

Tested with Vaadin 13.0.3.

TatuLund:
Fixes #1270 
Fixes https://github.com/vaadin/flow-components/issues/2383
Fixes https://github.com/vaadin/flow-components/issues/464


tomivirkki:
> fixes: #2411

points to this PR

TatuLund:
> points to this PR

@tomivirkki thanks, fixed

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/treegrid/it/TreeGridBasicFeaturesIT.java
```diff
@@ -283,7 +283,6 @@ public void expanded_nodes_stay_expanded_when_parent_expand_state_is_toggled() {
         Assert.assertEquals(9, getTreeGrid().getRowCount());
 
         getTreeGrid().expandWithClick(7);
-        getTreeGrid().expandWithClick(8);
```
tomivirkki, line 286: Does this mean that an existing test breaks with the change in place?

There should also be a new test that 1. fails without the change 2. passes with the change

TatuLund: My understanding is that existing test is faulty and this line should not be there. The bug I am fixing has allowed the existing test to pass. I have asked internally confirmation about this from the original author of the test.

TatuLund: > There should also be a new test that 1. fails without the change 2. passes with the change

Done

tomivirkki: After the change, the Grid on the test page behaves quite unexpectedly in general. There may be applications out there that also rely on the current item identification logic (based on hashCode) and after we'd release this change, they could also start acting weird.

After the change:

https://user-images.githubusercontent.com/1222264/150156899-c0737864-1a7f-4a13-b793-2fa076e57026.mp4

Before the change:

https://user-images.githubusercontent.com/1222264/150156919-a3b2cbb4-60c8-4c69-a274-3f682980f055.mp4

Also, if we change the item ID logic in `TreeGrid`, shouldn't we also change it in `Grid` (and any other component that uses items) for consistency?

TatuLund: Thanks for this, it gave me an idea to check one thing.

TatuLund: I think the problem is in test page

https://github.com/vaadin/flow-components/blob/fix2383/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/treegrid/it/TreeGridBasicFeaturesPage.java#L170

There is a bit rigged identity logic in the test entity, and there are more than one bean with id == null. And that is actually causing the problem. Thus the test bean is not realistic and this should be fixed.

tomivirkki: Are you still working on this? The fix to `TreeGrid` lgtm but this change to the test case isn't correct. As you suggested, the test fix should be done to the `HierarchicalTestBean` identity logic.

TatuLund: Lets see, if it works now

tomivirkki (approved):

PR #2416, opened 2021-12-09
fix: fallback to super.getLocale() if no custom locale is set

### LINKED ISSUE #2379: getLocale on DatePicker overwritten

sissbruecker:
## Description

Extended version of #2380 that covers DatePicker, TimePicker and DateTimePicker. Also refactors the components to pull the locale on demand, rather than setting the custom locale fields to the UI locale on attach.

Opening this as a separate PR for now, in case we don't want to go ahead with the refactoring.

Fixes #2379 

## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
@@ -329,7 +329,11 @@ public class DatePicker extends GeneratedVaadinDatePicker<DatePicker, LocalDate>
      */
     @Override
     public Locale getLocale() {
-        return locale;
+        if (locale != null) {
+            return locale;
+        } else {
+            return super.getLocale();
+        }
     }
 
     @Override
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
@@ -71,7 +71,6 @@ public class DatePicker extends GeneratedVaadinDatePicker<DatePicker, LocalDate>
     };
 
     private Locale locale;
-    private String languageTag;
 
     private LocalDate max;
     private LocalDate min;
@@ -305,20 +304,6 @@ public class DatePicker extends GeneratedVaadinDatePicker<DatePicker, LocalDate>
     public void setLocale(Locale locale) {
         Objects.requireNonNull(locale, "Locale must not be null.");
         this.locale = locale;
-        // For ill-formed locales, Locale.toLanguageTag() will append subtag
-        // "lvariant" to it, which will cause the client side
-        // Date().toLocaleDateString()
-        // fallback to the system default locale silently.
-        // This has been caught by DatePickerValidationPage::invalidLocale test
-        // when running on
-        // Chrome(73+)/FireFox(66)/Edge(42.17134).
-        if (!locale.toLanguageTag().contains("lvariant")) {
-            languageTag = locale.toLanguageTag();
-        } else if (locale.getCountry().isEmpty()) {
-            languageTag = locale.getLanguage();
-        } else {
-            languageTag = locale.getLanguage() + "-" + locale.getCountry();
-        }
         requestI18nUpdate();
     }
 
@@ -340,14 +325,7 @@ public class DatePicker extends GeneratedVaadinDatePicker<DatePicker, LocalDate>
     protected void onAttach(AttachEvent attachEvent) {
         super.onAttach(attachEvent);
         initConnector();
-        if (locale == null) {
-            getUI().ifPresent(ui -> setLocale(ui.getLocale()));
-        } else if (languageTag != null) {
-            requestI18nUpdate();
-        }
-        if (i18n != null) {
-            requestI18nUpdate();
-        }
+        requestI18nUpdate();
         FieldValidationUtil.disableClientValidation(this);
     }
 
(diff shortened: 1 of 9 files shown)


vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
```diff
@@ -329,21 +314,18 @@ public void setLocale(Locale locale) {
...
         initConnector();
-        if (locale == null) {
-            getUI().ifPresent(ui -> setLocale(ui.getLocale()));
-        } else if (languageTag != null) {
-            requestI18nUpdate();
-        }
-        if (i18n != null) {
-            requestI18nUpdate();
-        }
+        requestI18nUpdate();
```
sissbruecker, line 328: All of the cases above should trigger `requestI18nUpdate()`, setting the custom locale field to UI locale is not necessary anymore.

vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow/src/main/java/com/vaadin/flow/component/datetimepicker/DateTimePicker.java
```diff
@@ -148,8 +148,6 @@ public DateTimePicker(LocalDateTime initialDateTime) {
         addToSlot(datePicker, "date-picker");
         addToSlot(timePicker, "time-picker");
 
-        setLocale(UI.getCurrent().getLocale());
```
sissbruecker, line 151: Setting the custom locale field to UI locale is not necessary anymore, DatePicker and TimePicker will automatically use the current UI locale internally.

vaadin-time-picker-flow-parent/vaadin-time-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/timepicker/tests/TimePickerDetachAttachPageIT.java
```diff
@@ -15,36 +15,61 @@
...
         // Detaching and attaching time picker
-        WebElement toggleAttach = findElement(By.id("toggle-attached"));
         toggleAttach.click();
         toggleAttach.click();
 
         assertTimePickerIsValidOnTab();
     }
 
+    @Test
+    public void formatShouldRespectLocaleAfterDetachAndReattach() {
```
sissbruecker, line 57: This PR also fixes a bug in the TimePicker where the time format from the locale would not be reapplied when detaching and reattaching. See changes in `TimePicker.onAttach`.

vaadin-time-picker-flow-parent/vaadin-time-picker-flow/src/main/java/com/vaadin/flow/component/timepicker/TimePicker.java
```diff
@@ -59,11 +62,11 @@
     private static final String PROP_AUTO_OPEN_DISABLED = "autoOpenDisabled";
 
     private Locale locale;
-    private transient DateTimeFormatter dateTimeFormatter;
```
sissbruecker, line 62: Unused code, same with `initializeAndReturnFormatter` below.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
```diff
@@ -408,6 +390,24 @@ private void executeI18nUpdate() {
...
         }
 
+        // For ill-formed locales, Locale.toLanguageTag() will append subtag
+        // "lvariant" to it, which will cause the client side
+        // Date().toLocaleDateString()
+        // fallback to the system default locale silently.
+        // This has been caught by DatePickerValidationPage::invalidLocale test
+        // when running on
+        // Chrome(73+)/FireFox(66)/Edge(42.17134).
+        Locale appliedLocale = getLocale();
```
sissbruecker, line 400: Locale is now pulled on demand, when updating I18N settings in connector.

vaadin-time-picker-flow-parent/vaadin-time-picker-flow/src/main/java/com/vaadin/flow/component/timepicker/TimePicker.java
```diff
@@ -416,7 +407,37 @@ public void setLocale(Locale locale) {
...
+            }
+            pendingLocaleUpdate = ui.beforeClientResponse(this, context -> {
+                pendingLocaleUpdate = null;
+                executeLocaleUpdate();
+            });
+        });
+    }
+
+    private void executeLocaleUpdate() {
+        Locale appliedLocale = getLocale();
```
sissbruecker, line 430: Locale is now pulled on demand, when updating locale settings in connector.

hdamr (approved):
Looks good.

PR #2433, opened 2021-12-16
test: Remove usage of Shadowroot methods

caalador:

PR #2434, opened 2021-12-16
chore: replace findInShadowroot with TB method

ZheSun88:
added mothed to shared module
update the findInShadowroot usage


ZheSun88:
let us close this in favor of #2433

Issue #2461, opened 2021-12-23
TreeGrid.expandRecursivelly(..) does not fire ExpandEvent


TatuLund:
Vaadin 14.8.0

Use the following code to demonstrate the issue

Expected: See Grid and Label with text "Expanded"

Actual: Label is not shown

```
@Route("tree-grid-cols")
public class TreeGridColumns extends VerticalLayout {

    TreeGrid<String> grid;

    public TreeGridColumns() {
        grid = new TreeGrid<>();
        grid.setWidth("350px");
        grid.setMaxHeight("300px");

        grid.addComponentHierarchyColumn(value -> new Label(value))
                .setHeader("Name").setAutoWidth(true);
        grid.addComponentColumn(value -> new Label(new Date().toString()))
                .setHeader("Datum");
        grid.addComponentColumn(value -> new Label(new Date().toString()))
                .setHeader("Datum 2");

        grid.addExpandListener(event -> {
            Label label = new Label("Expanded");
            label.setId("expanded");
            add(label);
        });

        TreeData<String> treeData = new TreeData<>();
        treeData.addRootItems("A");
        treeData.addItems("A", Arrays.asList("A 1", "B 1"));
        treeData.addItems("A 1", Arrays.asList(
                "Child A which may has a long title", "Child B", "Child C"));
        treeData.addItems("B 1",
                Arrays.asList("Child D", "Child E which may has a long title",
                        "Child C which may has a long title"));
        grid.setTreeData(treeData);

        grid.expandRecursively(Collections.singleton("A"), 2);

        add(grid);
    }
 }
```

TatuLund:
/**
     * Expands the given items and their children recursively until the given
     * depth.
     * <p>
     * {@code depth} describes the maximum distance between a given item and its
     * descendant, meaning that {@code expandRecursively(items, 0)} expands only
     * the given items while {@code expandRecursively(items, 2)} expands the
     * given items as well as their children and grandchildren.
     * <p>
     * **This method will <i>not</i> fire events for expanded nodes.**
     *
     */

JavaDoc says it is not even supposed to be fired. This is however quite counterintuitive as the normal expand method does fire the expand event. Thusly issue is still valid as "enhancement" and I did relabel this so.

Now, there is API design question, what ExpandEvent should offer as items when fired from expandRecursively?

[...]

PR #2462, opened 2021-12-23
feat: Fire ExpandEvent in expandRecursively

TatuLund:
fixes: #2461

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/treegrid/it/TreeGridExpandRecursivelyPage.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/treegrid/it/TreeGridExpandRecursivelyPage.java
@@ -49,7 +49,7 @@ public class TreeGridExpandRecursivelyPage extends VerticalLayout {
             add(label);
         });
 
-        grid.addExpandListener(event -> {
+        grid.addCollapseListener(event -> {
             Span label = new Span("Collapsed");
             label.setId("collapsed");
             add(label);
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/treegrid/it/TreeGridExpandRecursivelyPage.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/treegrid/it/TreeGridExpandRecursivelyPage.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2000-2019 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component.treegrid.it;
+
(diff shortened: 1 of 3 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/treegrid/it/TreeGridExpandRecursivelyPage.java
```diff
@@ -0,0 +1,65 @@
...
+
+        grid.addComponentHierarchyColumn(value -> new Label(value))
+                .setHeader("Name").setAutoWidth(true).setFlexGrow(0);
+        grid.addComponentColumn(value -> new Label(new Date().toString()))
+                .setHeader("Datum");
+        grid.addComponentColumn(value -> new Label(new Date().toString()))
+                .setHeader("Datum 2");
+
+        grid.addExpandListener(event -> {
+            Label label = new Label("Expanded");
```
tomivirkki, line 46: Please don't use `Label` element for text. See the description for V10+ Label at https://vaadin.com/api/platform/10.0.19/com/vaadin/flow/component/html/Label.html

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -796,6 +796,10 @@ public void expandRecursively(Stream<T> items, int depth) {
     public void expandRecursively(Collection<T> items, int depth) {
         getDataCommunicator()
                 .expand(getItemsWithChildrenRecursively(items, depth));
+        // Call with client roundtrip as expansion takes a while.
+        // This ensures that event is fired after exapnsion is complete.
+        getElement().executeJs("return 1;").then(result -> fireEvent(
+                new ExpandEvent<T, TreeGrid<T>>(this, false, items)));
```
tomivirkki, line 802: The `items` here only contains the items at the root of the recursive expansion. Sub-items that may also get expanded aren't included in the event.

tomivirkki, line 802: The JavaDoc for this method says `This method will <i>not</i> fire events for expanded nodes.`

TatuLund: I will use null instead

TatuLund: I changed the issue type to enhancement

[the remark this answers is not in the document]
TatuLund: I will remove the delay

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/treegrid/it/TreeGridExpandRecursivelyPage.java
TatuLund: Will change to Span

yuriy-fix:
Thank you for your contribution!
The original issue was fixed with this fix: https://github.com/vaadin/flow-components/pull/2540. Closing as enhancements is not needed anymore.

TatuLund:
Eventually there could be other use cases, which would make this feature useful.

tomivirkki:
> Eventually there could be other use cases, which would make this feature useful.

The `ExpandEvent` type doesn't fit well for the case of recursively expanding items. The suggested convention to use `null` for the `items` property doesn't sound right, because then the event doesn't really tell you anything else than "something in the grid was expanded recursively". If an event is needed for recursive expansion, there should probably be a dedicated event type for it with perhaps `rootItems` property. As such, the PR is also breaking, because methods that explicitly documented not to fire `ExpandEvent` events, would start firing them.

Issue #1086, opened 2021-05-21
Message "There seems to be an error in the Vaadin Combo Box" when using combo boxes in a Grid


eriklumme:
### Description

When having a `Grid<Integer>`, and a column defined as such:

```
for (int c = 0; c <numberOfColumns; c++) {
    grid.addComponentColumn(createComboBoxFactory(c))
            .setAutoWidth(true).setHeader("Component-Combo-Box" + c);
}
```

with the factory defined as such

```
private final List<String> comboBoxItems = Arrays.asList("One", "Two", "Three", "Four");

private ValueProvider<Integer, Component> createComboBoxFactory(int columnIndex) {
    return i -> {
        ComboBox<String> comboBox = new ComboBox<>(String.format("Component-Combo-Box-%d-%d", i, columnIndex));
        comboBox.setItems(comboBoxItems);
        comboBox.addValueChangeListener(e -> Notification.show(
                String.format("Selected value %s in row %d column %d", e.getValue(), i, columnIndex)));
        return comboBox;
    };
}
```

When generating a `Grid` on a button click, the following error message is displayed in the browser console, once for each column:

![image](https://user-images.githubusercontent.com/29859656/96599380-88213880-12f8-11eb-8470-ec8b4fa82934.png)

### Expected outcome
No errors are shown.

### Actual outcome
The error is shown.

### Browsers Affected
Browser: Chrome 86.0.4240.75
Vaadin version: 14.4.1

gianluca-valentini:
I have the same problems using Vaadin 14.6.5
but with different messages:
![image](https://user-images.githubusercontent.com/3224592/125096958-eb950580-e0d5-11eb-9e20-d0179b8febc1.png)

![image](https://user-images.githubusercontent.com/3224592/125097002-f780c780-e0d5-11eb-944d-3ac6f80f5b97.png)

![image](https://user-images.githubusercontent.com/3224592/125097127-14b59600-e0d6-11eb-8a64-ed385d2d287e.png)

Every time I add one Combo to my Grid, that 3 messages are showed in the browser

PR #2487, opened 2022-01-10
fix: patch _isItemSelected once combobox is finalized

tomivirkki:
Fixes https://github.com/vaadin/flow-components/issues/1086

The `comboBoxConnector` needs to patch the `<vaadin-combo-box-scroller>`'s `__isItemSelected` function in order to add custom logic to it.

Previously, the [patching took place](https://github.com/vaadin/flow-components/blob/d4d675d6730c64fbd58586ed59b5c4ba16f08d2b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js#L245) once `customElements.whenDefined('vaadin-combo-box')` had resolved. The timing is problematic since the Polymer-based `<vaadin-combo-box>` may have not been yet finalized even if the Custom Element type was defined. In order for a Polymer element type to get fully finalized, at least one instance of it must have been attached to the DOM.

This PR changes the timing so that the function gets patched not until the `<vaadin-combo-box>` element emits an "opened-changed" event, which indicates that the Custom Element type has been defined AND finalized.

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
@@ -239,7 +239,10 @@ import { ComboBoxPlaceholder } from '@vaadin/combo-box/src/vaadin-combo-box-plac
                 comboBox.$server.confirmUpdate(id);
             });
 
-            customElements.whenDefined('vaadin-combo-box').then(tryCatchWrapper(() => {
+            comboBox.addEventListener('opened-changed', () => {
+                // Patch once the instance is ready and vaadin-combo-box has
+                // been finalized (i.e. opened-changed is emitted)
+
                 const isItemSelected = comboBox.$.dropdown._scroller.__isItemSelected;
                 // Override comboBox's _isItemSelected logic to handle remapped items
                 comboBox.$.dropdown._scroller.__isItemSelected = (item, selectedItem, itemIdPath) => {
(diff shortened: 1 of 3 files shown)


vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -239,7 +239,10 @@ import { ComboBoxPlaceholder } from '@vaadin/combo-box/src/vaadin-combo-box-plac
                 comboBox.$server.confirmUpdate(id);
             });
 
-            customElements.whenDefined('vaadin-combo-box').then(tryCatchWrapper(() => {
+            comboBox.addEventListener('opened-changed', () => {
```
yuriy-fix, line 242: Nit: to align with other listeners, should probably be wrapped in `tryCatchWrapper` as well.

yuriy-fix (approved):
One nit, otherwise LGTM

web-padawan (commented):
The `addEventListener()` approach looks good. Note, we also have similar code below checking for `comboBox.$` that probably needs to be updated accordingly.

https://github.com/vaadin/flow-components/blob/ae392ddb64f9f01080d65e6d8ce5aa6aea8f0abd/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js#L262-L276

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
tomivirkki: Done.

tomivirkki:
> The addEventListener() approach looks good. Note, we also have a similar code below checking for comboBox.$ that probably needs to be updated accordingly. But we can consider doing that in a separate PR.

Would prefer keeping that separate since it requires refactoring for the existing `enableClientValidation` logic

PR #2547, opened 2022-01-24
feat: allow silent configuration updates

sissbruecker:
## Description

Adds a method to all configuration classes to allow updating them silently, without dispatching change events. This is useful for updating them based on client-side changes, where we don't want to send the changes back to the client again.


hdamr (approved):
Looks good to me.

PR #2554, opened 2022-01-24
feat: implement map feature flag

### LINKED ISSUE #2495: [map] Feature flag
### Motivation

While the map component is in development, and in experimental state, it should be hidden behind a feature flag.

Specific implementation needs to be clarified, but workable solutions would be:
- for the web component, look for a global variable
  - only define the custom element if it exists
  - or throw error when initialising the component
- for the Flow component, look for a Java property, only define global variable in browser if exists

### Acceptance criteria
- Map component can only be used when feature flag is enabled

sissbruecker:
## Description

Adds support for the new map feature flag to the map component.

Fixes #2495 

## Type of change

- [x] Feature


hdamr (approved):
LGTM.

Issue #1667, opened 2019-09-06
checkBox.setLabelAsHtml method is not behaving as expected!


arunt4ever:
Vaadin Version: 14.0.3

when I try to set the checkbox label using this method (checkBox.setLabelAsHtml) and then if I try to access the label using checkBox.getLabel() I am not getting anything back!

```java
Checkbox checkBox = new Checkbox(todoField.getValue());
checkBox.addValueChangeListener(valueChangeEvent->{
   if(valueChangeEvent.getValue()){
	   checkBox.setLabelAsHtml("<strike>"+checkBox.getLabel()+"</strike>");
   }else{
	   checkBox.setLabel(checkBox.getLabel());
   }

});
```

web-padawan:
The method was removed in Vaadin 24 in #4078. Closing as stale.

PR #2558, opened 2022-01-25
fix: only use custom label when setting HTML

### LINKED ISSUE #2551: [checkbox] Label property defined in template is ignored when using `@Id`
### Description

There is an IT covering usage `<vaadin-checkbox>` in Polymer template with `@Id`:

https://github.com/vaadin/flow-components/blob/b3155936bbc319396b1021721f3eecd35b77d815/vaadin-checkbox-flow-parent/vaadin-checkbox-flow-integration-tests/frontend/inject-checkbox.js#L8

When updating the template to use `label` property instead of the default slot, it is ignored"

Screenshot 2022-01-24 at 16 17 23

### Expected outcome

The checkbox should have correct label `"Accept"`.

### Actual outcome

The label provided in template is getting overridden.

### Environment

- Vaadin 22 (stable) and 23 (alpha)

web-padawan:
## Description

Currently, the custom `<label>` is always created even if there is one provided by the user via template.
This PR changes it to use `label` property instead, and only create `<label>` when using `.setLabelAsHtml()`.

Also, updated the IT for the checkbox in template to ensure the `label` attribute value is preserved.

Fixes #2551

## Type of change

- Bugfix

### DIFF
--- a/vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/Checkbox.java
+++ b/vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/Checkbox.java
@@ -35,8 +35,6 @@ import com.vaadin.flow.dom.PropertyChangeListener;
 public class Checkbox extends GeneratedVaadinCheckbox<Checkbox, Boolean>
         implements HasSize, HasLabel {
 
-    private final Label labelElement = appendLabelElement();
-
     private static final PropertyChangeListener NO_OP = event -> {
     };
 
@@ -138,7 +136,7 @@ public class Checkbox extends GeneratedVaadinCheckbox<Checkbox, Boolean>
      */
     @Override
     public void setLabel(String label) {
-        labelElement.setText(label);
+        getElement().setProperty("label", label == null ? "" : label);
     }
 
     /**
@@ -153,7 +151,11 @@ public class Checkbox extends GeneratedVaadinCheckbox<Checkbox, Boolean>
      *            the label html to set
      */
     public void setLabelAsHtml(String htmlContent) {
-        labelElement.getElement().setProperty("innerHTML", htmlContent);
+        // Create and add a new slotted label
+        Label label = new Label();
+        label.getElement().setAttribute("slot", "label");
+        label.getElement().setProperty("innerHTML", htmlContent);
+        getElement().appendChild(label.getElement());
     }
 
     /**
--- a/vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/Checkbox.java
+++ b/vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/Checkbox.java
@@ -120,14 +120,6 @@ public class Checkbox extends GeneratedVaadinCheckbox<Checkbox, Boolean>
                 .get().getText();
     }
 
-    private Label appendLabelElement() {
-        // Create and add a new slotted label
-        Label label = new Label();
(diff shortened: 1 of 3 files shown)


vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/Checkbox.java
```diff
@@ -138,7 +128,7 @@ private Label appendLabelElement() {
      */
     @Override
     public void setLabel(String label) {
-        labelElement.setText(label);
+        getElement().setProperty("label", label == null ? "" : label);
```
tomivirkki, line 131: I'm guessing `getLabel()` wouldn't work after this change because it tries to find a `<label>` element from inside the component but now there's no server-side representation for it.

web-padawan: Updated to use client-side property, same as for other components like `TextField`.

vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/Checkbox.java
```diff
@@ -153,7 +141,11 @@ public void setLabel(String label) {
      *            the label html to set
      */
     public void setLabelAsHtml(String htmlContent) {
-        labelElement.getElement().setProperty("innerHTML", htmlContent);
+        // Create and add a new slotted label
```
tomivirkki, line 144: We should call `setLabel("")` here. Otherwise after `setLabel("foo"); setLabelAsHtml("bar")`, `getLabel()` would return "foo". Related https://github.com/vaadin/flow-components/issues/1667

tomivirkki (changes requested):

vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/Checkbox.java
web-padawan: Done.

vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/Checkbox.java
```diff
@@ -153,7 +141,11 @@ public void setLabel(String label) {
...
+        Label label = new Label();
+        label.getElement().setAttribute("slot", "label");
+        label.getElement().setProperty("innerHTML", htmlContent);
+        getElement().appendChild(label.getElement());
```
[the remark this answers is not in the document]
web-padawan: Updated to only create label once.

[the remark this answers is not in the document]
tomivirkki: If we append the label element here, the Web Component results in state:
```
<vaadin-checkbox><label>html</label></vaadin-checkbo>
```
If `setLabel("new label")` is called afterwards, the result is:
```
<vaadin-checkbox label="new label"><label>html</label></vaadin-checkbo>
```
That would probably render "html", even though it should say "new label"?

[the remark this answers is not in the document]
web-padawan: Updated `setLabel()` to remove custom label in case it's connected.

tomivirkki (approved):

yuriy-fix (approved):

PR #2560, opened 2022-01-25
feat: display geographic features on the map

### LINKED ISSUE #2512: [map] High-level API for adding / removing features
### Motivation

OpenLayers provides an extensive API to add custom features to a map, using a combination of vector layers, vector sources, features, geometry and styles. While we want to provide low-level access to this API at some point, in a first step we want to provide a high-level API for managing a set of features.

For this the map could provide, and initialize by default, a high-level feature layer that abstracts all these low-level constructs from OpenLayers into something that's easier to use. The layer should allow to add and remove specific types of features. A feature is again a high-level abstraction of several OL APIs that are needed to display for example a point, or a marker with an image.

The first feature we want to provide is for displaying points. Specific functionality around the point feature is TBD, but will likely include positioning, size, stroke color, background color.

### Acceptance criteria
- Map creates a feature layer on initialization
- Map provides access to the feature layer
- Feature layer allows to add and remove features
- Map provides a point feature
- Point feature is configurable, functionality TBD

### LINKED ISSUE #2514: [map] Marker feature
### Motivation

Building on #2512 , we want to provide another high-level feature for markers. A marker feature should show up as an icon on the map. The marker should provide a default icon / image. Additionally it should be possible to provide custom images through a URL.

### Acceptance criteria
- Marker feature exists
- Marker feature can be added to feature layer
- Marker feature displays a default icon
- Marker feature allows to configure a custom icon through a URL
- Marker feature allows to configure the anchor / offset for the custom icon

sissbruecker:
## Description

Allows to display geographic features on the map:
- add a high-level feature layer that manages map features
- Map provides a default feature layer
- adds a high-level marker feature type for easily displaying an icon at a coordinate
- marker feature uses a default marker image, allows to specify custom images

Fixes #2512 
Fixes #2514 

## Type of change

- [x] Feature


vaadin-map-flow-parent/vaadin-map-flow/src/main/resources/META-INF/resources/frontend/vaadin-map/synchronization/styles.js
```diff
@@ -0,0 +1,101 @@
...
+    target = new Stroke();
+  }
+
+  target.setColor(source.color);
+  target.setWidth(source.width);
+
+  return target;
+}
+
+function synchronizeImageStyle(target, source, _context) {
```
hdamr, line 32: `synchronizeImageStyle` it looks like this is not added to the lookup table in the `index.js` file.

sissbruecker: That was intentional. The Image style is only a base class in OpenLayers, so we don't want it to be used directly in a configuration. Similarly the Java class pendant is declared as abstract, so it can not be instantiated. So the only use-case for `synchronizeImageStyle` is to be called from sync functions of sub-classes (see `synchronizeIcon`), which is why we don't need to expose it.

hdamr (approved):

Issue #2528, opened 2022-01-18
RadioButtonGroup ItemEnabledProvider no more applied after calling setEnabled(...) and setRenderer(...)


SonReimer:
After Calling bothe methods  setEnabled(...) and setRenderer(...) on RadioButtonGroup in a e.g. ClickListener a previously configured ItemEnabledProvider will no more work correctly (Click twice). 
Setting enabled=false initially leads to the same problem (Click once).
A workaround is to set the ItemEnabledProvider again.
This occures only when calling both methods given below together.

[RadioButtonGroupView.java.txt](https://github.com/vaadin/flow-components/files/7888938/RadioButtonGroupView.java.txt)

I think it has something to do with the internal uodateButton-method, but I could not figure out, what's going wrong.

Expected :

![Expected](https://user-images.githubusercontent.com/46707079/149937202-05d13fa7-3e84-490f-9288-7e796c358f01.JPG)

Shown

![Shown](https://user-images.githubusercontent.com/46707079/149937239-9d6a6289-7e26-404e-aec4-b596ffb9eecb.JPG)

PR #2563, opened 2022-01-25
fix: enforce radio button web component disabled state

tomivirkki:
Fixes [#2528](https://github.com/vaadin/flow-components/issues/2528)


vaadin-radio-button-flow-parent/vaadin-radio-button-flow/src/main/java/com/vaadin/flow/component/radiobutton/RadioButtonGroup.java
```diff
@@ -570,16 +569,11 @@ private void updateEnabled(RadioButton<T> button) {
...
-            button.getElement().setProperty("disabled",
-                    disabled ? Boolean.TRUE.toString() : null);
-        } else {
-            button.setDisabled(disabled);
-        }
+        button.setDisabled(disabled);
+        // When enabling a disabled radio group, individual button Web Components that should
+        // remain disabled (due to itemEnabledProvider), may end up rendering as enabled.
+        // Enforce the Web Component state using JS.
+        button.getElement().executeJs("this.disabled = $0", disabled);
```
tomivirkki, line 576: For some reason, simply setting the `disabled` property state for the button element isn't enough to get it to reflect properly on the Web Component (there's probably an issue in Flow enabled state handling). Added a JS call to enforce the state.

There were some earlier workarounds that aimed to tackle the same issue, removed them in favor of this approach.

web-padawan (approved):

PR #2576, opened 2022-01-26
feat: add map component behind feature flag

sissbruecker:
## Description

Adds the map component from the feature branch to master.

## Type of change

- [X] Feature


vaadin-map-flow-parent/vaadin-map-flow-integration-tests/pom.xml
```diff
@@ -0,0 +1,186 @@
...
+            <activation>
+                <property>
+                    <name>!skipJetty</name>
+                </property>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.eclipse.jetty</groupId>
+                        <artifactId>jetty-maven-plugin</artifactId>
```
tomivirkki, line 166:
```suggestion
                        <artifactId>jetty-maven-plugin</artifactId>
                        <configuration>
                            <scanIntervalSeconds>5</scanIntervalSeconds>
                        </configuration>
```

manolo:
Validation passes now, please fix sonar and promote from draft to PR

manolo (approved):

vaadin-map-flow-parent/vaadin-map-flow/src/main/resources/META-INF/resources/frontend/vaadin-map/mapConnector.js
```diff
@@ -0,0 +1,66 @@
...
+
+        const context = { synchronize };
+
+        synchronize(target, configuration, context);
+        // TODO: layers don't render on initialization in some cases, needs investigation
+        mapElement.configuration.updateSize();
+      },
+    };
+
+    mapElement.configuration.on("moveend", (_event) => {
```
tomivirkki, line 43: It's possible for `mapElement.configuration` to be `undefined` at this point. For example, if you don't have a Map on the page beforehand and add one inside a Dialog (gets first added as a virtual child).

The error has to do with init timing in the `<vaadin-map>` Web Component: `configuration` property should be available right after the constructor has run, (`ready()` is too late since it will invoke only once the component class has finalized)

sissbruecker: Great find, opened a bug ticket for now: https://github.com/vaadin/flow-components/issues/2595

Issue #2601, opened 2022-02-01
[map] Add support for custom tile grids


sissbruecker:
### Motivation

We should support configuring custom [tile grids](https://openlayers.org/en/latest/apidoc/module-ol_tilegrid_TileGrid-TileGrid.html) to define how the grid of tiles is set up. This allows to define the extent of the grid, the tile size, the origin, and how all these values align with the maps zoom level. This may be necessary for certain map services or specific use-cases.

PR #2616, opened 2022-02-08
[map] Add support for custom tile grids

hdamr:
Adds basic support for setting tile grid for sources.

Fixes: [#2601](https://github.com/vaadin/flow-components/issues/2601)


sissbruecker:
We decided to shelve this feature for now.

PR #2618, opened 2022-02-08
fix: clear selected key when item selected

### LINKED ISSUE #2615: ComboBox Issue - two selected items(v22)
### Description

The error looks similar to #1399, but it affects Vaadin 22 and is very easy to reproduce.

I have a ComboBox that already has a value set from the beginning. When ComboBox#setValue is called with the set value in the request where a new value is selected, two check marks appear in the ComboBox popup after the next open.

### Expected outcome

The ComboBox should always contain only one check mark.

### Live Demo (optional)

image

```
        comboBox = new ComboBox<String>();

        comboBox.setItems("1", "2", "3");
        comboBox.setValue("1");
            value = e.getValue();
            comboBox.setValue(value);
        });

        add(comboBox);
```

### Steps to reproduce

Select Item 2 in the ComboBox. Open Popup of ComboBox again.

### Environment

Vaadin 22
Windows and Linux

### Browsers Affected

- [X] Chrome
- [X] Firefox

did not test further browsers

web-padawan:
## Description

Updated to use `selected-item-changed` event for checking if `_selectedKey` is outdated.

Fixes #2615

## Type of change

- Bugfix

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/main/java/com/vaadin/flow/component/combobox/test/ComboBoxPage.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/main/java/com/vaadin/flow/component/combobox/test/ComboBoxPage.java
@@ -62,6 +62,7 @@ public class ComboBoxPage extends Div {
         createWithUpdateProvider();
         createWithValueChangeListener();
         createWithUpdatableValue();
+        createWithUpdateOnValueChange();
         createWithPresetValue();
         createWithButtonRenderer();
         setLabelGeneratorAfterValue();
@@ -197,6 +198,20 @@ public class ComboBoxPage extends Div {
         message.setId("updatable-combo-message");
         button.setId("updatable-combo-button");
         add(combo, message, button);
+    }
+
+    private void createWithUpdateOnValueChange() {
+        ComboBox<String> combo = new ComboBox<>();
+        combo.setItems("1", "2", "3");
+        combo.setValue("2");
+
+        combo.addValueChangeListener(e -> {
+            String value = e.getValue();
+            combo.setValue(value);
+        });
+
+        combo.setId("update-on-change-combo");
+        add(combo);
     }
 
     private void setLabelGeneratorAfterValue() {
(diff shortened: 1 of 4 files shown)


vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/main/java/com/vaadin/flow/component/combobox/test/ComboBoxPage.java
```diff
@@ -199,6 +200,20 @@ private void createWithUpdatableValue() {
         add(combo, message, button);
     }
 
+    private void createWithUpdateOnValueChange() {
```
tomivirkki, line 203: nit: let's avoid adding new components on these IT pages / old demo pages. Having to render unnecessary component instances makes the tests run slower and difficult to debug. Ideally, one page would contain just 1 component instance. Not a blocker for this PR.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -260,8 +260,11 @@ import { ComboBoxPlaceholder } from '@vaadin/combo-box/src/vaadin-combo-box-plac
                 }
             }), { once: true });
 
-            comboBox.addEventListener('value-changed', tryCatchWrapper((event) => {
-                if (comboBox._selectedKey && !event.detail.value) {
+            comboBox.addEventListener('selected-item-changed', tryCatchWrapper((event) => {
```
tomivirkki, line 263: The `_selectedKey` property is originally [set](https://github.com/vaadin/flow-components/blob/8dc595c13137613db970412956a38757272a7989/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L1671) in ComboBox (on requested range change). Because of that, something about updating the value in the connector feels wrong.

Would it make sense to turn it into a synchronized property and have it update on requested range change and value change. Something like
```java
getElement().setProperty("_selectedKey",
  getValue() != null ? getKeyMapper().key(getValue()) : "");
```

I think it would resolve this issue also because the property would update on selection change.

web-padawan: Removed the logic from the connector and used `addValueChangeListener` instead.

tomivirkki (approved):

PR #2621, opened 2022-02-09
fix: avoid temporarily clearing the component renderer

### LINKED ISSUE #2363: Row height does not adjust after setting new data in Grid when using AppLayout
Grid row height does not adjust after setting new data, see picture

image

Vaadin 21.0.x

Use the following app to replicate, click the button to set new data provider with new items

Use this sequence to replicate
- Click add button
- Scroll to bottom
- Click add button again

[grid-row-height-demo.zip](https://github.com/vaadin/flow-components/files/7569660/grid-row-height-demo.zip)

The workaround is to do delayed call to `grid.getElement().executeJs("this.__virtualizer.flush()");`

tomivirkki:
Fixes #2363

The `Virtualizer` (also used by `<vaadin-grid>`) [uses](https://github.com/vaadin/web-components/blob/87bbb0e333cd037c006c6cf279f574aa28cedd01/packages/component-base/src/virtualizer-iron-list-adapter.js#L34) a native `ResizeObserver` internally to detect and react to item height changes; in the case of `Grid`, it observes height changes on the body row (`<tr>`) elements. Typically it catches row height changes just fine and updates the row positions accordingly.

In this specific case, a `<flow-component-renderer>` on each row gets assigned a new element reference due to a dynamic data update, resulting in the [old element getting removed](https://github.com/vaadin/flow-components/blob/8dc595c13137613db970412956a38757272a7989/vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/flow-component-renderer.js#L66) and the [new element getting added _asynchronously_](https://github.com/vaadin/flow-components/blob/8dc595c13137613db970412956a38757272a7989/vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/flow-component-renderer.js#L67).

Between these events, (after clearing the content and before the new element gets asynchronously added), the `Grid` [requests](https://github.com/vaadin/web-components/blob/87bbb0e333cd037c006c6cf279f574aa28cedd01/packages/grid/src/vaadin-grid.js#L494) the `Virtualizer` to flush, resulting in the [row positions getting recalculated ](https://github.com/vaadin/web-components/blob/87bbb0e333cd037c006c6cf279f574aa28cedd01/packages/component-base/src/virtualizer-iron-list-adapter.js#L118)with _empty_ `<flow-component-renderer>`s in place (meaning, with reduced row height)

After that, the new content is added and the row height increases back to what it was before. One could assume that the `ResizeObserver` would catch this change in height and fix the row positions. But while all this happens asynchronously, the height actually gets restored exactly to what it was and the delay is fast enough so that the `ResizeObserver` simply ignores the temporary size change. As a result, the rows end up not getting re-measured and re-positioned with the proper content in place.

This PR changes `flow-component-renderer` so that it doesn't clear its content while waiting for the new element, but always [uses](https://github.com/vaadin/flow-components/blob/8dc595c13137613db970412956a38757272a7989/vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/flow-component-renderer.js#L69) `replaceChild` if it includes a discarded child element beforehand.

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridDetailsRowIT.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridDetailsRowIT.java
@@ -25,7 +25,6 @@ import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
 
 import java.util.List;
-import java.util.Locale;
 
 @TestPath("vaadin-grid/grid-details-row")
 public class GridDetailsRowIT extends AbstractComponentIT {
@@ -69,10 +68,10 @@ public class GridDetailsRowIT extends AbstractComponentIT {
                 .findElements(By.tagName("flow-component-renderer"));
         Assert.assertEquals(3, detailsElement.size());
 
-        // detail on row 0 is empty
-        assertElementHasNoButton(detailsElement.get(0));
-        // detail on row 1 is empty
-        assertElementHasNoButton(detailsElement.get(1));
+        // detail on row 0 is not displayed
+        assertElementNotDisplayed(detailsElement.get(0));
+        // detail on row 1 is not displayed
+        assertElementNotDisplayed(detailsElement.get(1));
         // detail on row 3 contains a button
         assertElementHasButton(detailsElement.get(2), "Person 4");
     }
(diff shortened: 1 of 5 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridDetailsRowIT.java
```diff
@@ -25,7 +25,6 @@
 import org.openqa.selenium.WebElement;
 
 import java.util.List;
-import java.util.Locale;
```
tomivirkki, line 28: Changes to GridDetilaRow tests are a cherry-pick of https://github.com/vaadin/flow-components/pull/2611/commits/57c2b3de08df7e4e0a32a0f889dbdad8142b0172 and not directly related to this PR

web-padawan (approved):
LGTM. The old logic is indeed confusing, it's better to only clear node once.

PR #2493, opened 2022-01-11
fix: Add custom field when column has been added

TatuLund:
fixes: https://github.com/vaadin/web-components/issues/3264

PR #2624, opened 2022-02-09
fix: add custom field when column has been added (#2493)

yuriy-fix:
Backport https://github.com/vaadin/flow-components/pull/2493 to Vaadin 21


web-padawan:
Vaadin 21 support has ended, we no longer update it.

PR #2638, opened 2022-02-11
feat: add default icons for map markers

sissbruecker:
## Description

Adds final version of the default icons for map markers.


DiegoCardoso (approved):
LGTM

PR #2650, opened 2022-02-11
fix: prevent duplicate feature click events for overlapping markers

sissbruecker:
## Description

Currently clicking on a location that contains overlapping features, results in one click event per feature. This fixes it by sending only a single event per layer that contains clicked features, for the top-most feature in that layer at that position. This is possible because OpenLayers actually returns the features for a clicked position sorted by their display order (either the order in which they are added, or using their z-index), so the first returned feature is the one that is displayed on top.

Theoretically we could also only send a single event for all layers, but I'm not sure if there would be use-cases where the clickable feature would be below another one in a higher layer. Besides the Flow component already supports registering feature clicks for specific layers only.

## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-map-flow-parent/vaadin-map-flow/src/main/resources/META-INF/resources/frontend/vaadin-map/mapConnector.js
+++ b/vaadin-map-flow-parent/vaadin-map-flow/src/main/resources/META-INF/resources/frontend/vaadin-map/mapConnector.js
@@ -1,5 +1,5 @@
 import { synchronize } from "./synchronization";
-import { getLayerForFeature } from "./util";
+import { findFirstFeaturePerLayer } from "./util";
 
 (function () {
   function init(mapElement) {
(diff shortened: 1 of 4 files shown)


vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/events/MapClickEvent.java
```diff
@@ -50,6 +65,19 @@ public Coordinate getCoordinate() {
...
+     * List of map features at the clicked position. Can be used to distinguish
+     * whether the click was on the background, or on a feature. The features
+     * are sorted by their display order, meaning the top-most feature is the
+     * first item in the list.
+     *
+     * @return the list of features at the clicked position, or an empty list
+     *         otherwise
+     */
+    public List<FeatureEventDetails> getFeatures() {
+        return features;
```
DiegoCardoso, line 78: Thinking of whether it should return an unmodifiable list or not.

vaadin-map-flow-parent/vaadin-map-flow/src/main/resources/META-INF/resources/frontend/vaadin-map/mapConnector.js
```diff
@@ -62,34 +62,51 @@ import { getLayerForFeature } from "./util";
...
+      // returns the features sorted in the order that they are displayed,
+      // with the front-most feature as the first result, and the
+      // back-most feature as the last result
+      const pixelCoordinate = event.pixel;
+      const featuresAtPixel =
+        mapElement.configuration.getFeaturesAtPixel(pixelCoordinate);
+      // Create tuples of features and the layer that they are in
+      const featuresAndLayers = featuresAtPixel.map((feature) => {
+        const layer = getLayerForFeature(
+          mapElement.configuration.getLayers().getArray(),
```
DiegoCardoso, line 76: Not sure how costly this is, but I believe this array could be stored outside of the loop.

DiegoCardoso (changes requested):

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/events/MapClickEvent.java
sissbruecker: Sure, can't hurt. Done.

vaadin-map-flow-parent/vaadin-map-flow/src/main/resources/META-INF/resources/frontend/vaadin-map/mapConnector.js
sissbruecker: It just [accesses the internal array](https://github.com/openlayers/openlayers/blob/v6.12.0/src/ol/Collection.js#L167), the OL Collection is basically an observable wrapper around an array. The only benefit for extracting a variable here might be readability, but I don't think it improves much.

DiegoCardoso: That's fine, then.

DiegoCardoso (approved):
LGTM

PR #2663, opened 2022-02-14
test: fix simultaneous multi-upload tests

Artur-:


alvarezguille:
How's the test broken? Does it fail locally? The referred issue seems to be fixed now.

Artur-:
It fails locally

sissbruecker:
Restored the tests and fixed the multi-file upload setup. It seems that when running the tests locally, the upload was so quick that it completed before the Selenium commands for the next upload could be scheduled. As a workaround I added a method to the testbench element for uploading multiple files at once, which internally disables auto upload. That should approximate the desired behavior close enough.

@alvarezguille Maybe you can take a look at the changes?

sissbruecker:
Converted to draft, going to spend more time on this (after vacation, in about a week) to check if we can fix the tests to also run in CI.

alvarezguille (approved):

alvarezguille:
Not a must for this PR, but next we should do the same change in `NonImmediateUploadIT`

alvarezguille (approved):

Issue #2549, opened 2022-01-24
[map] Optimize client-side memory usage


sissbruecker:
### Describe your motivation

Currently, even when reusing some instances in the Flow component (e.g. a Style, an Icon), we always create new instances in the places where these are used on the client side. For example each client-side marker gets its own instance of an Icon, even though its always the same icon, with the same ID. This in turn means we use much more memory than we would need to, especially for image related resources, which seem to duplicate the whole image buffer again and again. We should optimize this to reuse instances somehow, based on ID or some other reference.

### Proposal

A simple solution could be to add a cache to the client-side synchronization logic, that maps IDs to configuration object instances. We then update the sync logic to only create new instances if the cache does not contain an entry for that ID.

Issue #2548, opened 2022-01-24
[map] Optimize synchronization payload


sissbruecker:
### Motivation

Currently the Flow map component synchronizes the whole configuration whenever any property within the configuration hierarchy changes. That means for every change, we always transmit the whole configuration JSON to the browser and we always traverse the whole configuration hierarchy on the client side to synchronize it to the OpenLayers model. Especially when adding many markers, then updating a single one should not trigger a resync of all markers. We should optimize the synchronization to only update the parts that have changed.

### Proposal

A simple solution could be to extend the existing observable mechanism, that the configuration classes use, to provide more fine-grained change events. For example we could extract the following events:
 - Layer added
 - Layer updated
 - Layer removed
 - Feature added
 - Feature updated
 - Feature removed

Based on these events we add custom synchronization logic to sync only that part of the configuration tree.

PR #2664, opened 2022-02-14
perf: implement dirty checking for map configuration sync

sissbruecker:
## Description

Currently, on any change, the map will synchronize its complete configuration. Also the synchronization mechanism doesn't have any capabilities of reusing client-side instances if they relate to the same Java instance. This can lead to two types of performance issues:
- when the configuration hierarchy is quite large (e.g. several hundred markers), then changing a single property somewhere will transfer the whole hierarchy as JSON to the client again. For use-cases where the map is frequently updated that can lead to a lot of data being transfered
- when reusing objects, such as marker icons, on the Java side, the client will still create separate OL icon instance for each usage. Especially for icons, this can be problematic, as OL stores the image data for each icon in a buffer, meaning that using the same image a 100 times will create 100 image buffers

This change tackles these issues by introducing two new concepts:
- Dirty-checking: On the Java side, configuration objects now also track whether they have been changed (marked "dirty") since the last synchronization, and they keep track of nested configuration objects. That allows to walk through the hierarchy and only collect the instances that have changes
- Client-side lookup for OL instances: On the client side, the component creates a lookup (using a JS Map) to keep track of every single OL instance used by the component. This allows to synchronize every configuration object individually, and it also allows to retrieve existing instances. The lookup will by default keep only weak references to these instances, so that these can still be garbage collected by the browser if they are not needed anymore. If the browser does not support weak references, a fallback implementation is used that does not release references. Currently all major browsers, except Opera should support [weak references](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef).

Fixes https://github.com/vaadin/flow-components/issues/2548
Fixes https://github.com/vaadin/flow-components/issues/2549


vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/AbstractConfigurationObject.java
```diff
@@ -45,6 +51,29 @@ public void setId(String id) {
 
     public abstract String getType();
 
+    protected void addChild(AbstractConfigurationObject configurationObject) {
+        children.add(configurationObject);
```
sissbruecker, line 55: In order to retrieve changes from the whole hierarchy, we also need to keep track of each reference that is used in an object. Maybe addReference / removeReference might be a better name.

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/AbstractConfigurationObject.java
```diff
@@ -84,4 +119,13 @@ public void update(Runnable updater, boolean notifyChanges) {
             this.notifyChanges.remove();
         }
     }
+
+    public void collectChanges(
+            Consumer<AbstractConfigurationObject> changeCollector) {
+        children.forEach(child -> child.collectChanges(changeCollector));
```
sissbruecker, line 125: Ensures that nested references are synced first, so that they are available when synchronizing this instance.

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/Configuration.java
```diff
@@ -39,6 +42,8 @@ public String getType() {
      *
      * @return the list of layers managed by this map
      */
+    @JsonIdentityInfo(generator = ObjectIdGenerators.PropertyGenerator.class, property = "id")
+    @JsonIdentityReference(alwaysAsId = true)
```
sissbruecker, line 46: All nested objects are now just serialized as IDs. Since we sync every object separately and then just wire up the references, we don't need the full object anymore. In fact it would be wasteful to use the full object, as we would include some objects several times in the JSON that is sent to the client.

vaadin-map-flow-parent/vaadin-map-flow/src/main/resources/META-INF/resources/frontend/vaadin-map/synchronization/index.js
```diff
@@ -73,12 +73,8 @@ function synchronizeFeature(target, source, context) {
     target = new Feature();
   }
 
-  target.setGeometry(
-    context.synchronize(target.getGeometry(), source.geometry, context)
-  );
-  target.setStyle(
-    context.synchronize(target.getStyle(), source.style, context)
-  );
+  target.setGeometry(context.lookup.get(source.geometry));
```
sissbruecker, line 76: Nested objects are now synchronized separately, so all we need to do here is just get their reference from the lookup.

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/AbstractConfigurationObject.java
```diff
@@ -45,43 +51,77 @@ public void setId(String id) {
...
     }
 
-    public void removePropertyChangeListener(PropertyChangeListener listener) {
-        this.propertyChangeSupport.removePropertyChangeListener(listener);
+    protected void deepMarkAsDirty() {
+        if (!trackObjectChanges.get())
+            return;
+        dirty = true;
+        children.forEach(AbstractConfigurationObject::deepMarkAsDirty);
     }
```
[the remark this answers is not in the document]
DiegoCardoso: We can just document it for now, so it is clear that it was intended to work like that.

vaadin-map-flow-parent/vaadin-map-flow/src/main/resources/META-INF/resources/frontend/vaadin-map/synchronization/util.js
```diff
@@ -21,27 +21,26 @@ export function convertToSizeArray(size) {
...
+    collection
       .getArray()
-      .find((item) => item.id === jsonItem.id);
-    const syncedItem = options.synchronize(existingItem, jsonItem, options);
-    // Add item if it didn't exist before
-    if (!existingItem) {
-      collection.insertAt(index, syncedItem);
-    }
-  });
+      .some((existingItem, index) => existingItem.id !== updatedIds[index]);
```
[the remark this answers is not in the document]
DiegoCardoso: Ok, thanks for the explanation. For some reason, I ignored the `!==` sign. 🤦🏻

DiegoCardoso (approved):
LGTM 🎉

PR #2690, opened 2022-02-17
perf: preload visible children for expanded tree grid items

### LINKED ISSUE #2648: [grid] Implement TreeGrid eager fetch mode
Change the logic of sending requested ranges to the client, so that flow counterpart checks if children are expanded before sending the parent items to the client.
If some of the items are expanded flow counterpart prepares and sends its children within the same roundtrip.
It should be applied for both the initial rendering and on scrolling.
Ranges should be sent considering the viewport capacity, so extra ranges are not sent if the viewport is already fulfilled.

tomivirkki:
Closes #2648 
Fixes https://github.com/vaadin/flow-components/issues/2713

This PR optimizes the `TreeGrid` to "preload" child item data aiming to cover the whole viewport with items ideally on a single roundtrip. The optimization affects the first load as well as any subsequent request initiating from the Web Component (for example, due to scrolling).

The current order of events:
1. Client requests some items
2. Servers sends the items to the client
3. Client notices one of the items in the viewport is expanded and request child items for it
4. Server sends child items for the expanded item
...
This loop continues level-by-level until the viewport is fully covered resulting in bad UX and unnecessary data transfer:

https://user-images.githubusercontent.com/1222264/154704350-d4ea6adc-b386-452c-882c-22d1160eac81.mp4

After the change:
1. Client requests some items
2. Servers prepares to send the items to the client (includes them in the response)
3. _Server_ notices one of the items in the viewport is expanded and includes its children in the response
4. This loop continues level-by-level until the viewport is fully covered and only then the response is sent back to the client

With enough child item data to cover the user's current viewport the client is able to synchronously render the full hierarchy and no additional data requests are necessary:

https://user-images.githubusercontent.com/1222264/154704376-f3b433c9-9f3e-4618-9f4f-0adc67c2a5d2.mp4

### Reviewer notes:

- You can pass `expandedRootIndexes`, `sortDirection`, `nodesPerLevel` and `depth` query parameters to the IT test view. Example:
`http://localhost:8080/vaadin-grid/treegrid-preload/?expandedRootIndexes=0,2&sortDirection=ascending`


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -372,14 +368,23 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
...
+              (firstIndex, size) => grid.$connector.beforeParentRequest(firstIndex, size, params.parentItem.key),
+              page,
+              parentUniqueKey
+            );
           }
-          grid.$connector.fetchPage(
-            (firstIndex, size) => grid.$connector.beforeParentRequest(firstIndex, size, params.parentItem.key),
-            page,
-            parentUniqueKey
-          );
```
tomivirkki, line 382: No data request should take place if cached data is already found. Moved this inside the else block where it belongs.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -372,14 +368,23 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
             page = Math.min(page, Math.floor(cache[parentUniqueKey].size / grid.pageSize));
 
             callback(cache[parentUniqueKey][page], cache[parentUniqueKey].size);
+            
+            // Flush after the callback to have the grid rows up-to-date
+            updateAllGridRowsInDomBasedOnCache();
+            // Prevent sub-caches from being created (& data requests sent) for items
+            // that may no longer be visible
+            ensureSubCacheQueue = [];
```
tomivirkki, line 376: During the rendering cycle, some expanded items that eventually end up outside the viewport may temporarily be inside the viewport. These lines are essential in making sure no unnecessary data requests get made once the hierarchy is completely rendered.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -136,10 +136,6 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
           itemkey: grid.getItemId(targetCache.items[scaledIndex]),
           level: targetCache.getLevel()
         });
-        // sort by ascending scaledIndex and level
-        ensureSubCacheQueue.sort(function(a, b) {
-          return a.scaledIndex - b.scaledIndex || a.level - b.level;
-        });
```
tomivirkki, line 142: This sorting logic didn't really add up: `scaledIndex` is an item's index under its own level/cache, so using it for sorting universally doesn't really make sense. Removing this invalid sorting logic is essential for making the preloading feature work.

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/treegrid/it/TreeGridHugeTreeIT.java
```diff
@@ -105,7 +105,6 @@ public void collapsed_subtrees_outside_of_cache_stay_expanded() {
         // assuming buffer to match visible row count
         int assumedCachedSize = (grid.getLastVisibleRowIndex()
                 - grid.getFirstVisibleRowIndex()) * 3;
-        waitUntil(b -> grid.getRowCount() >= assumedCachedSize);
```
tomivirkki, line 108: Removed a bad assumption in this failing test case.

web-padawan (commented):
LGTM overall. Tested the IT view with various query parameters and here is the only observation:

https://user-images.githubusercontent.com/10589913/155098353-ffc961d4-e974-4058-8426-3cefe3ed2b24.mp4

When there is only one expanded root index, pressing Sort button triggers several data requests, depending on the index (3 requests in this example with `?expandedRootIndexes=2`, or 4 requests for `?expandedRootIndexes=1`).

When all the root items are collapsed, there are still 2 requests for sorting, so I guess it's how this feature is supposed to work.

DiegoCardoso:
Tested here a bit, and so far the changes look good.

Checked how it behaves with scroll to index. I added `depth=8` to the query parameters, so we could see more items on the grid.

Both tests are calling `grid.scrollToIndex(650)` and running with 4G emulation.

This is running on master:

https://user-images.githubusercontent.com/262432/155150823-fab7367f-d914-40fc-b793-a9a363af6590.mp4

And this is after the changes:

https://user-images.githubusercontent.com/262432/155150928-b7cc9074-79f5-4bef-8381-c4d4dfda6668.mp4

You can see that in the **current implementation** it stops before getting to the desired index. In fact, there's an error on which you cannot scroll past that index (`317`). The grid ends up on a weird state:

https://user-images.githubusercontent.com/262432/155151462-3161aaa4-82ed-40b8-9707-00cb4e50db43.mp4


I noticed that on the current implementation, it is a bit quicker to get to the same level than on the proposed one. I mean, if you check both videos, the first one gets to the 300-ish level quicker than the second one.

tomivirkki:
@DiegoCardoso Can you try the newest revision? It behaves very similarly to the current implementation in this case. 

> You can see that in the current implementation it stops before getting to the desired index. In fact, there's an error on which you cannot scroll past that index (317).

Using `scrollToIndex` is [not recommended](https://vaadin.com/docs/latest/ds/components/tree-grid/#best-practices) with `TreeGrid` because it doesn't work deterministically. If you scroll to end right away, you will end up with smaller final index than what you'd get if you visited expanded children somewhere in the middle.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -367,6 +367,8 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
             // workaround: sometimes grid-element gives page index that overflows
             page = Math.min(page, Math.floor(cache[parentUniqueKey].size / grid.pageSize));
 
+            // Ensure grid isn't in loading state when the callback executes
+            ensureSubCacheQueue = [];
```
DiegoCardoso, line 371: I see the commit message and the change, but I cannot make the connection between them. Could you explain it shortly?

tomivirkki: Sorry, should've added a better explanation for the fix.

There was a [failing test case](https://teamcity.vaadin.com/viewLog.html?buildId=112823&buildTypeId=VaadinFlowComponents_PullRequestValidation) related to recalculating column auto widths. The item hierarchy was preloaded and rendered synchronously (as desired).There's a [check](https://github.com/vaadin/web-components/blob/4c388aeec6623869c70896c909799804fc95d0f9/packages/grid/src/vaadin-grid.js#L538) in `<vaadin-grid>` that prevents running `_recalculateColumnWidths` in case the cache is in loading state. The `gridConnector` [overrides](https://github.com/vaadin/flow-components/blob/8cf147695dbe55a1b4589b62013c14d0ee96d496/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js#L39-L43) `isLoading` to also check the `ensureSubCacheQueue` and in this case it wasn't empty so `_recalculateColumnWidths` would not get invoked.

To fix this, we're emptying the queue here before invoking the callback to make sure `_recalculateColumnWidths` is not blocked by the `ensureSubCacheQueue` array (which gets emptied again right after the callback).

DiegoCardoso: Thanks, it's clear now.

sissbruecker:
Did some testing as well and it looks really solid 👍

Found one issue where I'm not sure if it's expected or not. When scrolling down, it seems that for all nodes that were visible at the scroll position, the first level of children is loaded, regardless if the first node that is fetched has so many descendants that it wouldn't be necessary to load the children of sibling nodes.

See the recording here:

https://user-images.githubusercontent.com/357820/155331068-8c9c959f-8e22-4ed5-9b38-c7355b0ea886.mov

On the first scroll we see that `/0/1/1/1` and `/0/1/1/2` are visible, but do not have their children loaded. Theoretically it should be enough to only load the descendants of `/0/1/1/1`. However on the second scroll, we can see that the direct children of `/0/1/1/2` have also already been loaded. Same happens with sub-sequent scrolls in that it will always load one level for all visible nodes, see how on the third scroll, all of `/0/2/1/0`, `/0/2/1/1`, `/0/2/1/2` already have their children loaded.

sissbruecker:
@tomivirkki Should we also tag https://github.com/vaadin/flow-components/issues/2713 as fixed in the OP?

tomivirkki:
> Found one issue where I'm not sure if it's expected or not. When scrolling down, it seems that for all nodes that were visible at the scroll position, the first level of children is loaded, regardless if the first node that is fetched has so many descendants that it wouldn't be necessary to load the children of sibling nodes.

Yeah this is expected behavior. The logic doesn't really discriminate how many items/hierarchy level get loaded, but relies on the existing (paged) loading behavior. So in case of the of `/0/1`, it only has 3 child items and all of them (`/0/1/1/0`, `/0/1/1/1` amd `/0/1/1/2`) get sent to the client. But of these three, child items are only sent for `/0/1/1/0` etc

> Should we also tag https://github.com/vaadin/flow-components/issues/2713 as fixed in the OP?

Done

sissbruecker (approved):

PR #2692, opened 2022-02-17
feat: add convenience methods for modifying the viewport

### LINKED ISSUE #2682: [map] Add convenience APIs for manipulating the viewport
### Describe your motivation

The map's viewport is defined with an instance of the `View` class, and in order to change the viewport developers have to first get this instance in order to make changes. For example setting the zoom level requires you to do: `map.getView().setZoom(4)`.

The map component could provide shortcut methods to change the viewport, such as `map.setZoom(4)`. These methods would just delegate to the internal view object. We probably don't need to cover all properties of the view, but likely candidates would be: center, zoom, rotation.

sissbruecker:
## Description

Adds convenience methods for setting the center and the zoom level of the viewport to the map, in order to make these easier to use and to discover.

Fixes #2682 

## Type of change

- [x] Feature


DiegoCardoso (approved):
💯

PR #2712, opened 2022-02-23
feat: Enable smoother update of the TreeGrid when using TreeData

TatuLund:
DataCommunicator.reset() will cause collapse - expand cycle and thus flicker when setEnabled(..) is called. In case of TreeData we can avoid this.

This change is neutral with regards integration tests as the final outcome will not change.

Fixes: #2713


tomivirkki:
> The issue is most likely already fixed by https://github.com/vaadin/flow-components/pull/2690

Confirmed. Closing.

PR #2744, opened 2022-03-01
fix: re-enable lazy loading when reducing page size

### LINKED ISSUE #2736: ComboBox doesn't fetch the next page
### Description

For certain ComboBox pageSize values only the first page is fetched

### Expected outcome

After scrolling to the bottom the next page should be fetched and all items should  end up being displayed at some point.

### Actual outcome

Only the first X items are shown, according to the pageSize. But it only happens for specific pageSize values

```
// Bug doesn't happen:

﻿add(comboBox);
﻿
﻿List<String> items = new ArrayList<>();﻿
﻿for(int i=0; i<53; i++) {
    ﻿items.add("Item #" + i);
﻿}
﻿
﻿comboBox.setDataProvider(new ListDataProvider<>(items));
﻿comboBox.setPageSize(20);


// Bug happens:

﻿add(comboBox);
﻿
﻿List<String> items = new ArrayList<>();﻿
﻿for(int i=0; i<50; i++) {
    ﻿items.add("Item #" + i);
﻿}
﻿
﻿comboBox.setDataProvider(new ListDataProvider<>(items));
﻿comboBox.setPageSize(20);
```

### Environment

- Vaadin version: v14.8.4 / v22.0.6

sissbruecker:
## Description

The combo box connector has an internal mode where it disables lazy loading (or rather all data provider requests) if the page size is less than or equal the total number of items. This PR fixes a bug where reducing the page size would not disable that mode.

Fixes #2736 

## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/main/java/com/vaadin/flow/component/combobox/test/LazyLoadingPage.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/main/java/com/vaadin/flow/component/combobox/test/LazyLoadingPage.java
@@ -66,6 +66,8 @@ public class LazyLoadingPage extends Div {
         createCallbackDataProviderWhichReturnsZeroItems();
         addSeparator();
         createComboBoxWithCustomPageSizeAndLazyLoading();
+        addSeparator();
+        createComboBoxWithDisabledLazyLoading();
     }
 
     private void createListDataProviderWithStringsAutoOpenDisabled() {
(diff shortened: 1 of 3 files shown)


tomivirkki (approved):

tomivirkki:
I also tested the new example with a very small page size (2) and it results in a similar issue = no items are loaded after a certain index and you get a warning saying `Attempted to fetch more items from server than allowed in one go: number of items requested '22', maximum items allowed '20'`

This, however, isn't a component bug but rather an issue in DataCommunicator which [limits](https://github.com/vaadin/flow/blob/138fc89d9cd51bc65b4af2439fe7c779429cc460/flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java#L1446) the amount of fetched items depending on the page size used in the component.

PR #2766, opened 2022-03-03
perf: improve tree grid viewportRemaining reset

tomivirkki:
Follow up for https://github.com/vaadin/flow-components/pull/2690

The previous logic using `UI.getCurrent().beforeClientResponse` wasn't ideal for the purpose of resetting the `viewportRemaining` once / request. Changed to use and compare current `VaadinRequest` instead. This cuts down `DataProvider` access.

Example page: http://localhost:8080/vaadin-grid/treegrid-preload?expandedRootIndexes=0
Before:
Screenshot 2022-03-03 at 14 22 59

After:
Screenshot 2022-03-03 at 14 22 29


sissbruecker:
This definitely reduces the number of data provider requests, and I could not find any issues so far.

However one thing I noticed is that #2690 in general doubles the amount of data provider requests (now in one round-trip, previously over multiple round-trips). For each parent node, we load the children from the data provider to check if we should include them. Then the data communicator itself requests these children again from the data provider. When using in-memory data that might not be an issue. But if you use a database as backend that would increase the load by a factor of 2, unless you have some caching in between. I'm a bit worried that this might introduce a performance regression.

Ideally the data communicator would not need to fetch the same data a second time. For example, the data communicator would have an API that does the same as our `recursiveSetParentRequestedRange`, or the data provider might be wrapped with a cache (per request?), though it would probably be more complex than that.

tomivirkki:
> For each parent node, we load the children from the data provider to check if we should include them. Then the data communicator itself requests these children again from the data provider.

I have a patch for this locally, wanted to keep these performance tweaks separate for minimal PRs.

sissbruecker:
> I have a patch for this locally, wanted to keep these performance tweaks separate for minimal PRs.

Awesome, looking forward to see how you managed to solve that.

sissbruecker (approved):

PR #2776, opened 2022-03-04
perf: avoid duplicate requests in tree grid item preloading

tomivirkki:
Follow up for #2690
Addresses https://github.com/vaadin/flow-components/pull/2766#issuecomment-1058285846

This PR removes the [`recursiveSetParentRequestedRange`](https://github.com/vaadin/flow-components/blob/ee484322e319e7f9312fb1fa3d104e789fb9af61/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java#L176) method which was causing duplicate dataProvider requests to occur as described [here](https://github.com/vaadin/flow-components/pull/2766#issuecomment-1058285846). The updated item preloading implementation doesn't manually request for items at all but instead reuses those passed to the `TreeGridUpdateQueue`.

Example page: http://localhost:8080/vaadin-grid/treegrid-preload?expandedRootIndexes=0
Before:
Screenshot 2022-03-03 at 14 22 29

After:
Screenshot 2022-03-04 at 11 19 24


vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/treegrid/it/TreeGridExpandDataRequestIT.java
```diff
@@ -17,14 +17,6 @@ public void before() {
 
         setupTreeGrid();
         findElement(By.id("LoggingDataProvider")).click();
-
```
tomivirkki, line 20: Reverted https://github.com/vaadin/flow-components/commit/f81f79e2c133191c660687428a1f662b4f27ef4b#diff-aa65b3cdd8b798c2156938985958fc695fac474a1cf728b0e633fcb4b4f7acaa since it's no longer valid after this update

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -139,6 +140,7 @@ public void commit(int updateId, String parentKey, int levelSize) {
         // Approximated size of the viewport. Used for eager fetching.
         private final int EAGER_FETCH_VIEWPORT_SIZE_ESTIMATE = 40;
         private int viewportRemaining = 0;
+        private List<JsonValue> queuedParents = new ArrayList<>();
```
sissbruecker, line 143:
```suggestion
        private final List<JsonValue> queuedParents = new ArrayList<>();
```
Since this is cleared instead of setting a new instance, it can be final. `updateQueueFactory` above could also be made final.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -155,56 +157,37 @@ public TreeGridUpdateQueue startUpdate(int sizeChange) {
...
-
-                if (viewportRemaining > 0 && isExpanded(child)) {
-                    int childLength = Math.max(
-                            EAGER_FETCH_VIEWPORT_SIZE_ESTIMATE, getPageSize());
-
-                    // There's still room left in the viewport and the child is
-                    // expanded. Set parent requested range for the child.
-                    getDataCommunicator().setParentRequestedRange(0,
-                            childLength, child);
+                        break;
```
sissbruecker, line 184: Got stuck a bit on this part until I realized that requesting children triggers another while loop, and that would then just keep processing the parent siblings that might be left in the queue. Maybe we could add some explanation here.

There is one scenario where this fails, which is if the parent is expanded, but has no children. I'm not 100% certain how the data provider contract works, but I can reproduce it by overriding the following methods in the custom `LazyHierarchicalDataProvider` in `TreeGridPreloadPage`, and then opening http://localhost:8080/vaadin-grid/treegrid-preload?expandedRootIndexes=1:
```java
                    @Override
                    public boolean hasChildren(HierarchicalTestBean item) {
                        if(item.getId().equals("/0/1/1/0")) {
                            return true;
                        }
                        return super.hasChildren(item);
                    }

                    @Override
                    public int getChildCount(HierarchicalQuery<HierarchicalTestBean, Void> query) {
                        if(query.getParentOptional().isPresent() && query.getParent().getId().equals("/0/1/1/0")) {
                            return 0;
                        }
                        return super.getChildCount(query);
                    }

                    @Override
                    public Stream<HierarchicalTestBean> fetchChildren(HierarchicalQuery<HierarchicalTestBean, Void> query) {
                        if(query.getParentOptional().isPresent() && query.getParent().getId().equals("/0/1/1/0")) {
                            return Stream.of();
                        }
                        return super.fetchChildren(query);
                    }
```

That leads to multiple round-trips then:

Bildschirmfoto 2022-03-04 um 13 33 22

That wouldn't be a huge problem, but maybe we can avoid that somehow.

tomivirkki: > Maybe we could add some explanation here.

Added the documentation

> There is one scenario where this fails, which is if the parent is expanded, but has no children

I guess this would be quite an edge case and multiple roundtrips wouldn't be too serious of an issue if this happens to occur. Nonetheless, added a check for `hasChildren`.

tomivirkki: > Nonetheless, added a check for hasChildren.

After some more thinking, decided to revert this. Depending on how `hasChildren` is implemented in the `DataProvider`, an additional server roundtrip on a very rare occasion is probably a lesser evil than running `hasChildren` for each child item every time.

sissbruecker: Seems reasonable, let's keep it like that.

sissbruecker (approved):

PR #2778, opened 2022-03-04
fix: remove server-side modality on outside click

sissbruecker:
## Description

Fixes Dialog to remove server-side modality when closing the dialog on outside click, in case it has been added manually to the UI. 

Fixes a regression from #2608

## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-dialog-flow-parent/vaadin-dialog-flow-integration-tests/src/main/java/com/vaadin/flow/component/dialog/tests/ModalityDialogsPage.java
+++ b/vaadin-dialog-flow-parent/vaadin-dialog-flow-integration-tests/src/main/java/com/vaadin/flow/component/dialog/tests/ModalityDialogsPage.java
@@ -37,22 +37,31 @@ public class ModalityDialogsPage extends Div {
 
         Dialog nonModalDialog = setupNonModalDialog();
 
-        NativeButton modalButton = new NativeButton("Open modal dialog",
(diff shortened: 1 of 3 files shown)


tomivirkki (approved):

web-padawan (approved):

jouni:
The issue seems to persist at https://vaadin.com/docs/latest/ds/components/dialog

tomivirkki:
@jouni I tested this yesterday and saw the issue was still there even though `23.0.1` was deployed.

Started investigating it just now and the issue seems to be gone for some reason even though there have been no changes to docs. Can you confirm?

jouni:
Still a problem with the Java examples. Tested in Safari and Firefox, tried to "force refresh" to make sure cache was not affecting it.

https://user-images.githubusercontent.com/66382/157638174-e7a57c66-2662-4d3e-9578-b25554dce001.mov

tomivirkki:
Dang, I wonder if I was accidentally testing the TS version after all :/

tomivirkki:
Copied the example source to a local project and the issue doesn't produce:

https://user-images.githubusercontent.com/1222264/157665053-ac0ef718-ee78-4966-8564-5da57b3fd742.mp4

I wonder if this is an embedded UI -only issue

tomivirkki:
Filed the issue here https://github.com/vaadin/flow/issues/13275

PR #2788, opened 2022-03-04
refactor: remove avatar demo

tomivirkki:
Part of #193


web-padawan (changes requested):
ITs are failing because of removed image resources:

> java.lang.AssertionError: Received error message in browser log console right after opening the page, message: [2022-03-04T15:17:38.552Z] [SEVERE] http://172.16.1.48:8080/VAADIN/dynamic/resource/0/34fcdb4f-b8ca-4699-823c-54218ae50a49/user%2B.png - Failed to load resource: the server responded with a status of 500

tomivirkki:
> ITs are failing because of removed image resources:

Thanks, so it seems. Restored the image.

PR #2824, opened 2022-03-08
refactor: use SecureRandom instead of Random on iron-list view

DiegoCardoso:
To make SonarCloud happy, this PR removes the usage of `Random` on `IronListView`.


tomivirkki (approved):

vaadin-virtual-list-flow-parent/vaadin-virtual-list-flow-integration-tests/src/main/java/com/vaadin/flow/component/virtuallist/tests/VirtualListViewPage.java
```diff
@@ -46,7 +47,7 @@
 
     private static final List<String> LIST_OF_BOOKS;
     static {
-        Faker faker = Faker.instance(new Random(42));
+        Faker faker = Faker.instance(new SecureRandom());
```
tomivirkki, line 50: nit: not using a seed makes the content different on every pageload. Probably not an issue in this case.

https://user-images.githubusercontent.com/1222264/157391699-81f610cd-c555-4809-908c-3e55adc71214.mp4

DiegoCardoso: Yeah, I noticed that. Not a big deal on this case, as we don't use any visual tests here.

But I also noticed that the seed doesn't have the same effect on `SecureRandom` as it has no `Random`:
image

`SonarCloud` complained about the seed as well (with a warning on this case)

PR #2846, opened 2022-03-11
fix: In getBBox, return 0 height if 0 width

aroneous:
Returning a non-0 height for an empty text box (which should consume no
space) causes Highcharts to reserve vertical space in the chart.

## Description

When exporting charts with the `SVGGenerator`, vertical space was being allocated above and below the plot area to hold title, subtitle, and caption, even if no text was set for these elements.

This was due to the implementation of `getBBox` in `jsdom-exporter.js`: even for text boxes with 0 width, it would return a non-zero height (based on heuristics of the font size associated with the element). This appeared to cause Highcharts to allocate vertical space, even though no text would be rendered.

Fix is to return a height of 0 if the algorithm computes a width of 0.

Fixes https://github.com/vaadin/flow-components/issues/2907

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


sissbruecker:
Updated a comment to mention the zero-height behavior, and created and linked to an issue for this specific bug.

sissbruecker (approved):

Issue #2870, opened 2022-03-15
ComboBoxElement#selectByText fails if one navigates in value change event


TatuLund:
If I have a TestBench test of a ComboBox, that happens to be my navigation menu, the selectByText fails and throws StaleElementReferenceException as ComboBox is already detached due navigation when ComboBoxElement attempts to close the popup. See test case and test ui below to reproduce.

```
public class ComboBoxNavigationIT extends AbstractViewTest {

    public ComboBoxNavigationIT() {
        super("combobox-navigation");
    }

    @Test
    public void comboBoxNavigationTest() {
        ComboBoxElement combo = $(ComboBoxElement.class).first();
        combo.selectByText("Four");
        findElement(By.id("dummy-view"));
    }
}

```
```
@Route("combobox-navigation")
public class ComboBoxNavigationView extends Div {

    public ComboBoxNavigationView() {
        ComboBox<String> comboBox = new ComboBox<>();
        comboBox.setItems("One", "Two", "Three", "Four");
        comboBox.addValueChangeListener(event -> {
            UI.getCurrent().navigate("combobox-navigation-out");
        });

        add(comboBox);
    }

    @Route("combobox-navigation-out")
    public static class DummyView extends Div {
        public DummyView() {
            setId("dummy-view");
            setText("Nothing here");
        }
    }
}
```

sissbruecker:
After discussing this in the team we think that the root problem here is the actual implementation of the `ComboBoxElement.selectByText` method, which basically "programs" the web component to change the value. This in turn leads to the combo box not closing automatically, and the subsequent call to close the popup leads to the exception.

It would be more appropriate if the method would use actual user-inputs, such as entering the value to select, and then pressing enter, which would close the dropdown automatically.

PR #2871, opened 2022-03-15
fix: Catch StaleElementReferenceException in selectByText

TatuLund:
Fixes https://github.com/vaadin/flow-components/issues/2870


tomivirkki:
> Fixes: https://github.com/vaadin/flow-components/issues/2870

Please use format "Fixes #1234" or "Closes #1234" in the description to make Github & Zenhub automation work properly. See GitHub documentation for the feature. https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

sissbruecker:
I don't consider using the combo box as a navigation menu a common use-case, and as such I don't think we need to add logic to let `ComboBoxElement` handle this case. Closing the PR. As an alternatively you can catch this exception in your test, or write a custom `selectByText` implementation that does not close the popup.

Issue #2691, opened 2022-02-17
DatePicker triggers `value-changed` when attached


probert94:
### Description

In an application I am working on, I use `UI#push` to force an update of the client site state (in my case I want to show a loading indicator). Now I noticed, that the value of displayed `DatePicker` is wrongly set to `null`.
I found that this is caused by a `ValueChangeEvent` coming from the client.
After some debugging, I found that a `value-changed` event is triggered during the `ready` of the webcomponent, because the value changed from `undefined` to `""`.
If the value on the server is still the same at this time, no `ValueChangeEvent` will be triggered but if the value on the server changed in the mean time, a `ValueChangeEvent` with `fromClient=true` is triggered and the server side value is changed to `null`.
This can even result in a difference between the client side value and the server side value.

The `value-changed` event on the client also happens for `TextField` (and probably other fields as well) but the server side `ValueChangeEvent` only seems to happen with `DatePicker`.

### Expected outcome

There should be no `ValueChangeEvent` with `fromClient=true` unless the user manually changed the value on the client side.

### Actual outcome

There is an initial `value-changed` event on the client, which might create a `ValueChangeEvent` on the server, if the server value changed in the mean time.

### Live Demo (optional)

Notification that shows, when the server receives a `ValueCHangeEvent` from client:
![image](https://user-images.githubusercontent.com/18258317/154485610-2975660e-39b2-4a3e-9881-41d51a14ecf5.png)
It triggers immediately after opening the page, basically as soon as the `Datepicker` is attached.

Pressing the `Show values` button shows the server side values:
![image](https://user-images.githubusercontent.com/18258317/154485790-c66d8c5d-6950-4830-a5ba-b91ce0a7880f.png)
The date on the server is now `null`, while the client shows the expected value (the current date).

### Minimal reproducible example
```
  private DatePicker dp;
  private TextField tf;

  public EmptyView() {
      dp = new DatePicker();
      dp.addValueChangeListener(e -> {
          if (e.isFromClient())
              Notification.show("Date change from client: " + e.getValue());
      });
      tf = new TextField();
      tf.addValueChangeListener(e -> {
          if (e.isFromClient())
              Notification.show("Text change from client: " + e.getValue());
      });
      Button btn = new Button("Show values", e -> {
         Notification.show("Date: " + dp.getValue());
         Notification.show("Text: " + tf.getValue());
      });
      add(dp, tf, btn);
  }

  @Override
  protected void onAttach(AttachEvent attachEvent) {
[...]

PR #2878, opened 2022-03-16
fix: ensure DatePicker initializes client side value property

vursen:
## Description

Fixes DatePicker such that to make it initialize the web component's `value` property with an empty string rather than leave it to be `null` when DatePicker is being instantiated with no value. This is needed to prevent the `value-changed` event on the server side which happens as a result of Polymer converting `null` to `""`.

It pretty much reproduces the fix made for Select a while ago in https://github.com/vaadin/flow-components/commit/14676def81188ec9a674bfe202ebac8d71b65067#diff-1c027c9076652ce3a2c641ac7407c10cc3a87846a37ee152c27aa9fe59aad609R126-R129

The same fix will be made for TimePicker, but in a separate PR.

Part of #2691

## Type of change

- [x] Bugfix

## Checklist

### DIFF
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
@@ -83,10 +83,6 @@ public class DatePicker extends GeneratedVaadinDatePicker<DatePicker, LocalDate>
      */
     public DatePicker() {
         this((LocalDate) null, true);
-        // Trigger model-to-presentation conversion in constructor, so that
-        // the client side component has a correct initial value of an empty
-        // string
-        this.setPresentationValue(null);
     }
 
     /**
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
@@ -83,6 +83,10 @@ public class DatePicker extends GeneratedVaadinDatePicker<DatePicker, LocalDate>
      */
     public DatePicker() {
         this((LocalDate) null, true);
+        // Trigger model-to-presentation conversion in constructor, so that
+        // the client side component has a correct initial value of an empty
+        // string
+        this.setPresentationValue(null);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
```diff
@@ -83,6 +83,10 @@
      */
     public DatePicker() {
         this((LocalDate) null, true);
+        // Trigger model-to-presentation conversion in constructor, so that
+        // the client side component has a correct initial value of an empty
+        // string
```
sissbruecker, line 88: Can't remember anymore if/why this was an appropriate wording for Select, but maybe we should rather hint at the fact that this avoids the initial value-changed event from the WC?

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
```diff
@@ -83,6 +83,10 @@
      */
     public DatePicker() {
```
sissbruecker, line 84: Seeing how the constructor in `GeneratedDatePicker` only calls `setPresentationValue` if the initial value is not null, there are several constructor variants where `setPresentationValue` still might not be called:

For example the following test fails:
```java
    @Test
    public void initialNullValue_valuePropertyHasEmptyString() {
        // Assume we load a nullable LocalDate value from somewhere and initialize the date picker with it
        LocalDate initialDateFromDatabase = null;
        DatePicker picker = new DatePicker((LocalDate) null);
        Assert.assertNull(picker.getValue());
        Assert.assertEquals("", picker.getElement().getProperty("value"));
    }
```

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/GeneratedVaadinDatePicker.java
```diff
@@ -1217,9 +1217,7 @@ protected void removeAll() {
     public GeneratedVaadinDatePicker(T initialValue, T defaultValue,
```
vursen, line 1217: This method doesn't appear to be used anywhere, for note.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
vursen: Good catch, updated the PR.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
vursen: Decided to fix the issue in `Vaadin Generated DatePicker`. The comment is no longer necessary as the fix looks natural there. Let me know if you think it is still needed.

sissbruecker (approved):

PR #2883, opened 2022-03-17
fix: ensure TimePicker initializes client side value property

vursen:
## Description

Fixes TimePicker so that it initializes the web component's `value` property with an empty string rather than leaves it to be `null` when TimePicker is being instantiated with no value. This is needed to prevent the `value-changed` event on the server side which comes from the client side as a result of Polymer converting `null` to `""`.

Part of #2691

## Type of change

- [x] Bugfix

## Checklist


vaadin-time-picker-flow-parent/vaadin-time-picker-flow/src/main/java/com/vaadin/flow/component/timepicker/GeneratedVaadinTimePicker.java
```diff
@@ -868,8 +868,13 @@ protected Registration addInvalidChangeListener(
             boolean isInitialValueOptional) {
         super("value", defaultValue, elementPropertyType, presentationToModel,
                 modelToPresentation);
+        // Only apply initial value if the element does not already have a value
+        // (this can be the case when binding to an existing element from a Lit
+        // template), or if isInitialValueOptional enforces setting the initial
+        // value, which is the case when calling a DatePicker constructor with a
```
web-padawan, line 874:
```suggestion
        // value, which is the case when calling a TimePicker constructor with a
```

PR #2919, opened 2022-03-22
feat: forward ConfirmDialog class name to overlay

### LINKED ISSUE #2909: Update ConfirmDialog to fix usage of HasStyle
## Description

Currently, `ConfirmDialog` implements `HasStyle` but its methods do not work.
The reason is that CSS classes need to be set on the overlay, as described in #2856

Part of https://github.com/vaadin/platform/issues/2801

## Proposed solution

1. Copy CSS classes from the component to the overlay
2. Override methods for setting styles as unsupported

web-padawan:
## Description

Added logic to copy `className` from `vaadin-confirm-dialog` to its overlay element for styling.

Fixes #2909

## Type of change

- Feature

### DIFF
--- a/vaadin-confirm-dialog-flow-parent/vaadin-confirm-dialog-flow/src/main/java/com/vaadin/flow/component/confirmdialog/ConfirmDialog.java
+++ b/vaadin-confirm-dialog-flow-parent/vaadin-confirm-dialog-flow/src/main/java/com/vaadin/flow/component/confirmdialog/ConfirmDialog.java
@@ -16,6 +16,7 @@ package com.vaadin.flow.component.confirmdialog;
  * #L%
  */
 
+import com.vaadin.flow.component.AttachEvent;
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.ComponentEvent;
 import com.vaadin.flow.component.ComponentEventListener;
@@ -43,6 +44,7 @@ import com.vaadin.flow.shared.Registration;
 @NpmPackage(value = "@vaadin/confirm-dialog", version = "23.0.2")
 @NpmPackage(value = "@vaadin/vaadin-confirm-dialog", version = "23.0.2")
 @JsModule("@vaadin/confirm-dialog/src/vaadin-confirm-dialog.js")
+@JsModule("./confirmDialogConnector.js")
 public class ConfirmDialog extends Component
         implements HasSize, HasStyle, HasOrderedComponents {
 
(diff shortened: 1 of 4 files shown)


sissbruecker:
Instead of introducing the connector, I wonder if we shouldn't rather attempt to solve this in the WC. An alternative might be:
- introduce something like `overlayClass`, `overlayStyle` properties to the dialog components
- have the WCs delegate the property values to `overlay.class`, `overlay.style`
- in the Flow component override the `HasStyle` implementation to use the delegate properties

That would bring the additional benefit that developers using the plain WC, or Hilla, would also have a simple API to modify classes / styles for the overlay.

krissvaa:
If class is added when creating dialog and then the dialog is opened without attaching, the class is missing from overlay.
Screenshot 2022-03-22 at 17 44 18

https://user-images.githubusercontent.com/101052651/159523293-c1289320-5194-400b-8d44-d0655a39b6de.mp4

web-padawan:
> If class is added when creating dialog and then the dialog is opened without attaching, the class is missing from overlay.

Good catch, fixed. Also updated the test intended to cover this to make it actually work as expected.

vaadin-confirm-dialog-flow-parent/vaadin-confirm-dialog-flow/src/main/java/com/vaadin/flow/component/confirmdialog/ConfirmDialog.java
```diff
@@ -126,6 +128,17 @@ public void updateHeight() {
...
+        super.onAttach(attachEvent);
+        initConnector();
+    }
+
+    private void initConnector() {
+        getElement().executeJs(
+                "window.Vaadin.Flow.confirmDialogConnector.initLazy(this)");
+    }
+
     private boolean autoAddedToTheUi;
```
krissvaa, line 143: As `getStyle()` is not supported from `HasStyle` with this implementation then probably it should be documented

```suggestion
/**
     * @throws UnsupportedOperationException
     *             Adding styles to dialog overlay are not supported
     */
    @Override
    public Style getStyle() {
        throw new UnsupportedOperationException();
    }
```

web-padawan: Done.

DiegoCardoso (approved):

PR #2928, opened 2022-03-23
refactor!: remove fluent option APIs from Map

### LINKED ISSUE #2849: [map] Remove fluent APIs
### Describe your motivation

Currently the Map component uses a fluent API for building options. While it works, it has some downsides for cases where option classes need to be extended. For example the option APIs are split into an abstract `BaseOptions` class used for sub-classing and then a non-abstract `Options` for actual use ([example](https://github.com/vaadin/flow-components/blob/master/vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/source/XYZSource.java#L72-L84)). Also constructors have to use the `BaseOptions` variant, which is confusing for API discoverability, as `BaseOptions` is not usable by developers ([example](https://github.com/vaadin/flow-components/blob/master/vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/source/XYZSource.java#L35)).
There is also a matter of consistency, where we try to avoid adding fluent APIs because it seems that was tried before, and then removed afterwards (see https://github.com/vaadin/flow/pull/2147, https://github.com/vaadin/flow/issues/486).

For now we easiest thing we can do to solve these issues is to simply remove the fluent API usage in the Map component. This should be fine as it is still in experimental state and breaking changes are acceptable. The current uses of a fluent API are:
- `ImageStyle.BaseOptions` and sub-classes (https://github.com/vaadin/flow-components/blob/master/vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/style/ImageStyle.java#L115)
- `Source.BaseOptions` and sub-classes (https://github.com/vaadin/flow-components/blob/master/vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/source/Source.java#L103)

We can consider adding a fluent API again at a later point if we find better patterns around its usage.

sissbruecker:
Removes all fluent option APIs from the Map component. The fluent APIs required some workarounds for sub-classing, which were hard to follow and caused DX concerns. We will reconsider adding fluent APIs if better patterns emerge for implementing them in class hierarchies.

Closes #2849


vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/source/VectorSource.java
```diff
@@ -63,10 +63,6 @@ public void removeFeature(Feature feature) {
         removeChild(feature);
     }
 
-    protected static class BaseOptions<T extends BaseOptions<T>>
-            extends Source.BaseOptions<T> {
-    }
-
-    public static class Options extends BaseOptions<Options> {
+    protected static class Options extends Source.Options {
```
vursen, line 66: Looks like a typo was made here:

```suggestion
    public static class Options extends Source.Options {
```

vursen (changes requested):

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/source/VectorSource.java
vursen: I would also suggest adding a couple of tests for `VerctorSource`. It seems there are no tests for it at all.

sissbruecker: I've fixed the visibility of `VectorSource.Options`, and also updated the visibility of options for abstract classes to protected, as discussed.

Testing this is not as straightforward. It's common practice to put unit tests into the same package as the source file, which also means that the test has access to protected internals of the tested class. I think it would be a bit overblown to add a separate unit test for ensuring the visibility of the options classes, and discovering errors like these is why we have code reviews. However I added missing tests for ensuring that the individual source classes correctly adopt their properties from the options.

vursen: I meant testing only public classes such as `VectorSource.Options` (if it should be a public class after all).

vursen (approved):

PR #2937, opened 2022-03-25
chore: Improve Javadocs 

tarekoraby:
The new descriptions of the components are mostly taken from their descriptions in the [docs](https://vaadin.com/docs/latest/ds/components). 

For example, instead of the Button docs saying
```
/**
 * Server-side component for the <code>vaadin-button</code> element.
 */
```

It will be changed to say
```
/**
 * The Button component allows users to perform actions. It comes in several
 * style variants, and supports icons in addition to text labels.
 */
```

### DIFF
--- a/vaadin-accordion-flow-parent/vaadin-accordion-flow/src/main/java/com/vaadin/flow/component/accordion/Accordion.java
+++ b/vaadin-accordion-flow-parent/vaadin-accordion-flow/src/main/java/com/vaadin/flow/component/accordion/Accordion.java
@@ -35,8 +35,25 @@ import com.vaadin.flow.component.dependency.NpmPackage;
 import com.vaadin.flow.shared.Registration;
 
 /**
- * Accordion is a a vertically stacked set of expandable panels. Only one panel
- * can be opened at a time.
+ * Accordion is a vertically stacked set of expandable panels. It reduces
+ * clutter and helps maintain the user’s focus by showing only the relevant
+ * content at a time.<br>
+ * <br>
+ *
+ * Accordion consists of stacked panels, each composed of two parts: a summary
+ * and a content area. Only one panel can be expanded at a time.<br>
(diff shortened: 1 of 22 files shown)


tarekoraby:
This PR doesn't include the full list of components yet. But I'd like to see if it's ready to be merged so that I can estimate the amount of effort needed to complete the full list of components.

vaadin-accordion-flow-parent/vaadin-accordion-flow/src/main/java/com/vaadin/flow/component/accordion/Accordion.java
```diff
@@ -35,8 +35,25 @@
 import com.vaadin.flow.shared.Registration;
 
 /**
- * Accordion is a a vertically stacked set of expandable panels. Only one panel
- * can be opened at a time.
+ * Accordion is a vertically stacked set of expandable panels. It reduces
+ * clutter and helps maintain the user’s focus by showing only the relevant
+ * content at a time.<br>
+ * <br>
+ *
```
sissbruecker, line 42:
```suggestion
 * content at a time.
 * <p>
```
For separating JavaDoc into paragraphs we usually use HTML paragraphs. Also no need for blank lines in between.

Please update the other occurrences accordingly.

sissbruecker (changes requested):
Thanks for the effort, this is definitely an improvement.

However I think we should be more careful about consistency, for example there are still several components with a leading introduction of:
```
Server-side component for the {@code vaadin-checkbox} element.
```
I think we can resolve these in some way. I would suggest to take a look at these together, feel free to ping me for a organizing a pairing session.

tarekoraby:
For the record, we agreed to take the task of improving the docs of the rest of the components in a separate PR.

sissbruecker (approved):

Issue #2935, opened 2022-03-25
Overriding DataProvider::getId not enough for Grid Selection


KasparScherrer:
### Description

Sometimes we have an editable Grid for Entities that have an equals/hashCode method that compares all its properties.
For simplicity let's assume a `Person` class with `id`, `firstName` and `lastName` properties. When the user edits the firstName, the hashCode of the person changes, which would lead to issues with the DataProvider. What I can do to avoid these issues is override `DataProvider::getId` to return `person.getId()` instead of the whole person.

However, when the Grid allows multi selection, this is not enough, as the SelectionModel is based on a `LinkedHashSet<T>` which - as the name suggests - is based on the hashCode implementation of the object.

What this means is when I select a Person, edit their firstName, then unselect again, the unselect will no longer work.  **The Person in question will now always remain in the HashSet of Selected Items, as the hashCode is no longer the same.** 
Also, no SelectionEvent is fired anymore for this specific Person. All this is because in the Method **`AbstractGridMultiSelectionModel::deselectFromClient`, the command `selected.remove(item)` does not work and returns false** (again, because of hashCode).

The easy solution for us devs would be to change the `Person::equals` and `Person::hashCode` to only include the id and not all of the properties, but in my case this is not possible because the equals and hashCode are critical for other parts of our logic. I must not touch the equals and hashCode implementations.

So what I had to do was creating a DTO class where i was able to change the equals and hashCode to my liking. But it's not really pretty, as I need to convert my list of received Persons to a List of PersonDTO just for this Vaadin Grid usage.

What I would like is if you could change your (de-)selection logic to use the `DataProvider::getId` method for the HashSet of selected items. Thank you

### Expected outcome
As long as I override `DataProvider::getId` to return something unique and non-editable, I expect user-originated item selections to work even though some properties used in equals/hashCode have changed.

### Actual outcome

User-originated Selections no longer work after editing properties (on a selected item) that are used for equals/hashCode, even though I override `DataProvider::getId`

### Live Demo

![vaadin_grid_selection_issue](https://user-images.githubusercontent.com/13610722/160101570-ebb91f59-d5fa-4d89-937b-6f2a6a31679e.gif)

### Minimal reproducible example

```
@Route(value = "test", layout = MainView.class)
public class TestView extends VerticalLayout {

    public TestView() {
        Grid<Person> personGrid = new Grid<>(Person.class, false);

        personGrid.addComponentColumn(person -> {
            TextField textField = new TextField();
[...]

PR #2947, opened 2022-03-29
fix: ensure GridMultiSelectionModel uses identity given by DataProvider

TatuLund:
- Change to use LinkedHasMap internally
- Refactor unit test to be based on Set instead of Array
- Add unit test for case 2935
- Modify GridViewSelectionIT assert to be order-indenpendent, as value is Set

Fixes: #1476
Fixes: #2935


sissbruecker:
@TatuLund There is a related issue that we want to fix sooner rather than later: https://github.com/vaadin/flow-components/issues/2935. I think long term it would make sense to refactor the multi selection model to not store the selection as a HashSet, but rather use another data structure. Maybe a HashMap, where keys are the IDs provided by the data provider. WDYT?

TatuLund:
>  Maybe a HashMap, where keys are the IDs provided by the data provider. WDYT?

It looks to be quite feasible option. I already drafted the code in my project. I could update the PR.

sissbruecker (approved):

PR #2957, opened 2022-03-30
chore: attempt to make IronListIT more reliable

sissbruecker:
## Description

`IronListIT.listWithComponentRendererWithBeansAndPlaceholder_scrollToBottom_placeholderIsShown` fails randomly, it seems there is a race condition between checking if the list has rendered a placeholder after scrolling, and the Flow request resolving before that with updated data (in which case no placeholders are needed). 

The problem could be that the first check for a placeholder runs immediately after scrolling, at which point the list might not have re-rendered yet. The second check for a placeholder runs 20ms later, at which point the server-side roundtrip for new data could be completed. Attempt to fix this by adding a timeout for the first check, so that the list can re-render after scrolling.

Attempting this fix, because it affects the upgrade PRs for some reason:
- https://github.com/vaadin/flow-components/pull/2955
- https://github.com/vaadin/flow-components/pull/2911


sissbruecker:
At least when running locally, this consistently fixes the test, whereas without the fix the test fails more often than not.

vursen (approved):

vursen (approved):

PR #2962, opened 2022-03-31
chore: bump frontend maven plugin version (14.8)

sissbruecker:
## Description

Bumps the version of `frontend-maven-plugin` that is used in the `vaadin-date-picker-flow` build to be compatible with M1 Macs. The current version of the plugin tries to download the node releases that do not exist, which results in a 404:
```
[INFO] Downloading https://nodejs.org/dist/v14.17.4/node-v14.17.4-darwin-arm64.tar.gz to /Users/sascha/.m2/repository/com/github/eirslett/node/14.17.4/node-14.17.4-darwin-arm64.tar.gz
```

Reference: https://github.com/eirslett/frontend-maven-plugin/issues/952

### DIFF
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/pom.xml
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/pom.xml
@@ -157,7 +157,7 @@
             <plugin>
                 <groupId>com.github.eirslett</groupId>
                 <artifactId>frontend-maven-plugin</artifactId>
-                <version>1.12.1</version>
+                <version>1.11.3</version>
                 <configuration>
                     <nodeVersion>v14.17.4</nodeVersion>
                     <workingDirectory>date-fns</workingDirectory>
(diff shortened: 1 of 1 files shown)


vaadin-date-picker-flow-parent/vaadin-date-picker-flow/pom.xml
```diff
@@ -157,7 +157,7 @@
             <plugin>
                 <groupId>com.github.eirslett</groupId>
                 <artifactId>frontend-maven-plugin</artifactId>
-                <version>1.9.1</version>
+                <version>1.12.1</version>
```
sissbruecker, line 160:
```suggestion
                <version>1.11.3</version>
```

sissbruecker:
Updated to use `1.11.3` instead, since `1.12.x` requires Maven `3.6.x`, which we apparently don't use

vursen:
> that is used in the vaadin-date-picker-flow build

How could I test it? Is it about compiling Flow components?

vursen:
Confirmed that the PR fixes the error. Approved.

vursen (approved):

PR #461, opened 2020-11-27
fix: disable client side validation

DiegoCardoso:
- Disable client side validation
- Validation from Binder will happen upon Binder#validate() or if the field binding uses `withValidator`

fix vaadin/vaadin-radio-button#179

PR #2975, opened 2022-04-04
fix: disable client-side validation for CheckboxGroup

### LINKED ISSUE #2823: Validation message in CheckboxGroup does not appear directly on invalid selection and disappears when focusing out
### Description

Similar to the issue https://github.com/vaadin/vaadin-radio-button/issues/179 that affected RadioButtonGroups, the validation message for CheckboxGroups disappears when focusing out of the component. It also does not appear directly when an invalid selection is preselected after adding a Binder.

### Expected outcome

What did you expect to happen?

The validation message should appear directly if an invalid selection is done. It also should stay on-screen when changing focus from the CheckboxGroup. 

### Actual outcome

The validation message does not appear directly below a CheckboxGroup when an invalid selection is preset. It also disappears when focusing out of the CheckboxGroup.

This is a sample code for the checkbox group.
```java
CheckboxGroup<String> cbg = new CheckboxGroup<>();
cbg.setItems("valid 1", "valid 2", "invalid");

Binder<AtomicReference<Set<String>>> cbgBinder = new Binder<>();
cbgBinder.forField(cbg)
		.withValidator(values -> !values.contains("invalid"), "This selection is invalid!")
		.bind(AtomicReference::get, AtomicReference::set);

cbg.select("invalid");

this.add(cbg, new TextField());
```

The following is a sample reference code for the behaviour of a RadioButtonGroup, which behaves differently:
```java
RadioButtonGroup<String> rbg = new RadioButtonGroup<>();
rbg.setItems("valid 1", "valid 2", "invalid");

Binder<AtomicReference<String>> rbgBinder = new Binder<>();
rbgBinder.forField(rbg)
		.withValidator(value -> !"invalid".equals(value), "Value is invalid")
		.bind(AtomicReference::get, AtomicReference::set);

rbg.setValue("invalid");

this.add(rbg, new TextField());
```

### Steps to reproduce

1. Create a CheckboxGroup and set items .
2. Add a binder that displays an error message when the invalid selection is selected.
3. Programmatically set an invalid selection
4. Change focus to the textfield, e.g. by clicking on it or by tabbing out.

### Environment

- vaadin-checkbox-flow (14.8.1) / Vaadin Flow 14.8.1
- Windows 10

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [ ] Safari (not tested)
- [x] Edge
- [ ] IE 11 (not tested)
- [ ] iOS Safari (not tested)
- [ ] Android Chrome (not tested)

sissbruecker:
## Description

Disables client-side validation for the CheckboxGroup component, by overriding the client-side validation function and removing synchronization of the `invalid` property. The solution is adapted from a similar fix for RadioButtonGroup (https://github.com/vaadin/flow-components/pull/461). Note that the fix in RadioButtonGroup has changed quite a bit over time.

Fixes #2823 

## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-checkbox-flow-parent/vaadin-checkbox-flow-integration-tests/src/test/java/com/vaadin/flow/component/checkbox/tests/CheckboxGroupIT.java
+++ b/vaadin-checkbox-flow-parent/vaadin-checkbox-flow-integration-tests/src/test/java/com/vaadin/flow/component/checkbox/tests/CheckboxGroupIT.java
@@ -191,9 +191,6 @@ public class CheckboxGroupIT extends AbstractComponentIT {
                 "Correct error message should be shown after the button clicks",
                 "Field has been set to invalid from server side",
                 errorMessage.getText());
-
-        executeScript("arguments[0].value=['2'];", group);
-        verifyGroupValid(group, errorMessage);
     }
 
     @Test
--- a/vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/FieldValidationUtil.java
+++ b/vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/FieldValidationUtil.java
@@ -33,8 +33,7 @@ class FieldValidationUtil {
         // Wait until the response is being written as the validation state
         // should not change after that
         component.getUI().ifPresent(ui -> ui.beforeClientResponse(component,
-                        executionContext -> overrideClientValidation(
-                                component)));
+                executionContext -> overrideClientValidation(component)));
     }
 
     private static <T> void overrideClientValidation(
--- a/vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/FieldValidationUtil.java
+++ b/vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/FieldValidationUtil.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2000-2022 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component.checkbox;
+
+class FieldValidationUtil {
+    private FieldValidationUtil() {
+    }
+
+    static <T> void disableClientValidation(CheckboxGroup<T> component) {
+        // Since this method should be called for every time when the component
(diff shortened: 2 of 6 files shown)


vaadin-checkbox-flow-parent/vaadin-checkbox-flow-integration-tests/src/test/java/com/vaadin/flow/component/checkbox/tests/CheckboxGroupIT.java
```diff
@@ -191,9 +191,6 @@ public void groupHasLabelAndErrorMessage_setInvalidShowEM_setValueRemoveEM() {
                 "Correct error message should be shown after the button clicks",
                 "Field has been set to invalid from server side",
                 errorMessage.getText());
-
-        executeScript("arguments[0].value=['2'];", group);
-        verifyGroupValid(group, errorMessage);
```
sissbruecker, line 196: This test had the wrong assumption that setting a value on the client should make the component valid. That's not correct, as long as the component has been set to invalid, it should stay invalid.

vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/FieldValidationUtil.java
```diff
@@ -0,0 +1,54 @@
...
+                    "Component %s is not attached. Client side validation "
+                            + "should be disabled when the component is "
+                            + "attached, thus this method needs to be called "
+                            + "from the onAttach method of the component.",
+                    component));
+        }
+        // Wait until the response is being written as the validation state
+        // should not change after that
+        component.getUI().ifPresent(ui -> ui.beforeClientResponse(component,
+                executionContext -> overrideClientValidation(component)));
```
vursen, line 36: Looking at other similar places where the client-side validation is getting disabled for other components, it seems that they do it slightly differently:

```java
final StateNode componentNode = component.getElement().getNode();
componentNode.runWhenAttached(ui -> ui.getInternals().getStateTree()
        .beforeClientResponse(componentNode,
                executionContext -> overrideClientValidation(
                        component)));
```

Just want to make sure if there is any significant difference.

sissbruecker: Right, I simplified that bit, because the `runWhenAttached` wasn't necessary, as the assertion above checks that the component is already attached. Apart from that accessing the `StateNode` is not necessary, and neither is accessing the internal state tree to use `beforeClientResponse`.

vursen: OK 👍 It would be good to simplify that in other components at some point as well.

vursen:
I quickly looked around and noticed that ComboBox and CustomField might potentionally have the same problem as CheckboxGroup: both don't disable the client-side validation. I'm not sure if CustomField requires the fix right away as I heard it has some fundamental problems with the validation (?). However, as for ComboBox, it is probably worth fixing on a par with CheckboxGroup. 

What do you think?

vursen:
Ok, seems that ComboBox maintains its own way of disabling the client-side validation through the connector:

https://github.com/vaadin/flow-components/blob/b89541a6cec9ce8c0ebdb2e02a919debfcf3b514/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js#L272-L287

vursen (approved):

PR #3005, opened 2022-04-11
test: maxAttemps is hardcoded in AbstractParallelTest.java

manolo:


vaadin-flow-components-shared/src/main/java/com/vaadin/tests/AbstractParallelTest.java
```diff
@@ -20,7 +20,7 @@
 public abstract class AbstractParallelTest extends ParallelTest {
 
     @Rule
-    public RetryRule maxAttempts = new RetryRule(2);
+    public RetryRule maxAttempts = new RetryRule(Parameters.getMaxAttempts());
```
vursen, line 23: **nit:** This line could be dropped completely as it duplicates the one in `TestBenchTestCase`:

https://github.com/vaadin/testbench/blob/782616c2edda4f4bde4f905154dd7cc92f1463ea/vaadin-testbench-core/src/main/java/com/vaadin/testbench/TestBenchTestCase.java#L68-L76

manolo: good catch @vursen , done.

ZheSun88 (approved):

Issue #1690, opened 2021-05-21
TreeGrid content indent in Hierarchy Column


vlipovetskii:
How to change from java TreeGrid indent ("level-spacer") in Hierarchy column ?

rolfsmeds:
Sorry for the very very delayed response, hope this will still benefit someone :)

The indentation is controlled by the `--vaadin-grid-tree-toggle-level-offset` css property. It can be overriden by adding a stylesheet for the `vaadin-grid-tree-toggle` component, either with `@CssImport(some-stylesheet-name.css, themeFor="vaadin-grid-tree-toggle")` or, if using the new theme folder format, by adding a stylesheet called `vaadin-grid-tree-toggle.css` to the theme's `components` folder, and setting a new value for it there:

```
:host {
  --vaadin-grid-tree-toggle-level-offset: 1em;
}
```

TatuLund:
It would be even better if this could be controlled with theme attribute per instance basis, however it is currently not possible as TreeGrid does not set the theme attribute of the toggle.

```
:host([theme~="large"]) {
  --vaadin-grid-tree-toggle-level-offset: 4em;
}
```

vursen:
I'm a bit in doubt about whether the propagation of the TreeGrid theme to the toggle is a good idea. Looking at the tree toggle, it supports its own set of themes at least for the Lumo theme, so a conflict may potentionally occur? 

I guess it might be better to allow customizing `--vaadin-grid-tree-toggle-level-offset` directly at the grid style sheet level, which is currently not possible because of the way the tree toggle defines the default value for the level offset. That could be adjusted though:

```diff
registerStyles(
  'vaadin-grid-tree-toggle',
  css`
    :host {
-      --vaadin-grid-tree-toggle-level-offset: 2em;
+      --_vaadin-grid-tree-toggle-level-offset: var(--vaadin-grid-tree-toggle-level-offset, 2em);
```

rolfsmeds:
I think the proposal from @vursen makes sense. In the meantime, I just realized that the `vaadin-grid-tree-toggle` element is in light dom, and thus styleable most easily by placing the following css in a global stylesheet (e.g. `styles.css`):

```
vaadin-grid-tree-toggle {
    --vaadin-grid-tree-toggle-level-offset: 4em;
}

```
and even based on `vaadin-tree-grid`'s `theme` attrib as @TatuLund desires:

```
[...]

PR #3026, opened 2022-04-15
feat: Propagate theme to tree toggle in TreeGrid

TatuLund:
Fixes: #1690


vursen:
Based on Rolf's https://github.com/vaadin/flow-components/issues/1690#issuecomment-1102121554, I'm going to close this PR. 

The tree toggle is in the light DOM so that you can style it depending on the grid theme even without this change.

PR #3025, opened 2022-04-15
refactor: extract target connector, refactor items tree construction logic

vursen:
## Description

- [x] Extracted the context menu target connector into a separate file (before it was called the context menu connector which was confusing)
- [x] Refactored the items tree construction logic. The menu bar buttons are now saved from an unnecessary re-render that was previously caused by the double assignment of the `items` property: one in `ContextMenuConnector.generateItems` and the other one in `MenuBarConnector.updateButtons`.
- [x] Added JSDoc to some connector methods to clarify their purpose.

A part of #2134 

## Type of change

- [x] Internal

## Checklist

- [x] I have read the contribution guide: https://vaadin.com/docs-beta/latest/guide/contributing/overview/
- [x] I have added a description following the guideline.
- [x] The issue is created in the corresponding repository and I have referenced it.
- [x] I have added tests to ensure my change is effective and works as intended.
- [x] New and existing tests are passing locally with my change.
- [x] I have performed self-review and corrected misspellings.

PR #3027, opened 2022-04-15
fix: re-generate items on the hidden or disabled property change

### LINKED ISSUE #2134: MenuBar - MenuItem enabled state change does not apply properly when not visible
### Description

When on a MenuItem the setEnabled() called while being not visible (either directly or with parent.) the MenuItem does not apply the change properly (some times it doesn't change, and in some cases it stucks in an intermediate state being recolored, disabled, but still reacting to hover events by changing style). 

### Expected outcome

SetEnabled calls on invisible MenuBar/MenuItem components should apply correctly.

### Actual outcome

The MenuItem doesn't change at all, or only partially.

### Live Demo (optional)

A screenshot or video with issue reproduced.

```java
MenuBar menubar = new MenuBar();
MenuItem item = menubar.addItem("Click", ev -> System.out.println("Clicked"));
Div div = new Div(menubar);
add(div);
﻿add(new Button("Hide parent", ev->div.setVisible(false)));
add(new Button("Show parent", ev->div.setVisible(true)));
add(new Button("Enable button", ev->item.setEnabled(true)));
add(new Button("Disable button", ev->item.setEnabled(false)));
```

### Steps to reproduce

1. Hide the parent Div.
2. Disable the MenuItem.
3. Make the parent Div visible.
4. The MenuItem will appear in an incorrect state.

### Environment

Vaadin: 14.6.3 

### Browsers Affected
- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE 11
- [ ] iOS Safari
- [ ] Android Chrome

vursen:
## Description

This PR adds a MutationObserver to the menu bar connector that watches for the `hidden` and `disabled` attributes and triggers re-generating the menu bar items if an attribute change occurs.

It solves the issue where changing the visibility or disabled state of a menu item wouldn't be applied if an ancestor of the menu bar was hidden at the same time. 

Fixes #2134

Depends on #3025 

## Type of change

- [x] Bugfix

## Checklist

### DIFF
--- a/vaadin-menu-bar-flow-parent/vaadin-menu-bar-flow/src/main/resources/META-INF/resources/frontend/menubarConnector.js
+++ b/vaadin-menu-bar-flow-parent/vaadin-menu-bar-flow/src/main/resources/META-INF/resources/frontend/menubarConnector.js
@@ -31,8 +31,16 @@
         return;
       }
 
-      const observer = new MutationObserver(() => {
-        menubar.$connector.generateItems();
+      const observer = new MutationObserver((records) => {
+        const hasChangedAttributes = records.some((entry) => {
+          const oldValue = entry.oldValue;
+          const newValue = entry.target.getAttribute(entry.attributeName);
+          return oldValue !== newValue;
+        });
+
+        if (hasChangedAttributes) {
+          menubar.$connector.generateItems();
+        }
       });
 
       menubar.$connector = {
@@ -76,7 +84,8 @@
           // to sync the new attribute values with the corresponding properties in the items array.
           items.forEach((item) => {
             observer.observe(item.component, {
-              attributeFilter: ['hidden', 'disabled']
+              attributeFilter: ['hidden', 'disabled'],
+              attributeOldValue: true
             });
           });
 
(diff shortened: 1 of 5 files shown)


vaadin-menu-bar-flow-parent/vaadin-menu-bar-flow/src/main/resources/META-INF/resources/frontend/menubarConnector.js
```diff
@@ -31,6 +31,18 @@
         return;
       }
 
+      const observer = new MutationObserver((records) => {
+        const hasChangedAttributes = records.some((entry) => {
```
vursen, line 35: **note:** I had to add this check because the context-menu has a logic where it ensures the `disabled` attribute by re-assigning it which triggers the Mutation Observer here.

https://github.com/vaadin/web-components/blob/9b9fa8a6e26d6154cb06b27c1ddd3bda6e21e1df/packages/context-menu/src/vaadin-contextmenu-items-mixin.js#L229

vursen:
Fixing the current issue also revealed another one that I don't have a good workaround for: if you add a new menu item while an ancestor of the menu bar is hidden, the new item won't be properly attached to the DOM. In order to make it work, you will need to cause the menu bar to re-render manually after the ancestor becomes visible. 

There is no reliable way to pick up on that in the connector as well as no Flow API that would allow detecting the visibility change of an ancestor of a component. As long as there is no API, we cannot fix that issue, so I would suggest creating a ticket, but not considering that in the scope of this PR.

PR #3039, opened 2022-04-21
fix: throw exception instead of downloading node.js

### LINKED ISSUE #2837: [chart] svg-generator should not try to install nodejs if missing
### Description

Currently the svg-generator tries to access the internet to download and install nodejs on a server. This can lead to huge red flags in a secured environment.

### Expected outcome

Please fail with an error message that nodejs is missing, instead of trying to download it.

The [docs](https://vaadin.com/docs/latest/ds/components/charts/svg-generator) even state that "Node.js must be installed for the generator to work."

### Actual outcome

Nodejs is tried to be installed on a secured server environment. if the server has internet access - node is even installed (in .vaadin).

- export a chart on a system without nodejs

### Environment

- v22/23
- Any OS without nodejs

sissbruecker:
## Description

Prevent the SVG generator from downloading node.js if it can not be found, and throw an exception instead. First tries to locate a global node installation, and as a fallback looks for a node installation provided by Flow's frontend toolchain to cover development setups where developers don't use a global node installation.

Fixes #2837

## Type of change

- Bugfix

### DIFF
--- a/vaadin-charts-flow-parent/vaadin-charts-flow-svg-generator/src/main/java/com/vaadin/flow/component/charts/export/NodeRunner.java
+++ b/vaadin-charts-flow-parent/vaadin-charts-flow-svg-generator/src/main/java/com/vaadin/flow/component/charts/export/NodeRunner.java
@@ -13,23 +13,54 @@
 
 package com.vaadin.flow.component.charts.export;
 
+import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 
-import com.vaadin.flow.server.frontend.FrontendTools;
-import com.vaadin.flow.server.frontend.FrontendToolsSettings;
+import com.vaadin.flow.server.frontend.FrontendToolsLocator;
 import com.vaadin.flow.server.frontend.FrontendUtils;
 
 class NodeRunner {
 
+    private final FrontendToolsLocator frontendToolsLocator;
+
+    NodeRunner() {
+        this(new FrontendToolsLocator());
+    }
+
+    NodeRunner(FrontendToolsLocator frontendToolsLocator) {
+        this.frontendToolsLocator = frontendToolsLocator;
+    }
+
+    /**
+     * Returns the absolute file path to a Node.js executable.
+     * First tries to find a global installation, and as a fallback looks for an installation in the Vaadin home directory.
+     * Throws if no Node.js installation could be found.
+     * @return the absolute path to the Node.js executable
+     * @throws IllegalStateException when no Node.js installation could be found
+     */
+    String findNodeExecutable() {
+        // Try resolve global node installation
+        String nodeExecutableName = FrontendUtils.isWindows() ? "node.exe" : "node";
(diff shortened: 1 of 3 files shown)


vaadin-charts-flow-parent/vaadin-charts-flow-svg-generator/src/main/java/com/vaadin/flow/component/charts/export/NodeRunner.java
```diff
@@ -13,23 +13,54 @@
...
+            return nodeExecutableFile.getAbsolutePath();
+        }
+        // Try resolve installation from Vaadin home
+        // This covers development setups where developers rely on the Node.js installation provided by Flow's frontend toolchain
+        nodeExecutableFile = new File(FrontendUtils.getVaadinHomeDirectory(), FrontendUtils.isWindows() ? "node/node.exe" : "node/node");
+        if (frontendToolsLocator.verifyTool(nodeExecutableFile)) {
+            return nodeExecutableFile.getAbsolutePath();
+        }
+
+        throw new IllegalStateException("Node not found");
```
sissbruecker, line 57: The exception could be more explicit that SVG generator requires node, and where we tried to locate it.

sissbruecker: Improved the message.

vaadin-charts-flow-parent/vaadin-charts-flow-svg-generator/src/main/java/com/vaadin/flow/component/charts/export/NodeRunner.java
```diff
@@ -13,23 +13,64 @@
...
 
 class NodeRunner {
 
+    private final FrontendToolsLocator frontendToolsLocator;
+
+    NodeRunner() {
+        this(new FrontendToolsLocator());
+    }
+
+    NodeRunner(FrontendToolsLocator frontendToolsLocator) {
```
sissbruecker, line 32: Added this overload in order to pass in a mocked instance for testing.

vaadin-charts-flow-parent/vaadin-charts-flow-svg-generator/src/main/java/com/vaadin/flow/component/charts/export/NodeRunner.java
```diff
@@ -13,23 +13,64 @@
...
+     * Returns the absolute file path to a Node.js executable. First tries to
+     * find a global installation, and as a fallback looks for an installation
+     * in the Vaadin home directory. Throws if no Node.js installation could be
+     * found.
+     *
+     * @return the absolute path to the Node.js executable
+     * @throws IllegalStateException
+     *             when no Node.js installation could be found
+     */
+    String findNodeExecutable() {
```
sissbruecker, line 46: The logic here tries to mirror the resolving logic from the previously used `FrontendTools` class: https://github.com/vaadin/flow/blob/master/flow-server/src/main/java/com/vaadin/flow/server/frontend/FrontendTools.java#L446

vursen (approved):

PR #3045, opened 2022-04-22
chore: enable production settings only for merged ITs

sissbruecker:
## Description

Currently all Jetty configurations run with production settings by default. That is inconvenient for developers working with this repository as:
- production mode disables several useful features such as the dev mode tool, or the 404 fallback page listing available routes
- the frontend production bundle prevents from debugging JS code like connectors

This PR changes the setup to use development settings by default (no production mode, no production frontend bundle), and only enable those when running merged ITs. There might be other concerns that I'm not aware of, please address those in the review.


pom.xml
```diff
@@ -84,7 +84,7 @@
     </modules>
     <profiles>
         <profile>
-            <id>default</id>
+            <id>merged-it</id>
```
sissbruecker, line 87: This profile does not seem to have any relation with the other `default` profiles declared in the individual integration test modules, as it uses a different activation. Seemed appropriate to rename it.

tomivirkki (approved):
Works great by default locally

pom.xml
manolo: agree

manolo: agree

manolo (approved):

PR #3046, opened 2022-04-22
fix: Use incremental numeric id as a key for Tree Grid items

mshabarov:
## Description

Using Object::toString as an item's key for server-client communication makes item's data visible on the client side, so generated numeric id is used instead.

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -212,8 +214,12 @@ public UpdateQueueData getUpdateQueueData() {
         }
     }
 
-    private final ValueProvider<T, String> defaultUniqueKeyProvider = item -> getDataProvider()
-            .getId(item).toString();
+    private final AtomicLong uniqueKeyCounter = new AtomicLong(0);
+    private final Map<Object, Long> objectUniqueKeyMap = new HashMap<>();
+
+    ValueProvider<T, String> defaultUniqueKeyProvider = item -> String.valueOf(
```
sissbruecker, line 220: I'm in general a bit puzzled about what's going on here, specifically why `TreeGrid` and `HierarchicalDataCommunicator` have their own logic for generating default keys.

For `Grid`, `DataCommunicator` has a `KeyMapper` that generates an increasing number as unique key for each item. Also data communicator will ensure that key mapper gets cleaned up when data is removed.

For `TreeGrid`, `HierarchicalDataCommunicator` requires you to provide a unique key provider. This key provider is then used in a custom key mapper that extends from `KeyMapper` and uses the unique key provider to generate keys. Now we change the unique key provider to use an increasing number as key - which is the same thing the regular `KeyMapper` would have done. 

Why aren't we just using the default `KeyMapper` in `HierarchicalDataCommunicator`? The benefit here would be that we don't need to keep two maps of keys (one in `KeyMapper`, and one in `TreeGrid`). And the map in `KeyMapper` actually gets cleaned up when data changes, while the one that is being introduced here would just keep growing.

sissbruecker: Thanks, would be great if we can look into that. If it turns out to be more complicated, we can consider going ahead with this solution as a (maybe temporary) workaround.

sissbruecker: Merging the workaround for now until we can investigate whether we can reuse the default KeyMapper behavior.

sissbruecker (approved):

PR #3073, opened 2022-04-27
fix: Ensure that GridElement#select and GridElement#deselect emit event

### LINKED ISSUE #1598: Testbench: MultiSelect grid.select() doesn't fire a addSelectionListener listner
Grid  in `MultiSelect` mode doesn't fire a `addSelectionListener`  when selecting a row in Textbestbench, while assertion returns true for a selection. The same works for Grid in the default selection mode.

Grid:
```Grid<Person> grid = new Grid<>(Person.class);
//Comment out this line and listener is used
 grid.addSelectionListener(e->{
           System.out.println("Grid selection listener" + e.getAllSelectedItems());
  });
```
Test class:
```
 @Test
    public void selectGrid(){
        GridElement grid=$(GridElement.class).first();
        grid.select(0);
        //Doesn't work with listeners either
       // grid.getRow(1).click();
        //grid.getRow(0).getCell( grid.getColumn("First Name")).click();
        assertTrue(grid.getRow(0).isSelected());
    }
```

TatuLund:
Fixes #1598

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-testbench/src/main/java/com/vaadin/flow/component/grid/testbench/GridElement.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-testbench/src/main/java/com/vaadin/flow/component/grid/testbench/GridElement.java
@@ -331,15 +331,11 @@ public class GridElement extends TestBenchElement {
      *            the row to select
      */
     public void select(int rowIndex) {
-        WebElement checkbox = getCellContent(grid.getCell(rowIndex, 0));
-        checkbox.click();
+        if (!getRow(rowIndex).isSelected) {
+           WebElement checkbox = getCellContent(grid.getCell(rowIndex, 0));
+           checkbox.click();
+        }
     }
-
-    private WebElement getCellContent(GridTHTDElement cell) {
-        return (WebElement) executeScript(
-                "return arguments[0].firstElementChild.assignedNodes()[0].firstElementChild;",
-                cell);
-    }    
 
     /**
      * Selects the row.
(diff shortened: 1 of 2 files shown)


vaadin-grid-flow-parent/vaadin-grid-testbench/src/main/java/com/vaadin/flow/component/grid/testbench/GridElement.java
```diff
@@ -343,6 +344,8 @@ void select(GridTRElement row) {
         if (isMultiselect()) {
             executeScript("arguments[0].selectItem(arguments[1]._item);", this,
                     row);
+            executeScript("arguments[0].$server.select(arguments[1]._item.key)",
```
sissbruecker, line 347: I'm kind of wondering about the utility of this class if it just uses some internal API that calls the server directly. Maybe we should actually click the checkboxes, as was done in the test case.
I'm going to take a look at this.

TatuLund: > we should actually click the checkboxes, as was done in the test case.

I was implementing it that way first as you see from the history, but then I noticed that the method needs to work also in single select mode, when the checkbox is not there.

sissbruecker: Attempted an alternative fix here: https://github.com/vaadin/flow-components/pull/3080
We should probably add a usage of these methods in a single selection IT as well...

PR #3080, opened 2022-04-27
fix: make GridElement dispatch selection events

### LINKED ISSUE #1598: Testbench: MultiSelect grid.select() doesn't fire a addSelectionListener listner

sissbruecker:
## Description

Fixes `GridElement.select` and `GridElement.deselect` to properly dispatch change events on the server-side. The fix involves changing the implementation to click on the selection checkboxes, which triggers the same logic in the grid connector as it would when a real user interacts with the component.

Ideally we would also change the single-selection logic to use a click instead of setting the active item, but that can be done as a separate change.

Fixes #1598 

## Type of change

- Bugfix

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridSingleSelectionIT.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridSingleSelectionIT.java
@@ -43,12 +43,12 @@ public class GridSingleSelectionIT extends AbstractComponentIT {
 
     @Test
     public void selectFirstItemFromClient_deselectFirstItemFromClient_nothingSelected() {
-        grid.getCell("0").click();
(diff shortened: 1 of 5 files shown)


tomivirkki (approved):
This approach makes sense for end-to-end testing because it sticks to the actions the user can and would actually perform (clicking a checkbox).

PR #3103, opened 2022-04-29
chore: define feature flags for merged ITs

sissbruecker:
## Description

Adds an additional step to `mergeITs.js` to use a feature flag properties file that enables all experimental components.


sissbruecker:
Closing in favor of #3105

PR #3139, opened 2022-05-06
chore: remove checks for outdated IE versions

web-padawan:
## Description

We no longer support IE, so these checks can be dropped.

## Type of change

- Internal


manolo:
@web-padawan fixed couple of sonarcloud issues and rebased

sissbruecker:
@web-padawan @manolo Can this be merged?

PR #3146, opened 2022-05-09
refactor: deprecate addColumn(renderer, sortingProperties) API

tomivirkki:
This PR deprecates the `addColumn(renderer, sortingProperties)` API, which in its current form is merely an unnecessary shorthand for `addColumn(renderer).setSortProperty(sortingProperties)`.

The PR also eliminates some dead code from the method.

Note: let's merge this so that it lands with 23.2


sissbruecker:
@tomivirkki Is this still relevant?

tomivirkki:
Yes, it is. We agreed earlier that the approach should change slightly though:
- The `addColumn(renderer, sortProperties)` APIs should also be deprecated
- An example should be provided for how to achieve the same result without using the deprecated APIs

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridViewSortingPage.java
```diff
@@ -45,15 +46,20 @@ private void createSorting() {
...
+                "<div>${item.street}, number ${item.number}<br><small>${item.postalCode}</small></div>")
                 .withProperty("street",
                         person -> person.getAddress().getStreet())
                 .withProperty("number",
                         person -> person.getAddress().getNumber())
                 .withProperty("postalCode",
-                        person -> person.getAddress().getPostalCode()),
-                "street", "number").setHeader("Address");
+                        person -> person.getAddress().getPostalCode()))
+                .setComparator(addressComparator).setHeader("Address");
```
tomivirkki, line 62: This test page was using the deprecated API. Removed the `sortingProperties` parameter and added a comparator to make the column actually sortable (which it wasn't before).

yuriy-fix (approved):

PR #3154, opened 2022-05-10
fix: close a notification on detach

### LINKED ISSUE #3085: Notification showing multiple times when auto-added to reused modal
### Description

I have a modal dialog that I reuse, so I call `open` and `close` multiple times.
Now I added a `Notification` that is shown, when I open the `Dialog` in a certain context.
When I close the `Dialog` before the `Notification` is closed, the `Notification` is shown everytime I open the `Dialog` again.
To fix that, I have to wait for the `Notification` to close and close the `Dialog` after that.
The reason for this is, that the `Notification` is removed from the `Dialog` in the `opened-changed` listener. However, that listener is never notified, as the parent is already detached when it is fired.

The same problem _might_ occur in other circumstances, whenever an overlay is added to a modal and the modal is closed before the child-overlay is closed.

### Expected outcome

The `Notification` should not be shown when I reopen the dialog

```java
    private Dialog formDlg;
    private NumberField tfId;
    private TextField tfName;
    public MainView() {
        Grid<Person> grid = new Grid<>();
        grid.addColumn(p -> p.id).setHeader("ID");
        grid.addColumn(p -> p.name).setHeader("Name");
        grid.addColumn(new ComponentRenderer<>(p -> new Button(VaadinIcon.EDIT.create(), e -> {
            getOrCreateDialog(p).open();
            if (p.id == 1) {
                Notification.show("Cannot edit " + p.name);
                tfId.setReadOnly(true);
                tfName.setReadOnly(true);
            }
        })));
        grid.setItems(
            new Person(1, "John Doe"),
            new Person(2, "Jane Doe"),
            new Person(3, "Joe Ordinary"),
            new Person(4, "Mary Major")
        );
        add(grid);
    }
    private Dialog getOrCreateDialog(Person p) {
        if (formDlg == null) {
            formDlg = new Dialog();
            formDlg.add(new VerticalLayout(
                tfId = new NumberField("ID"),
                tfName = new TextField("Name")
            ));
        }
        tfId.setReadOnly(false);
        tfName.setReadOnly(false);
        tfId.setValue((double)p.id);
        tfName.setValue(p.name);
        return formDlg;
    }
    private static class Person {
        private final int id;
        private final String name;

        private Person(int id, String name) {
            this.id = id;
            this.name = name;
        }
    }
}
```

### Steps to reproduce

1. Add the above code to the main view of a new vaadin starter
2. Open the app and click the "edit" button on the first item in the grid. The `Notification` will be shown.
3. Close the dialog before the `Notification` disappears.
4. Open the `Dialog` by clicking the "edit" button on another item, the same `Notification` will be shown again.
5. Wait until the `Notification` disappears and close the `Dialog`.
6. Open the `Dialog` by clicking the "edit" button on another item. This time the `Notification` won't be shown.

### Environment

Vaadin version(s): 23.0.6
OS: Windows 10 x64

tomivirkki:
Fixes #3085

### DIFF
--- a/vaadin-notification-flow-parent/vaadin-notification-flow-integration-tests/src/main/java/com/vaadin/flow/component/notification/tests/NotificationView.java
+++ b/vaadin-notification-flow-parent/vaadin-notification-flow-integration-tests/src/main/java/com/vaadin/flow/component/notification/tests/NotificationView.java
@@ -76,7 +76,7 @@ public class NotificationView extends Div {
         Notification notification = Notification.show(
                 "This is a notification created with static convenience method");
         notification.setId("static-notification");
-        addCard("Notification using static convenience method", notification);
+        addCard("Notification using static convenience method");
     }
 
     private void createNotificationWithComponents() {
--- a/vaadin-notification-flow-parent/vaadin-notification-flow-integration-tests/src/main/java/com/vaadin/flow/component/notification/tests/NotificationView.java
+++ b/vaadin-notification-flow-parent/vaadin-notification-flow-integration-tests/src/main/java/com/vaadin/flow/component/notification/tests/NotificationView.java
@@ -77,7 +77,7 @@ public class NotificationView extends Div {
                 "This is a notification created with static convenience method");
         notification.setId("static-notification");
         addCard("Notification using static convenience method");
-        
+
     }
 
     private void createNotificationWithComponents() {
--- a/vaadin-notification-flow-parent/vaadin-notification-flow-integration-tests/src/main/java/com/vaadin/flow/component/notification/tests/NotificationView.java
+++ b/vaadin-notification-flow-parent/vaadin-notification-flow-integration-tests/src/main/java/com/vaadin/flow/component/notification/tests/NotificationView.java
@@ -76,8 +76,7 @@ public class NotificationView extends Div {
         Notification notification = Notification.show(
                 "This is a notification created with static convenience method");
         notification.setId("static-notification");
-        addCard("Notification using static convenience method");
-
+        addCard("Notification using static convenience method", notification);
     }
 
     private void createNotificationWithComponents() {
--- a/vaadin-notification-flow-parent/vaadin-notification-flow-integration-tests/src/main/java/com/vaadin/flow/component/notification/tests/NotificationView.java
+++ b/vaadin-notification-flow-parent/vaadin-notification-flow-integration-tests/src/main/java/com/vaadin/flow/component/notification/tests/NotificationView.java
@@ -76,7 +76,8 @@ public class NotificationView extends Div {
         Notification notification = Notification.show(
                 "This is a notification created with static convenience method");
         notification.setId("static-notification");
-        addCard("Notification using static convenience method", notification);
+        addCard("Notification using static convenience method");
+        
     }
 
     private void createNotificationWithComponents() {
--- a/vaadin-notification-flow-parent/vaadin-notification-flow/src/main/java/com/vaadin/flow/component/notification/Notification.java
+++ b/vaadin-notification-flow-parent/vaadin-notification-flow/src/main/java/com/vaadin/flow/component/notification/Notification.java
@@ -202,13 +202,14 @@ public class Notification extends GeneratedVaadinNotification<Notification>
         getElement().appendChild(templateElement);
         getElement().appendVirtualChild(container);
 
-        getElement().addEventListener("opened-changed", event -> removeAutoOpened());
+        getElement().addEventListener("opened-changed",
+                event -> removeAutoAdded());
     }
 
     /**
      * Removes the notification from its parent if it was added automatically.
      */
-    private void removeAutoOpened() {
+    private void removeAutoAdded() {
         if (autoAddedToTheUi && !isOpened()) {
             autoAddedToTheUi = false;
             getElement().removeFromParent();
@@ -535,7 +536,7 @@ public class Notification extends GeneratedVaadinNotification<Notification>
         // If the notificaiton gets detached, it needs to be marked as closed so
         // it won't auto-open when reattached.
         setOpened(false);
-        removeAutoOpened();
+        removeAutoAdded();
     }
 
     /**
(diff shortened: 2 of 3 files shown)


vaadin-notification-flow-parent/vaadin-notification-flow/src/main/java/com/vaadin/flow/component/notification/Notification.java
```diff
@@ -202,7 +202,7 @@ private void initBaseElementsAndListeners() {
         getElement().appendChild(templateElement);
         getElement().appendVirtualChild(container);
 
-        getElement().addEventListener("opened-changed", event -> {
+        addOpenedChangeListener(event -> {
```
tomivirkki, line 205: An "opened-changed" event only works if the Flow component doesn't get detached from its parent while the `<vaadin-notification>` Web Component is showing. Then we could rely on the DOM event to take care of getting this callback invoked.

But if the Flow component gets detached before the notification actually closes, the Web Component becomes an "orphan" without an active Flow counterpart and the following "opened-changed" event never reaches the callback.

vaadin-notification-flow-parent/vaadin-notification-flow/src/main/java/com/vaadin/flow/component/notification/Notification.java
```diff
@@ -524,6 +524,15 @@ public Registration addDetachListener(
         return super.addDetachListener(listener);
     }
 
+    @Override
+    protected void onDetach(DetachEvent detachEvent) {
+        super.onDetach(detachEvent);
+
+        // If the notificaiton gets detached, it needs to be marked as closed so
+        // it won't auto-open when reattached.
+        setOpened(false);
```
vursen, line 533: This will trigger the `openedChanged` listener above, right?

tomivirkki: That was the intention but it seems there was some timing issue, so I needed to change the logic. Let's see if the newest revision passes.

vaadin-notification-flow-parent/vaadin-notification-flow-integration-tests/src/main/java/com/vaadin/flow/component/notification/tests/NotificationView.java
```diff
@@ -76,7 +76,7 @@ private void createNotificationUsingStaticConvenienceMethod() {
         Notification notification = Notification.show(
                 "This is a notification created with static convenience method");
         notification.setId("static-notification");
-        addCard("Notification using static convenience method", notification);
```
tomivirkki, line 79: The test setup here was strange (yet supported): the notification was first auto-added to UI using `Notification.show()` but then the same instance got manually added to the card.

The assertion in the related test however was wrong. It assumed that the instance would automatically get removed from its parent on close even though it's manually re-added in the setup. This used to work because the [order of checking the condition and beforeClientResponse was kind of backwards](https://github.com/vaadin/flow-components/blob/5f95ad21b58855d71287cb1cefa716ab5f1e3ef6/vaadin-notification-flow-parent/vaadin-notification-flow/src/main/java/com/vaadin/flow/component/notification/Notification.java#L436-L437) (this PR fixes the order & adds a related test).

vaadin-notification-flow-parent/vaadin-notification-flow/src/main/java/com/vaadin/flow/component/notification/Notification.java
```diff
@@ -433,12 +445,14 @@ public void setOpened(boolean opened) {
                     + "That may happen if you call the method from the custom thread without "
                     + "'UI::access' or from tests without proper initialization.");
         }
-        if (opened && getElement().getNode().getParent() == null) {
-            ui.beforeClientResponse(ui, context -> {
+
+        ui.beforeClientResponse(ui, context -> {
+            if (opened && getElement().getNode().getParent() == null) {
```
sissbruecker, line 450: Should this use `getOpened`? `opened` might be stale at this point.

vaadin-notification-flow-parent/vaadin-notification-flow/src/main/java/com/vaadin/flow/component/notification/Notification.java
```diff
@@ -202,13 +202,25 @@ private void initBaseElementsAndListeners() {
...
-                getElement().removeFromParent();
-                autoAddedToTheUi = false;
-            }
+        getElement().addEventListener("opened-changed",
+                event -> removeAutoAdded());
+
+        addDetachListener(event -> {
+            // If the notification gets detached, it needs to be marked
+            // as closed so it won't auto-open when reattached.
+            setOpened(false);
```
sissbruecker, line 211: When adding a notification to a dialog, this does not actually close the notification in the browser. The detach listener is called, but the opened state is not synced to the client. Probably because the element was removed from the UI, and then it is not synced anymore.

Example:
```java
Dialog dialog = new Dialog();
dialog.setModal(true);

NativeButton openDialog = new NativeButton("Open", e -> {
    dialog.open();
    Notification.show("Hello", 0, Position.BOTTOM_START); // Should add it to the modal dialog
});
add(openDialog);
```

`Notification.show` is just a shortcut here. A more realistic scenario is to create a persistent Notification with a button to close it, and add it to the dialog. After closing the dialog, the notification can not be interacted with anymore. Doesn't seem to matter if the dialog is modal or not.

sissbruecker: Just to be clear, this is not necessarily related to the current bug / fix, but it was confusing to see the state being changed to closed, and then still have the notification in the browser.

vaadin-notification-flow-parent/vaadin-notification-flow/src/main/java/com/vaadin/flow/component/notification/Notification.java
tomivirkki: Done

vaadin-notification-flow-parent/vaadin-notification-flow/src/main/java/com/vaadin/flow/component/notification/Notification.java
tomivirkki: This is indeed similar, but still a different issue. Reported here https://github.com/vaadin/web-components/issues/3823

sissbruecker (approved):

PR #3163, opened 2022-05-11
fix: Cleanup disabledListener on detach

### LINKED ISSUE #3157: Memory leak

TatuLund:
Fixes vaadin/flow#13851

### DIFF
--- a/vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java
+++ b/vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java
@@ -19,6 +19,7 @@ import com.vaadin.flow.component.AttachEvent;
 import com.vaadin.flow.component.ClickEvent;
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.ComponentEventListener;
+import com.vaadin.flow.component.DetachEvent;
 import com.vaadin.flow.component.HasEnabled;
 import com.vaadin.flow.component.HasSize;
 import com.vaadin.flow.component.html.Image;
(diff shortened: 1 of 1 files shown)


vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java
```diff
@@ -418,8 +419,21 @@ private void doDisableOnClick() {
...
+
+    @Override
+    protected void onDetach(DetachEvent detachEvent) {
+        disableListener.remove();
+    }
+
     @Override
     protected void onAttach(AttachEvent attachEvent) {
+        if (disableListener == null) {
+            disableListener = addClickListener(buttonClickEvent -> {
```
knoobie, line 431: This could be problematic; because now (after re-attach) the disableOnClick listener runs after the other listener and therefor does not stop the click propagation anymore.. is there a way to register the clicklistener as "first" again?

sissbruecker: Yeah that is the challenging part here. I also wonder why this specific listener creates a leak in the first place. We have lots of components creating listeners on themselves. So we either have dozens of memory leaks, or there is some specific issue with this one. One noticable difference is that this component stores the registration as a field which other components do not do.

sissbruecker:
Is there a reason to duplicate the listener from here, or can this one just be removed: https://github.com/vaadin/flow-components/blob/07dac45a77438136ef235c3a72baa4aff73822ba/vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java#L50-L56

vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java
TatuLund: Based on comment by @Artur- in the ticket, this PR might not be the proper way to fix the problem and actually provides only a partial help.

sissbruecker:
Closing this, as the memory leak seems to result from Flow internals: https://github.com/vaadin/flow-components/issues/3157#issuecomment-1124895351

PR #3168, opened 2022-05-12
refactor: extract ComboBoxBase from ComboBox

sissbruecker:
Part of https://github.com/vaadin/platform/issues/2961

This is the first part of a refactoring to extract a `ComboBoxBase` class that should be used as common base class for `ComboBox` and `MultiSelectComboBox`. The intention for this class is to hold all logic common to both components, such as property wrappers, event listeners, and data provider logic. The only logic that should remain in the specific classes should be related to handling the value, and other component-specific things like theme variants or specific properties like I18N.

For now this change contains the following:

- Introduce the new `ComboBoxBase` class, with generic types, and a constructor signature that should allow both components to make use of it. The main difference here is to make the value property name, and the value property type generic.
- Make `ComboBox` extend from `ComboBoxBase` instead of `GeneratedVaadinComboBox`
- Extract all basic web component property wrappers, and event listeners, from both `ComboBox` and `GeneratedVaadinComboBox`, into the new base class. The remaining logic will be moved in later refactorings.
- Tried to add/improve the JavaDoc of the property wrappers, many of those were overly verbose with web-component related topics, or did not explain what the property actually does. Feel free to make suggestions for improvements.
- Removes the usage of `GeneratedVaadinComboBox`
  - We agreed that we want to gradually get rid of these classes, and it seemed like a bad choice to introduce the new base class as an abstraction on top of the existing base class. It is debatable whether `GeneratedVaadinComboBox` should be considered public API, but in my opinion it is not.
  - As mentioned above, moved all property wrapper that were still relevant to the new base class, although most had been overriden by `ComboBox` anyway
  - Removed all property wrappers that were not used, all of them had protected visibility
  - Removed all methods that called remote functions, none of them were used, all of them had protected visibility
  - Removed all event listeners that were not used, only `addCustomValueListener` remained, which has been moved to the new base class
  - Copied the nested `CustomValueSetEvent` class into a separate file, made the existing one extend from the new class, and marked it as deprecated.
  - Marked all the other nested event classes as deprecated. Nothing more to do here, they are not used anywhere else in our API.
  - The nested event classes, and the `GeneratedVaadinComboBox` class, need to be kept around for now, as the event classes are marked as public and might be in use in parameter or variable declarations. They should be removed in the next major.


vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/RequiredComboboxIT.java
```diff
@@ -76,7 +76,6 @@ private void assertValidationError(ComboBoxElement comboBox) {
                 comboBox.getAttribute("invalid"));
 
         TestBenchElement error = comboBox.$("[part='error-message']").first();
-        Assert.assertTrue(error.getSize().getHeight() > 0);
```
sissbruecker, line 79: At least locally this was flaky, and it seems redundant to the other assertions.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -472,7 +483,7 @@ private void refreshValue() {
         JsonObject json = Json.createObject();
         json.put("key", keyMapper.key(value));
         dataGenerator.generateData(value, json);
-        setSelectedItem(json);
+        getElement().setPropertyJson("selectedItem", json);
```
sissbruecker, line 486: It seems there was no need to keep the `selectedItem` property around, so I inlined this usage.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -102,11 +105,17 @@
...
+@NpmPackage(value = "@vaadin/vaadin-combo-box", version = "23.1.0-beta1")
+@JsModule("@vaadin/combo-box/src/vaadin-combo-box.js")
+@JsModule("@vaadin/polymer-legacy-adapter/template-renderer.js")
 @JsModule("./flow-component-renderer.js")
 @JsModule("./comboBoxConnector.js")
-public class ComboBox<T> extends GeneratedVaadinComboBox<ComboBox<T>, T>
-        implements HasSize, HasValidation,
-        HasDataView<T, String, ComboBoxDataView<T>>,
+public class ComboBox<T> extends ComboBoxBase<ComboBox<T>, T, T> implements
+        HasSize, HasValidation, HasDataView<T, String, ComboBoxDataView<T>>,
```
sissbruecker, line 118: We will want to move some of these interfaces to the base class as well, but maybe we should wait until all corresponding implementations have moved.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxBase.java
```diff
@@ -0,0 +1,344 @@
...
+ * @param <TItem>
+ *            Type of individual items that are selectable in the combo box
+ * @param <TValue>
+ *            Type of the selection / value of the extending component
+ */
+public abstract class ComboBoxBase<TComponent extends ComboBoxBase<TComponent, TItem, TValue>, TItem, TValue>
+        extends AbstractSinglePropertyField<TComponent, TValue>
+        implements HasStyle, Focusable<TComponent> {
+
+    private static final String PROP_AUTO_OPEN_DISABLED = "autoOpenDisabled";
```
web-padawan, line 27: nit: personally I don't see any reason for this specific property to be extracted into a constant.
IMO we can inline it for now, same as it's done for any other properties e.g. `placeholder` etc.

We might want to extract this into an interface e.g. `HasAutoOpen` later on.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxBase.java
```diff
@@ -0,0 +1,344 @@
...
+     */
+    public boolean isRequired() {
+        return super.isRequiredIndicatorVisible();
+    }
+
+    /**
+     * Whether the component should block user input that does not match the
+     * configured pattern
+     */
+    public boolean isPreventInvalidInput() {
```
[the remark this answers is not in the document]
web-padawan: Yes, IMO these should be only in `ComboBox`.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/events/CustomValueSetEvent.java
```diff
@@ -0,0 +1,30 @@
+package com.vaadin.flow.component.combobox.events;
```
web-padawan, line 1: Let's add a license header comment to this new file.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxBase.java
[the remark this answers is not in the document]
sissbruecker: Moved them to `ComboBox`

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxBase.java
```diff
@@ -0,0 +1,344 @@
+package com.vaadin.flow.component.combobox;
```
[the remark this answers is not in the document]
sissbruecker: Thanks, done

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxBase.java
sissbruecker: Agree, done

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/events/CustomValueSetEvent.java
sissbruecker: Thanks, done

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/GeneratedVaadinComboBox.java
```diff
@@ -1259,220 +101,21 @@ public String getFilter() {
...
-     *            a function that accepts this component and a property value
-     *            and returns a model value
-     * @param modelToPresentation
-     *            a function that accepts this component and a model value and
-     *            returns a property value
-     * @param isInitialValueOptional
-     *            if {@code isInitialValueOptional} is {@code true} then the
-     *            initial value is used only if element has no {@code "value"}
-     *            property value, otherwise element {@code "value"} property is
-     *            ignored and the initial value is set
```
web-padawan, line 1427: Just to clarify, is this logic added by https://github.com/vaadin/vaadin-combo-box-flow/pull/394 no longer necessary?

sissbruecker: From my understanding not. It seems this construct for initializing the presentation value was auto-generated for all field components. However for ComboBox the initial value was hard-coded to `null`, so the condition below could never pass.
Besides the presentation value for an empty/null value in `ComboBox` is `null` as well, so there is no need to initialize the presentation value otherwise.

web-padawan: Thanks, that makes sense 👍

web-padawan (approved):

PR #3172, opened 2022-05-13
fix: return empty string if no summary text defined

### LINKED ISSUE #3171: Details getSummaryText throws NullPointerExceptions if summary was not set.
### Describe your motivation

I wanted to implement this code, where I compare the old and new values of summary text. 
```
if (!Objects.equals(newSummaryText, detail.getSummaryText())) {
  detail.setSummaryText(newSummaryText);
}
```
Then I've noticed the getSummaryText throws a nullpointerexception because the summary was not set before. I had to do like this:
```
    if (detail.getSummary() == null || !Objects.equals(newSummaryText, detail.getSummaryText())) {
      detail.setSummaryText(newSummaryText);
    }
```

I think getSummaryText() should return an empty string (or null) if the summary was not set before.
or/and
Details.setSummaryText should do this comparsion before doing any component creation and logic.

_No response_

### Additional context

_No response_

DiegoCardoso:
## Description
Returns empty string instead of NPE when calling `getSummaryText` before any summary text is defined.

Decided to use empty string instead of `null` to make it consistent with `setSummaryText` which sets the content to `""` when called with `null`.

Fixes #3171

## Type of change

- [x] Bugfix
- [ ] Feature

### DIFF
--- a/vaadin-details-flow-parent/vaadin-details-flow/src/main/java/com/vaadin/flow/component/details/Details.java
+++ b/vaadin-details-flow-parent/vaadin-details-flow/src/main/java/com/vaadin/flow/component/details/Details.java
@@ -81,7 +81,7 @@ public class Details extends Component
      * Initializes a new Details using the provided summary.
      *
      * @param summary
-     *                the summary component to set.
+     *            the summary component to set.
      * @see #setSummaryText(String)
      */
     public Details(String summary) {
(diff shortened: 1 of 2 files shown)


vaadin-details-flow-parent/vaadin-details-flow/src/main/java/com/vaadin/flow/component/details/Details.java
```diff
@@ -81,7 +81,7 @@ public Details() {
      * Initializes a new Details using the provided summary.
      *
      * @param summary
-     *            the summary component to set.
+     *                the summary component to set.
```
web-padawan, line 84: Formatting looks a bit off, when I run `mvn formatter:format` these changes get reverted.

web-padawan (changes requested):

web-padawan (approved):

PR #3185, opened 2022-05-13
refactor: preserve item keys server-side

sissbruecker:
## Description

WIP

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxDataCommunicator.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxDataCommunicator.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2000-2022 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component.combobox;
+
+import com.vaadin.flow.data.provider.ArrayUpdater;
+import com.vaadin.flow.data.provider.DataCommunicator;
+import com.vaadin.flow.data.provider.DataGenerator;
+import com.vaadin.flow.data.provider.KeyMapper;
+import com.vaadin.flow.function.SerializableConsumer;
+import com.vaadin.flow.internal.StateNode;
+import elemental.json.JsonArray;
+
+import java.util.HashSet;
+import java.util.Set;
+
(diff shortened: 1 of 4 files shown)


web-padawan (approved):

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxDataCommunicator.java
```diff
@@ -0,0 +1,100 @@
...
+            } else {
+                // Mark item for removal as soon as it is not selected anymore
+                itemsMarkedForRemoval.add(item);
+            }
+        }
+
+        @Override
+        public String key(TItem item) {
+            // Unmark item when it becomes active again
+            itemsMarkedForRemoval.remove(item);
```
tomivirkki, line 71: Is it safe to remove the item from the set whenever its key is requested?

sissbruecker: You mean if `HashSet.remove` will throw an exception if the item does not exist? [It does not](https://docs.oracle.com/javase/8/docs/api/java/util/HashSet.html#remove-java.lang.Object-).

tomivirkki: I was thinking about a scenario where the `key` method gets called for some other reason.

For example (probably not a realistic one):
1. Select an item
2. Filter so that items, other than than the selected one, get removed. The selected item gets added to the `itemsMarkedForRemoval` set.
3. Call `comboBox.setValue(comboBox.getValue())` -> this causes `key()` to get [invoked with the selected value](https://github.com/vaadin/flow-components/blob/979be8617db3dcf65d138b6d5d551a7d2c887d19/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L489) -> selected value is removed from the `itemsMarkedForRemoval` set.

Still trying to wrap my head around this to understand if this is a potential problem or not.

sissbruecker: Thanks for clarifying. If we add a step 4, where we finally change the selected value to something else, then item would indeed not be removed from the key mapper. It can only get removed afterwards if the data communicator pushes an update that includes that item, after which the data communicator would request to remove that item when confirming the update.
I think this issue also exists in the current version of the component - basically calling `key` outside of a data communicator update will not clean up the item.

One option I can think of is to make use of the previous selection, and try to remove keys for the previous selected items. But I fear that could interfere with data communicator updates. Do you have any ideas?

In general I feel that this is a minor issue for the combo box, as the number of items that we keep references to should be small - it's the items that have been selected, in combo box, or multi select combo box. One point where they should definitely get cleaned up is when setting a new data provider, the data communicator calls `removeAll` on the key mapper in that case, which is not changed in the custom key mapper.

tomivirkki: Couldn't come up with a better solution. As you pointed out, the risk of memory leak here is really small so it's good to go IMO.

tomivirkki (approved):

PR #3217, opened 2022-05-24
chore!: Make shared base package name be more specific

mshabarov:
## Description

Makes shared module package name to be more specific.
com.vaadin.flow.component package name already used in flow-server
so it produces a conflicts when used in OSGi environments.
Also, export directive was missing, which also produces errors.

Fixes https://github.com/vaadin/base-starter-flow-osgi/issues/224

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

- [ ] I have read the contribution guide: https://vaadin.com/docs-beta/latest/guide/contributing/overview/

### DIFF
--- a/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/base/HasClearButton.java
+++ b/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/base/HasClearButton.java
@@ -13,7 +13,7 @@
  * License for the specific language governing permissions and limitations under
  * the License.
  */
-package com.vaadin.flow.component;
+package com.vaadin.flow.component.shared.base;
 
 import com.vaadin.flow.component.HasElement;
 
--- a/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/test/java/com/vaadin/flow/component/HasClearButtonTest.java
+++ b/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/test/java/com/vaadin/flow/component/HasClearButtonTest.java
@@ -4,6 +4,8 @@ import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 
+import com.vaadin.flow.component.shared.base.HasClearButton;
+
 public class HasClearButtonTest {
 
     private TestComponent component;
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
@@ -21,22 +21,13 @@ import java.util.Objects;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import com.vaadin.flow.component.ComponentEventListener;
-import com.vaadin.flow.component.HasClearButton;
-import com.vaadin.flow.component.HasHelper;
-import com.vaadin.flow.component.HasLabel;
-import com.vaadin.flow.component.HasSize;
-import com.vaadin.flow.component.HasTheme;
-import com.vaadin.flow.component.HasValidation;
 import com.vaadin.flow.component.Tag;
-import com.vaadin.flow.component.combobox.events.CustomValueSetEvent;
 import com.vaadin.flow.component.dependency.JsModule;
 import com.vaadin.flow.component.dependency.NpmPackage;
 import com.vaadin.flow.data.provider.DataCommunicator;
 import com.vaadin.flow.data.provider.DataKeyMapper;
 import com.vaadin.flow.data.provider.DataProvider;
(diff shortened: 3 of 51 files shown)


vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/base/HasClearButton.java
```diff
@@ -13,7 +13,7 @@
  * License for the specific language governing permissions and limitations under
  * the License.
  */
-package com.vaadin.flow.component;
+package com.vaadin.flow.component.shared.base;
```
sissbruecker, line 16: Not happy with the package name in general. Are there any alternatives in how we use OSGI to keep the current package name? This will impact DX, as these classes are public API for developers, and it's going to be weird that some mixin interfaces have to be imported from `com.vaadin.flow.component` and others from `com.vaadin.flow.component.shared.base`. Besides the name is very non-descript for being so specific.

If we have to change it, why use `com.vaadin.flow.component.shared.base` instead of `com.vaadin.flow.component.shared`? If package names create collisions with OSGI, how can we be sure that the new name will not conflict with anything from Flow in the future?

sissbruecker (changes requested):

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/test/java/com/vaadin/flow/component/HasClearButtonTest.java
```diff
@@ -4,6 +4,8 @@
 import org.junit.Before;
 import org.junit.Test;
 
+import com.vaadin.flow.component.shared.base.HasClearButton;
```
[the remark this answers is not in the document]
mshabarov: Done

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/base/HasClearButton.java
mshabarov: Renamed to `com.vaadin.flow.component.shared`.
OSGi exports bundle context by specifying the packages. 
I don't know how to tell OSGi to which bundle to look, if it's possible at all.
It can be so that OSGi finds the package match and looks for a class in the first bundle containing such a package.
Having commonly used classes in *.shared package is not a bad idea IMO.

mshabarov: It would be good to have some sanity test checking that no packages from different JARs have such a conflicts, but I don't know so far how to do it rather than run the OSGi starter project and check if it's started well.

sissbruecker: We discussed this in the team and for now it seems using a different package is the only option short-term. Reusing the same package between multiple OSGI bundles is not intended (see https://developer.ibm.com/articles/osgi-demystified-part-4-dont-bungle-your-bundles/), and while there are workarounds (see same article) it seems they would be challenging to apply to our module structure.

In terms of DX we still think developers would expect interfaces like `HasClearButton` to be in the same package as `HasLabel`, so we might want to revisit this for a future major, for example by actually combining all the `com.vaadin.flow.components` classes in the same Maven module, in either repository.

sissbruecker (approved):

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/base/HasClearButton.java
mshabarov: Yes, the similar question is raised in SO https://stackoverflow.com/a/47327573/16650110 .
Alternative is to use OSGi Fragments (SO provides a link to docs).
But this solution looks a bit unpredictable on the first glance and would rely on the OSGi container.
The support of this Fragment feature might change in the future and we again fall back to the same issue.
At least this Fragments approach needs an investigation and tests. 

I agree that the package name is better to be same for mixins.
A random idea: could we transfer them to the Flow?
Technically the shared module doesn't depend on components, so it's possible.

sissbruecker: Agree, fragments don't seem to be a great solution, and we might run into issues with how our projects depend on each other.

> A random idea: could we transfer them to the Flow?
That would be an option, or there might also be an option to move `com.vaadin.flow.components` to `flow-components`. There are some concerns in that the interfaces we introduced are more specific to Vaadin components, while the ones in the Flow repository are more of a general nature. So we felt this requires more thought, and we should use the separate package for now, and deal with a breaking change later (which would have limited impact, since the new interfaces are probably not going to get a lot of use).

mshabarov:
`VirtualListIT` failed two times. It passes on my local. Looks unrelated to changes made.

PR #3231, opened 2022-05-25
test: migrate some ITs from FW8

tomivirkki:
## Description

Part of https://github.com/vaadin/components-team-tasks/issues/604

Migrate a few ITs from FW8 repo.

Note: 
- There are a few tests tagged with `@Ignore`. Most of them point to opened tickets and I decided to keep them, so it is easier to simply remove the `@Ignore` annotation while fixing these tickets.


DiegoCardoso (approved):
💪🏻

PR #3238, opened 2022-05-30
refactor: add multi select combo box integration tests

sissbruecker:
## Description

Adds a `MultiSelectComboBoxElement` TestBench element, and integration tests for the `MultiSelectComboBox`. The tests are mostly intended to cover the differences compared to the regular combo box component regarding value handling and the client-side connector.
The PR also introduces some necessary changes to the connector in order to support both components. The new multi select combo box filtering ITs are supposed to cover that change.

Part of https://github.com/vaadin/platform/issues/2961

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/MultiSelectComboBoxRefreshValueIT.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/MultiSelectComboBoxRefreshValueIT.java
@@ -1,0 +1,57 @@
+package com.vaadin.flow.component.combobox.test;
+
+import com.vaadin.flow.component.combobox.testbench.MultiSelectComboBoxElement;
+import com.vaadin.flow.testutil.TestPath;
+import com.vaadin.testbench.TestBenchElement;
+import com.vaadin.tests.AbstractComponentIT;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Ignore;
+import org.junit.Test;
+
+import java.util.List;
+import java.util.Set;
+
+@TestPath("vaadin-multi-select-combo-box/refresh-value")
+public class MultiSelectComboBoxRefreshValueIT extends AbstractComponentIT {
+    private MultiSelectComboBoxElement comboBox;
+    private TestBenchElement changeItemLabelGenerator;
+    private TestBenchElement changeItemData;
+
+    @Before
+    public void init() {
+        open();
+        comboBox = $(MultiSelectComboBoxElement.class).waitForFirst();
+        changeItemLabelGenerator = $("button")
+                .id("change-item-label-generator");
+        changeItemData = $("button").id("change-item-data");
+    }
+
+    @Test
+    public void selectItems_changeItemLabelGenerator_valueUpdated() {
+        comboBox.selectByText("Item 1");
+        comboBox.selectByText("Item 10");
+        changeItemLabelGenerator.click();
+
+        assertSelectedItems(Set.of("Custom Item 1", "Custom Item 10"));
+    }
+
+    @Test
+    @Ignore("https://github.com/vaadin/flow-components/issues/3239")
+    public void selectItems_changeItemData_valueUpdated() {
+        comboBox.selectByText("Item 1");
+        comboBox.selectByText("Item 10");
+        changeItemData.click();
+
+        assertSelectedItems(Set.of("Updated Item 1", "Updated Item 10"));
+    }
+
+    private void assertSelectedItems(Set<String> items) {
+        List<String> selectedTexts = comboBox.getSelectedTexts();
+        Assert.assertEquals("Number of selected items does not match",
+                items.size(), selectedTexts.size());
+        items.forEach(item -> Assert.assertTrue(
+                "Selection does not include item: " + item,
+                selectedTexts.contains(item)));
+    }
+}
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
@@ -17,11 +17,16 @@ import { ComboBoxPlaceholder } from '@vaadin/combo-box/src/vaadin-combo-box-plac
 
         comboBox.$connector = {};
 
-        // <vaadin-multi-select-combo-box> wraps a regular combo box internally,
-        // to reuse logic between both components we need to identify the
-        // element that implements the data provider mixin
-        const dataProviderMixin =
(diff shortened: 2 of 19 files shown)


vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -54,17 +60,17 @@ import { ComboBoxPlaceholder } from '@vaadin/combo-box/src/vaadin-combo-box-plac
         const clearPageCallbacks = (pages = Object.keys(pageCallbacks)) => {
           // Flush and empty the existing requests
           pages.forEach((page) => {
-            pageCallbacks[page]([], comboBox.size);
+            pageCallbacks[page]([], dataProviderMixin.size);
```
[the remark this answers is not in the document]
sissbruecker: Seems to be public for combo box (via data provider mixin), and we have some tests for combo-box that verify that you can set a predefined size. So yeah, I think we should adopt this.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -221,7 +227,7 @@ import { ComboBoxPlaceholder } from '@vaadin/combo-box/src/vaadin-combo-box-plac
         comboBox.$connector.reset = tryCatchWrapper(function () {
           clearPageCallbacks();
           cache = {};
-          comboBox.clearCache();
+          dataProviderMixin.clearCache();
```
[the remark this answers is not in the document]
sissbruecker: Agree, that makes sense.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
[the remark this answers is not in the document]
web-padawan: Submitted the PR to add `size` and `clearCache()`: https://github.com/vaadin/web-components/pull/3961

vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/MultiSelectComboBoxRefreshValueIT.java
```diff
@@ -0,0 +1,57 @@
...
+        comboBox.selectByText("Item 1");
+        comboBox.selectByText("Item 10");
+        changeItemLabelGenerator.click();
+
+        assertSelectedItems(Set.of("Custom Item 1", "Custom Item 10"));
+    }
+
+    @Test
+    @Ignore("https://github.com/vaadin/flow-components/issues/3239")
+    public void selectItems_changeItemData_valueUpdated() {
```
sissbruecker, line 41: Disabled for now, see https://github.com/vaadin/flow-components/issues/3239

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -306,11 +312,11 @@ import { ComboBoxPlaceholder } from '@vaadin/combo-box/src/vaadin-combo-box-plac
         // and submitting the filtered items to specified callback.
         // The filter used is the one from combobox, not the lastFilter stored since
         // that may not reflect user's input.
-        const performClientSideFilter = tryCatchWrapper(function (page, callback) {
+        const performClientSideFilter = tryCatchWrapper(function (page, filter, callback) {
           let filteredItems = page;
 
-          if (comboBox.filter) {
-            filteredItems = page.filter((item) => comboBox.$connector.filter(item, comboBox.filter));
```
sissbruecker, line 313: Had to change this, as the `filter` was only synchronized to `vaadin-multi-select-combo-box` after the data provider request, so the filter value was always lagging one character behind.

vaadin-combo-box-flow-parent/vaadin-combo-box-testbench/src/main/java/com/vaadin/flow/component/combobox/testbench/MultiSelectComboBoxElement.java
```diff
@@ -0,0 +1,193 @@
...
+     * Attempts to select an item from the popup by matching the label. Throws
+     * an {@link IllegalArgumentException} if the popup does not contain an item
+     * with the specified label. Does nothing if the item is already selected.
+     *
+     * @param label
+     *            The label of the item to select from the popup
+     * @throws IllegalArgumentException
+     *             if there is no item with the specified label
+     */
+    public void selectByText(String label) {
```
sissbruecker, line 90: Tried to align this, and the following methods, with the existing `HasSelectByText` interface that `ComboBoxElement` implements, even though the naming is not ideal. I tried to make it more clear in the API docs that this uses item labels. Implementing `HasSelectByText` is not possible, as this only handles a single string value, so the best we can do is provide something similar for users already familiar with `ComboBoxElement`.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -17,6 +17,12 @@ import { ComboBoxPlaceholder } from '@vaadin/combo-box/src/vaadin-combo-box-plac
 
         comboBox.$connector = {};
 
+        // <vaadin-multi-select-combo-box> wraps a regular combo box internally,
+        // to reuse logic between both components we need to identify the
+        // element that implements the data provider mixin
+        const dataProviderMixin =
+          comboBox.localName === 'vaadin-multi-select-combo-box' ? comboBox.$.comboBox : comboBox;
```
web-padawan, line 24: IIRC, when using `PolymerTemplate` and `@Id`, the `comboBox.$` might be not yet initialized because web component does not yet have shadow root created. See https://github.com/vaadin/vaadin-combo-box-flow/pull/223 

It would be nice to add ITs for `PolymerTemplate` and `LitTemplate` to the `template` folder just to verify that this does not cause bugs, and refactor this logic accordingly if needed.

tomivirkki: Good catch. This would indeed throw in the case a `ComboBox` is added as a virtual child (as `Dialog` content for example). The shadow root would not get created before the custom element class is finalized.

sissbruecker: Thanks for catching this, that did indeed not work. Updated the connector to resolve the data provider mixin element on demand. Added ITs for adding the component as a virtual child, through a Lit template, and through a Polymer template.

web-padawan (approved):

PR #3240, opened 2022-05-30
fix: do not use merged regions index from POI

### LINKED ISSUE #3186: [spreadsheet] Merge cells functionality is not working correctly
### Description

_Found when migrating MergeTests from FW8 to flow-components_

It seems that the merge functionality is broken on Spreadsheet, when trying to merge/unmerge cells from the context menu.

At a first glance it looks like the state of merged cells is not being correctly saved because merging cells work only on the second try (and only if the cells are different from the first try). And because if this inconsistency, trying to unmerge cells lead to unexpected results, like other cells being unmerged instead of the ones currently selected.

https://user-images.githubusercontent.com/262432/168300853-3cde50b1-5afb-4cb9-b1ee-6e609ef5ad95.mp4

### Expected outcome

User is able to merge/unmerge cells as expected

```java
var spreadsheet = new Spreadsheet();
add(spreadsheet);
```

### Steps to reproduce

1. Run the spreadsheet IT module with `mvn jetty:run`
2. Click on "Create new" button
3. Select two or more cells and then, click with the right button and select "Merge cells"
4. Repeat step (3) a few more times
5. Right click on a merged cell and select "Unmerge cells"
6. Repeat step (5) a few more times

### Environment

Vaadin version(s): 23.1-SNAPSHOT
OS: mac OS

### Browsers

Chrome, Edge

tomivirkki:
Fixes #3186 

While https://github.com/vaadin/flow-components/pull/3215 does some nice cleanup for the code, I feel like the test coverage is not yet sufficient enough for such a refactor.

This PR fixes the issue with fewer changes.

### DIFF
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/Spreadsheet.java
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/Spreadsheet.java
@@ -3316,7 +3316,7 @@ public class Spreadsheet extends Component implements HasComponents, HasSize,
 
     private void createMergedRegionIntoSheet(CellRangeAddress region) {
         Sheet sheet = getActiveSheet();
-        int addMergedRegionIndex = sheet.addMergedRegion(region);
+        sheet.addMergedRegion(region);
         MergedRegion mergedRegion = new MergedRegion();
         mergedRegion.col1 = region.getFirstColumn() + 1;
         mergedRegion.col2 = region.getLastColumn() + 1;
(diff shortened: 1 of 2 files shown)


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/Spreadsheet.java
```diff
@@ -3316,7 +3316,7 @@ public void addMergedRegion(CellRangeAddress region)
 
     private void createMergedRegionIntoSheet(CellRangeAddress region) {
         Sheet sheet = getActiveSheet();
-        int addMergedRegionIndex = sheet.addMergedRegion(region);
```
tomivirkki, line 3319: Seems that the index received from the POI API (`addMergedRegion`) does not work well as an index for the local `mergedRegions` list:
- The index from `addMergedRegion` always increases when new regions are added, even if some regions have been removed earlier.
- The local `mergedRegion` always reflects the true set of merged regions

DiegoCardoso (approved):
Nice 🎉

PR #3242, opened 2022-05-30
fix: do not trigger item click listener for prevented click

### LINKED ISSUE #3177: Cannot use Select in editable Grid
### Description

When clicking on the select button inside of an editable grid, the select component opens and closes immediately

### Expected outcome

I would expect that the click event does not propagate (and the ItemClickEvent is not fired), so that the select remains open.

```
    Binder<Bean> binder = new Binder<>(Bean.class);
    Select<String> select = new Select<>();
    select.setItems("foo", "bar");

    Grid<Bean> grid = new Grid<>();
    grid.addColumn(Bean::getValue).setEditorComponent(select);

    grid.addItemClickListener(ev -> grid.getEditor().editItem(ev.getItem()));
    binder.forField(select).bind(Bean::getValue, Bean::setValue);
    grid.getEditor().setBinder(binder);
    grid.setItems(new Bean());

    add(grid);
 ```
select

### Steps to reproduce

1. Double click the item row in order to enter edition mode
2. Click the select button in order to open the overlay

### Environment

Vaadin version(s): 23.0.9

### Browsers

_No response_

### PR DESCRIPTION
**Pull request by web-padawan**
## Description

Fixes #3177

This PR updates item click listener to not fire in case if `event.preventDefault()` was called for the `click` event on the client.
It's especially relevant for the Grid editor because of the following logic that closes the editor:

https://github.com/vaadin/flow-components/blob/6aefadf2db230a0a120c3b25122aea79c9d9d5c2/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/editor/EditorImpl.java#L67

https://github.com/vaadin/flow-components/blob/6aefadf2db230a0a120c3b25122aea79c9d9d5c2/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/editor/EditorImpl.java#L199-L205

Clarification about https://github.com/vaadin/flow-components/issues/3177#issuecomment-1125849666: the [`isFocusable` check](https://github.com/vaadin/flow-components/blob/6aefadf2db230a0a120c3b25122aea79c9d9d5c2/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js#L1131) in the connector prevents click listener from firing on `<input>` element. But the toggle button of `vaadin-select` and `vaadin-combo-box` is not focusable, so it caused click listener to fire.

## Type of change

- Bugfix

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridViewClickListenersPage.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridViewClickListenersPage.java
@@ -18,7 +18,7 @@ package com.vaadin.flow.component.grid.it;
 import com.vaadin.flow.component.grid.Grid;
 import com.vaadin.flow.component.grid.Grid.SelectionMode;
 import com.vaadin.flow.component.html.Div;
-import com.vaadin.flow.component.html.NativeButton;
+import com.vaadin.flow.component.html.Span;
 import com.vaadin.flow.data.bean.Person;
 import com.vaadin.flow.data.renderer.ComponentRenderer;
 import com.vaadin.flow.router.Route;
@@ -40,10 +40,10 @@ public class GridViewClickListenersPage extends LegacyTestView {
         grid.addColumn(Person::getFirstName).setHeader("Name");
         grid.addColumn(Person::getAge).setHeader("Age");
         grid.addColumn(new ComponentRenderer<>(person -> {
-            NativeButton button = new NativeButton("Action");
-            button.getElement().executeJs(
+            Span span = new Span("Action");
+            span.getElement().executeJs(
                     "$0.addEventListener('click', e => e.preventDefault())");
-            return button;
+            return span;
         })).setHeader("Action");
 
         // Disable selection: will receive only click events instead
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridViewClickListenersPage.java [hunk 1]
```diff
@@ -37,6 +39,12 @@ private void createClickListener() {
         grid.setItems(getItems());
         grid.addColumn(Person::getFirstName).setHeader("Name");
         grid.addColumn(Person::getAge).setHeader("Age");
+        grid.addColumn(new ComponentRenderer<>(person -> {
+            NativeButton button = new NativeButton("Action");
+            button.getElement().executeJs(
+                    "$0.addEventListener('click', e => e.preventDefault())");
```
tomivirkki, line 45: The event gets blocked by something else because the test case passes even without this event listener in place.

Looks like it's because `button` is listed [here](https://github.com/vaadin/web-components/blob/1f9603fcba7d56f16b49e211339f802cd51ae7c0/packages/grid/src/vaadin-grid-active-item-mixin.js#L111)

tomivirkki (changes requested):

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridViewClickListenersPage.java
web-padawan: Good catch, thanks. Updated to use `span` instead.

tomivirkki (approved):

PR #3167, opened 2022-05-12
refactor: improve performance of AbstractGridMultiSelectionModel#isSelected

pepijnve:
## Description

Query the presence of the ID of the given item in the `selected` Map instead of iterating over the value set.
Please list all relevant dependencies in this section and provide summary of the change, motivation and context.

Closes #3166

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

- [x] I have read the contribution guide: https://vaadin.com/docs-beta/latest/guide/contributing/overview/
- [x] I have added a description following the guideline.
- [x] The issue is created in the corresponding repository and I have referenced it.
- [ ] I have added tests to ensure my change is effective and works as intended.
- [ ] New and existing tests are passing locally with my change.
- [x] I have performed self-review and corrected misspellings.

#### Additional for `Feature` type of change

- [ ] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.

PR #3247, opened 2022-05-30
Add AbstractGridMultiSelectionModel#getSelectedItemIds

### LINKED ISSUE #3246: Allow subclasses to access the key set of AbstractGridMultiSelectionModel#selected
### Describe your motivation

The subclassed selection model we use in our application implements logic to fetch a recent copy of each selected item based on the item id rather than using the (possible stale) copy that's referenced by the selection model. To avoid unnecessary work it would be nice to be able to get hold of the key set of the internal selection hash map.

Add a protected final method

```
protected final Set<Object> getSelectedItemIds() {
        return Collections.unmodifiableSet(this.selected.keySet());
}
```

Implement the same logic as (stream for brevity)

```
getSelectedItems().stream().map(getGrid().getDataCommunicator().getDataProvider()::getId).collect(Collectors.toSet())
```

This will effectively do the same thing, but causes quite a bit of unnecessary work.

### Additional context

_No response_

pepijnve:
## Description

Add a method, usable by subclasses, that provides access to the internal set of selected item ids.

Fixes #3246

## Type of change

- [ ] Bugfix
- [x] Refactoring
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridMultiSelectionModel.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridMultiSelectionModel.java
@@ -159,6 +159,17 @@ public abstract class AbstractGridMultiSelectionModel<T>
                 .unmodifiableSet(new LinkedHashSet<>(selected.values()));
     }
 
+    /**
+     * Returns an unmodifiable view of the selected item ids.
+     *
+     * The returned Set may be a direct view of the internal data structures of this class.
+     * A defensive copy should be made by callers when iterating over this Set and modifying
+     * the selection during iteration to avoid ConcurrentModificationExceptions.
+     */
+    protected final Set<T> getSelectedItems() {
+        return Collections.unmodifiableSet(this.selected.keySet());
+    }
+
     @Override
     public Optional<T> getFirstSelectedItem() {
         return selected.values().stream().findFirst();


vursen:
Hello @pepijnve, thank you for your contribution! 

However, adding a protected helper method that is not used by the class doesn't sound right to me. I guess we could make it a public API that would also require test coverage, but then it becomes a feature that is not a high priority now. 

Given that there is a workaround, I'm closing this PR.

pepijnve:
Closing one-sidedly is a bit blunt, no? A workaround exists, but is quite suboptimal in terms of performance. I only added this for completeness' sake.
Our application deals with very large grids and selection has proven to be quite the bottleneck (see #3167, and https://github.com/vaadin/web-components/issues/3844). This PR is the last piece of the optimisation puzzle. The actual real workaround we're currently using is to access the internal selected HashMap via reflection, which is something I would like to be able to get rid of.

I understand your comment about the protected method that's not being used by the class itself. I've updated AbstractGridMultiSelectionModel to use the accessor method instead of calling `keySet` directly. That 'solves' this problem, but perhaps it's a bit contrived.

I've added some test cases for the added method in the meantime as well.

yuriy-fix:
Dear @pepijnve ,

Provided protected method in this PR suits this particular case and not intended for wide use with other use-cases. It might easily get removed later during refactoring by future developers because of the reasons mentioned.

It might make sense to come up with a public API that would help in this case and also cover it with tests.
If you want, we could proceed with this option and try to prepare a proper Acceptance Criteria for the feature request.

yuriy-fix:
Dear @pepijnve ,

Provided protected method in this PR suits this particular case and not intended for wide use with other use-cases. It might easily get removed later during refactoring by future developers because of the reasons mentioned.

It might make sense to come up with a public API that would help in this case and also cover it with tests.
If you want, we could proceed with this option and try to prepare a proper Acceptance Criteria for the feature request.

pepijnve:
@yuriy-fix FYI, I've massaged the code a bit further so that the added methods are actually used by the implementation itself as well. That change and the added tests are not showing up here since the PR has been marked as closed.

They're still marked as protected for now. The goal here is improved performance. For that reason I want to avoid creating a defensive copy up front. That differs a bit from what `getSelectedItems()` does though. Having two very similar public methods that differ in this regard felt like a bad idea.

yuriy-fix:
Dear @pepijnve,

Thank you for your effort!

We can make an explicit review for the changes you've mentioned (improved performance where methods are used by the implementation). Feel free to open a PR and mention me on that one.

willemv:
@yuriy-fix As requested, I've created a new PR #3438

PR #3250, opened 2022-05-31
fix: ensure the hidden grid sorters are reset

### LINKED ISSUE #1008: Grid shows invalid sort order numbering after hiding previously sorted column
### Description
Vaadin Flow 19.0.2

When hiding a column that was previously part of a multi-sort (but was removed from it during the same action), the remaining columns get wrong sort numbering in header.
[Code example](https://github.com/vaadin/vaadin-grid/files/6270464/HelloWorldView.java.txt)
image
image

### Expected outcome
Sort numbering resets properly.

### Actual outcome
Sort numbering is wrong.

### Steps to reproduce
1. Add a `Grid` component to a page.
2. Two or more sortable columns with headers.
3. Enable multi-sort `grid.setMultiSort(true)`
4. Sort by two of the columns.
5. Hide one of the columns sorted by.
6. Sort by just one column.

### Workaround
```
// after changing which columns are ordered by
if (columnShouldBeVisible) {
    column.setVisible(true);
} else {
    UI.getCurrent().getPage().executeJs("").then(result -> column.setVisible(false));
}
```

### Browsers Affected
- [x] Chrome
- [x] Firefox

web-padawan:
## Description

Current logic for updating sorters in the connector only updates `vaadin-grid-sorter` elements that are added to the DOM.
When the column gets hidden while also calling `grid.sort(null)`, the connector does not update its sorter, because of not checking the actually applied sorters stored in `grid._sorters`. This causes the problem with restoring sort direction.

Fixes #1008

## Type of change

- Bugfix

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridViewSortingIT.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridViewSortingIT.java
@@ -128,11 +128,10 @@ public class GridViewSortingIT extends AbstractComponentIT {
         assertSortMessageEquals(QuerySortOrder.asc("firstName").build(), true);
 
         clickElementWithJs(toggleFirstColumnButton);
-        assertSortMessageEquals(QuerySortOrder.desc("firstName").build(),
-                false);
+        assertSortMessageEquals(Collections.emptyList(), false);
 
         clickElementWithJs(toggleFirstColumnButton);
-        assertSortMessageEquals(QuerySortOrder.asc("firstName").build(), false);
+        assertSortMessageEquals(Collections.emptyList(), false);
     }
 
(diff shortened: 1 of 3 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridViewSortingIT.java
```diff
@@ -116,6 +116,24 @@ public void gridWithSorting_invertAndResetSortings() {
 
     }
 
+    @Test
+    public void gridWithSorting_toggleColumnVisibilityAndReset() {
```
tomivirkki, line 120: The change itself works well, but the test case passes even without the fix in place

web-padawan: Thanks, added the check for the actual `direction` attribute on the `vaadin-grid-sorter` element.

tomivirkki (approved):

PR #3280, opened 2022-06-03
refactor: remove shared state from avatar tests

sissbruecker:
## Description

Remove shared state from avatar unit tests, and inline some constants.


vaadin-avatar-flow-parent/vaadin-avatar-flow/src/test/java/com/vaadin/flow/component/avatar/tests/AvatarGroupTest.java
```diff
@@ -21,56 +21,52 @@
...
-    private String abbr = "fb";
-    private String imgUrl = "https://vaadin.com/";
-    private Integer colorIndex = 3;
+    private AvatarGroup avatarGroup;
+    private AvatarGroupItem avatarGroupItem1;
+    private AvatarGroupItem avatarGroupItem2;
 
-    private List<AvatarGroupItem> items = new ArrayList<>();
+    @Before
+    public void setUp() {
```
vursen, line 36: **nit:** `init` seems to be a more common naming.

```suggestion
    public void init() {
```

vursen (approved):

vaadin-avatar-flow-parent/vaadin-avatar-flow/src/test/java/com/vaadin/flow/component/avatar/tests/AvatarGroupTest.java
sissbruecker: As discussed, `setup` is more common for unit tests in general, while `init` is more common for ITs, probably because `setup` is already used in the class hierarchy.

PR #3292, opened 2022-06-06
fix: update sorters only if grid is attached

### LINKED ISSUE #1672: Grid sorting errors in dialog
### Description

I place a Grid (with multi sorting enabled) inside a Dialog, and set sorting on multiple columns of the Grid.
I set the same sorting each time before opening the Dialog.
I tested it with a Dialog manually attached to the UI before opening it, and with a Dialog that is not attached to the UI.

### Vaadin version

I'm using Vaadin 14.0.2

### Expected behavior

Each time the Dialog opens the Grid is sorted in the way the program sets it.
It should not matter if the Dialog is manually attached to the UI or not. Also should not matter how many times the dialog is opened.

### Actual behavior

- When the Dialog is manually attached to UI before opening it:
  - opening the first time: there is no sorting
  - opening the second time: the column header shows sorting, but items are not sorted
  - opening the third time: the items are sorted, but the column order of the sorting is wrong (backwards; sorts the last sorted column first)
- When the Dialog is not attached manually to the UI:
  - opening the first time: sorting as expected
  - opening the second time: there is no sorting 
  - opening the third time: there is no sorting and the following error is shown: "(TypeError) : Cannot read property 'setSorterDirections' of undefined"

The second part is already reported in vaadin/vaadin-grid-flow#747, but I include it here for the sake of completeness.

In the following example the "X asc, Y asc" sorting is set before each dialog open (see code below):

GridSortingInDialog

### Code to reproduce

```
package test;

import java.util.Arrays;
import java.util.concurrent.atomic.AtomicInteger;
import java.util.stream.IntStream;

import com.vaadin.flow.component.Composite;
import com.vaadin.flow.component.dialog.Dialog;
import com.vaadin.flow.component.grid.Grid;
import com.vaadin.flow.component.grid.GridSortOrder;
import com.vaadin.flow.component.html.Label;
import com.vaadin.flow.component.orderedlayout.FlexComponent.Alignment;
import com.vaadin.flow.data.provider.SortDirection;

@Route(value = "GridSortingInDialog")
public class GridSortingInDialog extends Composite<VerticalLayout> {

	public GridSortingInDialog() {
		getContent().setDefaultHorizontalComponentAlignment(Alignment.CENTER);
		// Grid in pre-attached Dialog:
		init(true);
		// grid in non-attached Dialog
		init(false);
	}

	private void init(boolean attachDialog) {
		Grid<Data> grid = new Grid<>(Data.class);
		grid.setWidthFull();
		grid.setMultiSort(true);
		grid.setSortableColumns("x", "y");
		grid.setItems(IntStream.range(0, 100).mapToObj(i -> new Data()));
		sort(grid);

		VerticalLayout vl = new VerticalLayout();
		vl.setSpacing(true);
		Dialog dialog = new Dialog(vl);
		dialog.setWidth("20rem");
		dialog.setHeight("30rem");

		if (attachDialog) {
			vl.add(new Label("Pre-attached dialog"));
			getContent().add(dialog);
		} else {
			vl.add(new Label("Standard dialog"));
		}
		vl.add(grid);

		String btnLabel = "Open" + (attachDialog ? " pre-attached" : "") + " dialog";
		AtomicInteger openingCounter = new AtomicInteger();
		Button btn = new Button(btnLabel);
		btn.addClickListener(evt -> {
			sort(grid);
			dialog.open();
			btn.setText(btnLabel + " (" + openingCounter.incrementAndGet() + ")");
		});

		getContent().add(btn);
	}

	private static void sort(Grid<Data> grid) {
		grid.sort(Arrays.asList(
			new GridSortOrder<>(grid.getColumnByKey("x"), SortDirection.ASCENDING),
			new GridSortOrder<>(grid.getColumnByKey("y"), SortDirection.ASCENDING)));
	}

	public static class Data {

		private int x = (int) Math.round(Math.random() * 10);
		private int y = (int) Math.round(Math.random() * 10);

		public int getX() {
			return x;
		}

		public int getY() {
			return y;
		}
	}
}
```

web-padawan:
## Description

Added a check to ensure calling `grid.sort()` while grid is detached does not throw a JS error upon reattach.

Fixes #1672

## Type of change

- Bugfix

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/DetachReattachPage.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/DetachReattachPage.java
@@ -28,7 +28,7 @@ public class DetachReattachPage extends Div {
     public DetachReattachPage() {
         Grid<String> grid = new Grid<String>();
         grid.setItems("A", "B", "C");
-        grid.addColumn(x -> x).setHeader("Col");
+        grid.addColumn(x -> x).setHeader("Col").setSortable(true);
 
         grid.setSelectionMode(Grid.SelectionMode.SINGLE);
 
@@ -62,7 +62,13 @@ public class DetachReattachPage extends Div {
         toggleDetailsVisibleOnClick
                 .setId("toggle-details-visible-click-button");
 
+        NativeButton resetSortingButton = new NativeButton("Reset sorting",
+                e -> {
+                    grid.sort(null);
+                });
+        resetSortingButton.setId("reset-sorting-button");
+
         add(btnAttach, btnDetach, btnDisallowDeselect, addItemDetailsButton,
-                toggleDetailsVisibleOnClick, grid);
+                toggleDetailsVisibleOnClick, resetSortingButton, grid);
     }
 }
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/DetachReattachIt.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/DetachReattachIt.java
@@ -95,6 +95,25 @@ public class DetachReattachIt extends AbstractComponentIT {
 
     }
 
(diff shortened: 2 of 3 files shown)


vursen (approved):

vursen:
Verified that it is enough to just skip a client-side update of the sorter while the component is detached: after the grid is attached, it ensures the same logic is called.

PR #3312, opened 2022-06-09
refactor: set data provider related properties on the host element

vursen:
## Description

Pretty much all the properties related to data provider are available directly on the host combo-box / multi-select-combo-box element so there seems to be no reason for setting them through internal elements bypassing the host element. 

~~The only case when accessing internal elements is reasonable is when calling data provider methods such as `clearCache()`.~~ UPD: The multi-select-combo-box's host element already provides `clearCache()`.

Related to https://github.com/vaadin/web-components/pull/4019#pullrequestreview-1001168796

## Type of change

- [x] Internal

## Checklist

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
@@ -79,7 +79,7 @@ import { ComboBoxPlaceholder } from '@vaadin/combo-box/src/vaadin-combo-box-plac
             // filter based on comboBox.filter. While later we only filter clientside data.
 
             if (cache[0]) {
-              performClientSideFilter(cache[0], callback);
+              performClientSideFilter(cache[0], params.filter, callback);
               return;
             } else {
               // If client side filter is enabled then we need to first ask all data
@@ -284,7 +284,7 @@ import { ComboBoxPlaceholder } from '@vaadin/combo-box/src/vaadin-combo-box-plac
           let data = cache[page];
 
           if (comboBox._clientSideFilter) {
-            performClientSideFilter(data, callback);
+            performClientSideFilter(data, comboBox.filter, callback);
           } else {
             // Remove the data if server-side filtering, but keep it for client-side
             // filtering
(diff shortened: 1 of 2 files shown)


vursen:
Some multi-select-combo-box tests failed apparently because the `filteredItems` property of the internal combo-box is only 1-way bound to the host element.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -316,11 +301,10 @@ import { ComboBoxPlaceholder } from '@vaadin/combo-box/src/vaadin-combo-box-plac
         // The filter used is the one from combobox, not the lastFilter stored since
         // that may not reflect user's input.
         const performClientSideFilter = tryCatchWrapper(function (page, callback) {
-          const dataProviderMixin = getDataProviderMixin();
           let filteredItems = page;
 
-          if (dataProviderMixin.filter) {
-            filteredItems = page.filter((item) => comboBox.$connector.filter(item, dataProviderMixin.filter));
+          if (comboBox.filter) {
+            filteredItems = page.filter((item) => comboBox.$connector.filter(item, comboBox.filter));
```
sissbruecker, line 307: This change brings up a problem that the filter property is synchronized **after** the data provider call, which causes the filtered results to lag one character behind:

https://user-images.githubusercontent.com/357820/173076458-6779f213-44f1-4aa2-9b62-02736668457a.mp4

I had the same issue initially when adapting the connector, and had a solution where I passed the actual filter from the data provider callback into the `performClientSideFilter` call. I reverted that because it was not necessary anymore after using the data provider element directly. So maybe can try something like that as an alternative here.

vursen: Done.

vursen: Seems that the change broke some other tests.

vursen: This is fixed.

web-padawan (approved):

PR #3319, opened 2022-06-10
fix: informative exception thrown if grid row index is out of bounds #3111

### LINKED ISSUE #3111: GridElement.getRow() fails with NPE on non-existing row
### Description

Calling `GridElement.getRow(0)` on an empty Grid fails with a NPE:
```java
java.lang.NullPointerException
	at com.vaadin.flow.component.grid.testbench.GridElement.getRow(GridElement.java:194)
	at com.vaadin.flow.component.grid.testbench.GridElement.select(GridElement.java:343)
```
It should either return null, or even better fail with a more informative exception (e.g. `IndexOutOfBoundsException`)

### Expected outcome

`IndexOutOfBoundsException` should have been thrown

```java

    public MainView() {
        Grid<String> grid = new Grid<String>();
        grid.setItems();
        grid.setId("grid");
        grid.addColumn(it -> it).setHeader("foo");
        add(grid);
    }
}

public class MainViewIT extends AbstractViewTest {
    @Test
    public void myTest() {
        GridElement grid = $(GridElement.class).first();
        grid.select(0);
    }
}
```

### Steps to reproduce

Paste the classes into the "skeleton starter flow" project, then run the IT tests

### Environment

Vaadin version(s): 14.7.5
Reproduced on Firefox, untested on other browsers

### Browsers

Firefox

ugur-vaadin:
Calling GridElement.getRow(0) for an empty grid caused a NPE. The issue was also present when calling GridElement.getRow(int index) for any index that was out of the current range of row indexes.

This issue is a TestBench issue, therefore the impact is low.

This change introduces an index validation before trying to get the relevant row. If the input index is not within the current range of grid row indexes, an IndexOutOfBoundsException is thrown with a message that also specifies the expected index range. 
Since GridElement.getRow(int index) now can throw an IndexOutOfBoundsException for invalid index, this change was also reflected onto TreeGridElement.hasRow(int row). The logic in that method depended on the assumption that if there is no row, the method GridElement.getRow(int index) will throw a NPE. With this change, it was changed to Exception in order to account for the new exception.

Fixes #3111

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-testbench/src/main/java/com/vaadin/flow/component/grid/testbench/GridElement.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-testbench/src/main/java/com/vaadin/flow/component/grid/testbench/GridElement.java
@@ -181,9 +181,15 @@ public class GridElement extends TestBenchElement {
      *
      * @param rowIndex
      *            the row index
+     * @throws IndexOutOfBoundsException
+     *            if no row with given index exists
      * @return the tr element for the row
      */
-    public GridTRElement getRow(int rowIndex) {
+    public GridTRElement getRow(int rowIndex) throws IndexOutOfBoundsException {
+        int rowCount = getRowCount();
+        if (rowIndex < 0 || rowIndex >= rowCount) {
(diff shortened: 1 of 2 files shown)


yuriy-fix (approved):
Looks great! Please make sure build is passing before merging.

yuriy-fix (approved):

PR #3340, opened 2022-06-14
fix: update initialSheetSelection on reload

### LINKED ISSUE #3230: [spreadsheet] Selection state is lost on sheet change
### Description

When you select cells on a sheet, leave the sheet, and then return back to the sheet, the selection state is lost.

https://user-images.githubusercontent.com/1222264/170250577-f487b1e4-2f5c-4cb8-84e0-547aaa0564c9.mp4

tomivirkki:
Fixes #3230

### DIFF
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/Spreadsheet.java
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/Spreadsheet.java
@@ -3789,7 +3789,12 @@ public class Spreadsheet extends Component implements HasComponents, HasSize,
         clearSheetOverlays();
         topLeftCellCommentsLoaded = false;
 
-        setReload(true);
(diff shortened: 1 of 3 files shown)


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/Spreadsheet.java
```diff
@@ -3985,15 +3990,7 @@ protected SpreadsheetClientRpc getRpcProxy() {
         return clientRpc;
     }
 
-    /*
-     * (non-Javadoc)
-     *
-     * @see com.vaadin.ui.AbstractComponent#beforeClientResponse(boolean)
-     */
-    // @Override
-    public void beforeClientResponse(boolean initial) {
```
tomivirkki, line 3994: The root problem here was that this method never got called. In V8, it was automatically called because the method was overridden from AbstractComponent. For Flow Spreadsheet, I needed to add an explicit invocation.

DiegoCardoso (approved):

PR #3343, opened 2022-06-15
 fix: cleanup orphan sorters after reordering columns

### LINKED ISSUE #2102: setColumnOrder() on Grid in Vaadin Flow causes the getSortOrder() list to increment infinitely in the event of the sort listener [2 days]
### Description

This was opened as requested by Tatu Lund (https://github.com/TatuLund) from the following StackOverFlow question: https://stackoverflow.com/questions/69096844/setcolumnorder-on-grid-in-vaadin-flow-seems-to-cause-the-getsortorder-list-t

I have also created a Github repo which you can run to see the bug/issue directly at: https://github.com/FollowSteph/VaadinGridBugReport To replicate just change the sort order once then click on the Reset button a number of times. Then change the sort order once more and you'll find that the List of GridSortOrder just keeps going up to match the number of times the reset button is clicked (multiplied by the number of GridSortOrder passed to grid.sort() where instead it should be two after the Reset button is clicked. This only happens if the grid.setColumnOrder() is also called.

To quote the stackoverflow question (there's a lot more details in the question): "The main issue is that the GridSortOrder list just keeps incrementing in size for the event in the sort listener rather than replacing the sortorder list. I believe this is a bug within Vaadin 14-21 because it only happens if I call setColumnOrder() in the reset method below (removing that call it works as expected). In other words calling setColumnOrder() seems to add the list of GridSortOrder to the event in the sort listener when it should be replacing it. This is actually also true if you do grid.getSortOrder() in the listener, meaning it's not just the event but also the grid that is compromised." Then further "If I remove grid.setColumnOrder() from the above then the List<GridSortOrder> works as expected and is two in the event listener. " And "With that in mind if you debug grid.setColumnOrder() it seems to call updateClientSideSorterIndicators(sortOrder) which is what I believe is the cause and is where it's being added to the GridSortOrder list."

### Expected outcome

See above.

### Actual outcome

See above.

### Live Demo (optional)

https://github.com/FollowSteph/VaadinGridBugReport

https://github.com/FollowSteph/VaadinGridBugReport

### Steps to reproduce

See above along with sample code. 

### Environment

Vaadin 14-21. 

### Browsers Affected

All browsers seem to be affected.

web-padawan:
## Description

This PR fixes the edge case that occurs when you set the same column order twice in a row.
See https://github.com/vaadin/flow-components/issues/2102#issuecomment-1154860141 for an overview of the issue.

In short, when we detach and immediately re-attach columns so that they end up in the DOM order as before, the  `MutationObserver` for children is not triggered, and the `grid._sorters` array is not cleaned properly.

UPD: this PR was originally about not setting the same column order twice, but that would break the valid behavior (restore the correct order on the server after reordering on the client). So I ended up with the connector refactoring instead.

Closes #2102

## Type of change

- Bugfix

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridViewSortingPage.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridViewSortingPage.java
@@ -53,14 +53,15 @@ public class GridViewSortingPage extends LegacyTestView {
                 .thenComparing(
                         (Person person) -> person.getAddress().getNumber());
 
-        grid.addColumn(LitRenderer.<Person> of(
-                "<div>${item.street}, number ${item.number}<br><small>${item.postalCode}</small></div>")
-                .withProperty("street",
-                        person -> person.getAddress().getStreet())
-                .withProperty("number",
-                        person -> person.getAddress().getNumber())
-                .withProperty("postalCode",
-                        person -> person.getAddress().getPostalCode()))
+        Grid.Column<Person> addressColumn = grid
+                .addColumn(LitRenderer.<Person> of(
+                        "<div>${item.street}, number ${item.number}<br><small>${item.postalCode}</small></div>")
+                        .withProperty("street",
+                                person -> person.getAddress().getStreet())
+                        .withProperty("number",
+                                person -> person.getAddress().getNumber())
+                        .withProperty("postalCode",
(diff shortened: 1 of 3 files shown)


web-padawan (commented):
Looks like this breaks another IT: `clientReorderColumns_serverResetOrder`. Taking a look.

web-padawan:
Updated the PR to cleanup orphan sorters in the connector, this seems to be a proper fix for the original issue.

web-padawan:
Closing in favor of https://github.com/vaadin/web-components/pull/4068. Confirmed that the web-component fix addressees the issue.

PR #3358, opened 2022-06-17
fix: do not refresh datacommunicator with previously selected item on select

### LINKED ISSUE #3229: Selected record not refreshed in Grid with refreshAll

tomivirkki:
Fixes #3229


web-padawan (approved):
The problematic logic originates from https://github.com/vaadin/vaadin-grid-flow/pull/77 but I was unable to understand the need for it based on the review comments to that PR. Removing this code looks reasonable to me.

PR #3383, opened 2022-06-23
feat: automatically include utility styles when using LumoUtility class

TatuLund:
This would greatly help DX of the feature as now you need to add Lumo utility css in your project yourself before this Java API is functional. If we have the importing in the class, the css is included automatically when you use the API.


vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/LumoUtility.java
```diff
@@ -18,6 +18,8 @@
 /**
  * Contains the definition for all the CSS utility classes provided by Lumo.
  */
+@CssImport("./lumo-utility-classes.css")
+@NpmPackage(value = "lumo-css-framework", version = "^4.0.10")
```
knoobie, line 25: If I remember correctly, this npm package isn't the current lumo utility classes. It was more like an experiment / proof of concept / idea by @anezthes before it was officially added inside the web component's repo

web-padawan (changes requested):
Please update to use correct version of Lumo utility classes:

```java
@NpmPackage(value = "@vaadin/polymer-legacy-adapter", version = "23.2.0-alpha2")
@JsModule("@vaadin/vaadin-lumo-styles/utility.js")
```

vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/LumoUtility.java
web-padawan: Thanks. The original `lumo-css-framework` npm package should be probably deprecated to make it more clear.

sissbruecker:
Outcome from grooming:
- check that styles are not duplicated if the application also has the utility module enabled through `theme.json`. If this would result in duplicates that would be a blocker for this change
- improve JavaDoc to note that just importing this class somewhere will import all CSS utility classes and define them in the global namespace, possible conflicting with user-defined classes
- update documentation to state that just importing this class is enough (instead of configuring `theme.json`), note that just importing this class somewhere will import all CSS utility classes and define them in the global namespace, possible conflicting with user-defined classes

sissbruecker:
Did some testing around this:
- The approach would indeed add duplicate declarations of the utility classes if the `utility` module is enabled in `theme.json` . Theoretically it would be possible to check if the CSS has already been added, but that would have to be aligned with the logic that the generated Flow frontend code uses to check for duplicates.
- The `@JsModule` annotation does not work correctly for production builds. That might be some optimization in the Java compiler or a limitation of Flow, but as long as the application only accesses a string constant such as `LumoUtility.Border.ALL`, the referenced frontend file is never loaded in the browser. On the other hand it works if the class is instantiated somewhere in the application (`new LumoUtility()`).

web-padawan:
> The @JsModule annotation does not work correctly for production builds. That might be some optimization in the Java compiler or a limitation of Flow, but as long as the application only accesses a string constant such as LumoUtility.Border.ALL, the referenced frontend file is never loaded in the browser.

This sounds like a blocker to me. Let's discuss with the Flow team what could we do.

web-padawan:
Regarding enums, scanning for cases like `VaadinIcon.CALENDAR.create()` was added in https://github.com/vaadin/flow/pull/5649.
I'm not sure what can we do about this, as we don't invoke any method when adding utility classes.

Legioth:
The reference to `LumoUtility` should be present in the applicatinon's compiled bytecode, so I guess it's just a matter of adding support for that particular pattern in Flow's bytecode scanner.

Legioth:
> should be present in the applications's compiled bytecode

Actually not. It seems like the compiler is in the simple cases inlining the string constant. This can be worked around by making the string a non-constant expression by using a no-op operation such as `public static final String AROUND = "content-around".trim()`. Even with this workaround, it seems like the scanner would still also have to be supplemented to look at annotations on the outer class (i.e. `LumoUtility`) and not only on the inner class (e.g. `LumoUtiliy.BORDER`).

web-padawan (commented):
One more use case that is not covered by this PR is injecting Lumo CSS into shadow DOM, which is done by Flow when using `"lumoImports"` - see https://github.com/vaadin/flow/pull/9842. 

This problem is not specific to this PR though, because existing `lumo-includes.ts` file does the same for color and typography.

It just means that if someone removes `utility` from `lumoImports`, they won't be able to use Lumo utility classes e.g. in case of exported web components.

Legioth:
The reference to the outer class can be worked around in this PR without needing any changes in the Flow scanner by introducing an artificial type reference to the outer class. Taken together, a working structure for the inner classes would thus be like this:

```java
public static final class Accessibility {
  private final LumoUtility dummyBackReference = null;
  public static final String SCREEN_READER_ONLY = "sr-only".trim();
  private Accessibility() {
  }
}
```

TatuLund:
@Legioth thanks!

Legioth:
The back reference is probably not enough. I think it's also necessary to make it so that the compiler doesn't view the field values as compile-time constants that can be inlined. That's why there was also a dummy `.trim()` on the string value in my example.

Now when I think about it, it might be clearer to have something like this:
```java
public class LumoUtility {
  private static String notConstant(String value) {
      return value;
  }
  
  public static final class Accessibility {
    public static final String SCREEN_READER_ONLY = notConstant("sr-only");
    private Accessibility() {
    }
  }
}
```

In this way, the back reference is automatic (through referencing the method declared in `LumoUtility`)  and it also provides a natural location to document the purpose of `notConstant` compared to the more opaque approach with a `.trim()` invocation that doesn't seem to have any purpose without a separate explanation for each line.

TatuLund:
@Legioth Ok, I applied that change to PR.

vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/LumoUtility.java
```diff
@@ -15,17 +15,15 @@
...
 
-import com.vaadin.flow.component.dependency.JsModule;
-import com.vaadin.flow.component.dependency.NpmPackage;
-
 /**
  * Contains the definition for all the CSS utility classes provided by Lumo.
  */
-@NpmPackage(value = "@vaadin/vaadin-lumo-styles", version = "23.2.0-alpha2")
-@JsModule("@vaadin/vaadin-lumo-styles/utility.js")
-@JsModule("./lumo-utility.ts")
```
web-padawan: @TatuLund Why have you removed these annotations?

vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/LumoUtility.java
```diff
@@ -34,52 +32,47 @@ private LumoUtility() {
...
         private Accessibility() {
         }
 
     }
 
     /**
-     * Classes for distributing space around and between items along a flexbox’s
-     * cross axis or a grid’s block axis. Applies to flexbox and grid layouts.
+     * Classes for distributing space around and between items along a
+     * flexboxâ€™s cross axis or a gridâ€™s block axis. Applies to flexbox and
```
web-padawan, line 47: Please revert this change, it contains some incorrect characters.

vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/LumoUtility.java
```diff
@@ -34,52 +32,47 @@ private LumoUtility() {
...
 
         private AlignContent() {
         }
 
     }
 
     /**
-     * Classes for aligning items along a flexbox’s cross axis or a grid’s block
-     * axis. Applies to flexbox and grid layouts.
+     * Classes for aligning items along a flexboxâ€™s cross axis or a gridâ€™s
```
web-padawan, line 66: Please revert this change, it contains some incorrect characters.

vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/LumoUtility.java
```diff
@@ -238,15 +233,13 @@ private BoxShadow() {
 
     /**
      * Classes for setting the box sizing property of an element. Box sizing
-     * determines whether an element’s border and padding is considered a part
+     * determines whether an elementâ€™s border and padding is considered a part
```
web-padawan, line 236: Same as above, `’` is replaced with incorrect character.

vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/LumoUtility.java
```diff
@@ -1461,57 +1415,54 @@ private Vertical() {
...
+        public static final String STICKY = notConstant("sticky");
 
         private Position() {
         }
 
     }
 
     /**
-     * Classes for setting an element’s text alignment.
+     * Classes for setting an elementâ€™s text alignment.
```
web-padawan, line 1430: Same as above, `’` is replaced with incorrect character.

vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/LumoUtility.java
```diff
@@ -1461,57 +1415,54 @@ private Vertical() {
...
+        public static final String JUSTIFY = notConstant("text-justify");
 
         private TextAlignment() {
         }
 
     }
 
     /**
-     * Classes for setting an element’s text color.
+     * Classes for setting an elementâ€™s text color.
```
web-padawan, line 1445: Same as above, `’` is replaced with incorrect character.

web-padawan (changes requested):

vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/LumoUtility.java
TatuLund: By accident, fixed

vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/LumoUtility.java
TatuLund: Done and fixed

vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/LumoUtility.java
```diff
@@ -15,11 +15,21 @@
...
+
 /**
  * Contains the definition for all the CSS utility classes provided by Lumo.
  */
+@NpmPackage(value = "@vaadin/vaadin-lumo-styles", version = "23.2.0-alpha2")
+@JsModule("@vaadin/vaadin-lumo-styles/utility.js")
+@JsModule("./lumo-utility.ts")
 public final class LumoUtility {
 
+    private static String notConstant(String value) {
```
web-padawan, line 29: Let's add a comment based on https://github.com/vaadin/flow-components/pull/3383#issuecomment-1205044363 to explain the purpose of this.

TatuLund: Done

web-padawan (commented):
Let's also update Javadoc as suggested in https://github.com/vaadin/flow-components/pull/3383#issuecomment-1164284621

> improve JavaDoc to note that just importing this class somewhere will import all CSS utility classes and define them in the global namespace, possible conflicting with user-defined classes

vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/LumoUtility.java
```diff
@@ -63,29 +82,29 @@ private AlignContent() {
...
+        public static final String STRETCH = notConstant("items-stretch");
 
         private AlignItems() {
         }
 
     }
 
     /**
-     * Classes for overriding individual items' align-item property. Applies to
+     * Classes for overriding individual items’ align-item property. Applies to
```
web-padawan, line 97:
```suggestion
     * Classes for overriding individual items' align-item property. Applies to
```

vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/resources/META-INF/resources/frontend/lumo-utility.ts
```diff
@@ -0,0 +1,15 @@
...
+
+var hasStyles = false;
+const styleTags = document.head.getElementsByTagName('style');
+for (var i = 0; i < styleTags.length; i++) {
+  if (styleTags[i].sheet === utility.styleSheet) {
+    hasStyles = true;
+    break;
+  }
+}
+if (!hasStyles) {
```
sissbruecker, line 11: I'm still getting duplicate style declarations when enabling the utility module in `theme.json`. I'm not sure if the check here works correctly with the `injectGlobalCss` generated by Flow, which creates new stylesheets rather than reusing imported one. The other problem is that `injectGlobalCss` uses a different duplicate check than the one implemented here, so if `injectGlobalCss` runs after this code, you will get duplicates anyway.

TatuLund: Should this be handled by documentation? I.e. by telling not to use importing via theme.json when using Flow.

Artur-: With the latest changes in Flow, you can remove this TS file and instead import `utility-global` in the Java file. As Flow imports this same file, you should not get duplicates anymore

TatuLund: Done

sissbruecker (approved):
Tested this again, and seems to work well after the recent change. Approved 👍

sissbruecker:
@leluna Please create an issue, a discussion here has little visibility for the team.

PR #425, opened 2020-11-23
feat: Add column in GridContextMenu#dynamicContentHandler

Springrbua:
This PR adds the possibility to dynamically create the context menu based on the clicked column.

For this, the dynamicContentHandler has been changed to a `SerializableBiPredicate<T, Column<T>>` but I kept the `setDynamicContentHandler(SerializablePredicate<T>)` to not break existing apps. Only the corresponding getter (`getDynamicContentHandler`) returns the `SerializableBiPrediacte<T, Column<T>>` now.
In the demo, the `Column` is always `null`, as none of the `Column`s have an ID. So the behavior is the same as in the `GridContextMenuOpenedEvent`.

Unfortunately, I was not able to run the integration tests but I will create a separate issue for that.

PR #3386, opened 2022-06-23
feat: add column to GridContextMenu#dynamicContentHandler

### LINKED ISSUE #1939: Add API for column in DynamicContentHandler how the Grid
Current API with GridContextMenu is like follows

```
contextMenu.setDynamicContentHandler(item -> {
     ...
});
```

This makes possible to populate the GridContextMenu based on item clicked.

However sometimes there is need dynamically populate the menu based on the specific column clicked as well.

Currently this is not possible since only method to get column info is in GridContextMenuOpenedEvent, which fired after DynamicContentHandler is processed.

```
contextMenu.addGridContextMenuOpenedListener(event -> {
    columnId event.getColumnId();
    ...
});

```

Thus there is a need for API like using BiFunction in the handler

```
contextMenu.setDynamicContentHandler((item,columnId) -> {
     ...
});
```

Note, this was possible in Vaadin 8.

### LINKED ISSUE #2258: GridContextMenu dynamicContentHandler called before _contextMenuTargetColumnId is set by Grid.updateContextMenuTargetItem
### Description

In Grid there is a method updateContextMenuTargetItem which updates the key and column Id from client side using properties. However this method is called and thus column ID is populated after the attached GridContextMenu's dynamicContentHandler has been called. Therefore it is impossible to build a context menu which related to a specific column in the Grid. This is especially necessary for Grid Header related Context Menus (like sorting, filtering, data ordering, etc.).

### Expected outcome

I expect the clicked column ID to be populated before the dynamicContentHandler will be called, so assembling a Context Menu can relay on column ID information.

### Actual outcome

Currently I have to click the column twice, because the first click has either no column id (empty string at first click ever) or the previously clicked column's ID.

### Steps to reproduce

1. Create a grid with data and some columns, set ID for each column (like. column.setId("col1");)
2. Attach a context menu with a dynamic content handler
3. Put this line in the context handler: String columnId = <grid_variable>.getElement().getProperty("_contextMenuTargetColumnId"); -> write columnId to log or system out
4. Click around in the table, and check the columnId the first click ever will be empty string, every other click will show the previous clicked column
5. Using debugger set a breakpoint on the dynamicContentHandler and on the Grid.updateContextMenuTargetItem method. You will notice, the dynamicContentHandler will be called before the updateContextMenuTargetItem.

### Environment

- Vaadin version: 21.0.2
- OS. Linux, chrome

web-padawan:
## Description

WIP, depends on #3385 and needs to be rebased once that PR is merged.

Fixes #1939
Closes #2258

## Clarification

The implementation is based on #425 and is based on `column.id` that we also use in `GridContextMenuOpenedEvent`.

Ideally, to make it aligned with the other Grid events, we should use `column._flowId` instead, e.g. like this:

https://github.com/vaadin/flow-components/blob/30abd294a99e798b59e655f4b833e990ffca2dd6/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/ColumnResizeEvent.java#L64-L67

Unfortunately, it's not possible to use `column.getInternalId()` as it's protected and not available in `GridContextMenu`.

## Type of change

- Feature

## Checklist

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/contextmenu/GridContextMenu.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/contextmenu/GridContextMenu.java
@@ -288,7 +288,7 @@ public class GridContextMenu<T> extends
             final T item = grid.getDataCommunicator().getKeyMapper().get(key);
             final Grid.Column<T> column = grid.getColumns().stream()
                     .filter(col -> columnId.equals(col.getId().get()))
-                    .findFirst().get();
+                    .findFirst().orElse(null);
             return getDynamicContentHandler().test(item, column);
         }
 
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/contextmenu/GridContextMenu.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/contextmenu/GridContextMenu.java
@@ -24,6 +24,7 @@ import com.vaadin.flow.component.contextmenu.ContextMenuBase;
 import com.vaadin.flow.component.contextmenu.MenuManager;
 import com.vaadin.flow.component.grid.Grid;
 import com.vaadin.flow.function.SerializableBiFunction;
+import com.vaadin.flow.function.SerializableBiPredicate;
 import com.vaadin.flow.function.SerializablePredicate;
 import com.vaadin.flow.function.SerializableRunnable;
 import com.vaadin.flow.shared.Registration;
@@ -41,7 +42,7 @@ public class GridContextMenu<T> extends
         ContextMenuBase<GridContextMenu<T>, GridMenuItem<T>, GridSubMenu<T>>
         implements HasGridMenuItems<T> {
 
-    private SerializablePredicate<T> dynamicContentHandler;
+    private SerializableBiPredicate<T, Grid.Column<T>> dynamicContentHandler;
 
     /**
      * Event that is fired when a {@link GridMenuItem} is clicked inside a
@@ -223,7 +224,7 @@ public class GridContextMenu<T> extends
      * @return the callback function that is executed before opening the context
      *         menu, or {@code null} if not specified.
      */
-    public SerializablePredicate<T> getDynamicContentHandler() {
+    public SerializableBiPredicate<T, Grid.Column<T>> getDynamicContentHandler() {
         return dynamicContentHandler;
     }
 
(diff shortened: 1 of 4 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/contextmenu/GridContextMenu.java
```diff
@@ -223,7 +224,7 @@ public Registration addGridContextMenuOpenedListener(
      * @return the callback function that is executed before opening the context
      *         menu, or {@code null} if not specified.
      */
-    public SerializablePredicate<T> getDynamicContentHandler() {
+    public SerializableBiPredicate<T, Grid.Column<T>> getDynamicContentHandler() {
```
sissbruecker, line 227: This would be a breaking change. Probably falls under the definition of a minor one, so I would agree with putting it into a minor, but I feel we should document it somehow.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/contextmenu/GridContextMenu.java
```diff
@@ -257,10 +282,14 @@ public void setDynamicContentHandler(
     protected boolean onBeforeOpenMenu(JsonObject eventDetail) {
         Grid<T> grid = (Grid<T>) getTarget();
         String key = eventDetail.getString("key");
+        String columnId = eventDetail.getString("columnId");
```
sissbruecker, line 285: Using the column ID is consistent with what you get from the opened change event, but requires an ID to be set on the column. Ideally we would always provide a column. I can imagine developers not familiar with the opened changed event getting confused why the column is null, and then having to dig around how to make it show up.

As the method for accessing the column by internal ID has protected visibility and is not accessible from this class, maybe we can introduce a construct to expose some internals from the Grid to be used by related components. Taking inspiration from `UI.getInternals`, we could add something like this to the Grid:
```java
    /**
     * Helper class for exposing internal Grid functionality to related components, such as {@link GridContextMenu}
     * <p>
     * For internal use only. May be renamed or removed in a future release.
     * @param <T> the item type of the grid
     */
    public static class GridInternals<T> {
        private final Grid<T> grid;

        public GridInternals(Grid<T> grid) {
            this.grid = grid;
        }

        public Column<T> getColumnByInternalId(String id) {
            return grid.getColumnByInternalId(id);
        }
    }

    /**
     * Exposes internals of the Grid to be used by other Grid-related Vaadin components.
     * <p>
     * For internal use only. May be renamed or removed in a future release.
     * @return helpers for accessing internals
     */
    public GridInternals<T> getInternals() {
        return new GridInternals<>(this);
    }
```

We could also re-use this to provide the column instance in the opened change event, and deprecate the `columnId` property.

web-padawan:
Closing this PR for now. As discussed within a team and agreed with @sissbruecker we won't work on this any time soon. Let's still keep the branch around in case there is some interest in this at some point.

netsrotr:
We have interest on that, it should be part of the API anyway, instead poking around with a workaround hack.

PR #3388, opened 2022-06-23
refactor: make unintentional public spreadsheet api private or protected

### LINKED ISSUE #3348: [spreadsheet] Unintentional public API
The Flow Spreadsheet class seems to have [quite a lot of public API](https://github.com/vaadin/flow-components/blob/883c18b3e468bedcf7292741e37b940802c08b5d/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/Spreadsheet.java#L320-L830) that's probably there unintentionally.

Screenshot 2022-06-15 at 16 43 35

ugur-vaadin:
## Description

There are public Flow Spreadsheet APIs that were not present in v8 Spreadsheet. 

This change converts the newly introduced public APIs into private or protected, depending on the usage within the package.

Fixes #3348

## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] Refactor

## Checklist


tomivirkki:
There's no need to keep any of the unused private API. I guess the methods were added for debugging purposes but not intended for the final product.

Screenshot 2022-08-09 at 15 55 49

Some of the methods referenced in the [issue](https://github.com/vaadin/flow-components/issues/3348) were not covered.

ugur-vaadin:
Removed unused private methods. 

Converted getColGroupingMax and getRowGroupingMax from public into protected.

The method getRowH is kept protected in order to maintain the consistency with getColW. It can still be made private though. What do you think?

Some of the methods mentioned in the issue are now used within the Spreadsheet class or other classes:
getCols
getColGroupingMax
getRowGroupingMax
isColGroupingInversed
isRowGroupingInversed
getDefRowH
getDefColW
getRowH
getColW
getSheetNames
getCellStyleToCSSStyle
getRowIndexToStyleIndex
getColumnIndexToStyleIndex
getLockedColumnIndexes
getLockedRowIndexes
getShiftedCellBorderStyles
getConditionalFormattingStyles
getVerticalScrollPositions
getHorizontalScrollPositions
getComponentIDtoCellKeysMap

tomivirkki:
> Some of the methods mentioned in the issue are now used in other classes in the package:

If removal isn't an option for some methods, rather make them package protected. No need to add this strange undocumented API available for users.

> Some of the methods referenced in the issue were not covered

The screenshot in the description only included a subset of the unintentional API. Please see the link in the description for full listing.

ugur-vaadin:
Addressed the previous comments.

Also deleted some unused fields.

One thing to note is while there is a public method `getDescription`, there is no setter for it. At this state, it can only return an empty string. Is this field still in use? If so, I think we should introduce a setter or any other way to set the description. If not, we can maybe deprecate it.

vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/Spreadsheet.java
```diff
@@ -283,538 +245,429 @@ public void setWidth(String width) {
...
         return colGroupingInversed;
     }
 
-    public boolean isRowGroupingInversed() {
+    protected boolean isRowGroupingInversed() {
         return rowGroupingInversed;
     }
 
-    public float getDefRowH() {
+    protected float getDefRowH() {
```
tomivirkki, line 287: > If removal isn't an option for some methods, rather make them package-private. No need to add this strange undocumented API available for users.

Why are these API still protected? Were they protected in V8 TabSheet?

ugur-vaadin: The protected ones are now used in other classes. For this example, it is used in SpreadsheetFactory.

tomivirkki: SpreadsheetFactory is in the same package as SpreadSheet, so it can be package-private, right? (to not expose this undocumented internal API for extensions)

tomivirkki:
> One thing to note is while there is a public method getDescription, there is no setter for it. At this state, it can only return an empty string. Is this field still in use? If so, I think we should introduce a setter or any other way to set the description. If not, we can maybe deprecate it.

The description API was used in V8 to add a tooltip to components.

Screenshot 2022-08-16 at 11 00 06

It's no longer there in the Flow components so it can be removed from SpreadSheet also.

vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/Spreadsheet.java
ugur-vaadin: Sure, done.

ugur-vaadin:
The description field is now removed.

tomivirkki (approved):

PR #3406, opened 2022-06-28
feat: make field components implement HasValidator interface

### LINKED ISSUE #3381: Implement HasValidator interface
### Describe your motivation

All input components should implement `HasValidator` interface and provide `getDefaultValidator` method.
Tests for usage together with `Binder` should be added.

DiegoCardoso:
## Description

- Add `HasValidator<T>` to field components and make them override `getDefaultValidator` method
- Make field components return string containing the reason for the validation error
- Extract some common logic to an util class on the shared module

Fixes #3381 

## Type of change

- [ ] Bugfix
- [X] Feature

## Checklist

- [X] I have read the contribution guide: https://vaadin.com/docs-beta/latest/guide/contributing/overview/
- [X] I have added a description following the guideline.
- [X] The issue is created in the corresponding repository and I have referenced it.
- [X] I have added tests to ensure my change is effective and works as intended.
- [X] New and existing tests are passing locally with my change.
- [X] I have performed self-review and corrected misspellings.

- [X] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -13,7 +13,7 @@
     <url>https://vaadin.com/components</url>
     <properties>
         <flow.version>23.2-SNAPSHOT</flow.version>
-        <testbench.version>8.1.0.alpha8</testbench.version>
+        <testbench.version>8.1.0.alpha10</testbench.version>
         <maven.compiler.source>11</maven.compiler.source>
         <maven.compiler.target>11</maven.compiler.target>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/main/java/com/vaadin/flow/component/combobox/test/RefreshEmptyLazyDataProviderPage.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/main/java/com/vaadin/flow/component/combobox/test/RefreshEmptyLazyDataProviderPage.java
@@ -1,0 +1,36 @@
+package com.vaadin.flow.component.combobox.test;
+
+import com.vaadin.flow.component.combobox.ComboBox;
+import com.vaadin.flow.component.html.Div;
+import com.vaadin.flow.component.html.NativeButton;
+import com.vaadin.flow.data.provider.AbstractBackEndDataProvider;
+import com.vaadin.flow.data.provider.Query;
+import com.vaadin.flow.router.Route;
+
+import java.util.stream.Stream;
+
+@Route("vaadin-combo-box/refresh-empty-lazy-data-provider")
+public class RefreshEmptyLazyDataProviderPage extends Div {
+    public RefreshEmptyLazyDataProviderPage() {
+        ComboBox<String> comboBox = new ComboBox<>();
+        comboBox.setItems(new AbstractBackEndDataProvider<>() {
+            @Override
+            protected Stream<String> fetchFromBackEnd(
+                    Query<String, String> query) {
+                return Stream.of();
+            }
+
+            @Override
+            protected int sizeInBackEnd(Query<String, String> query) {
+                return 0;
+            }
+        });
+
+        NativeButton refreshDataProvider = new NativeButton("Refresh", e -> {
+            comboBox.getDataProvider().refreshAll();
+        });
+        refreshDataProvider.setId("refresh-data-provider");
+
+        add(comboBox, refreshDataProvider);
+    }
+}
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/RefreshEmptyLazyDataProviderIT.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/RefreshEmptyLazyDataProviderIT.java
@@ -1,0 +1,40 @@
+package com.vaadin.flow.component.combobox.test;
+
+import com.vaadin.flow.component.combobox.testbench.ComboBoxElement;
+import com.vaadin.flow.testutil.TestPath;
+import com.vaadin.testbench.TestBenchElement;
+import com.vaadin.tests.AbstractComponentIT;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+@TestPath("vaadin-combo-box/refresh-empty-lazy-data-provider")
+public class RefreshEmptyLazyDataProviderIT extends AbstractComponentIT {
+    private ComboBoxElement comboBox;
+    private TestBenchElement refreshDataProvider;
+
+    @Before
+    public void init() {
+        open();
+        comboBox = $(ComboBoxElement.class).waitForFirst();
+        refreshDataProvider = $("button").id("refresh-data-provider");
+    }
+
+    // Regression test for:
+    // https://github.com/vaadin/flow-components/issues/3432
+    @Test
+    public void open_close_refreshDataProvider_open_overlayIsHiddenAndLoadingStateIsCleared() {
+        comboBox.openPopup();
+        comboBox.closePopup();
+
+        refreshDataProvider.click();
+        comboBox.openPopup();
+
+        // Verify the overlay is closed
+        // When there are no items to display then the overlay is hidden, but
+        // the opened state of the combo box is still true, so we can't check
+        // the opened state. Instead, we test that there is no overlay element.
+        Assert.assertFalse($("vaadin-combo-box-overlay").exists());
+        Assert.assertFalse(comboBox.getPropertyBoolean("loading"));
+    }
+}
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxDataController.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxDataController.java
@@ -255,6 +255,19 @@ class ComboBoxDataController<TItem>
      * Called when the client-side connector requests data
      */
     void setRequestedRange(int start, int length, String filter) {
+        // If the filter is null, which indicates that the combo box was closed
+        // before, then reset the data communicator to force sending an update
+        // to the client connector. This covers an edge-case when using an empty
+        // lazy data provider and refreshing it before opening the combo box
+        // again. In that case the data provider thinks that the client should
+        // already be up-to-date from the refresh, as in both cases, refresh and
+        // empty data provider, the effective requested size is zero, which
+        // results in it not sending an update. However, the client needs to
+        // receive an update in order to clear the loading state from opening
+        // the combo box.
+        if (lastFilter == null) {
+            dataCommunicator.reset();
+        }
         dataCommunicator.setRequestedRange(start, length);
         filterSlot.accept(filter);
     }
--- a/vaadin-crud-flow-parent/vaadin-crud-flow-integration-tests/src/test/java/com/vaadin/flow/component/crud/tests/CustomGridIT.java
+++ b/vaadin-crud-flow-parent/vaadin-crud-flow-integration-tests/src/test/java/com/vaadin/flow/component/crud/tests/CustomGridIT.java
@@ -10,7 +10,6 @@ import com.vaadin.flow.component.crud.testbench.CrudElement;
 import com.vaadin.flow.component.grid.testbench.GridElement;
 import com.vaadin.flow.component.textfield.testbench.TextFieldElement;
 import com.vaadin.testbench.TestBenchElement;
-import com.vaadin.testbench.parallel.BrowserUtil;
 
 public class CustomGridIT extends AbstractParallelTest {
 
@@ -36,10 +35,6 @@ public class CustomGridIT extends AbstractParallelTest {
         lastNameField.setValue("Otto");
         crud.getEditorSaveButton().click();
 
-        if (BrowserUtil.isIE(getDesiredCapabilities())) {
-            return;
-        }
-
         Assert.assertFalse(crud.isEditorOpen());
         Assert.assertEquals("Otto",
                 $(GridElement.class).first().getCell(0, 2).getText());
@@ -58,10 +53,6 @@ public class CustomGridIT extends AbstractParallelTest {
         Assert.assertEquals("Sayo", lastNameField.getValue());
         lastNameField.setValue("Otto");
         crud.getEditorCancelButton().click();
-
-        if (BrowserUtil.isIE(getDesiredCapabilities())) {
-            return;
-        }
 
         ConfirmDialogElement confirmCancel = crud.getConfirmCancelDialog();
         Assert.assertEquals("Discard changes", confirmCancel.getHeaderText());
--- a/vaadin-crud-flow-parent/vaadin-crud-flow-integration-tests/src/test/java/com/vaadin/flow/component/crud/tests/EventHandlingIT.java
+++ b/vaadin-crud-flow-parent/vaadin-crud-flow-integration-tests/src/test/java/com/vaadin/flow/component/crud/tests/EventHandlingIT.java
@@ -12,7 +12,6 @@ import com.vaadin.flow.component.crud.testbench.CrudElement;
 import com.vaadin.flow.component.grid.testbench.GridElement;
 import com.vaadin.flow.component.textfield.testbench.TextFieldElement;
 import com.vaadin.testbench.TestBenchElement;
-import com.vaadin.testbench.parallel.BrowserUtil;
 
 public class EventHandlingIT extends AbstractParallelTest {
 
@@ -159,13 +158,6 @@ public class EventHandlingIT extends AbstractParallelTest {
         Assert.assertFalse(lastNameField.hasAttribute("invalid"));
 
         crud.getEditorSaveButton().click();
-
-        if (BrowserUtil.isIE(getDesiredCapabilities())) {
-            // TODO(oluwasayo): Investigate why editor sometimes doesn't
-            // disappear on first click in IE
-            // especially when server-side validation is involved
-            return;
-        }
 
         Assert.assertFalse(crud.isEditorOpen());
         Assert.assertEquals("Oladeji",
--- a/vaadin-crud-flow-parent/vaadin-crud-flow-integration-tests/src/test/java/com/vaadin/flow/component/crud/tests/ProtectedBackendIT.java
+++ b/vaadin-crud-flow-parent/vaadin-crud-flow-integration-tests/src/test/java/com/vaadin/flow/component/crud/tests/ProtectedBackendIT.java
@@ -1,7 +1,6 @@
 package com.vaadin.flow.component.crud.tests;
 
 import org.junit.Assert;
-import org.junit.AssumptionViolatedException;
 import org.junit.Before;
 import org.junit.Test;
 import org.openqa.selenium.NoSuchElementException;
@@ -11,7 +10,6 @@ import com.vaadin.flow.component.confirmdialog.testbench.ConfirmDialogElement;
 import com.vaadin.flow.component.crud.testbench.CrudElement;
 import com.vaadin.flow.component.grid.testbench.GridElement;
 import com.vaadin.flow.component.textfield.testbench.TextFieldElement;
-import com.vaadin.testbench.parallel.BrowserUtil;
 
 public class ProtectedBackendIT extends AbstractParallelTest {
 
@@ -54,11 +52,6 @@ public class ProtectedBackendIT extends AbstractParallelTest {
 
     @Test
     public void tryModify() {
-        if (BrowserUtil.isIE(getDesiredCapabilities())) {
-            throw new AssumptionViolatedException(
-                    "Skipped IE11, cause textfield setValue doesn't make the editor dirty");
-        }
-
         CrudElement crud = $(CrudElement.class).waitForFirst();
 
         crud.openRowForEditing(0);
--- a/vaadin-crud-flow-parent/vaadin-crud-flow-integration-tests/src/test/java/com/vaadin/flow/component/crud/tests/TemplateSupportIT.java
+++ b/vaadin-crud-flow-parent/vaadin-crud-flow-integration-tests/src/test/java/com/vaadin/flow/component/crud/tests/TemplateSupportIT.java
@@ -10,7 +10,6 @@ import com.vaadin.flow.component.grid.testbench.GridElement;
 import com.vaadin.flow.component.orderedlayout.testbench.VerticalLayoutElement;
 import com.vaadin.flow.component.textfield.testbench.TextFieldElement;
 import com.vaadin.testbench.ElementQuery;
-import com.vaadin.testbench.parallel.BrowserUtil;
 
 public class TemplateSupportIT extends AbstractParallelTest {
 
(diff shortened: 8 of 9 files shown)


vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/TextFieldValidationSupport.java
```diff
@@ -63,18 +66,37 @@ void setPattern(String pattern) {
...
+        var requiredValidation = ValidationUtils.checkRequired(required, value,
+                field.getEmptyValue());
+        if (requiredValidation.isError()) {
+            return requiredValidation;
+        }
+
         final boolean isMaxLengthExceeded = value != null && maxLength != null
                 && value.length() > maxLength;
+        if (isMaxLengthExceeded) {
+            return ValidationResult.error(ValidationError.MAX_LENGTH_EXCEEDED);
```
vursen, line 82: **thought:** Considering that these validation errors are defined in a shared module, I guess it might make sense to maintain their corresponding validation checks in the shared `ValidationUtils` module as well, for the sake of consistency. What do you think?

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/ValidationError.java
```diff
@@ -0,0 +1,17 @@
+package com.vaadin.flow.component.shared;
+
+import java.io.Serializable;
+
+public final class ValidationError implements Serializable {
+    public static final String REQUIRED = "VALIDATION_REQUIRED_ERROR";
```
knoobie, line 6: Make sure those non-explaining error messages are not overriding errors provided by the binder (like asRequired("My awesome name of the field is empty"). 

Additionally, if those strings are ever shown to the end-user there should be a way to translate them. All fields should provide an i18n mapping for the possible errors and they outcome.

DiegoCardoso: > Make sure those non-explaining error messages are not overriding errors provided by the binder (like asRequired("My awesome name of the field is empty").

Hmmm. That's a valid point. Currently the `HasValidator` with `Binder` works so that the `getDefaultValidator()` is called before the validation constraints from `Binder` are called. So if `asRequired()` is called on the binding definition, that will effectively set the component's `required` property to `true` and the validation will fail with the `VALIDATION_REQUIRED_ERROR` message. From `getDefaulltValidator()`, I can't see a way of checking if there's a custom message defined on `asRequired()`, though. 
I also don't think we can't remove the required check from the component (even when used with `Binder`), since my assumption is that it should work if the required is set on the component `setRequired()/`setRequiredIndicatorVisible()`. So I need to investigate how to proceed here.

> Additionally, if those strings are ever shown to the end-user there should be a way to translate them. All fields should provide an i18n mapping for the possible errors and their outcome.

Good point. There should be a way of defining custom messages.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/TextFieldValidationSupport.java
DiegoCardoso: I thought about doing that on first place, but later thought that those checks are not shared with any other component at the moment, so I decided to keep them in place. But maybe, it's a better idea to move them to `ValidationUtils` as well.

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/ValidationError.java
knoobie: Just a thought: I don't think it's possible to fix this behavior on your side if getDefaultValidator runs before the binder validations. I think this should be fixed / done inside flow - and the getDefaultValidator() should run AFTER the binder validation; meaning the getDefaultValidator() should be registered once the field and it's validations are bind. This allows to stop the validations once the binder fails and uses this message instead of the messages / validations provided later by the field

DiegoCardoso: We decided to exclude required from the validation done by `getDefaultValidator()`, so if one wants to use binder, then only `asRequired()` should be used.

knoobie: That races multiple contains.. short question: Can the field validation be disabled? We have moved all our validations to the binder over the years, because the client side wasn't  secure and intuitive.

Just another example we currently use:
 - DatePicker:.setMax(LocalDate.now()); is used to disable to get the user instant visible feedback that this date isn't selectable, but the user can still type tomorrow in.. resulting on a validation on the server with a proper error message like "Birthday can't be set to the future." 
 
 With this change, what error message is shown? The server side error message or the field validation? I would expect the later.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/test/java/com/vaadin/flow/component/textfield/tests/NumberFieldValidationTest.java
```diff
@@ -0,0 +1,108 @@
...
+import com.vaadin.flow.component.shared.ValidationError;
+import com.vaadin.flow.component.textfield.NumberField;
+import com.vaadin.flow.data.binder.Binder;
+import com.vaadin.flow.data.binder.BindingValidationStatusHandler;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.Objects;
+
+public class NumberFieldValidationTest {
```
vursen, line 12: **suggestion:** I would suggest we mention `Binder` in the test's class name considering that it only focuses on binder validation and doesn't test the default constraint validation.

```java
public class NumberFieldBinderValidationTest {
  ...
}
```

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/ValidationError.java
DiegoCardoso: First, thanks for all the feedback so far, they have been raising some internal conversation about the feature.

For now, we want to build a MVP to make this behaviour change on the fields to make client and flow constraints part of the binder validation. Later on, we will make the adjustments needed, like the ones you suggested.

With that said, we have been considering have an API to disable component validation and also to deliver these changes under a feature flag, so that we can release it and validate the new behaviour with more customers before making it the default behaviour.

knoobie: > deliver these changes under a feature flag

This would be really good! Thanks for the discussion! 

Just to clarify: I'm totally with the addition of the feature, because I've requested this as "bug" 2 years ago (https://github.com/vaadin/flow/issues/7767) - but it feels the currently MVP is behind the feature richness of the binder in terms of user and developer experience, that's why I wanted to mention it. Additionally because of the issue mentioned, we have migrated all our Vaadin 8 applications to the Binder only paradigm in V23.1 - where it would be a huge effort to adopt to a surprising change like this. 

Something similar happened to us with an addtional Binder feature / behaviour change in the past, where it did something we didn't expect and had problems with the behavior change, because we heavily heavily really on Binder as the stable foundation of all our applications (validation and security) - https://github.com/vaadin/flow/issues/10342

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/test/java/com/vaadin/flow/component/textfield/tests/NumberFieldValidationTest.java
DiegoCardoso: I moved them to a new package, so it will be a bit more organized.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/test/java/com/vaadin/flow/component/textfield/binder/AbstractTextFieldValidationTest.java
```diff
@@ -0,0 +1,130 @@
...
+    @Captor
+    private ArgumentCaptor<BindingValidationStatus<?>> statusCaptor;
+
+    @Mock
+    private BindingValidationStatusHandler statusMock;
+
+    @Before
+    public void init() {
+        MockitoAnnotations.openMocks(this);
+    }
```
vursen, line 51: **suggestion:** 

If we instantiate the field in the `init` method:

```suggestion
    public void init() {
        MockitoAnnotations.openMocks(this);
        
        field = getField();
    }
```

...we can simplify `getField` in subclasses:

```diff
protected HasValue<?, BigDecimal> getField() {
-  if (field == null) {
-      field = new BigDecimalField();
-  }
-  return field;
+  return new BigDecimalField();
}
```

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/test/java/com/vaadin/flow/component/textfield/binder/EmailFieldValidationTest.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    protected void setEmptyValue() {
+        getField().setValue("");
+    }
+
+    @Override
+    protected SerializablePredicate<? super String> getValidator() {
+        return value -> Objects.equals(value, "")
```
vursen, line 47:
```suggestion
        return value -> value.equals("") || value.contains("@example.com")
```

vursen (commented):
The first portion of comments.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/test/java/com/vaadin/flow/component/datepicker/DatePickerBinderValidationTest.java
```diff
@@ -0,0 +1,150 @@
...
+        Assert.assertTrue(statusCaptor.getValue().isError());
+
+    }
+
+    @Test
+    public void elementWithConstraints_binderValidationNotMet_binderValidationFails() {
+        var field = getFieldWithValidation(status -> {
+            Assert.assertTrue(status.isError());
+            Assert.assertEquals(BINDER_FAIL_MESSAGE,
+                    status.getMessage().orElse(""));
```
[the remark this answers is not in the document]
DiegoCardoso: Removed.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/ComboBoxBinderValidationTest.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Before
+    public void init() {
+        MockitoAnnotations.openMocks(this);
+    }
+
+    @Test
+    public void elementWithConstraints_binderValidationNotMet_binderValidationFails() {
+        var field = getFieldWithValidation();
```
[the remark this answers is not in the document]
DiegoCardoso: Applied the suggestion here and on the similar classes.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/ComboBoxBinderValidationTest.java
```diff
@@ -0,0 +1,115 @@
...
+        field.setValue(null);
+
+        Mockito.verify(statusHandlerMock, Mockito.times(2))
+                .statusChange(statusCaptor.capture());
+        Assert.assertFalse(statusCaptor.getValue().isError());
+
+    }
+
+    @Test
+    public void elementWithConstraints_validValue_validationOk() {
```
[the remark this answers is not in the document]
DiegoCardoso: Done.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
```diff
@@ -65,8 +69,11 @@
 @NpmPackage(value = "date-fns", version = "2.28.0")
 public class DatePicker extends GeneratedVaadinDatePicker<DatePicker, LocalDate>
         implements HasSize, HasValidation, HasHelper, HasTheme, HasLabel,
-        HasClearButton, HasAllowedCharPattern {
+        HasClearButton, HasAllowedCharPattern, HasValidator<LocalDate> {
 
+    public static final String VALIDATION_REQUIRED_ERROR = "VALIDATION_REQUIRED_ERROR";
+    public static final String VALIDATION_GREATER_THAN_MAX_ERROR = "VALIDATION_GREATER_THAN_MAX_ERROR";
+    public static final String VALIDATION_SMALLER_THAN_MIN_ERROR = "VALIDATION_SMALLER_MIN_THAN_ERROR";
```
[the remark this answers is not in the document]
DiegoCardoso: Thanks. Removed.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/test/java/com/vaadin/flow/component/datepicker/DatePickerBinderValidationTest.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    @Test
+    public void elementRequiredFromBinder_emptyField_binderValidationFail() {
+        var field = getFieldWithValidation(status -> {
+            if (status.getField().isEmpty()) {
+                Assert.assertTrue(status.isError());
+                Assert.assertEquals(REQUIRED_MESSAGE,
+                        status.getMessage().orElse(""));
+            }
```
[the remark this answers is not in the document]
DiegoCardoso: Removed.

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/ValidationUtils.java
```diff
@@ -0,0 +1,99 @@
...
+import com.vaadin.flow.data.binder.ValidationResult;
+
+import java.util.Objects;
+
+/**
+ * Util methods for component validation
+ */
+public class ValidationUtils {
+
+    private ValidationUtils() {
```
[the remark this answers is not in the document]
DiegoCardoso: Renamed.

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/ValidationUtils.java
```diff
@@ -0,0 +1,99 @@
...
+     * @param value
+     *            the current value set on the component
+     * @param emptyValue
+     *            the empty value for the component
+     * @return <code>Validation.ok()</code> if the validation pass,
+     *         <code>Validation.error()</code> otherwise
+     * @param <T>
+     *            the type of the component value
+     */
+    public static <T> ValidationResult checkRequired(boolean required, T value,
```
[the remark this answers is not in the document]
DiegoCardoso: Done.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/TextFieldValidationSupport.java
```diff
@@ -63,18 +64,34 @@ void setPattern(String pattern) {
...
+        }
+
         final boolean isMinLengthNotReached = value != null && minLength != null
                 && value.length() < minLength;
+        if (isMinLengthNotReached) {
+            return ValidationResult.error("");
+        }
+
         // Only evaluate if necessary.
         final BooleanSupplier doesValueViolatePattern = () -> value != null
```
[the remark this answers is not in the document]
DiegoCardoso: Yeah, makes sense. Removed the wrapper.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/test/java/com/vaadin/flow/component/textfield/binder/AbstractTextFieldValidationTest.java
DiegoCardoso: Somewhat applied the proposed change. Let me know what you think.

vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow/src/test/java/com/vaadin/flow/component/datetimepicker/DateTimePickerBinderValidationTest.java
```diff
@@ -0,0 +1,126 @@
...
+        }
+    }
+
+    @Before
+    public void init() {
+        MockitoAnnotations.openMocks(this);
+    }
+
+    @Test
+    public void elementWithConstraints_componentValidationNotMet_elementValidationFails() {
```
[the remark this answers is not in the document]
DiegoCardoso: In theory, all the components now will rely on the client side validation. Like `min`/`max` will be checked on a `NumberField` instance first on the client and then on the server. So, that would make sense to add the ITs. At the other hand, the server side validation for the same constraints should work on the same way as the client side, so it may be beneficial to have unit tests to check them.

One thing it got me thinking is how the client side validation status will play along with the server side check. For instance, the developer sets `max` to `NumberField` and the user inserts a number greater than `max`. The client side validation will show it as invalid and so the event will send this information to the server. When `checkValidity` on the server is called, will the method say that the component is invalid based on the client status or will it check the constraints again and only rely on the client side status when all server side constraints pass?

vaadin-time-picker-flow-parent/vaadin-time-picker-flow/src/test/java/com/vaadin/flow/component/timepicker/tests/TimePickerBinderValidationTest.java
```diff
@@ -0,0 +1,125 @@
...
+        field.setRequiredIndicatorVisible(true);
+        binder.validate();
+
+        Mockito.verify(statusHandlerMock).statusChange(statusCaptor.capture());
+        Assert.assertFalse(statusCaptor.getValue().isError());
+    }
+
+    @Test
+    public void elementWithConstraints_validValue_validationPasses() {
+        getFieldWithValidation();
```
vursen, line 95:
```suggestion
        attachBinderToField();
```

vaadin-time-picker-flow-parent/vaadin-time-picker-flow/src/test/java/com/vaadin/flow/component/timepicker/tests/TimePickerBinderValidationTest.java
```diff
@@ -0,0 +1,125 @@
...
+import org.mockito.Captor;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+
+import java.time.LocalTime;
+
+public class TimePickerBinderValidationTest {
+    private static final String BINDER_FAIL_MESSAGE = "BINDER_VALIDATION_FAILED";
+    private static final String REQUIRED_MESSAGE = "REQUIRED";
```
vursen, line 20:
```suggestion
    private static final String BINDER_REQUIRED_MESSAGE = "REQUIRED";
```

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/test/java/com/vaadin/flow/component/textfield/binder/BigDecimalFieldValidationTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+import com.vaadin.flow.component.HasValue;
+import com.vaadin.flow.component.textfield.BigDecimalField;
+import com.vaadin.flow.function.SerializablePredicate;
+import org.junit.Ignore;
+
+import java.math.BigDecimal;
+
+public class BigDecimalFieldValidationTest
+        extends AbstractTextFieldValidationTest<BigDecimal> {
```
vursen, line 11: How about we also add a generic type argument for the field's class? This way we could avoid duplicate private `field` properties and instead have one single protected property in the abstract class.

```suggestion
        extends AbstractTextFieldValidationTest<BigDecimal, BigDecimalField> {
```

vursen (commented):
A couple of more suggestions to keep everything consistent.

vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow/src/test/java/com/vaadin/flow/component/datetimepicker/DateTimePickerBinderValidationTest.java
```diff
@@ -0,0 +1,124 @@
...
+        Assert.assertFalse(statusCaptor.getValue().isError());
+
+    }
+
+    private Binder<Bean> attachBinderToField() {
+        return attachBinderToField(false);
+    }
+
+    private Binder<Bean> attachBinderToField(boolean isRequired) {
+        field.setMax(LocalDateTime.now().plusDays(1));
```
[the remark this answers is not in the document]
DiegoCardoso: Good point. Moved to `init()`.

vaadin-time-picker-flow-parent/vaadin-time-picker-flow/src/test/java/com/vaadin/flow/component/timepicker/tests/TimePickerBinderValidationTest.java
DiegoCardoso: Done.

vaadin-time-picker-flow-parent/vaadin-time-picker-flow/src/test/java/com/vaadin/flow/component/timepicker/tests/TimePickerBinderValidationTest.java
DiegoCardoso: Done.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/test/java/com/vaadin/flow/component/textfield/binder/BigDecimalFieldValidationTest.java
DiegoCardoso: Makes sense. Refactored.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/test/java/com/vaadin/flow/component/textfield/binder/AbstractTextFieldValidationTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+    @Before
+    public void init() {
+        MockitoAnnotations.openMocks(this);
+        field = getField();
+    }
+
+    @Test
+    public void elementWithConstraints_componentValidationNotMet_elementValidationFails() {
+        setupFieldWithValidation();
```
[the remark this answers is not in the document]
DiegoCardoso: Done.

vursen (approved):

vursen:
It would probably make sense to implement `HasValidator` also for `RadioGroup` and `CheckboxGroup` as they will need `ValidationStatusChangeListener` to trigger the binder validation on the `validated` event.

PR #3429, opened 2022-07-04
feat: integrate client-side validation into TextField

vursen:
## Description

- [x] Implement `HasClientValidation`.
- [x] Implement `ClientValidationUtil`.
- [x] Implement `ValidationStatusChangeEventListener`.
- [x] Add ITs ensuring the constraint validation is run on the `validated` event. 
- [x] Add ITs ensuring the binder validation is run on the `validated` event.
- [x] Add ITs ensuring only the server can set `TextField` to valid.
- [x] Add ITs for basic constraint validation.
- [x] Add ITs for basic constraint validation when used through binders.

The integration tests are implemented as abstract classes so that they can be reused later for `NumberField`, `PasswordField`, and so on.

Depends on 
- https://github.com/vaadin/flow-components/pull/3490
- https://github.com/vaadin/flow-components/pull/3501
- https://github.com/vaadin/flow-components/pull/3629

## Type of change

- [x] Feature

## Checklist


vursen:
Some ITs have failed which is apparently due to the fact that we don't trigger the binder validation on the `validated` event. 

Consider the following scenario:

1. The field doesn't pass the binder validation so it gets the `invalid` attribute.
2. The user triggers a `blur` event which leads to firing a `validated` event.
3. The field passes the constraint validation but the binder validation is not run so the `invalid` attribute gets removed (INCORRECT).

vursen:
I actually encountered a more serious challenge which is that the `validated` event may happen during the web component's initialization which may result in an `invalid` reset.

Example:

```java
textField = new TextField();
textField.setValue("Some not empty value");
textField.setInvalid(true);
add(textField)
```

What happens here is Polymer triggers value observers (because of the provided initial value) during the web component's initialization which subsequently triggers a `validated` event (because of [this logic](https://github.com/vaadin/web-components/blob/80863025b4386507bc74780020248ed1c532f1e5/packages/field-base/src/input-field-mixin.js#L121-L127) in the web component). As a result, Flow resets the `invalid` state since everything is valid in terms of the Flow validation (no constraints, no binder and etc).

I'm trying to fix that in the web components.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/TextField.java
```diff
@@ -81,6 +83,8 @@ private TextField(boolean isInitialValueOptional) {
         setValueChangeMode(ValueChangeMode.ON_CHANGE);
 
         addValueChangeListener(e -> validate());
+
+        addClientValidatedEventListener(e -> validate());
```
DiegoCardoso, line 87: There's still one missing piece, which is to fire the `ValidationStatusChangeEvent` when the client side validation changes. Testing how it works, it seems that once the event is dispatched, the binder will take the validation status sent on the event to set the component is valid/invalid. So, the `getDefaultValidator()` is not called.

The thing I am still thinking about is how the `validate` method comes into play here. It needs to be called so that the component being used without binder will get the invalid state correctly set. But when binder is being used, we would probably need to call `getValidationSupport().isInvalid(getValue())` to get the validation result and send it to binder with the event.

vursen: Good point. My first impression of `ValidationStatusChangeEventListener` was that we will need to subscribe **the binder** to `ClientValidatedEvent` by adding `ValidationStatusChangeEventListener` on `ClientValidatedEvent`.

vursen: So, the final flow will be the following:

1. `ValueChange` triggers constraint validation and then binder validation.
2. `ClientValidated` triggers constraint validation and then binder validation by invoking `ValidationStatusChangeEventListener` listeners.

DiegoCardoso: I actually had a wrong assumption here:

> So, the getDefaultValidator() is not called.

My mistake was because I placed the breakpoint on the method body instead of the lambda's. What I can understand now is:
1. Client side validation fires `validated` event
2. Flow component listens the event and fires `ValidationStatusChangeEventListener`
3. Binder listens to this event and calls its `validate()` method
4. Eventually, `validate()` method will call the lambda created on `getDefaultValidator()`.
5. Component's `checkValidity()` method is called.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/TextField.java
```diff
@@ -490,6 +498,16 @@ public Validator<String> getDefaultValidator() {
         return (value, context) -> getValidationSupport().checkValidity(value);
     }
 
+    @Override
+    public Registration addValidationStatusChangeListener(
+            ValidationStatusChangeListener<String> listener) {
+        return addClientValidatedEventListener(event -> {
+            listener.validationStatusChanged(
+                    new ValidationStatusChangeEvent<String>(this,
+                            !isInvalid()));
```
vursen, line 507: The purpose of the second parameter of `ValidationStatusChangeEvent` is not very clear since it is not used anywhere.

yuriy-fix (approved):

PR #3449, opened 2022-07-08
fix: Force rendering the icon after a change

### LINKED ISSUE #2924: [map] Changing marker icons does not work
### Description

When setting a new icon on a marker using `MarkerFeature.setIcon(Icon)` the new icon is not displayed on the map. Only after an additional change, like moving the viewport, or changing the markers coordinates, the marker is rendered with the new icon. The issue seems to be that the OpenLayers `Style` class that holds the icon is not observable ([source](https://github.com/openlayers/openlayers/blob/7f12698a9143798dd42032174c2b13084acc4593/src/ol/style/Style.js#L333-L340)), so replacing the icon does not trigger a change event in OpenLayers, and does not notify the map that it should re-render.

### Workaround

Triggering any other change in the marker itself will cause it to be synchronized to the client as well, which then also triggers a re-render of the map. One example would be to just reassign its coordinates:
```java
marker.setIcon(icon);
marker.setCoordinates(marker.getCoordinates());
```

### Expected outcome

Changing an icon on a marker should immediately render the marker with the icon.

### Actual outcome

The new icon is not rendered, unless another change in the map triggers a rerender.

```java
Map map = new Map();
add(map);

MarkerFeature marker = new MarkerFeature();
map.getFeatureLayer().addFeature(marker);

NativeButton changeMarkerIcon = new NativeButton("Change Marker Icon", e -> {
    // Using a default icon here, could be any custom icon
    marker.setIcon(MarkerFeature.POINT_ICON);
});
add(changeMarkerIcon);
```

### Steps to reproduce

1. Run the example above
2. The marker should initially use the default pin icon
3. Click the button to replace the marker icon with the default point icon
4. The marker is still displayed using the pin icon

### Environment

- v23

### Browsers Affected

N/A

blafoo:
## Description

A marker is not rendered after an icon change until the map is rerendered, e.g. by a map move or adding another marker.
Rerendering from the client side can't be achived thus a rerendering must be started from the server. Marking all childs of the FeatureMarker as dirty leads to such a rerendering.

Fixes #2924

## Type of change

- [x ] Bugfix
- [ ] Feature

## Checklist

- [x ] I have read the contribution guide: https://vaadin.com/docs-beta/latest/guide/contributing/overview/
- [x ] I have added a description following the guideline.
- [x ] The issue is created in the corresponding repository and I have referenced it.
- [1 ] New and existing tests are passing locally with my change.
- [ x] I have performed self-review and corrected misspellings.

1 I can't run the IT tests locally even without any changes -> Gave up trying to create a chromedriver instance
 
### DIFF
--- a/vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/feature/MarkerFeature.java
+++ b/vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/feature/MarkerFeature.java
@@ -149,5 +149,8 @@ public class MarkerFeature extends PointBasedFeature {
     public void setIcon(Icon icon) {
         Objects.requireNonNull(icon);
         getStyle().setImage(icon);
+
+        // force rerendering of the map by re-syncing the geometry (coordinates)
+        deepMarkAsDirty();
     }
 }


blafoo:
@czp13 There is already (a non-working) IT: vaadin-map/marker-feature. The test is fixed (missing resource) in my PR 3550. A testbench test won't work here because the change is not displayed by the browser. The object itself is updated correctly and would return the new value.

PR #3451, opened 2022-07-11
feat!: use EPSG:4326 as default coordinate system

sissbruecker:
## Description

Changes the default coordinate system / user projection of the Map component to `EPSG:4326`, a.k.a. GPS coordinates. That means that all coordinates passed to, or returned from the public API are assumed to be in this projection. Internally the component will automatically convert coordinates into the configured view projection.

This is a breaking change, and all coordinates that were specified in the previous default (`EPSG:3857`) need to be changed to use `EPSG:4326`. A separate change will allow setting the user projection back to `EPSG:3857`. Also applications that only use `Coordinate.fromLonLat` are going to be covered by another separate change that removes the conversion to `EPSG:3857`, and just creates coordinates in `EPSG:4326` (the new default).

### DIFF
--- a/vaadin-map-flow-parent/vaadin-map-flow-integration-tests/src/main/java/com/vaadin/flow/component/map/demo/SamplerView.java
+++ b/vaadin-map-flow-parent/vaadin-map-flow-integration-tests/src/main/java/com/vaadin/flow/component/map/demo/SamplerView.java
@@ -1,8 +1,9 @@
-package com.vaadin.flow.component.map;
+package com.vaadin.flow.component.map.demo;
 
 import com.vaadin.flow.component.html.Div;
 import com.vaadin.flow.component.html.NativeButton;
 import com.vaadin.flow.component.html.Span;
+import com.vaadin.flow.component.map.Map;
 import com.vaadin.flow.component.map.configuration.Coordinate;
 import com.vaadin.flow.component.map.configuration.feature.MarkerFeature;
 import com.vaadin.flow.component.map.configuration.layer.FeatureLayer;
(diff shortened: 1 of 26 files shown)


vaadin-map-flow-parent/vaadin-map-flow-integration-tests/src/main/java/com/vaadin/flow/component/map/demo/SamplerView.java
```diff
@@ -1,8 +1,9 @@
-package com.vaadin.flow.component.map;
+package com.vaadin.flow.component.map.demo;
```
sissbruecker, line 1: Moved this view to demos, and renamed to `SamplerView` as it was only used as a manual testing playground, and is not used in an IT.

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/Coordinate.java
```diff
@@ -19,10 +19,9 @@
...
 /**
- * Represents map coordinates in a specific projection. Which projection the
- * coordinates are in is not known by the coordinate itself, and developers must
- * ensure themselves to use the same projection between coordinates, the map
- * viewport, and map sources.
+ * Represents map coordinates in a specific projection. Coordinates must be
+ * specified in the map's user projection, which by default is
+ * {@code EPSG:4326}, also referred to as GPS coordinates.
  */
 public class Coordinate {
```
sissbruecker, line 26: This class also has a `fromLonLat` method that needs changes, but I'm going to address that in a follow-up PR.

vaadin-map-flow-parent/vaadin-map-testbench/src/main/java/com/vaadin/flow/component/map/testbench/MapElement.java
```diff
@@ -38,8 +38,10 @@ public void clickAtCoordinates(double x, double y) {
                 "return arguments[0].configuration.getPixelFromCoordinate([arguments[1], arguments[2]])",
                 this, x, y);
 
-        int clickX = startLeft + pixelCoordinates.get(0).intValue();
-        int clickY = startTop + pixelCoordinates.get(1).intValue();
+        int clickX = startLeft
+                + (int) Math.round(pixelCoordinates.get(0).doubleValue());
+        int clickY = startTop
+                + (int) Math.round(pixelCoordinates.get(1).doubleValue());
```
sissbruecker, line 44: With all coordinates now being run through a conversion I ran into some JS number precision issue with pixel values (which are returned as a fractional number on the client). Applied a rounding, otherwise converting into an int would just truncate the number.

DiegoCardoso (approved):

PR #3454, opened 2022-07-13
feat: set projection for coordinates used in public API

sissbruecker:
## Description

Allows to set a different user projection for all coordinates passed to, or received from the public API. This allows to use coordinates in a projection other than `EPSG:4326`, which is the default user projection. Currently the feature is limited to `EPSG:4326` and `EPSG:3857` projections, a separate PR will add support for defining custom projection. The user projection applies to all maps in a browser page, it is not possible to configure this setting for individual maps.

For testing, [epsg.io](https://epsg.io/) is a handy website for looking up projections, and converting coordinates.

### DIFF
--- a/vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/Map.java
+++ b/vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/Map.java
@@ -112,8 +112,11 @@ public class Map extends MapBase {
     public static void setUserProjection(String projection) {
         UI ui = UI.getCurrent();
         if (ui == null || ui.getPage() == null) {
-            throw new IllegalStateException(
-                    "Setting a user projection requires a current UI, and a page.");
+            throw new IllegalStateException("UI instance is not available. "
+                    + "It means that you are calling this method "
+                    + "out of a normal workflow where it's always implicitly set. "
+                    + "That may happen if you call the method from the custom thread without "
+                    + "'UI::access' or from tests without proper initialization.");
         }
         UI.getCurrent().getPage().executeJs(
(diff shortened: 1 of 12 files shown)


vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/Map.java
```diff
@@ -70,6 +80,46 @@ public class Map extends MapBase {
...
+     * this setting.
+     *
+     * @param projection
+     *            the user projection to use for all public facing API
+     */
+    public static void setUserProjection(String projection) {
+        UI ui = UI.getCurrent();
+        if (ui == null || ui.getPage() == null) {
+            throw new IllegalStateException(
+                    "Setting a user projection requires a current UI, and a page.");
```
web-padawan, line 116: Consider using a bit more detailed error description to align with other components, example:

https://github.com/vaadin/flow-components/blob/48830df666083a2ee34e625daa1ca3f7f40489a3/vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java#L734-L738

I'm also not sure if `ui.getPage()` can be `null` as we only check for `ui == null` elsewhere.

web-padawan (approved):

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/Map.java
sissbruecker: Updated the exception message. Let's keep the check for the page, in other parts of the code we don't necessarily use the page instance, and the check doesn't really hurt.

web-padawan (approved):

PR #3456, opened 2022-07-13
fix: force render on style changes

sissbruecker:
## Description

As stated in the OpenLayers [documentation](https://openlayers.org/en/latest/apidoc/module-ol_style_Style-Style.html), changing properties of a style, or one of its children will not automatically trigger a render:
> Any changes made to the style or its children through set*() methods will not take effect until the feature or layer that uses the style is re-rendered.

This change forces an automatic render of all layers when a style or a possible child of a style is changed.

I started working on this because there was a [concern](https://github.com/vaadin/flow-components/issues/2924#issuecomment-1076103409) that we might want to introduce a breaking change to fix this, by making all style classes immutable. However as the issue was fixable by forcing a re-render that should not be necessary.

Fixes https://github.com/vaadin/flow-components/issues/2924

## Type of change

- Bugfix

### DIFF
--- a/vaadin-map-flow-parent/vaadin-map-flow-integration-tests/src/main/java/com/vaadin/flow/component/map/ModifyStylePage.java
+++ b/vaadin-map-flow-parent/vaadin-map-flow-integration-tests/src/main/java/com/vaadin/flow/component/map/ModifyStylePage.java
@@ -1,0 +1,63 @@
+package com.vaadin.flow.component.map;
+
+import com.vaadin.flow.component.html.Div;
+import com.vaadin.flow.component.html.NativeButton;
+import com.vaadin.flow.component.html.Span;
+import com.vaadin.flow.component.map.configuration.Coordinate;
+import com.vaadin.flow.component.map.configuration.feature.MarkerFeature;
+import com.vaadin.flow.component.map.configuration.style.Icon;
+import com.vaadin.flow.router.Route;
+
+@Route("vaadin-map/modify-style")
+public class ModifyStylePage extends Div {
+    public ModifyStylePage() {
+        Map map = new Map();
+
+        MarkerFeature marker = new MarkerFeature(new Coordinate(0, 0),
+                createIcon());
+        map.getFeatureLayer().addFeature(marker);
+
+        NativeButton setImage = new NativeButton("Set image", e -> {
+            Icon icon = createIcon();
+            marker.getStyle().setImage(icon);
+        });
+        setImage.setId("set-style-image");
+
+        NativeButton setImageScale = new NativeButton("Set image scale",
+                e -> marker.getStyle().getImage().setScale(5));
+        setImageScale.setId("set-image-scale");
+
+        Span renderCount = new Span("0");
+        renderCount.setId("render-count");
+        // Register event listener for the rendercomplete event, and increase
+        // render count
+        //@formatter:off
+        renderCount.getElement().executeJs(
+                "const map = $0;" +
+                        "map.configuration.on('rendercomplete', () => {" +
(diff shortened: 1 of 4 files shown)


vaadin-map-flow-parent/vaadin-map-flow-integration-tests/src/main/java/com/vaadin/flow/component/map/ModifyStylePage.java
```diff
@@ -0,0 +1,63 @@
...
+        // render count
+        //@formatter:off
+        renderCount.getElement().executeJs(
+                "const map = $0;" +
+                        "map.configuration.on('rendercomplete', () => {" +
+                        "  const nextValue = parseInt(this.textContent) + 1;" +
+                        "  this.textContent = nextValue.toString();" +
+                        "});",
+                map.getElement()
+        );
```
DiegoCardoso, line 42: I am assuming that this test is aiming to verify that even if `forceRender` is called multiple times, the debouncing mechanism makes the render to be called only once (in most cases). However, on my tests removing the debouncing control from `forceRender` doesn't change the number of renders, so it seems that OL has itself some control to avoid multiple renders.
I think it's still important to have the debounce, since it avoids looping through the layers multiple times unnecessarily.

vaadin-map-flow-parent/vaadin-map-flow/src/main/resources/META-INF/resources/frontend/vaadin-map/mapConnector.js
```diff
@@ -35,6 +35,26 @@ openLayersSetUserProjection('EPSG:4326');
...
+       * and do not trigger change events, for example Style objects or any of their children.
+       * In these cases this method can be called from the synchronization functions of these
+       * objects.
+       * This method will trigger a debounced render of the map by firing a change event from
+       * each layer. We simply render all layers as a sync. function does not know which layer
+       * its synced object is in. Even if the change event is fired from multiple layers, this
+       * only results in a single render of the map.
+       */
+      forceRender() {
+        if (this._forceRenderTimeout) return;
```
DiegoCardoso, line 50: nit: we don't seem to have any lint rules for the JS part of the project. On the web-component project, we have the [curly rule](https://eslint.org/docs/latest/rules/curly) that would fail here with `Expected { after 'if' condition`.

DiegoCardoso (commented):
No blockers found, but added some comments.

vaadin-map-flow-parent/vaadin-map-flow-integration-tests/src/main/java/com/vaadin/flow/component/map/ModifyStylePage.java
sissbruecker: The intention was actually to check that the map rendered at all after updating styles, which was not the case previously. Testing that this results in a single render is more of a side-effect. Another interesting thing is that rendering the map does not necessarily mean that a layer was re-rendered, these only re-render if they have been marked as changed. I added another assertion to the test that tries to verify that by checking that the layer's revision increased (which is was `layer.changed()` does).

Regarding the debounce: Yes, OpenLayers does some debouncing on its own, but I couldn't find any clear documentation on it. So I thought adding our own debounce as a safe-guard would make sense.

vaadin-map-flow-parent/vaadin-map-flow/src/main/resources/META-INF/resources/frontend/vaadin-map/mapConnector.js
sissbruecker: Updated the early return to use curly braces.

DiegoCardoso (approved):
LGTM

PR #3461, opened 2022-07-14
fix: add fromClient to ClientValidatedEvent cto

DiegoCardoso:
## Description

`boolean fromClient` is missing from the `ClientValidatedEvent` constructor signature as the second parameter, which causes the `valid` property to always be set to `true` when the event was coming from the client.

## Type of change

- [X] Bugfix
- [ ] Feature


vursen (approved):
Good catch, thank you. Approved 👍

PR #3477, opened 2022-07-19
refactor: deprecate Coordinate.fromLonLat

sissbruecker:
Since 23.2, the default coordinate system has been changed to EPSG:4326, which allows passing latitude and longitude into the constructor directly. Usages of this method should be replaced with a call to the constructor.

Since the default coordinate system has been changed to EPSG:4326, and the purpose of this method is to return coordinates in EPSG:3857, this method will not return correct results anymore. As a temporary measure, the coordinate system can be changed back to EPSG:3857 using Map.setUserProjection(String). Long-term, usages of this method should be replaced.

### DIFF
--- a/vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/Coordinate.java
+++ b/vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/Coordinate.java
@@ -87,6 +87,7 @@ public class Coordinate {
      *             {@link Map#setUserProjection(String)}. Long-term, usages of
      *             this method should be replaced.
      */
+    @Deprecated
     public static Coordinate fromLonLat(double longitude, double latitude) {
         return fromLonLat(longitude, latitude, Projection.EPSG_3857);
     }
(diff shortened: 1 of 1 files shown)


web-padawan (commented):
Should we also add `@Deprecated` annotation?

sissbruecker:
@web-padawan It should already be there on the methods: https://github.com/vaadin/flow-components/pull/3477/files#diff-b3e1a2d9a37b6e163cc75935a91865609cbba9ba2fdf2600e6fc07eda281a019R76. Or did you mean something else?

web-padawan:
I mean the annotation to be set on the method in addition to JavaDoc:

https://github.com/vaadin/flow-components/blob/fc878abb769255ec36b3be478c74f6dbc35f9c0a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L262

sissbruecker:
> I mean the annotation to be set on the method in addition to JavaDoc:
> 
> https://github.com/vaadin/flow-components/blob/fc878abb769255ec36b3be478c74f6dbc35f9c0a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L262

Right, added those.

web-padawan (approved):

DiegoCardoso (approved):

Issue #3420, opened 2022-07-01
`setValue` doesn't trigger validation for some TB field elements


vursen:
### Description

Currently, Field TestBench elements trigger `blur` and `change` on the component's host rather than the input when setting a value with `setValue()`, which results in validation not running.

https://github.com/vaadin/flow-components/blob/39b52fd4eba14cd4838a72eac5b78bc10e96dfd2/vaadin-text-field-flow-parent/vaadin-text-field-testbench/src/main/java/com/vaadin/flow/component/textfield/testbench/TextFieldElement.java#L35-L40

This is apparently an omission from the a11y slotted refactoring.

### Affected components

- TextFieldElement
- TextAreaElement
- EmailFieldElement
- IntegerFieldElement
- NumberFieldElement
- PasswordFieldElement
- BigDecimalFieldElement

### Expected outcome

`setValue()` of Field TextBench elements should trigger `blur` and `change` on the input rather than the host after a new value is set. 

Note, if `change` and `blur` are triggered on the input, then `setValue()` should also trigger an `input` event before the `change` event as otherwise the input's value won't be synced with the field's `value` property.

### Environment

Vaadin version(s): 21, 22, 23
OS: Mac OS

vursen:
I wonder if it would make sense to introduce a separate method `setInputValue` for emulating setting an input value (that would trigger `blur`, `change`, and so on) while the existing `setValue` would only set the `value` property on the component.

This is inspired by `DatePickerElement`:

https://github.com/vaadin/flow-components/blob/39b52fd4eba14cd4838a72eac5b78bc10e96dfd2/vaadin-date-picker-flow-parent/vaadin-date-picker-testbench/src/main/java/com/vaadin/flow/component/datepicker/testbench/DatePickerElement.java#L130-L140

https://github.com/vaadin/flow-components/blob/39b52fd4eba14cd4838a72eac5b78bc10e96dfd2/vaadin-date-picker-flow-parent/vaadin-date-picker-testbench/src/main/java/com/vaadin/flow/component/datepicker/testbench/DatePickerElement.java#L154-L163

yuriy-fix:
We would probably want to modify `setValue` to not trigger events (i.e. `blur`)

vursen:
The alternative way could be to ensure every field component provides a `sendKeys` method for emulating the user's input. 

[...]

PR #3484, opened 2022-07-20
feat!: provide a `setInputValue` method for TB field elements (WIP)

vursen:
## Description

This PR introduces a `setInputValue` method for TestBench field elements instead of an override of the `setValue` method. 

The problem with `setValue` is that it is currently overridden in the TB field elements to trigger `change` and `blur` events and it mistakenly triggers those events on the component rather than the input element which results in validation not happening. To fix this, it was decided to introduce a `setInputValue` method specifically for the purpose of emulating user input and remove the override of `setValue` so that it defaults to its original implementation defined by `HasStringValueProperty`.

Also, see https://github.com/vaadin/flow-components/issues/3420#issuecomment-1190086875 explaining an advantage of `setInputValue` over `sendKeys`.

A part of https://github.com/vaadin/flow-components/issues/3420

--- 

**Before:**

- `setValue()` sets the `value` property on the component and then triggers `change` and `blur` events on the component. 

**After:**

- `setValue()` only sets the `value` property on the component.
- `setInputValue()` sets the `value` property on the input element and then triggers `input` and `change` events.

## Type of change

- [x] Feature

## Checklist

### DIFF
--- a/vaadin-text-field-flow-parent/vaadin-text-field-testbench/src/main/java/com/vaadin/flow/component/textfield/testbench/BigDecimalFieldElement.java
+++ b/vaadin-text-field-flow-parent/vaadin-text-field-testbench/src/main/java/com/vaadin/flow/component/textfield/testbench/BigDecimalFieldElement.java
@@ -33,16 +33,18 @@ public class BigDecimalFieldElement extends TestBenchElement
         implements HasStringValueProperty, HasLabel, HasPlaceholder, HasHelper {
 
     /**
-     * Sets the given value for the slotted {@code input} element and then triggers
-     * {@code input} and {@code change} DOM events.
+     * Sets the given value for the slotted {@code input} element and then
+     * triggers {@code input} and {@code change} DOM events.
      *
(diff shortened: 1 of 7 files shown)


vursen:
> Why not fix setValue to .. set the value like if the user would? Instead of adding another method so you do not know which one to use

It is a good point about the confusion between the two methods. I was actually afraid that if we fix `setValue` then it will work differently for different components as by default it only sets the `value` property. Apart from that, there is already DatePicker that has both methods.

vursen:
As it turned out, `setValue` in its default implementation is actually useless for components in the `text-field` package. Since all of them are configured to use `ValueChangeMode.ON_CHANGE` by default, just updating the `value` property on the client-side doesn't cause it to sync on the server-side unless a `change` event is fired on the component. 

So, it doesn't make sense to introduce a new method. We should rather fix the old one. I'll open a new `fix!:` PR.

vursen:
Closing in favor of https://github.com/vaadin/flow-components/pull/3490.

PR #3490, opened 2022-07-21
fix!: make setValue emulate real user input

### LINKED ISSUE #3420: `setValue` doesn't trigger validation for some TB field elements

vursen:
## Description

This PR changes the `setValue` method for field components to make it emulate real user input firing `input` and `change` DOM events rather than just setting the `value` property on the host element. This is needed to allow using this method for testing validation behavior on invalid user input – the web components trigger validation on `change` DOM event but not all and not always on `value` property change.

Also, see https://github.com/vaadin/flow-components/issues/3420#issuecomment-1190086875 explaining why using `sendKeys` is not an option for `setValue`.

> **Warning**
> It is an altering behavior change.

Fixes #3420 

## Type of change

- [x] Bugfix

## Checklist

### DIFF
--- a/vaadin-text-field-flow-parent/vaadin-text-field-testbench/src/main/java/com/vaadin/flow/component/textfield/testbench/BigDecimalFieldElement.java
+++ b/vaadin-text-field-flow-parent/vaadin-text-field-testbench/src/main/java/com/vaadin/flow/component/textfield/testbench/BigDecimalFieldElement.java
@@ -32,11 +32,21 @@ import com.vaadin.testbench.elementsbase.Element;
 public class BigDecimalFieldElement extends TestBenchElement
         implements HasStringValueProperty, HasLabel, HasPlaceholder, HasHelper {
 
+    /**
+     * Emulates the user changing the value, which in practice means setting
+     * {@code value} of the {@code input} element to the given value and then
+     * triggering {@code input} and {@code change} DOM events.
+     *
+     * @param string
+     *            the value to set
+     */
     @Override
     public void setValue(String string) {
-        HasStringValueProperty.super.setValue(string);
-        dispatchEvent("change", Collections.singletonMap("bubbles", true));
-        dispatchEvent("blur");
+        TestBenchElement input = $("input").first();
+        input.setProperty("value", string);
+        input.dispatchEvent("input", Collections.singletonMap("bubbles", true));
+        input.dispatchEvent("change",
+                Collections.singletonMap("bubbles", true));
     }
 
 }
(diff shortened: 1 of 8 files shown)


vaadin-text-field-flow-parent/vaadin-text-field-testbench/src/main/java/com/vaadin/flow/component/textfield/testbench/BigDecimalFieldElement.java
```diff
@@ -32,11 +32,21 @@
...
     @Override
     public void setValue(String string) {
-        HasStringValueProperty.super.setValue(string);
-        dispatchEvent("change", Collections.singletonMap("bubbles", true));
-        dispatchEvent("blur");
+        TestBenchElement input = $("input").first();
+        input.setProperty("value", string);
+        input.dispatchEvent("input", Collections.singletonMap("bubbles", true));
+        input.dispatchEvent("change",
+                Collections.singletonMap("bubbles", true));
```
sissbruecker, line 49: Nit: We could consider extracting this in a static helper that receives the TestBench element. Don't have a strong opinion on this though, the code is rather simple.

vursen: Do you have an idea of possible naming for such a helper class?

sissbruecker: We can make it package-private, so naming isn't super important as it can be changed anytime. I would go with a more general `TextFieldElementHelper`, but your suggestion is fine too.

vursen: 👍  Done.

sissbruecker (approved):

DiegoCardoso (approved):

vursen:
I'm still not 100% sure if it is a good idea to override `setValue` to set the value through the input specifically considering that other elements such as `DatePickerElement` will have `setValue` and `setInputValue` because both make sense for it – one is to set a value in the ISO format, and the other one is to set a value in the final date format.

PR #3491, opened 2022-07-21
fix: refresh individual items in virtual list

### LINKED ISSUE #3487: Vaadin 23 VirtualList with 50 or more items is not refreshed, "TypeError: item is undefined" in virtualListConnector.js
### Description

When there are 50 or more items in a VirtualList, a ClickListener stops working after calling refreshItem and the item that should get refreshed (by refreshItem) is not refreshed.

Firefox says "TypeError: item is undefined" in virtualListConnector.js line 108.

Have a look at the screenshot here: https://stackoverflow.com/questions/73051597/vaadin-23-virtuallist-with-50-or-more-items-loses-clicklistener-and-is-not-refre?noredirect=1#comment129022996_73051597

### Expected outcome

I expect that an item in a VirtualList is refreshed after calling refreshItem.

Have a look at https://stackoverflow.com/questions/73051597/vaadin-23-virtuallist-with-50-or-more-items-loses-clicklistener-and-is-not-refre?noredirect=1#comment129022996_73051597

### Steps to reproduce

1. Add the code from https://stackoverflow.com/questions/73051597/vaadin-23-virtuallist-with-50-or-more-items-loses-clicklistener-and-is-not-refre?noredirect=1#comment129022996_73051597 into a view
2. click an item
3. Item should get refreshed, but it is not. (At least not without scrolling the list down and up.)

### Environment

Vaadin version(s): 23.1.3
OS: Ubuntu 22.04

### Browsers

Chrome, Firefox

sissbruecker:
## Description

When using a data provider with a size that exceeds the number of items that the virtual list fetches for a viewport, the connector stores `undefined` items as placeholders for items that have not been loaded. When updating individual items in the data provider, the connector iterates over all items, and tries to map updated to existing items - which fails if an existing item is `undefined` because it tries to access its key.
Updated the connector to check if the item is loaded before attempting to map updated to existing item.

Fixes #3487 

## Type of change

- Bugfix

### DIFF
--- a/vaadin-virtual-list-flow-parent/vaadin-virtual-list-flow-integration-tests/src/test/java/com/vaadin/flow/component/virtuallist/tests/VirtualListDataProviderRefreshIT.java
+++ b/vaadin-virtual-list-flow-parent/vaadin-virtual-list-flow-integration-tests/src/test/java/com/vaadin/flow/component/virtuallist/tests/VirtualListDataProviderRefreshIT.java
@@ -1,0 +1,60 @@
+package com.vaadin.flow.component.virtuallist.tests;
+
+import com.vaadin.flow.component.virtuallist.testbench.VirtualListElement;
+import com.vaadin.flow.testutil.TestPath;
+import com.vaadin.testbench.TestBenchElement;
+import com.vaadin.tests.AbstractComponentIT;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+@TestPath("vaadin-virtual-list/data-provider-refresh")
+public class VirtualListDataProviderRefreshIT extends AbstractComponentIT {
+    private VirtualListElement virtualList;
+    private TestBenchElement updateItemFive;
+    private TestBenchElement updateAllItems;
+
+    @Before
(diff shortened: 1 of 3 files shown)


vaadin-virtual-list-flow-parent/vaadin-virtual-list-flow-integration-tests/src/test/java/com/vaadin/flow/component/virtuallist/tests/VirtualListDataProviderRefreshIT.java
```diff
@@ -0,0 +1,60 @@
...
+        assertItemText(4, "Item5");
+
+        updateItemFive.click();
+
+        // Check updated render
+        assertItemText(4, "Item5 updated");
+    }
+
+    @Test
+    public void refreshAll_allItemsUpdated() {
```
sissbruecker, line 39: This was working before as it uses different methods in the connector, but couldn't hurt to have a test for this as well.

web-padawan (approved):

Issue #3428, opened 2022-07-04
SonarCloud validation problems


sissbruecker:
### Describe your motivation

With its current setup, the SonarCloud validation is causing a lot of noise, making it mostly useless. It reports issues in code that is not changed by the current PR, causing the validation to fail for every PR. Another problem is that some of the issues that are categorized as "Code smells" are marked as a "Blocker", which causes the validation to fail for problems that are arguably not that important. With so many issues reported on every PR, the tool loses its utility, which would be that devs fix the issues that their PR introduces, and instead the validation is basically ignored.

## Validation scope

By scope I refer to the code that should be checked during validation. In `web-components` the validation is set up so that only the modified code is checked, and the expectation is that the same should happen in `flow-components`. However as demonstrated in [this PR](https://github.com/vaadin/flow-components/pull/3413), changing a few lines of code in the select component results in SonarCloud [reporting over 1700 code smells](https://sonarcloud.io/project/issues?id=vaadin_flow-components&pullRequest=3413&resolved=false&types=CODE_SMELL) all over the codebase. The numbers are also not consistent between PRs, for example:
- [this PR](https://github.com/vaadin/flow-components/pull/3399) opened at around the same time reports [198 code smells](https://sonarcloud.io/project/issues?id=vaadin_flow-components&pullRequest=3399&resolved=false&types=CODE_SMELL)
- [this PR](https://github.com/vaadin/flow-components/pull/3422) opened around the same time reports [4387 code smells](https://sonarcloud.io/project/issues?id=vaadin_flow-components&pullRequest=3422&resolved=false&types=CODE_SMELL)

So it doesn't look like the whole codebase is checked every time, or only the recently changed code, but somehow a range of modules.

I would suggest to check the SonarCloud configuration and modify it so that only the changed code is checked for pull requests.

## Code smells reported as blockers

Code smells in SonarCloud have a severity, which can be "Blocker", which causes the validation to fail. Some of these blockers are:
- Add at least one assertion to this test case.
- End this switch case with an unconditional break, return or throw statement.
- Remove this "clone" implementation; use a copy constructor or copy factory instead.

While those are certainly valid points, it's arguable if these should really be categorized as blockers (=block validation / merging). For rules like `Add at least one assertion to this test case.` this is also not straight-forward as we have some tests where assertions are extracted into utility class methods, which the rule does not seem to recognize.

[...]

PR #3500, opened 2022-07-22
chore: disable selected sonar rules for integration tests

sissbruecker:
## Description

As an attempt to reduce noise for SonarCloud validation, this change disables some rules, that can cause validation to fail, for integration test modules. Test sources are already treated differently by SonarQube, however for integration tests we also have the test pages / test setups which run through the normal set of checks. As the test setups don't require as high a quality standard as code that gets released, we can drop some of the rules that don't bring any benefit.

Part of #3428


pom.xml
```diff
@@ -129,6 +129,28 @@
                 </pluginManagement>
             </build>
         </profile>
+        <profile>
+            <id>sonar-cloud</id>
+            <properties>
+                <!-- SonarQube rule exclusions for: Integration tests -->
+                <sonar.issue.ignore.multicriteria>e1,e2,e3,e4,e5</sonar.issue.ignore.multicriteria>
+                <!-- Disable: Public types, methods and fields (API) should be documented with Javadoc -->
```
sissbruecker, line 137: Reported as code smell with high severity, however we don't need to document test setups.

pom.xml
```diff
@@ -129,6 +129,28 @@
...
+                <sonar.issue.ignore.multicriteria.e1.ruleKey>java:S1176</sonar.issue.ignore.multicriteria.e1.ruleKey>
+                <sonar.issue.ignore.multicriteria.e1.resourceKey>**/*-integration-tests/**/*.java</sonar.issue.ignore.multicriteria.e1.resourceKey>
+                <!-- Disable: Constructors should only call non-overridable methods -->
```
sissbruecker, line 140: Reported as code smell with high severity. This is a problem in general with Flow, where you often call `Component.add` in the constructor. Unlikely to result in an actual issue in the simple test setup classes that we have.

pom.xml
```diff
@@ -129,6 +129,28 @@
...
+                <sonar.issue.ignore.multicriteria.e2.ruleKey>java:S1699</sonar.issue.ignore.multicriteria.e2.ruleKey>
+                <sonar.issue.ignore.multicriteria.e2.resourceKey>**/*-integration-tests/**/*.java</sonar.issue.ignore.multicriteria.e2.resourceKey>
+                <!-- Disable: String literals should not be duplicated -->
```
sissbruecker, line 143: Reported as code smell with high severity. Certainly a good practice in general, but not as relevant for test code IMO.

pom.xml
```diff
@@ -129,6 +129,28 @@
...
+                <sonar.issue.ignore.multicriteria.e3.ruleKey>java:S1192</sonar.issue.ignore.multicriteria.e3.ruleKey>
+                <sonar.issue.ignore.multicriteria.e3.resourceKey>**/*-integration-tests/**/*.java</sonar.issue.ignore.multicriteria.e3.resourceKey>
+                <!-- Disable: Inheritance tree of classes should not be too deep -->
```
sissbruecker, line 146: Reported as code smell with high severity. We have some test base classes that exceed this limit, but I don't see a lot of potential for issues here.

pom.xml
```diff
@@ -129,6 +129,28 @@
...
+                <sonar.issue.ignore.multicriteria.e4.ruleKey>java:S110</sonar.issue.ignore.multicriteria.e4.ruleKey>
+                <sonar.issue.ignore.multicriteria.e4.resourceKey>**/*-integration-tests/**/*.java</sonar.issue.ignore.multicriteria.e4.resourceKey>
+                <!-- Disable: Standard outputs should not be used directly to log anything -->
```
sissbruecker, line 149: Reported as code smell with high severity. We only have a few occurrences of this, but I think it's OK to just log something in a demo-like page.

tomivirkki (approved):
The build for this PR doesn't include any Java files https://sonarcloud.io/code?id=vaadin_flow-components&pullRequest=3500 Didn't find a way to test this but the configuration seems to be correct according to the documentation.

PR #3501, opened 2022-07-22
fix: delegate sendKeys calls to the input element

vursen:
## Description

This PR ensures `sendKeys` calls are delegated to the component's slotted input element as calling `sendKeys` directly on the component's host doesn't seem to be reliable.

- [x] TextFieldElement
- [x] BigDecimalFieldElement

Fixes https://github.com/vaadin/flow-components/issues/3502

## Type of change

- [x] Bugfix

## Checklist

### DIFF
--- a/vaadin-text-field-flow-parent/vaadin-text-field-flow-integration-tests/src/test/java/com/vaadin/flow/component/textfield/tests/ValueChangeModeIT.java
+++ b/vaadin-text-field-flow-parent/vaadin-text-field-flow-integration-tests/src/test/java/com/vaadin/flow/component/textfield/tests/ValueChangeModeIT.java
@@ -74,7 +74,7 @@ public class ValueChangeModeIT extends AbstractComponentIT {
 
     @Test
     public void testValueChangeModesForTextField() throws InterruptedException {
-        testValueChangeModes(textField.$("input").first(), "textfield");
+        testValueChangeModes(textField, "textfield");
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


vursen:
Somehow `sendKeys` works on `TextFieldElement` even without delegation... it is not clear though how it determines the input where to send keys.

web-padawan (commented):
I remember that we had a problem with `sendKeys` when adding slotted input, see https://github.com/vaadin/flow-components/pull/2037/commits/529f296717504c2ec280a61bf3ed3683f62620f1. 
It's unclear why it works with `TextField`, unlike other fields 🤔

vursen:
I tried `sendKeys` in a couple of self-written tests to see how it works in different cases and environments. My conclusion is that `sendKeys` doesn't always work reliably when calling it on something other than regular HTML input elements: looks like it sometimes works inconsistently on CI and a local setup. 

Besides, it looks like an oversight that TextFieldElement and BigDecimalFieldElement don't have delegation:

https://github.com/vaadin/flow-components/blob/6c7c6bb735c3fa4ed797d740d866635476a7043b/vaadin-text-field-flow-parent/vaadin-text-field-flow-integration-tests/src/test/java/com/vaadin/flow/component/textfield/tests/ValueChangeModeIT.java#L75-L78

If change `textField.$("input").first()` to `textField`, this test starts failing without delegation.

So I would suggest that we go with this PR after all.

web-padawan (approved):
Looks like we faced the problem with `sendKeys` when updating some components like `NumberField` in  https://github.com/vaadin/flow-components/pull/2037. However, the `TextField` itself was only updated in #2203 and we adjusted tests in https://github.com/vaadin/flow-components/pull/2203/commits/5c3c1eb05ac4981f364bafbabcf85dd51864c927 instead of adding `sendKeys`. Let's add it now.

Also, IMO this could be labelled as a bug fix without `!` - it doesn't look like a breaking change.

vursen:
> Also, IMO this could be labelled as a bug fix without ! - it doesn't look like a breaking change.

Makes sense since it already kind of works as we found out. Done.

DiegoCardoso (approved):

PR #3524, opened 2022-07-28
feat: add HasInputValue interface

yuriy-fix:
Introduces `HasInputValue` which can be implemented by a component to get the populated state of the internal input's value.

Requires https://github.com/vaadin/web-components/pull/4276

### DIFF
--- a/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/HasInputValue.java
+++ b/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/HasInputValue.java
@@ -33,8 +33,8 @@ public interface HasInputValue extends HasElement {
      * @return <code>true</code> if the input's value is populated,
      *         <code>false</code> otherwise
      */
-    @Synchronize(property = "__inputValuePopulated", value = "input-value-populated-changed")
+    @Synchronize(property = "_hasInputValue", value = "has-input-value-changed")
     default boolean isInputValuePopulated() {
-        return getElement().getProperty("__inputValuePopulated", false);
+        return getElement().getProperty("_hasInputValue", false);
     }
 }
--- a/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/HasInputValue.java
+++ b/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/HasInputValue.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2000-2022 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component.shared;
+
+import com.vaadin.flow.component.HasElement;
+import com.vaadin.flow.component.Synchronize;
+
+/**
+ * Mixin interface for components that has an internal input element. Intended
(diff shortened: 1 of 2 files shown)


vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/HasInputValue.java
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * Gets the populated state of the input's value, which is {@code false} by
+     * default.
+     *
+     * @return <code>true</code> if the input's value is populated,
+     *         <code>false</code> otherwise
+     */
+    @Synchronize(property = "_hasInputValue", value = "has-input-value-changed")
+    default boolean isInputValuePopulated() {
```
DiegoCardoso, line 37: Should it be also renamed to align with the property name?

yuriy-fix: Renamed to `isInputValue` 👌

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/HasInputValue.java
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * Gets the populated state of the input's value, which is {@code false} by
+     * default.
+     *
+     * @return <code>true</code> if the input's value is populated,
+     *         <code>false</code> otherwise
+     */
+    @Synchronize(property = "_hasInputValue", value = "has-input-value-changed")
+    default boolean isInputValue() {
```
DiegoCardoso, line 37: The name feels a bit odd to read, IMO. Could we change it to `hasInputValue`?
```suggestion
    default boolean hasInputValue() {
```

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/HasInputValue.java
```diff
@@ -0,0 +1,40 @@
...
+import com.vaadin.flow.component.HasElement;
+import com.vaadin.flow.component.Synchronize;
+
+/**
+ * Mixin interface for components that has an internal input element. Intended
+ * for internal use.
+ *
+ * @author Vaadin Ltd
+ */
+public interface HasInputValue extends HasElement {
```
knoobie, line 27: I'm normally a big fan of interfaces and re-usability but I have the feeling that this (as stated in the web component) is a private API and should not be exposed as public methods and therefore not be an interface, instead I would suggest to add this method inside all components as protected of even private API so that flow users aren't bothered with it.

The name `hasInputValue` or `isInputValue` is so generic that a lot of people probably going to use it just to check if there is something inside the field, expecting that `setValue("something")` changes it to `true` even tho the web component states "input element has a non-empty value entered by the user".

yuriy-fix: I have had similar fears. 

Hoped that **not** having type specified parameter type (unlike `HasValue`), documentation mentioned that it's `Intended for internal use` and not having a setter would discourage users from using this interface instead of the proper one. Initial naming was `isInputValuePopulated` which is probably more descriptive name in terms of what it is  returning.

However, I would agree that it still would be available for the users and somebody could probably use it by mistake.
I do appreciate your feedback as an actual user of our product! 🙇  I will change the approach in the PR in order to avoid confusions.

knoobie: I would argue that for most people any Java API is public that they can access, even tho there is Javadocs stating otherwise (which most people probably don't read) 

Like people typing "textField.value" in their IDE and use the first method that pop ups and works.. with this in mind moving this somewhere else sounds the best to reduce public API cluttering and improve DX. 

Another option could be to rename it so something totally technical like "isWebComponentUserPopulated" which sounds so obscure that probably less people are going to find and even use it.. but it perfectly describes it's meaning ;) But that feels like a way of "improved DX by obscurity" 😅

yuriy-fix:
Closing in favour of introducing private methods in individual components. Example: https://github.com/vaadin/flow-components/pull/3550

PR #3545, opened 2022-08-01
docs: deprecate legacy IronResizeEvent APIs

web-padawan:
## Description

The `IronResizableBehavior` has been removed from `vaadin-split-layout` in https://github.com/vaadin/web-components/pull/3200.
So this API should be removed as well, as it's not really working anymore. Let's deprecate it for removal in V24.

## Type of change

- Deprecation


yuriy-fix (commented):
We can deprecate the generated class entirely following example with `ComboBox` https://github.com/vaadin/flow-components/blob/55b03b15cf3bd3f36dc332f035e00ec7d0ea59bd/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/GeneratedVaadinComboBox.java#L24

Related task: https://github.com/vaadin/components-team-tasks/issues/606

web-padawan:
> We can deprecate the generated class entirely

Sure, but I'd prefer to do that in a separate PR for cleaner commit history. Also we might want to move stuff from `GeneratedSplitLayout` to `SplitLayout` itself, to make generated class unused before deprecating it.

yuriy-fix (approved):

PR #3563, opened 2022-08-08
feat: add setDefaultMultiSortPriority to Grid

web-padawan:
## Description

Related to https://github.com/vaadin/platform/issues/3052

Added a static method to configure default multi-sort priority as proposed in https://github.com/vaadin/platform/issues/3052#issuecomment-1202325061
See https://github.com/vaadin/web-components/pull/4305 where the web component API for this was added.

## Type of change

- Feature

## Checklist

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/DefaultMultiSortPriorityIT.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/DefaultMultiSortPriorityIT.java
@@ -36,8 +36,7 @@ public class DefaultMultiSortPriorityIT extends AbstractComponentIT {
 
     @Test
     public void setDefaultMultiSortPriority_addGrid_priorityIsUsed() {
-        findElement(By.id("btn-set-msp-append")).click();
-        findElement(By.id("btn-add-grid")).click();
+        findElement(By.id("btn-set-add")).click();
 
         GridElement grid = $(GridElement.class).id("multi-sort-priority-grid");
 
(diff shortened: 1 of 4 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -415,6 +416,40 @@ public enum MultiSortPriority {
...
+        UI ui = UI.getCurrent();
+        if (ui == null || ui.getPage() == null) {
+            throw new IllegalStateException("UI instance is not available. "
+                    + "It means that you are calling this method "
+                    + "out of a normal workflow where it's always implicitly set. "
+                    + "That may happen if you call the method from the custom thread without "
+                    + "'UI::access' or from tests without proper initialization.");
+        }
+        UI.getCurrent().getPage().executeJs(
+                "window.Vaadin.Flow.gridConnector.setDefaultMultiSortPriority($0)",
```
sissbruecker, line 449: For the Map component we needed the `executeJs` call, and the UI scope constraints, because the state is effectively stored in a JS library running in the browser page, and we wanted to be able to differentiate per UI / per user.

However for this use-case, I feel we could make this a lot simpler by making the setting global by storing the default in a static property, and then applying it in the `Grid` constructor by setting the respective attribute. Using the `setDefaultMultiSortPriorityAppend` function of the web component does not seem necessary here.

web-padawan: Good point, thanks. Updated as you suggested and reverted the change to the connector.

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/DefaultMultiSortPriorityIT.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Before
+    public void init() {
+        open();
+    }
+
+    @Test
+    public void setDefaultMultiSortPriority_addGrid_priorityIsUsed() {
+        findElement(By.id("btn-set-msp-append")).click();
+        findElement(By.id("btn-add-grid")).click();
```
sissbruecker, line 40: This test actually fails, because the grid is already created before the button click. I think the test passes in CI because it is re-run after a failure, and a second run passes because the static state still applies after the first test run.

I think there are two issues here:
1. The test setup should create the grid instance after changing the setting - this can easily be fixed by creating the grid in the click handler, rather than just add it
2. While that would fix the test setup, it leaves the application in a non-default state, which could have side effects on other tests, so I think the setting should be reverted to its default. One option would be to use a test teardown method, however in this case that would require another Selenium interaction which seems brittle. Instead I would suggest we add a button that does everything in one step: apply the setting, create the grid, then revert the setting. Preferrably with some comment why we revert the setting immediately.

Or we just add a unit test for this, which would be easier to control.

web-padawan: > add a button that does everything in one step: apply the setting, create the grid, then revert the setting

Thanks, updated the test.

sissbruecker (approved):

tomivirkki (approved):

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -3152,6 +3170,12 @@ public void setMultiSort(boolean multiSort, MultiSortPriority priority) {
                 "Multi-sort priority must not be null");
         setMultiSort(multiSort);
 
+        if (priority != defaultMultiSortPriority) {
```
tomivirkki: This check may be problematic when changing the priority dynamically (checking the value agains the default one, but not the effective one)

web-padawan: Thanks, changed to always update the priority to avoid such case.

PR #3576, opened 2022-08-10
refactor!: make setInputValue simulate real user input

vursen:
## Description

This PR changes the `setInputValue` method of `DatePickerElement` to make it simulate real user input by sending the respective characters + <kbd>Enter</kbd> via the Selenium `sendKeys` API. The clearing of the input's value is now done by sending <kbd>Shift</kbd> + <kbd>Home</kbd> and then <kbd>Backspace</kbd> instead of clearing the input's `value` property, see https://github.com/vaadin/flow-components/pull/3576#discussion_r945471738.

In particular, this PR is needed to allow using the method for testing validation behavior on invalid user input – the web component trigger validation on `change` DOM event but not on `_inputValue` property change.

I also removed explicit opening and closing of the overlay as it doesn't seem to be necessary since `DatePicker` already automatically opens on input and closes on <kbd>Enter</kbd>, but let's see if the tests will pass.

> **Warning**
> It is an altering behavior change.

Related to #3420 
Related to #3496 

## Type of change

- [x] Bugfix

## Checklist

### DIFF
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-testbench/src/main/java/com/vaadin/flow/component/datepicker/testbench/DatePickerElement.java
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-testbench/src/main/java/com/vaadin/flow/component/datepicker/testbench/DatePickerElement.java
@@ -22,11 +22,11 @@ import com.vaadin.testbench.TestBenchElement;
 import com.vaadin.testbench.elementsbase.Element;
 
 import java.time.LocalDate;
-import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 
 import org.openqa.selenium.By;
+import org.openqa.selenium.Keys;
 
 /**
  * A TestBench element representing a <code>&lt;vaadin-date-picker&gt;</code>
@@ -155,18 +155,18 @@ public class DatePickerElement extends TestBenchElement
     }
 
     /**
-     * Opens the overlay, sets the value to the inner input element as a string
-     * and closes the overlay. This simulates the user typing into the input and
-     * triggering an update of the value property.
+     * Emulates the user changing the value. This in practice means clearing
+     * {@code value} of the {@code input} element and then setting the new value
+     * by sending the respective characters + Enter via the Selenium
+     * {@code sendKeys} API.
+     *
+     * @param string
+     *            the value to set
      */
     public void setInputValue(String value) {
-        this.open();
         var input = $("input").first();
-        input.setProperty("value", value);
-        input.dispatchEvent("input", Collections.singletonMap("bubbles", true));
-        input.dispatchEvent("change",
-                Collections.singletonMap("bubbles", true));
-        this.close();
+        input.setProperty("value", "");
+        input.sendKeys(value, Keys.ENTER);
     }
 
     /**
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-testbench/src/main/java/com/vaadin/flow/component/datepicker/testbench/DatePickerElement.java
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-testbench/src/main/java/com/vaadin/flow/component/datepicker/testbench/DatePickerElement.java
@@ -22,6 +22,7 @@ import com.vaadin.testbench.TestBenchElement;
 import com.vaadin.testbench.elementsbase.Element;
 
 import java.time.LocalDate;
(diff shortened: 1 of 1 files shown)


vaadin-date-picker-flow-parent/vaadin-date-picker-testbench/src/main/java/com/vaadin/flow/component/datepicker/testbench/DatePickerElement.java
```diff
@@ -152,14 +154,18 @@ protected String getValue() {
...
+     * {@code sendKeys} API.
+     *
+     * @param string
+     *            the value to set
      */
     public void setInputValue(String value) {
-        this.open();
-        setProperty("_inputValue", value);
-        this.close();
+        var input = $("input").first();
```
sissbruecker, line 166: I think using `var` is untypical in this codebase, let's write out the type instead.

sissbruecker (commented):
I think the behavioral change is probably fine, but I wonder if we really need this? For devs using this API, why they would need to trigger the client-side validation? If the purpose is just internal testing, we can just use sendKeys in our tests, rather than using `setInputValue`.

vursen:
> If the purpose is just internal testing, we can just use sendKeys in our tests, rather than using setInputValue.

We can, even though it will be quite clunky because we will have to clear the input element every time, but:

> but I wonder if we really need this

It is actually not only about validation. The previous JSDoc stated that the method simulates the user typing something into the input. This is not quite accurate if `input` and `change` events are missing. Also, there can potentially be logic in DatePicker on those events that will be not invoked with the previous implementation, so I thought it is better to make the method closer to real world usage.

sissbruecker:
Thinking some more about this I think it's OK to change this. I was a bit concerned about the overlay opening and closing part, but I can't fathom what purpose that had in the original implementation, as the overlay was closed again before returning from this method, so there would be no chance to interact with it afterwards.

Let's maybe rather tag this as a refactor though?

vursen:
> Let's maybe rather tag this as a refactor though?

It could be treated as a refactoring, yeah. I just took the same approach as for https://github.com/vaadin/flow-components/pull/3490 but I don't mind if we rename it to `refactor!:`. @yuriy-fix do you have any objections / other opinions?

vursen:
> I was a bit concerned about the overlay opening and closing part, but I can't fathom what purpose that had in the original implementation

I guess that was just an old way of committing value. DatePicker commits the value on close but in order to close it, you have to open it first. :)

vaadin-date-picker-flow-parent/vaadin-date-picker-testbench/src/main/java/com/vaadin/flow/component/datepicker/testbench/DatePickerElement.java
vursen: 👍 Done.

sissbruecker (approved):

vursen:
**side note:** `setInputValue` doesn't seem to be the best naming for a method that actually **commits** the value by closing DatePicker or pressing <kbd>Enter</kbd>. I guess eventually it makes sense to drop it in favor of `getInputElement()`, see https://github.com/vaadin/flow-components/issues/3589 for more details.

vaadin-date-picker-flow-parent/vaadin-date-picker-testbench/src/main/java/com/vaadin/flow/component/datepicker/testbench/DatePickerElement.java
```diff
@@ -154,14 +155,18 @@ protected String getValue() {
...
+     *
+     * @param string
+     *            the value to set
      */
     public void setInputValue(String value) {
-        this.open();
-        setProperty("_inputValue", value);
-        this.close();
+        TestBenchElement input = $("input").first();
+        input.setProperty("value", "");
```
vursen, line 168: As I found, when calling this method with an empty string, no `input` event will be fired with the current approach of clearing the value.

vursen: Fixed by https://github.com/vaadin/flow-components/pull/3576/commits/7657174dcc69ac13aea9324cd786e2d4cd05ded7

yuriy-fix (approved):

PR #3597, opened 2022-08-15
refactor!: put getDefaultValidator implementation behind a feature flag

vursen:
## Description

The PR puts the `getDefaultValidator` implementation behind the feature flag `forceFieldValidation` as we agreed in https://github.com/vaadin/flow/pull/14285#issuecomment-1209102605.

- [x] DatePicker
- [x] TimePicker 
- [x] DateTimePicker 
- [x] IntegerField
- [x] NumberField
- [x] EmailField
- [x] TextArea

A follow-up to #3381 

## Type of change

- [x] Refactor

## Checklist

### DIFF
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
@@ -24,6 +24,7 @@ import java.util.Objects;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import com.vaadin.experimental.FeatureFlags;
 import com.vaadin.flow.component.AttachEvent;
 import com.vaadin.flow.component.ComponentEventListener;
 import com.vaadin.flow.component.shared.HasAllowedCharPattern;
@@ -461,7 +462,11 @@ public class DatePicker extends GeneratedVaadinDatePicker<DatePicker, LocalDate>
 
     @Override
     public Validator<LocalDate> getDefaultValidator() {
-        return (value, context) -> checkValidity(value);
+        if (getFeatureFlags().isEnabled(FeatureFlags.ENFORCE_FIELD_VALIDATION)) {
+            return (value, context) -> checkValidity(value);
+        }
+
+        return Validator.alwaysPass();
     }
 
     @Override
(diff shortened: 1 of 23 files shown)


vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
```diff
@@ -744,6 +750,18 @@ public void removeThemeVariants(DatePickerVariant... variants) {
...
+    /**
+     * Gets the feature flags for the current UI.
+     * <p>
+     * Exposed with protected visibility to support mocking
+     *
+     * @return the current set of feature flags
+     */
+    protected FeatureFlags getFeatureFlags() {
+        return FeatureFlags
+                .get(UI.getCurrent().getSession().getService().getContext());
```
knoobie, line 762:
```suggestion
                .get(VaadinService.getCurrent().getContext());
```

Even tho I have a feeling that this could create a problem / bottle neck if e.g. the context isn't available and it called like literally every-time a field is instantiated, which could be hundred times on a single view. Should this value be cached somewhere and not queried every time again? r.h. `FieldFeatureFlagUtils::isEnforceFieldValidationEnabled` as static method somewhere which can cache the value in a thread local if you wanna "over-engineer" it, or just a simple static boolean

vursen: Looking at the implementation of `FeatureFlags.get`, it seems to cache feature flags to an attribute of the Vaadin context object, so getting these flags here should be quite cheap. Do you rather mean that calling `VaadinService.getCurrent().getContext()` can be not efficient?

knoobie: I'm thinking more about the call to `VaadinService.getCurrent().getContext()` but I could be totally wrong

knoobie: I'm also hesitent to use `UI::getCurrent` or other `getCurrent` implementations like above, because I've seen issue with it in the past e.g. exceptions inside the `DatePicker` constructor (where I can't find the issue)

yuriy-fix: `VaadinService.getCurrent()` would return an existing instance from a Map data structure, so it shouldn't be a big deal, and then `getContext()` is always returning (not exactly a classic singleton instance, but still) a single existing instance of `VaadinContext`, so it shouldn’t be an expensive operation either.

knoobie: Thanks for the update - this looks better :) This comment can be closed

yuriy-fix (approved):

PR #3438, opened 2022-07-06
 Add AbstractGridMultiSelectionModel#getSelectedItemIds

willemv:
## Description

Add a method, usable by subclasses, that provides access to the internal set of selected item ids.

Fixes #3246

This is a resubmission of #3247 as requested by @yuriy-fix in the last comment there.

## Type of change

* [ ]  Bugfix
* [x]  Refactoring
* [ ]  Feature

## Checklist
* [x]  I have read the contribution guide: https://vaadin.com/docs-beta/latest/guide/contributing/overview/
* [x]  I have added a description following the guideline.
* [x]  The issue is created in the corresponding repository and I have referenced it.
* [x]  I have added tests to ensure my change is effective and works as intended.
* [x]  New and existing tests are passing locally with my change.
* [x]  I have performed self-review and corrected misspellings.

#### Additional for `Feature` type of change
* [ ]  Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.

PR #3598, opened 2022-08-15
feat: introduce getSelectedItemIds

yuriy-fix:
Based on https://github.com/vaadin/flow-components/pull/3438.
WIP (documentation updates)

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridMultiSelectionModel.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridMultiSelectionModel.java
@@ -163,10 +163,11 @@ public abstract class AbstractGridMultiSelectionModel<T>
 
     /**
      * Returns an unmodifiable view of the selected item ids.
-     *
-     * The returned Set may be a direct view of the internal data structures of this class.
-     * A defensive copy should be made by callers when iterating over this Set and modifying
-     * the selection during iteration to avoid ConcurrentModificationExceptions.
+     * <p>
+     * The returned Set may be a direct view of the internal data structures of
+     * this class. A defensive copy should be made by callers when iterating
+     * over this Set and modifying the selection during iteration to avoid
+     * ConcurrentModificationExceptions.
      */
     protected Set<Object> getSelectedItemIds() {
         return Collections.unmodifiableSet(this.selected.keySet());
@@ -187,11 +188,7 @@ public abstract class AbstractGridMultiSelectionModel<T>
         if (item != null) {
             selected.put(itemId, item);
         }
-        doUpdateSelection(
-                selected,
-                Collections.emptyMap(),
-                false
-        );
+        doUpdateSelection(selected, Collections.emptyMap(), false);
     }
 
     @Override
@@ -204,11 +201,7 @@ public abstract class AbstractGridMultiSelectionModel<T>
         if (item != null) {
             deselected.put(itemId, item);
         }
-        doUpdateSelection(
-                Collections.emptyMap(),
-                deselected,
-                false
-        );
+        doUpdateSelection(Collections.emptyMap(), deselected, false);
         selectionColumn.setSelectAllCheckboxState(false);
     }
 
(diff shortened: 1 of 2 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridMultiSelectionModel.java
```diff
@@ -159,49 +161,67 @@ public Set<T> getSelectedItems() {
...
 
     @Override
     public void select(T item) {
-        if (isSelected(item)) {
+        Object itemId = getItemId(item);
+        if (isSelectedItemId(item)) {
             return;
         }
-        Set<T> selected = new HashSet<>();
+        Map<Object, T> selected = new HashMap<>();
```
knoobie, line 190: Should it be LinkedHashMap everywhere to make sure the selection order is preserved?

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridMultiSelectionModel.java
```diff
@@ -105,12 +105,13 @@ protected void remove() {
 
     @Override
     public void selectFromClient(T item) {
-        if (isSelected(item)) {
+        Object itemId = getItemId(item);
+        if (isSelectedItemId(itemId)) {
```
sissbruecker, line 109: I don't see any benefit in this change, it just makes the code more convoluted, duplicates the logic from `isSelected`, and confusing because it's not clear why this doesn't simply use `isSelected`. Same with most of the other changes here that now create map instances all over the place, while sets would just do the job.

If we want to provide `getSelectedItemIds` and `isSelectedItemId` we could just add them and mention in the JavaDoc that these are exposed for sub-classes. As protected methods are considered public API, and with proper documentation, there is basically no risk of these getting removed accidentally.

yuriy-fix: That is the intention and the reason why I have overtaken the PR from the contributor.
I will add needed notes to the documentation and clean up unnecessary changes.

sissbruecker: Sounds good thanks. I'll resolve this then.

willemv: Just FYI, there is one benefit to these changes: the item ID is calculated only once. In the original code, the `isSelected(item)` will itself call `getItemId` again.

This is only a slight benefit, to be sure, but I thought to mention it.

As said, this is just FYI, I'm certainly fine keeping the original code here.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridMultiSelectionModel.java
yuriy-fix: Refactored.

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/AbstractGridMultiSelectionModelSelectedItemsTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+    private void verifySelection(String... values) {
+        Assert.assertEquals(Set.of(values), selectionModel.getSelectedItems());
+        Assert.assertEquals(
+                Stream.of(values).map(grid.getDataProvider()::getId)
+                        .collect(Collectors.toSet()),
+                selectionModel.getSelectedItemIds());
+        for (String value : values) {
+            Assert.assertTrue(selectionModel.isSelected(value));
+            Assert.assertTrue(selectionModel.isSelected(value));
```
sissbruecker, line 179: Looks like this line is duplicated.

sissbruecker (approved):

PR #3602, opened 2022-08-18
fix: prevent adding overlay to UI when forwarding happens in beforeEnter

### LINKED ISSUE #3414: LoginOverlay remains in the DOM after detach
### Description

When a route extends LoginOverlay, the login overlay is not removed from the DOM when navigating away.

For instance if you redirect from a login view to a hello world view, the `vaadin-login-overlay` element remains in the DOM and covers the hello world view

### Expected outcome

The Hello world view should be shown

```java
@Route(value = "login")
public class LoginView extends LoginOverlay implements BeforeEnterObserver {
    public void beforeEnter(BeforeEnterEvent event) {
            event.forwardTo(HelloWorldView.class);
   }
}

@Route("")
public class HelloWorldView extends Div {
   public HelloWorldView() {
     add (new Text("Hello"));
   }
}
```

### Steps to reproduce

Open /login

### Environment

Vaadin: 23.1.2
Flow: 23.1.2
Java: Homebrew 11.0.12
OS: aarch64 Mac OS X 12.4
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/103.0.0.0 Safari/537.36

### Browsers

Issue is not browser related

ugur-vaadin:
## Description

The issue was caused by registrations regarding adding the component to UI persisting even after a forwarding/rerouting happens in `beforeEnter`.

This PR introduces an `AfterNavigationListener` for programmatic navigation events. Whenever such an event occurs, the listener removes the previously added registration regarding adding the component to UI.

Applies to ConfirmDialog, Dialog, LoginOverlay, and Notification.

Fixes #3414

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/LoginOverlay.java
+++ b/vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/LoginOverlay.java
@@ -131,6 +131,10 @@ public class LoginOverlay extends AbstractLogin implements HasStyle {
                     .beforeClientResponse(ui, context -> {
                         ui.addToModalComponent(this);
                         autoAddedToTheUi = true;
+                        if (afterProgrammaticNavigationListenerRegistration != null) {
+                            afterProgrammaticNavigationListenerRegistration
+                                    .remove();
+                        }
                     });
             if (ui.getSession() != null) {
                 afterProgrammaticNavigationListenerRegistration = ui
--- a/vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/LoginOverlay.java
+++ b/vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/LoginOverlay.java
@@ -29,6 +29,9 @@ import com.vaadin.flow.component.UI;
 import com.vaadin.flow.component.dependency.JsModule;
 import com.vaadin.flow.component.dependency.NpmPackage;
 import com.vaadin.flow.dom.Style;
+import com.vaadin.flow.internal.StateTree;
+import com.vaadin.flow.router.NavigationTrigger;
+import com.vaadin.flow.shared.Registration;
 
 /**
  * Server-side component for the {@code <vaadin-login-overlay>} component.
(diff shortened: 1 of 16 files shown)


vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/LoginOverlay.java
```diff
@@ -122,10 +127,22 @@ private UI getCurrentUI() {
     private void ensureAttached() {
         if (getElement().getNode().getParent() == null) {
             UI ui = getCurrentUI();
-            ui.beforeClientResponse(ui, context -> {
-                ui.addToModalComponent(this);
-                autoAddedToTheUi = true;
```
tomivirkki, line 127: Could this be fixed by wrapping this inside `if (this.isAttached()) { ... }`?

ugur-vaadin: Using that check also works for this case but causes another test to fail. I will investigate that.

ugur-vaadin: It fails for the self-attach test case, because at the time the this block is executed, the overlay is not attached

vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/LoginOverlay.java
```diff
@@ -122,10 +127,22 @@ private UI getCurrentUI() {
...
-            });
+            StateTree.ExecutionRegistration addToUiRegistration = ui
+                    .beforeClientResponse(ui, context -> {
+                        ui.addToModalComponent(this);
+                        autoAddedToTheUi = true;
```
tomivirkki, line 133: Should we add `afterProgrammaticNavigationListenerRegistration.remove()` here also? Otherwise, the listener will stay there, invoking on every navigation event, until a programmatic navigation occurs, right?

ugur-vaadin: Definitely. Updated the PR.

tomivirkki (approved):

PR #3649, opened 2022-09-01
feat: add HasTooltip interface

tomivirkki:
## Description

Adds the `HasTooltip` interface to `vaadin-flow-components-base`. Components implementing the interface will get the following API:
- `Tooltip setTooltip(String text)`
  - Creates a `<vaadin-tooltip slot="tooltip" text="text">` element and adds it inside the target component
  - Returns the `Tooltip` handle that can be used to configure and control the tooltip
- `String getTooltip()`
  - Returns the `text` property that was set using `setTooltip(String text)`

Excluded from this PR (to be submitted separately):
- `HasTooltip` integration to components

## Type of change

- Feature

## Note

- This PR is targeting the `feat/tooltip` branch, not `master`.

### DIFF
--- a/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/HasTooltip.java
+++ b/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/HasTooltip.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2022 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component.shared;
+
+import com.vaadin.flow.component.HasElement;
+
+/**
+ * Mixin interface for components that support a tooltip.
+ *
+ * @author Vaadin Ltd
+ */
+public interface HasTooltip extends HasElement {
+
+    /**
+     * Sets a tooltip text for the component.
+     *
+     * @param text
+     *            The tooltip text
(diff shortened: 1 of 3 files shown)


vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/HasTooltip.java
```diff
@@ -0,0 +1,52 @@
...
+        tooltip.setText(text);
+        return tooltip;
+    }
+
+    /**
+     * Gets the tooltip text of the component.
+     *
+     * @return The tooltip text
+     */
+    default String getTooltip() {
```
web-padawan, line 46: The original idea was to get the reference to `Tooltip` so that user can set `position` etc/
Would it be a better idea to use `getTooltipText()` for this instead?

tomivirkki: Will fix. I wonder if it would be unexpected if `getTooltip` returns a different type than what you pass to `setTooltip`?

web-padawan: I'm not sure. @yuriy-fix @rolfsmeds WDYT about `setTooltip` / `getTooltip` API?

tomivirkki:
```
var t1 = field.setTooltip("foo")
var t2 = field.setTooltip("bar")
// Is t2 the same as t1 but with the text changed?

var t3 = field.setTooltip(null);
// Should t3 still be the same Tooltip handle but only its text is set to null?

var t4 = field.getTooltip();
// Is t4 null or the ref to t1?

var t5 = field.setTooltip("baz");
// t5 still the same as t1?
```

tomivirkki: Fixed with the assumption that `HasTooltip` always retains the same `Tooltip` reference once `setTooltip` has been called.

rolfsmeds: That's... probably how I'd expect the API to work. I think :)

web-padawan (approved):

PR #3663, opened 2022-09-05
test: handle flaky tests

tomivirkki:


vaadin-iron-list-flow-parent/vaadin-iron-list-flow-integration-tests/src/test/java/com/vaadin/flow/component/ironlist/tests/IronListIT.java
```diff
@@ -201,7 +202,7 @@ public void listWithComponentRenderer() {
     }
 
     @Test
-
+    @Ignore("Flaky test")
```
tomivirkki, line 205: This test never fails locally, only on CI. Ignoring should be safe since IronList is deprecated and will likely no longer change.

web-padawan (approved):

Issue #2065, opened 2021-08-30
Extend Java API for Upload I18N


vursen:
There have been introduced three new properties to the Upload I18N object (see [vaadin/web-components#2316](https://github.com/vaadin/web-components/pull/2316/files#diff-e074f7c5f5dbe41fe0ae0d8b1851d02905b4ad867ae6c6c9606b84e315372073R428-R432)):

- `file.retry`
- `file.start`
- `file.remove`

All these properties should be added to the Flow counterpart ([UploadI18N.java](https://github.com/vaadin/flow-components/blob/master/vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/UploadI18N.java)) as well.

PR #3676, opened 2022-09-06
feat: add i18n translations for the file action buttons

vursen:
## Description

This PR allows configuring i18n translations for the file action buttons of the Upload component.

Fixes https://github.com/vaadin/flow-components/issues/2065

## Type of change

- [x] Feature

## Checklist

### DIFF
--- a/vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/UploadI18N.java
+++ b/vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/UploadI18N.java
@@ -29,6 +29,7 @@ public class UploadI18N implements Serializable {
     private Error error;
     private Uploading uploading;
     private Units units;
+    private File file;
 
     /**
      * Translations for dropping files.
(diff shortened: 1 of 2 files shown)


vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/UploadI18N.java
```diff
@@ -460,6 +461,78 @@ public Units setSize(List<String> size) {
         }
     }
 
+    /**
+     * file translations.
+     */
+    public static class File implements Serializable {
```
knoobie, line 467: Should this class be renamed to UploadFileI18n or FileI18n - I personally think this could clash with the really prominent import of File in java which could irritate people

vursen: Good point, I also had a hunch that File might cause a potential naming conflict. Let me rename it.

vursen: My only concern is that then it becomes not consistent with other classes.

knoobie: It's totally up to you guys :) just wanted to mention the possibility, there were some reports in the past in discord and stack overflow with different classes that matches e.g. the swing import that confused people

vursen: Considering that you can this class through the `UploadI18N` namespace to avoid conflicts, I guess we can keep this naming for now. @yuriy-fix what do you think?

yuriy-fix: I can understand the mentioned concerns and would avoid adding classes that matches other names.
However, in this particular case, the possibility of human error is quite limited and the class name is aligned with others available in the `UploadI18N`.
In case people would be facing issues we will reconsider names for the entire internationalization class.

yuriy-fix (approved):

PR #3688, opened 2022-09-09
chore: improve vite config for IT modules

sissbruecker:
## Description

Improves the default Vite config for integration test modules:
- disables dependency optimization to prevent caching connectors between application starts
- adds optional plugin for using local version of web-components


shared/web-components-vite-plugin.ts
```diff
@@ -0,0 +1,27 @@
+const path = require('path');
```
vursen, line 1:
```suggestion
import path from 'path';
```

sissbruecker: Done

vaadin-app-layout-flow-parent/vaadin-app-layout-flow-integration-tests/vite.config.ts
```diff
@@ -0,0 +1,27 @@
...
+  // https://vitejs.dev/config/
+
+  // Disable the Vite deps pre-bundling in the app to prevent Vite
+  // from caching connectors’ source code in the file system
+  optimizeDeps: {
+    disabled: true
+  },
+
+  // Use local version of web-components, disabled by default
+  // To use this un-comment the lines below and adapt the path to your local checkout
```
vursen, line 18: Let's maybe mention that it should be an absolute path to `node_modules`, because it can be not immediately noticeable from the path below.

vursen: > NOTE: Make sure that the path is an absolute one and that it points to the node_modules directory in the web components monorepo.

Like we have in the web components README.

sissbruecker: Done

vursen (approved):

PR #3712, opened 2022-09-13
fix: force refreshValue to update input value on the client-side

vursen:
## Description

[The earlier fix](https://github.com/vaadin/flow-components/pull/2340) by @sissbruecker doesn't cover the case when ComboBox is cleared in the same round-trip as a custom value is set, e.g when calling `comboBox.clear()` in a `customValueSet` listener. 

The issue manifests when trying to set a custom value on a combo-box without value. In that case, the [internal `customValueSet` listener](https://github.com/vaadin/flow-components/pull/2340/files#diff-53fde8312387fc148b8bc6f52de9306e11a962639c19c541bd2c761c5a24ec8cR309-R310) does have effect, setting `_inputElementValue` from an empty value to the custom value. However, since ComboBox is immediately cleared, it comes out like nothing has changed for the server in the end: `_inputElementValue` remains empty as it was, although it is not empty on the client-side.

This PR changes `refreshValue()` to force `_inputElementValue` client-side update by using the `executeJs` API.

Depends on 
- https://github.com/vaadin/flow-components/pull/3716

Fixes https://github.com/vaadin/flow-components/issues/1090

## Type of change

- [x] Bugfix

## Checklist

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridTooltipPage.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridTooltipPage.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2022 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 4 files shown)


vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -325,7 +329,7 @@ protected void refreshValue() {
         getDataGenerator().generateData(value, json);
         getElement().setPropertyJson(PROP_SELECTED_ITEM, json);
         getElement().setProperty(PROP_VALUE, keyMapper.key(value));
-        getElement().setProperty(PROP_INPUT_ELEMENT_VALUE,
+        getElement().executeJs("this._inputElementValue = $0",
```
vursen, line 332: **note:** I added this change more for the sake of consistency. It doesn't matter for the fix.

tomivirkki (approved):

PR #3717, opened 2022-09-14
feat: add setTooltipEnabled() to Avatar

web-padawan:
## Description

Added `setTooltipEnabled()` to change `withTooltip` property on the `vaadin-avatar`.

Note, the `Avatar` is not supposed to implement `HasTooltip` because its tooltip is always pre-configured to show the name, the abbreviation, or the "anonymous" string in case when neither of those is set by the user.

## Type of change

- Feature

### DIFF
--- a/vaadin-avatar-flow-parent/vaadin-avatar-flow/src/main/java/com/vaadin/flow/component/avatar/Avatar.java
+++ b/vaadin-avatar-flow-parent/vaadin-avatar-flow/src/main/java/com/vaadin/flow/component/avatar/Avatar.java
@@ -295,6 +295,28 @@ public class Avatar extends Component
         getElement().setProperty("colorIndex", colorIndex);
     }
 
+    /**
+     * Gets the enabled state of the avatar tooltip, which is {@code false} by
+     * default.
+     *
+     * @return <code>true</code> if the tooltip is shown on hover or focus,
+     *         <code>false</code> otherwise
+     */
+    public boolean isTooltipEnabled() {
+        return getElement().getProperty("withTooltip", false);
+    }
+
+    /**
+     * Sets the enabled of the avatar tooltip.
+     *
+     * @param tooltipEnabled
+     *            <code>true</code> to show the tooltip on hover or focus,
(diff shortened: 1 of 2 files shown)


tomivirkki (approved):

vaadin-avatar-flow-parent/vaadin-avatar-flow/src/main/java/com/vaadin/flow/component/avatar/Avatar.java
```diff
@@ -295,6 +295,28 @@ public void setColorIndex(Integer colorIndex) {
...
     }
 
+    /**
+     * Gets the enabled state of the avatar tooltip, which is {@code false} by
+     * default.
+     *
+     * @return <code>true</code> if the tooltip is shown on hover or focus,
+     *         <code>false</code> otherwise
+     */
+    public boolean isTooltipEnabled() {
```
tomivirkki, line 305: nit: why isn't the naming the same on the Flow & Web Components:
Web: "withTooltip"
Flow: "tooltipEnabled"

web-padawan: The API `setTooltipEnabled()` was requested in https://github.com/vaadin/web-components/issues/1606
Also it's likely that Flow users will look for `setTooltip` in the autocomplete.

As for the web component, I don't have a strong opinion. We could also use `tooltip-enabled`.
The current `withTooltip` property name was inspired by `withBackdrop` in the overlay.

Generally, we don't have consistent boolean property names in different web components 🤷‍♂️ 
E.g. we have `clearButtonVisible` and `revealButtonHidden`, but also `withBackdrop`.

tomivirkki: Yeah, there definitely are some goof-ups in the current names of some properties. No changes are needed in the scope of this PR now anyways.

PR #3725, opened 2022-09-15
feat: make setTooltip(null) remove existing tooltip

### LINKED ISSUE #3722: Add option to remove tooltip by using setTooltip(null)
### Describe your motivation

Currently, there is no option to remove tooltip after it has been set to the component.

Use `setTooltip(null)` to remove existing `Tooltip` and detach the element from DOM.

web-padawan:
## Description

Fixes #3722

Currently, `setTooltip(null)` clears the tooltip text but does not remove the tooltip. 
Added logic to remove it from DOM as that's what some users might expect.

## Type of change

- Feature


tomivirkki:
I assume this derives from https://github.com/vaadin/platform/discussions/3196#discussioncomment-3643504

I agree there needs to be a way for a developer to clear the tooltip so that it doesn't show when hovering over a component. However, the current `.setTooltip(null)` already achieves that.

Actually removing the `<vaadin-tooltip>` element might be an overkill for the purpose and probably doesn't make any difference from a Flow developer's point of view. Is there some good justification for the additional logic that I'm missing?

PR #3744, opened 2022-09-19
fix: clean up old content on tabsheet.add

tomivirkki:
## Description
Even though it's an edge case, it is techincally possible to add the same `Tab` instance to a `TabSheet` more than once and with different content. In this case, make sure any possible content previously related to the same tab gets cleaned up.

## Type of change

- Fix

### DIFF
--- a/vaadin-tabs-flow-parent/vaadin-tabs-flow/src/test/java/com/vaadin/flow/component/tabs/tests/TabSheetTest.java
+++ b/vaadin-tabs-flow-parent/vaadin-tabs-flow/src/test/java/com/vaadin/flow/component/tabs/tests/TabSheetTest.java
@@ -66,6 +66,23 @@ public class TabSheetTest {
     }
 
     @Test
+    public void addSameTabAgainWithNewContent_oldContentRemoved() {
+        // Add a tab with content
+        var content0 = new Span("Content 0");
+        var tab0 = tabSheet.add("Tab 0", content0);
+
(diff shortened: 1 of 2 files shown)


vaadin-tabs-flow-parent/vaadin-tabs-flow/src/test/java/com/vaadin/flow/component/tabs/tests/TabSheetTest.java
```diff
@@ -65,6 +65,23 @@ public void addTab_contentAdded() {
...
+
+        // Assert that the content is attached to the parent (the tab is
+        // selected)
+        Assert.assertEquals(tabSheet, content0.getParent().get());
+
+        // Add the same Tab instance again but with a new content component
+        tabSheet.add(tab0, new Span("Content 0"));
+
+        // Check that the old content is no longer attached to the parent
+        Assert.assertFalse(content0.getParent().isPresent());
```
knoobie, line 82: Just asking: would a developer expect that the "newly" added tab/content gets added last or at the position of the "old tab"? 

(I'm not sure myself as well, would be good to test the index tho)

tomivirkki: Added the test case.

web-padawan (approved):

Issue #3715, opened 2022-09-14
ComboBox with page size < default fails on fetching next pages


SonReimer:
### Description

Having a combobox with 1000 string items and page size set to values 10,20,25 (< default 50) only fetches the first 10 pages and fails on fetching subsequent pages.

I think this issue is related to #2736/#2744

### Expected outcome

I would expect the comboBox to fetch all pages when scrolling.

### Minimal reproducible example

    	ComboBox<String> comboBox = new ComboBox<>();
        add(comboBox);
    	List<String> items = new ArrayList<>();
        for (int i = 0; i < 1000; i++) {
        	items.add("item #" + i);  
        }
    	comboBox.setItems(items);
    	comboBox.setPageSize(25);

### Steps to reproduce

1. Vary the page size for the example above with values 10,20,25 and 49
2. When the comboBox shows up, scroll slowly down
3. for 10,20 and 25 the combBox shows only the first 10 pages (100,200,250 items) and a progress indicator shows up
    Sometimes I get the following error:
    TypeError: pageCallbacks[page] is not a function (http://localhost:8080/VAADIN/@fs/C:/projects/others/vaadinflow/my-app23/node_modules/.vite/deps/@vaadin_flow-frontend_comboBoxConnector__js.js?v=cbe367e4:47:30)
There seems to be an error in Vaadin Combo Box: callback is not a function Please submit an issue to https://github.com/vaadin/flow-components/issues/new/choose
4. For an unset page size or page size 49 it seems to work

### Environment

Vaadin version(s): 23.2.0
OS: Windows

### Browsers

Firefox

vursen:
Confirmed the bug. Looks like the server stops responding to `setRequestedRange` requests at some point.

https://user-images.githubusercontent.com/5039436/191422021-0ddbe4ab-7bbc-49ef-94d4-8b8864d7afe7.mov

vursen:
When reproducing the bug, I see a lot of errors in the console:

```
2022-09-22T11:05:15.589+03:00  WARN 75345 --- [nio-8080-exec-2] c.v.flow.data.provider.DataCommunicator  : Attempted to fetch more items from server than allowed in one go: number of items requested '165', maximum items allowed '150'.
2022-09-22T11:05:15.618+03:00  WARN 75345 --- [nio-8080-exec-5] c.v.flow.data.provider.DataCommunicator  : Attempted to fetch more items from server than allowed in one go: number of items requested '180', maximum items allowed '150'.
[...]

PR #3767, opened 2022-09-22
fix: make connector respect server-side max allowed range length

vursen:
## Description

There is a [server-side limitation](https://github.com/vaadin/flow/blob/6124a96af94be10cdb5749aa3384daa47003ed62/flow-data/src/main/java/com/vaadin/flow/data/provider/DataCommunicator.java#L332-L339) on how big the requested items range can be which equals 10 * pageSize. This PR ensures the ComboBox connector doesn't request ranges of а size exceeding 10 * pageSize to prevent exceptions and thus endless loading.

Fixes https://github.com/vaadin/flow-components/issues/3715

## Type of change

- [x] Bugfix

## Checklist


ugur-vaadin (approved):

tomivirkki:
Would fixing https://github.com/vaadin/flow/issues/14630 help in this case?

vursen:
> Edit: it would fix the issue in case the minimum is set to 500 items / request because that's also the current minimum active range length in the combobox connector

Right, it would be a more proper fix but we can keep a tests improvement part from this PR

vursen:
Extracted test improvements into https://github.com/vaadin/flow-components/pull/3799.

vursen:
Now that the test improvements part has got to `master` as part of #3799, I'm closing this PR.

Issue #1680, opened 2019-08-22
Value objects which change equals/hashCode lead to weird editing issues


mvysny:
Quoting a customer:

I've spent about 2 hours tracking down a problem involving GridPro (which might apply to any editable Grid) and I'd like to share my experience if you please.

I'm trying to use a GridPro to edit a Value Object (VO) class: a simple POJO with getters and setters per field.

Our application was originally developed under EJB2. When converting to EJB3, we understood at the time that every entity should return .equals() false when tested against a vanilla copy from the database if any field has been updated. This (misunderstanding?) has resulted in a lot of VO classes (many of which are used in entity class definitions) that compare all fields in the equals() method.

With GridPro, when you update the VO through a custom editor, it seems to use the getId() to identify which object has been updated [presumably by a call to refreshItem(vo)]. In the default DataProvider implementation, this didn't find the original row in the table because it uses the VO's .equals() method, and a field has been edited.
The result is that the VO becomes "detached" from the editor after the first edit: you can still edit the cell: but a custom renderer will display the original value, or the default renderer will display blank; and subsequent updates in the editor don't update the value. There is no error thrown (even though the presumed refreshItem() could not have found the item to refresh). This use of refreshItem() doesn't seem to be documented, so I was looking for the problem in the wrong place. (Most strictly, getId() documents that it's used to "determine equality between two items"; arguably it shouldn't be called here because there is only 1 item involved).

It's an easy fix (override DataProvider.getId()). But the behaviour of the custom editor is different to the built-in editors (which appear to work in this case), and it feels like there should be an error thrown or warning logged if the edited value suddenly "disappears" from the DataProvider. It was certainly not easy to identify the problem from the symptoms, even with the aid of a debugger.

As an aside: GridPro.createEditColumn() should not be protected. You cannot usefully override it, because whichever column you create would not then be added to the private idToColumnMap field - unless using a custom renderer.

This is my original description of the problem, which appeared as a framework bug:

I'm taking another look at GridPro, but I have a problem editing a numeric value using a "custom" editor (a NumberField); the rendered value disappears when I commit the edit.

Initially the value in the cell is 20.00
I can tell from the debugger that the renderer's ItemLabelGenerator is being called after the edit, the value set on the VO and returning the updated value (20.01). But the UI isn't updated to reflect the change.
Opening the editor again seems to return the updated value (20.01). If I edit this to 25.00 and commit the change, the display returns to 20.00 and the VO's setter is not called.
[...]

PR #3782, opened 2022-09-26
feat: warn about item id changes after editing field

sissbruecker:
## Description

When an item updater in GridPro modifies a field that is used in generating that item's data provider ID, that can lead to issues with the grid not refreshing individual items correctly.

This change adds a warning when an item updater modifies the item's data provider ID, which provides hints for the probable cause and how to fix it.

Closes https://github.com/vaadin/flow-components/issues/1680

## Type of change

-  Feature

### DIFF
--- a/vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/java/com/vaadin/flow/component/gridpro/GridPro.java
+++ b/vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/java/com/vaadin/flow/component/gridpro/GridPro.java
@@ -19,6 +19,7 @@ package com.vaadin.flow.component.gridpro;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


DiegoCardoso (approved):
The changes look good to me. 
I was trying to find how it could be tested by, e.g., checking the logger output, but it doesn't seem to be simple to do.

sissbruecker:
> The changes look good to me. I was trying to find how it could be tested by, e.g., checking the logger output, but it doesn't seem to be simple to do.

Sorry, I could have added that, since there isn't a test case. I used this view to verify it:
```java
@Route("vaadin-grid-pro/id-warning")
public class GridProIdWarningPage extends Div {
    public GridProIdWarningPage() {
        List<Item> items = List.of(
                new Item("A", "1"),
                new Item("B", "2"),
                new Item("C", "3")
        );

        GridPro<Item> gridPro = new GridPro<>();
        gridPro.addEditColumn(Item::getKey).text(Item::setKey).setHeader("Key");
        gridPro.addEditColumn(Item::getValue).text(Item::setValue).setHeader("Value");
        gridPro.setItems(items);
        add(gridPro);
    }

    private static class Item {
        private String key;
        private String value;

        public String getKey() {
            return key;
        }

        public void setKey(String key) {
            this.key = key;
        }

        public String getValue() {
            return value;
        }

        public void setValue(String value) {
            this.value = value;
        }

        public Item(String key, String value) {
            this.key = key;
            this.value = value;
        }

        @Override
        public boolean equals(Object o) {
            if (this == o) return true;
            if (o == null || getClass() != o.getClass()) return false;
            Item item = (Item) o;
            return Objects.equals(key, item.key) && Objects.equals(value, item.value);
        }

        @Override
        public int hashCode() {
            return Objects.hash(key, value);
        }
    }
}
```

When running the app one also needs to ensure that warnings are logged to console, for example by removing the `-q` option from the command generated by the `scripts/run.sh` script.

PR #3785, opened 2022-09-27
fix: downgrade ts syntax used in tooltip.ts

tomivirkki:
While upgrading `docs`, we discovered that `docs-app` doesn't yet support this modern TS syntax. Let's downgrade the syntax used in `tooltip.ts` for now.


web-padawan (approved):

tomivirkki:
FYI: [Updating Gatsby version in docs-app](https://github.com/vaadin/docs-app/pull/363) eliminates this limitation also

PR #3793, opened 2022-09-28
fix: reset data communicator on non sequential page request

### LINKED ISSUE #3678: [ComboBox] Regression: Endless items loading when opening the combo-box after selecting an item
### Description

Seems to be a regression caused by something in 23.2 since the issue is not reproducible in 23.1. 

Every once in a while, after selecting an item and opening the combo-box again, the combo-box gets stuck in the loading state. It must be noted though that the issue only occurs if you are scrolling the dropdown fast before making a selection.

https://user-images.githubusercontent.com/5039436/188892715-7503be39-3cad-41c4-84ef-61026789700b.mov

### Expected outcome

I would expect the items to be rendered when opening the combo-box after selecting any item.

```java
@Route(value = "combo-box")
public class ComboBoxView extends Div {
    public ComboBoxView() {
        List<String> items = new ArrayList<>();
        for (int i = 0; i < 1000; i++) {
            items.add("Item " + i);
        }
        comboBox.setItems(items);
        add(comboBox);
    }
}
```

### Steps to reproduce

1. Open the dropdown.
2. Scroll down fast somewhere to the middle of the list.
3. Select an item.
4. Open the dropdown again.

### Environment

Vaadin version(s): 23.2 and higher
OS: Mac OS

### Browsers

Issue is not browser related

vursen:
## Description

This PR makes the connector reset the ComboBox data communicator on non-sequential page request. This is needed to force the data communicator to resend the range in case there are several synchronous scroll jumps back and forth (can be programmatic scrolling), by the end of which ComboBox ends up at the same position as before but the cache has been already cleared.

- [x] Implementation
- [ ] Integration tests

Fixes #3678 

## Type of change

- [x] Bugfix

## Checklist

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/AbstractComboBoxIT.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/AbstractComboBoxIT.java
@@ -136,6 +136,9 @@ public class AbstractComboBoxIT extends AbstractComponentIT {
         comboBox.openPopup();
         executeScript("arguments[0]._scroller.scrollIntoView(arguments[1])",
                 comboBox, index);
+        // Wait for the scroller to request missing pages.
+        getCommandExecutor().getDriver()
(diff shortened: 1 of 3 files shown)


vursen:
The fix is not valid since it breaks other functionality as follows from the failing tests. Closing the PR.

PR #3798, opened 2022-09-29
fix: throw error if requested page or item count exceeds the Flow limit

### LINKED ISSUE #3708: Grid and TreeGrid: Empty item lines after 10 x pageSize items.
### Description

When I want to see all 600 root items in my demo application, I see only 20 because I set the page size to be 2.

### Expected outcome

I expect the page size to be only performance-related.

```
package com.vaadin.qa.views.helloworld;


import com.vaadin.flow.component.treegrid.TreeGrid;
import com.vaadin.flow.function.ValueProvider;
import com.vaadin.flow.router.RouteAlias;
import org.apache.commons.lang3.RandomStringUtils;

import java.util.ArrayList;
import java.util.Random;

@PageTitle("Hello World")
@Route(value = "hello")
@RouteAlias(value = "")

    private class EnverNode {
        private final String eins;
        private final String zwei;

        public EnverNode(){
            eins = RandomStringUtils.randomAlphabetic(15);
            zwei = RandomStringUtils.randomAlphabetic(15);
        }

        public String getEins(){
            return eins;
        }

        public String getZwei(){
            return zwei;
        }
    }

    public HelloWorldView() {
        Random random = new Random();

        setSizeFull();;

        TreeGrid<EnverNode> treeGrid = new TreeGrid<>();
        treeGrid.setWidthFull();
        treeGrid.setAllRowsVisible(true);

        treeGrid.setPageSize(2); // TODO BROKEN: You will only see N * 10 items, followed by blank lines.

        ArrayList<EnverNode> enverNodes = new ArrayList<>();
        for (int i=0; i < 600; i++){
            enverNodes.add(new EnverNode());
        }
        treeGrid.getTreeData().addRootItems(enverNodes);

        for (int i=0; i<12000; i++){
            treeGrid.getTreeData().addItem(enverNodes.get(random.nextInt(enverNodes.size())), new EnverNode());
        }

        ValueProvider<EnverNode, String> textProvider = (EnverNode::getEins);
        treeGrid.addHierarchyColumn(textProvider);

        add(treeGrid);
    }

}

```

### Steps to reproduce

Use Vaadin 23.2 standard demo application, use this view above as the 'hello' view.
`mvn spring-boot:run`

### Environment

Vaadin version(s): 23.2
OS: OSX
Tested: Firefox / Chrome latest

### Browsers

Chrome, Firefox, Issue is not browser related

ugur-vaadin:
## Description

A recent [change](https://github.com/vaadin/flow/pull/12003) in Flow has limited the number of pages in the response to 10. Also, there is an [issue](https://github.com/vaadin/flow/issues/14630) to add another constraint to the responses to have a maximum item count of 500. This PR ensures that these constraints are respected in the `Grid`. If more pages or items are requested from the, an exception will be thrown. 

Should be merged after the Flow [issue](https://github.com/vaadin/flow/issues/14630) is merged.

Fixes #3708 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -345,15 +345,20 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
           let firstPage = Math.max(0, firstNeededPage);
           let lastPage =
             parentKey !== root ? lastNeededPage : Math.min(lastNeededPage, Math.floor(grid.size / grid.pageSize));
+          let pageCount = lastPage - firstPage + 1;
+          let requestedItemCount = pageCount * grid.pageSize;
+          if ((!parentKey || parentKey === root) && (pageCount > 10 || requestedItemCount > 500)) {
+            throw 'Attempted to fetch more items from server than allowed in one go. ' +
+            'Maximum allowed page count is 10, maximum allowed item count is 500.';
```
tomivirkki, line 352: The problem is caused by using `setAllRowsVisible(true)` together with too many items.

The error message should be more informative about what actions the user should take to fix the issue:
1. Consider not using `setAllRowsVisible(true)` when you have a large amount of items (not only to cover this issue but also to avoid performance bottlenecks resulting from rendering thousands of DOM items).
2. If for some reason this is not an option, increase the page size of the grid so that rendering every item at once doesn't result in a request for over 10 pages

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -345,15 +345,20 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
           let firstPage = Math.max(0, firstNeededPage);
           let lastPage =
             parentKey !== root ? lastNeededPage : Math.min(lastNeededPage, Math.floor(grid.size / grid.pageSize));
+          let pageCount = lastPage - firstPage + 1;
+          let requestedItemCount = pageCount * grid.pageSize;
+          if ((!parentKey || parentKey === root) && (pageCount > 10 || requestedItemCount > 500)) {
+            throw 'Attempted to fetch more items from server than allowed in one go. ' +
```
tomivirkki, line 351: Should this error rather be thrown on the server-side component?

tomivirkki (changes requested):

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -3300,6 +3300,15 @@ private void confirmUpdate(int id) {
 
     @ClientCallable(DisabledUpdateMode.ALWAYS)
     private void setRequestedRange(int start, int length) {
+        if (length > 500 && length / getPageSize() > 10 && isAllRowsVisible()) {
+            throw new IllegalArgumentException(
+                    "Attempted to fetch more items from server than allowed in one go. "
+                            + "Maximum allowed page count is 10. Consider not using setAllRowsVisible(true) "
+                            + "when you have a large amount of items (not only to cover this issue but also "
+                            + "to avoid performance bottlenecks resulting from rendering thousands of DOM items). "
```
tomivirkki, line 3308: Could also mention the data payload transfer as a performance issue, something like: `...performance bottlenecks resulting from transferring the full item data set at once and then rendering an excess amount of DOM elements)...`

tomivirkki (approved):
LGTM

> Should be merged after the Flow https://github.com/vaadin/flow/issues/14630 is merged.

I think this change would already provide value for users, even before the DataProvider issue in Flow is fixed.

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridTest.java
```diff
@@ -111,4 +113,60 @@ public void dataProviderListeners_gridAttachedAndDetached_oldDataProviderListene
...
+        try {
+            Method method = Grid.class.getDeclaredMethod("setRequestedRange",
+                    int.class, int.class);
+            method.setAccessible(true);
+            method.invoke(grid, start, length);
+        } catch (NoSuchMethodException | SecurityException
+                | IllegalAccessException | IllegalArgumentException
+                | InvocationTargetException e) {
+            if (e.getCause() instanceof IllegalArgumentException) {
+                throw (IllegalArgumentException) e.getCause();
```
vursen, line 167: Did I read it correctly that this just re-throws `IllegalArgumentException`? Could it be then excluded from the union of exceptions above?

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridTest.java
```diff
@@ -111,4 +113,60 @@ public void dataProviderListeners_gridAttachedAndDetached_oldDataProviderListene
...
+            }
+            Assert.fail("Could not call Grid.setRequestedRange");
```
vursen, line 169: It would be helpful to know the cause of the failure here. Could we simply not handle these exceptions like [here](https://github.com/vaadin/flow-components/pull/3823/files#diff-e54ae4713dd2efeae52550fd49e826c987db76c85171182b6fda8e70c8ba4cc3R20-R28)? Does that approach have any negative consequences?

PR #3844, opened 2022-10-11
fix: allow setting filterable data providers with setItems

### LINKED ISSUE #3842: [Grid] `setItems` doesn't accept filterable data providers
### Description

The `setItems` method doesn't accept filterable data providers although the deprecated `setDataProvider` method does, so it becomes not possible to migrate from the deprecated method in some cases. 

An example is the Bakery app where [Grid](https://github.com/vaadin/bakery-app-starter-flow-spring/blob/9321112b9d317499a614981f9cd913e578a8d13d/src/main/java/com/vaadin/starter/bakery/ui/views/storefront/OrderPresenter.java#L51) on the storefront page is initialized with [`OrdersGridDataProvider`](https://github.com/vaadin/bakery-app-starter-flow-spring/blob/9321112b9d317499a614981f9cd913e578a8d13d/src/main/java/com/vaadin/starter/bakery/ui/dataproviders/OrdersGridDataProvider.java#L29) that implements `FilterablePageableDataProvider`. From the UX perspective, it is the case when there is an external field (search bar) that filters the grid.

### Expected outcome

The `setItems` method should accept filterable data providers.

```java
@Route(value = "grid")
    public GridView() {
        Grid<Person> grid = new Grid<Person>();
        PersonDataProvider dataProvider = new PersonDataProvider();
        grid.setItems(dataProvider);
        add(grid);
    }

    public class PersonDataProvider
            extends AbstractBackEndDataProvider<Person, String> {

        final List<Person> DATABASE = Arrays.asList(new Person("Anna"),
                new Person("John"), new Person("Michael"));

        @Override
        protected Stream<Person> fetchFromBackEnd(Query<Person, String> query) {
            int offset = query.getOffset();
            int limit = query.getLimit();
            String filter = query.getFilter().orElse("");

            return DATABASE.stream()
                    .filter(person -> person.getName().contains(filter))
                    .skip(offset).limit(limit);
        }

        @Override
        protected int sizeInBackEnd(Query<Person, String> query) {
            return (int) fetchFromBackEnd(query).count();
        }
    }

    public class Person {
        private String name;

        public Person(String name) {
            this.name = name;
        }

        public String getName() {
            return name;
        }

        public void setName(String name) {
            this.name = name;
        }
    }
}
```

### Steps to reproduce

---

### Environment

Vaadin version(s): 24
OS: MacOS

### Browsers

_No response_

vursen:
## Description

This PR allows setting filterable data providers with the `setItems` method to align with the deprecated `setDataProvider` method.

- [x] Implementation
- [ ] Unit tests 

Fixes #3842

## Type of change

- [x] Bugfix

## Checklist

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
@@ -2526,7 +2526,7 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
     }
 
     @Override
-    public GridDataView<T> setItems(DataProvider<T, Void> dataProvider) {
+    public GridDataView<T> setItems(DataProvider<T, ?> dataProvider) {
         setDataProvider(dataProvider);
         return getGenericDataView();
     }


vursen:
It turned out to be challenging to allow `setItems` to accept filterable data providers. Replacing `Void` with a wild card results in a name clash, so we decided to put off this issue for now.

PR #3862, opened 2022-10-13
docs: clarify usage of deprecated methods

### LINKED ISSUE #2762: The deprecation of setDataProvider is confusing
I can't classify it as a bug or feature.

I'm trying to migrate from Vaadin 14 to 23.

In Vaadin 23 in the Grid:
-  `setDataProvider` is deprecated in `Grid` and the Javadoc says "use instead one of the setItems methods"
but internally it's using `setDataProvider` which is confusing.

Also in the `TreeGrid`, `public void setDataProvider(DataProvider<T, ?> dataProvider)` is not deprecated but the new method `setItems` is deprecated.

It's very confusing.

I would suggest:
- Deprecate `public void setDataProvider(DataProvider<T, ?> dataProvider)` everywhere.
- Create an undeprecated internal method `setInternalDataProvider` and use this one internally.
- Use only `public void setDataProvider(HierarchicalDataProvider<T, ?> hierarchicalDataProvider)` in the `TreeGrid`, since you can't use anything else.
I don't have a good solution for the `setItems` in `TreeGrid`. Maybe create a new superclass `CommonGrid` that don't implement `HasDataView` ... and extends the `Grid` / `TreeGrid` from it. But it's probably harder than this.

sissbruecker:
## Description

Improve JavaDoc for deprecated methods in `TreeGrid` to clarify that they have been deprecated to indicate that they are not supported. I don't think any further action is necessary to resolve the linked issue.

Closes #2762

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
@@ -335,6 +335,21 @@ public class TreeGrid<T> extends Grid<T>
                 (ComponentEventListener) listener);
     }
 
+    /**
+     * Tree grid only supports hierarchical data providers. Use
+     * {@link #setDataProvider(HierarchicalDataProvider)} instead.
+     * <p>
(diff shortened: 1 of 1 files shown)


ugur-vaadin (approved):
All the changes seem pretty clear.

web-padawan (approved):

PR #3888, opened 2022-10-17
Test for Flow issue #14654 (Grid/DataProvider serialization)

tepi:
## Description

Adds test for https://github.com/vaadin/flow/issues/14654

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridDataProviderSerializableIT.java
```diff
@@ -0,0 +1,42 @@
+/*
```
tomivirkki, line 1: If this is only about Serialization, could it be handled with a [Junit test](https://github.com/vaadin/flow-components/blob/master/vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridSerializableTest.java) (instead of a TB test)?

tepi: Not really.. The two cases this tests only fail when the serialization is done in a real app - doing it from JUnit everything passes fine (which is probably why it has been missed even though a JUnit test already existed.

tomivirkki: Tried to reduce the test case into a unit test. With the following setup, I'm at least getting the same error message as with the TB test case, ptal

```java
@Test
public void gridWithoutColumnsIsSerializable() throws IOException, ClassNotFoundException {
    var grid = new Grid<String>();
    var layout = new Div(grid);
    grid.setItems("foo");

    var bos = new ByteArrayOutputStream();
    var oos = new ObjectOutputStream(bos);
    oos.writeObject(layout);
    oos.flush();
    
    var bis = new ByteArrayInputStream(bos.toByteArray());
    var ois = new ObjectInputStream(bis);
    ois.readObject();
}
```

tepi: Thanks! Looks good enough for now, even though it does not seem to break for the other related issue. I'll replace the TB test with a unit test.

tomivirkki (approved):

PR #3905, opened 2022-10-19
refactor: remove useless function from grid connector

### LINKED ISSUE #3829: Grid: Bug in gridConnector.js
### Description

This line 
808:  for (let i = 0; i < selectedItems; i++) {

does not seem to be ok in gridConnector.js
it has to be 
for (let i = 0; i < selectedItems.length; i++) {

### Expected outcome

Fix the bug

no example just review the code.

### Steps to reproduce

Load long grid with many pages of items - selection would mess up.

### Environment

Vaadin version(s): 23.2.0
OS:  no matter

### Browsers

_No response_

sissbruecker:
## Description

Removes the `isSelectedOnGrid` function which has a flawed implementation in that it always returns false, effectively doing nothing in the places where it's used.

Closes #3829

## Type of change

- Refactor

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -675,9 +675,9 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
             }
             cache[pkey][page] = slice;
 
-            grid.$connector.doSelection(slice.filter((item) => item.selected && !isSelectedOnGrid(item)));
+            grid.$connector.doSelection(slice.filter((item) => item.selected));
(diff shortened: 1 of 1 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -810,17 +810,6 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
           grid._cache.updateSize();
         });
 
-        const isSelectedOnGrid = function (item) {
-          const selectedItems = grid.selectedItems;
-          for (let i = 0; i < selectedItems; i++) {
```
sissbruecker, line 815: The loop condition `i < selectedItems` was flawed because `selectedItems` is an array, effectively never running the loop, making the function always return false.

web-padawan (approved):

PR #3913, opened 2022-10-19
fix: preserve components added to notification on reload

### LINKED ISSUE #3875: [@PreserveOnRefresh] Notification that contains other Components doesn't refresh correctly
### Description

When refreshing the browser while having an open Notification containing another Component, the Notification is there after the refresh, but it ist empty. Its Components are no longer visible. Somethimes the Notification even shows other Components inside itself that should be somewhere eles on the Page. This might be related to #3541 .

### Expected outcome

The Notification shows the same Componants after refresh as before.

new Notification(new Span("Hello World")).open;

### Steps to reproduce

1. Add the above snippet to a program.
2. Open in Browser.
3. Refresh the Browser(F5)

### Environment

Vaadin version(s): 23.2.4
OS:  Windows 10

### Browsers

Chrome, Edge, Issue is not browser related

sissbruecker:
## Description

When adding components to a notification, the notification adds a Flow component renderer to its internal template element, and configures the component renderer with the ID of the node that holds all elements added to the notification. When using preserve on refresh, the component renderer was not updated on reload and the node ID can now reference a different element, resulting in the notification card containing other elements from the page.

This change updates the component renderer on attach, so that it will reference the correct node after a reload.

Fixes #3875 

## Type of change

- Bugfix


vaadin-notification-flow-parent/vaadin-notification-flow/src/main/java/com/vaadin/flow/component/notification/Notification.java
```diff
@@ -65,13 +65,14 @@ public class Notification extends GeneratedVaadinNotification<Notification>
 
     private Registration afterProgrammaticNavigationListenerRegistration;
 
-    private SerializableConsumer<UI> deferredJob = new AttachComponentTemplate();
+    private SerializableConsumer<UI> configureTemplateJob = new ConfigureComponentRendererJob();
 
-    private class AttachComponentTemplate implements SerializableConsumer<UI> {
+    private class ConfigureComponentRendererJob
+            implements SerializableConsumer<UI> {
```
sissbruecker: Tried to make the naming here more meaningful.

vaadin-notification-flow-parent/vaadin-notification-flow/src/main/java/com/vaadin/flow/component/notification/Notification.java
```diff
@@ -130,8 +131,6 @@ static Position fromClientName(String clientName) {
      */
     public Notification() {
         initBaseElementsAndListeners();
-        getElement().getNode().runWhenAttached(ui -> ui
-                .beforeClientResponse(this, context -> deferredJob.accept(ui)));
```
sissbruecker, line 134: This is not needed anymore as the consumer is now always run in the attach handler.

vaadin-notification-flow-parent/vaadin-notification-flow/src/main/java/com/vaadin/flow/component/notification/Notification.java
```diff
@@ -574,16 +573,17 @@ public void removeThemeVariants(NotificationVariant... variants) {
...
     }
 
-    private void attachComponentTemplate() {
-        deferredJob = new AttachComponentTemplate();
-        getElement().getNode().runWhenAttached(ui -> ui
-                .beforeClientResponse(this, context -> deferredJob.accept(ui)));
+    private void configureComponentRenderer() {
+        configureTemplate = new ConfigureComponentRenderer();
+        getElement().getNode()
+                .runWhenAttached(ui -> configureTemplate.accept(ui));
```
sissbruecker, line 579: This call is still required, the component might already be attached when adding components to it.

Also it seems there is no need to defer to before client response.

tomivirkki (approved):

PR #3914, opened 2022-10-21
feat: introduce custom reference date to date picker

### LINKED ISSUE #1107: Convert 1-2 digit year input to range 1950-2049 for consistency with the web component
Currently the vaadin-date-picker web component converts 1-2 digit year input to the range 1950-2049 (you can try it here: https://cdn.vaadin.com/vaadin-date-picker/4.0.8/demo/#date-picker-basic-demos)

Some examples from the behaviour of the web component:
- When a year is input with 2 digits in range 0-49 is converted to 2000-2049. e.g. 19 -> 2019
- When a year is input with 2 digits in range 50-99 is converted to 1950-1999. e.g. 87 -> 1987
- When a year is input with 4 digits it is used as is. To explicitly select a date in the years 0-99 (before year 100) you need to prefix it with zeros e.g. `0012` for the year 12 C.E.

This can only work when there is a way to explicitly specify a year between 0-99 (to prevent conversion) on input e.g. all years before 1000 are always formatted/displayed with 4 digits prefixing with zeros when necessary. This is also what the web component does. It always automatically formats the displayed date so that the year is always prefixed with zeros up to 4 digits.

These updates were already implemented and tested in vaadin/vaadin-date-picker-flow#230 which was not merged (in favor of a more focused fix) but the code from that PR could be reused to make a new PR just for this issues.

The fix is a bit ugly (for converting the formatted year to be prefixed with zeros) because of how the Locale support is implemented as the connector here uses `Date.toLocaleDateString()` which has no options to force 4 digit year output.

Imo it's not good that the behaviour is so different (when inputting a date) when just using the vaadin-date-picker web component versus using the DatePicker from Flow.

Imo, if this is fixed soon, fixing this can be considered a bug fix (since so far inputting years before 1000 didn't work at all with the Flow component until vaadin/vaadin-date-picker-flow#231 which is not yet released), but if that fix for years before 1000 is released first, then maybe this would need a new minor or even major if the default behaviour is changed.

There's also the possibility to introduce this behaviour as configurable (in both web component and Flow component, or just Flow component) so that the developer could toggle 4 digit year formatting on, and with that, optionally enable the 2 digit year conversion (possibly with configurable target range).

ugur-vaadin:
## Description

The parsing and formatting behaviours for the web component and the Flow components were not aligned. Also, the same was true for locale or custom format based Flow date pickers. This PR ensures that the behaviour for all 3 are aligned and updated.

- Add custom reference date functionality. Instead of a fixed reference date of 1950, the default reference date can now be configured with a default value of the current date.

- Add API to DatePickerI18n for getting and setting the reference date.

- Add leading zero padding. This is for years between 0-999.

- Alter existing tests based on the changes. Add new tests.

Needs the web component [pull request](https://github.com/vaadin/web-components/pull/4792) to be merged first.

Partially fixes #1107 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
```diff
@@ -64,25 +65,71 @@ import dateFnsIsValid from 'date-fns/isValid';
             throw new Error('Array of custom date formats is null or empty');
           }
 
+          function _shortenFourDigitYearFormat(format) {
```
sissbruecker, line 68:
```suggestion
          function getShortYearFormat(format) {
```

Terminology could be improved to always use the same term like short year. No need to prefix any of the newly added functions with an underscore.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
```diff
@@ -64,25 +65,71 @@ import dateFnsIsValid from 'date-fns/isValid';
...
+            if (format.includes('yyyy') && !format.includes('yyyyy')) {
+              return format.replace('yyyy', 'yy');
+            }
+            if (format.includes('YYYY') && !format.includes('YYYYY')) {
+              return format.replace('YYYY', 'YY');
+            }
+            return undefined;
+          }
+
+          function _isShortFormat(format) {
```
sissbruecker, line 78:
```suggestion
          function isShortYearFormat(format) {
```

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
```diff
@@ -64,25 +65,71 @@ import dateFnsIsValid from 'date-fns/isValid';
...
-              const date = dateFnsParse(dateString, format, new Date());
+              // We first try to match the date with the shorter version.
+              const shorterFormat = _shortenFourDigitYearFormat(format);
+              if (shorterFormat) {
+                const shorterFormatDate = dateFnsParse(dateString, shorterFormat, referenceDate);
+                if (dateFnsIsValid(shorterFormatDate)) {
+                  let yearValue = shorterFormatDate.getFullYear();
+                  // The last parsed year check handles the case when a date with an actual year value is provided
+                  // with zero padding, but then got reformatted without the zeroes and parsed again.
+                  if (this._lastParsedYear && yearValue == this._lastParsedYear % 100) {
```
sissbruecker, line 106:
```suggestion
                  if (datePicker.$connector._lastParsedYear && yearValue === datePicker.$connector._lastParsedYear % 100) {
```
Using `this` here is kind of unpredictable, as it is not obvious what it is going to refer to when the function is called.

For comparing numbers prefer to use strict equality.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
```diff
@@ -64,25 +65,71 @@ import dateFnsIsValid from 'date-fns/isValid';
...
 
             return dateFnsFormat(date, format);
           }
 
           function parseDate(dateString) {
+            const referenceDate = _getReferenceDate();
             for (let format of formats) {
-              const date = dateFnsParse(dateString, format, new Date());
+              // We first try to match the date with the shorter version.
+              const shorterFormat = _shortenFourDigitYearFormat(format);
```
sissbruecker, line 99:
```suggestion
              const shortYearFormat = getShortYearFormat(format);
```

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
```diff
@@ -64,25 +65,71 @@ import dateFnsIsValid from 'date-fns/isValid';
...
+              if (shorterFormat) {
+                const shorterFormatDate = dateFnsParse(dateString, shorterFormat, referenceDate);
+                if (dateFnsIsValid(shorterFormatDate)) {
+                  let yearValue = shorterFormatDate.getFullYear();
+                  // The last parsed year check handles the case when a date with an actual year value is provided
+                  // with zero padding, but then got reformatted without the zeroes and parsed again.
```
sissbruecker, line 105: I figured out what this does, but only by removing the code and seeing which tests fail.

Maybe:
> The last parsed year check handles the case where a four-digit year is parsed, then formatted as a two-digit year, and then parsed again. In this case we want to keep the century of the originally parsed year, instead of using the century of the reference date.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/src/main/java/com/vaadin/flow/component/datepicker/DatePickerFormatPage.java
```diff
@@ -57,6 +67,11 @@ public DatePickerFormatPage() {
         setupSetLocaleAfterFormat();
         setupSetFormatAfterLocale();
         setupServerSideValueChange();
+        setupCustomReferenceDateAndLocaleWithTwoDigitsYear();
+        setupCustomReferenceDateAndLocaleWithFourDigitsYear();
```
sissbruecker, line 71: Let's move these two tests to `DatePickerLocalePage` / `DatePickerLocaleIT` as this test suite is for testing custom formats.

Maybe we should rename this suite to `DatePickerCustomFormatPage` / `DatePickerCustomFormatIT`.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/src/main/java/com/vaadin/flow/component/datepicker/DatePickerFormatPage.java
```diff
@@ -165,6 +180,85 @@ private void setupServerSideValueChange() {
         add(datePicker, btn);
     }
 
+    private void setupCustomReferenceDateAndLocaleWithTwoDigitsYear() {
+        DatePicker datePicker = new DatePicker(may13);
+        datePicker.setLocale(Locale.US);
```
sissbruecker, line 185: I'm getting a four-digit year here:

Bildschirmfoto 2022-11-07 um 13 37 22

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
```diff
@@ -64,25 +64,74 @@ import dateFnsIsValid from 'date-fns/isValid';
...
+                    year: yearValue
+                  };
+                }
+              }
+              const date = dateFnsParse(dateString, format, referenceDate);
 
               if (dateFnsIsValid(date)) {
+                let yearValue = date.getFullYear();
+                if (datepicker.$connector._lastParsedYear && yearValue % 100
+                  === datepicker.$connector._lastParsedYear % 100 && isShortYearFormat(format)) {
```
web-padawan, line 122:
```suggestion
                if (
                  datepicker.$connector._lastParsedYear &&
                  yearValue % 100 === datepicker.$connector._lastParsedYear % 100 &&
                  isShortYearFormat(format)
                ) {
```

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
```diff
@@ -64,25 +64,74 @@ import dateFnsIsValid from 'date-fns/isValid';
...
+              const shortYearFormat = getShortYearFormat(format);
+              if (shortYearFormat) {
+                const shortYearFormatDate = dateFnsParse(dateString, shortYearFormat, referenceDate);
+                if (dateFnsIsValid(shortYearFormatDate)) {
+                  let yearValue = shortYearFormatDate.getFullYear();
+                  // The last parsed year check handles the case where a four-digit year is parsed, then formatted
+                  // as a two-digit year, and then parsed again. In this case we want to keep the century of the
+                  // originally parsed year, instead of using the century of the reference date.
+                  if (datepicker.$connector._lastParsedYear && yearValue
+                    === datepicker.$connector._lastParsedYear % 100) {
```
web-padawan, line 107:
```suggestion
                  if (
                    datepicker.$connector._lastParsedYear &&
                    yearValue === datepicker.$connector._lastParsedYear % 100
                  ) {
```

web-padawan (commented):
Added some suggestions to keep Prettier JS formatting in the connector.
Tip: run "Format document" in VS Code to get the file formatted.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
```diff
@@ -92,9 +146,17 @@ import dateFnsIsValid from 'date-fns/isValid';
           };
         });
 
+        function _getReferenceDate() {
+          return datepicker.i18n.referenceDate ? new Date(datepicker.i18n.referenceDate.year,
+              datepicker.i18n.referenceDate.month - 1, datepicker.i18n.referenceDate.day) : new Date();
```
web-padawan, line 151:
```suggestion
          const { referenceDate } = datepicker.i18n;
          return referenceDate ? new Date(referenceDate.year, referenceDate.month - 1, referenceDate.day) : new Date();
```

vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/datepicker/DatePickerLocaleIT.java
```diff
@@ -186,6 +187,32 @@ public void datePickerWithLocale_setInputValue_blur_assertDisplayedValue() {
         assertNoWarnings();
     }
 
+    @Test
+    public void testLocaleBasedParsingShouldUseDefaultReferenceDateIfNotSpecified() {
```
sissbruecker, line 191: Test naming convention is different in this file, would be good to keep new tests consistent with the existing naming scheme.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/datepicker/DatePickerLocaleIT.java
```diff
@@ -212,4 +239,30 @@ private void assertNoWarnings() {
...
+            input.sendKeys(Keys.BACK_SPACE);
+        }
+        input.sendKeys(value);
+        input.sendKeys(Keys.ENTER);
+        getCommandExecutor().waitForVaadin();
+    }
+
+    private void assertInputOutputPair(String input, String expectedOutput) {
+        submitValue(input);
+        Assert.assertEquals(expectedOutput, $("span").id("output").getText());
```
sissbruecker, line 255: Using an output span is not really necessary for these tests, checking the parsed client value would be enough. The value can be accessed with `picker.getDate()`, see the existing tests above.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/datepicker/DatePickerLocaleIT.java
```diff
@@ -212,4 +239,30 @@ private void assertNoWarnings() {
                     logEntry));
         }
     }
+
+    private void submitValue(String value) {
```
sissbruecker, line 243: The other tests in this file just this to set the input value, maybe that's good enough?
https://github.com/vaadin/flow-components/blob/5daac2e5276271bb7776aab52ea5546124f01f13/vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/datepicker/DatePickerLocaleIT.java#L68-L69

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
```diff
@@ -446,6 +440,23 @@ private void executeI18nUpdate() {
...
     }
 
+    private JsonObject getI18nAsJsonObject() {
+        if (i18n == null) {
+            return null;
+        }
+        JsonObject i18nObject = (JsonObject) JsonSerializer.toJson(i18n);
+        // LocalDate objects have to be explicitly added to the serialized i18n
+        // object
+        // in order to be formatted correctly
```
sissbruecker, line 450:
```suggestion
        // object in order to be formatted correctly
```

vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/datepicker/DatePickerLocaleIT.java
ugur-vaadin: Addressed in the latest commits

vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/datepicker/DatePickerLocaleIT.java
ugur-vaadin: Addressed in the latest commits

vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/datepicker/DatePickerLocaleIT.java
```diff
@@ -186,6 +187,54 @@ public void datePickerWithLocale_setInputValue_blur_assertDisplayedValue() {
...
+    public void datePickerWithLocale_setInputValue_blur_defaultReferenceDateIsUsed() {
+        LocalDate now = LocalDate.now();
+        int currentYear = now.getYear();
+
+        int testYear1 = (currentYear + 51) % 100;
+        int adjustedYear1 = getAdjustedYear(currentYear, testYear1);
+
+        picker.setInputValue("02/27/" + testYear1);
+        picker.sendKeys(Keys.TAB);
+        Assert.assertEquals("2/27/" + adjustedYear1, picker.getInputValue());
```
sissbruecker, line 200: For parsing tests, using `getInputValue` in the assertion is not a good option, as in case the parsing fails, the input value will just stay the same, but the internal value will not change. It would be better to use `getDate` [as suggested](https://github.com/vaadin/flow-components/pull/3914#discussion_r1019149852).

If some tests are supposed to test the parse -> reformat cycle, then we can do assertions on both.

ugur-vaadin: In the newest commit, both assertions are used for the tests.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
```diff
@@ -1141,5 +1152,26 @@ public DatePickerI18n setCancel(String cancel) {
...
+            return referenceDate;
+        }
+
+        /**
+         * Sets the {@code referenceDate}.
+         *
+         * @param referenceDate
+         *            the date used to base relative dates on
+         * @return this instance for method chaining
+         */
```
sissbruecker, line 1171: Sorry, I missed this in the previous reviews. The API docs here are lacking information, refering to `{@code referenceDate}` is not helpful, and `the date used to base relative dates on` also leaves a lot of room for interpretation.

We should expand the docs a bit to indicate that the reference date is used to determine the century when parsing two-digit years, maybe provide an example, and information on how you can enter a year that actually only has two digits.

sissbruecker (approved):

web-padawan (approved):

PR #3922, opened 2022-10-24
fix: blur input when stopping spreadsheet cell edit

### LINKED ISSUE #3909: [spreadsheet] Editing a non-frozen cell changes scroll position on Safari
### Description

On Safari, if there's a value in one of the frozen panes, adding any value inside the non-frozen area makes the spreadsheet scroll to the very bottom. Note that this only happens on keyboard navigation.

https://user-images.githubusercontent.com/1222264/196698026-99cdd775-4c59-474e-afe2-cb7d7dc0db5f.mp4

### Expected outcome

The scroll position shouldn't change

Use the IT view for Spreadsheet

### Steps to reproduce

- Click "Freeze Pane" button
- Add a value to a cell on the frozen area
- Add a value to a cell on the non-frozen area

### Environment

Vaadin version(s): 23.3
OS: macOS

### Browsers

Safari

tomivirkki:
Fixes #3909 

The issue occurred when disposing of the `<input>` element that was used for editing the cell. This PR makes the `<input>` lose focus before disposal to avoid the undesired scroll position side effect.


DiegoCardoso (approved):
🎉

PR #3961, opened 2022-10-27
fix: include `CellData.value` on object serialization

### LINKED ISSUE #3911: [Spreadsheet] Frozen cells lose background color on scroll
### Description

If you have a Spreadsheet with frozen panes, the frozen cells lose their background color on scroll.

https://user-images.githubusercontent.com/1222264/196707169-49ae134e-2d07-44d5-a2d0-7d1bf62ba656.mp4

### Expected outcome

The cells' styling should not be affected by scrolling

Use the IT view for Spreadsheet

### Steps to reproduce

- Apply `freezepanels.xlsx` test file
- Open the "Copy" sheet
- Scroll the Spreadsheet

### Environment

Vaadin version(s): 23.3

### Browsers

Issue is not browser related

DiegoCardoso:
## Description

`SheetWidget` keeps a map with cached `CellData`, but a new entry is only added if the `CellData` object has the property `value` different from `null`. The serialization done to send data from the server to the client ignores properties that have null or empty values.

This breaks the following logic because `SheetWidget` won't put the entry to the map due to the missing property:

https://github.com/vaadin/flow-components/blob/5876423ab1e14d4888ea5f8d800c6bfca5e0d6f1/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/SheetWidget.java#L4193-L4197

Fixes #3911

## Type of change

- [X] Bugfix
- [ ] Feature

### DIFF
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/client/CellData.java
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/client/CellData.java
@@ -13,6 +13,8 @@ package com.vaadin.flow.component.spreadsheet.client;
  * #L%
  */
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+
 import java.io.Serializable;
 
 @SuppressWarnings("serial")
@@ -20,6 +22,7 @@ public class CellData implements Serializable {
 
     public int row;
     public int col;
+    @JsonInclude
     public String value;
     public String formulaValue;
     public String originalValue;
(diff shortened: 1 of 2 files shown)


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/client/CellData.java
```diff
@@ -13,13 +13,16 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+
 import java.io.Serializable;
 
 @SuppressWarnings("serial")
 public class CellData implements Serializable {
 
     public int row;
     public int col;
+    @JsonInclude
```
tomivirkki, line 25: Does this mean that `cd.value` property can never be `null`? In that case, the following branch would never get entered.
```java
 if (cd.value == null) { 
     cachedCellData.remove(key); 
 } else { 
```

DiegoCardoso: I don't know exactly on which conditions `CellData.value` can become `null` on the server side, so to test it out, I commented all code writing values to `cd.value` and checked the output of the serialization:

```json
[
	{"row":1,"col":5,"value":null,"cellStyle":"cs7"},
	{"row":1,"col":6,"value":null,"cellStyle":"cs7"},
	{"row":1,"col":7,"value":null,"cellStyle":"cs7"},
]
```

So, it will send the property serialized with the value as `null` in this case and the mentioned branch would be reached.

tomivirkki (approved):

PR #3978, opened 2022-10-28
chore: add module containing all components and their testbench dependencies

ZheSun88:
introduce a module that contains all components and their testbench dependencies. 
internal use only, so the idea is that only enable the `all-components` profile when doing snapshot releases.


vursen:
I checked with @ZheSun88 and it was decided not to proceed with this PR after all, so I'm closing it.

PR #4016, opened 2022-11-03
refactor: use date-fns for locale-based parsing and formatting

sissbruecker:
## Description

Changes the date picker connector to use date-fns for locale-based parsing and formatting instead of implementing custom logic.
- creates a date-dns pattern from the output of `Date.toLocaleDateString(locale)` instead of creating a regex for parsing
- re-uses the existing date-fns parsing and formatting logic using the created pattern, removes custom parsing and formatting logic for locale-based mode
- adds additional check for unsupported locales and falls back to `en-US` if a valid pattern could not be created
- restructure locale-based ITs to use a single date picker that allows changing the locale

## Type of change

- Refactoring

### DIFF
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/datepicker/DatePickerLocaleIT.java
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/datepicker/DatePickerLocaleIT.java
@@ -2,10 +2,13 @@ package com.vaadin.flow.component.datepicker;
 
 import java.time.LocalDate;
 import java.time.Month;
+import java.util.Collections;
 import java.util.List;
+import java.util.Locale;
 import java.util.logging.Level;
 import java.util.stream.Collectors;
 
+import com.vaadin.testbench.TestBenchElement;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
```diff
@@ -55,107 +26,40 @@ import dateFnsIsValid from 'date-fns/isValid';
           })
         );
 
-        const cleanString = tryCatchWrapper(function (string) {
-          // Clear any non ascii characters from the date string,
-          // mainly the LEFT-TO-RIGHT MARK.
-          // This is a problem for many Microsoft browsers where `toLocaleDateString`
-          // adds the LEFT-TO-RIGHT MARK see https://en.wikipedia.org/wiki/Left-to-right_mark
-          return string.replace(/[^\x00-\x7F]/g, '');
-        });
```
sissbruecker, line 64: This doesn't seem to be required anymore. It was added as part of https://github.com/vaadin/vaadin-date-picker-flow/pull/81 to fix an issue with Edge. Since then Edge has changed significantly to be based on Chrome, and the issue (using french locale) is not reproducable anymore in latest Edge - formatting and parsing return the same results as in Chrome.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/datepicker/DatePickerLocaleIT.java
```diff
@@ -20,40 +23,89 @@
...
-                .id("picker-with-value-and-polish-locale");
+        applyLocale(new Locale("pl", "PL"));
         Assert.assertEquals("Should display the value using Polish date format",
-                "3.05.2018", polishLocalePicker.getInputValue());
+                "3.05.2018", picker.getInputValue());
+
+        applyLocale(new Locale("bg"));
+        Assert.assertEquals(
+                "Should display the value using Bulgarian date format",
+                "3.05.2018 г.", picker.getInputValue());
```
sissbruecker, line 57: Added this to verify that the `г` character is not removed, which was the case previously when using `cleanString`

ugur-vaadin (approved):
Looks very clean and intuitive.

web-padawan (approved):

PR #4027, opened 2022-11-07
refactor: remove constructor with a single string parameter

### LINKED ISSUE #1276: String constructor should be used for label
`new TextField("Name");`  creates a `TextField` with `name` as the label
`new TextArea("Name");`  creates a `TextArea` with `name` as the label
`new PasswordField("Name");`  creates a `PasswordField` with `name` as the label
`new Checkbox("Name");` creates a `CheckBox` with `name` as the label
`new ComboBox("Name");` creates a `ComboBox` with `name` as the label

`new RichTextEditor("Name");`creates a `RichTextEditor` with `name` as the value

ugur-vaadin:
## Description

Constructors with a single string parameter should set the label for the component. Currently, RichTextEditor does not support labels but it has a single string parameter constructor which sets the value. For the behaviour to be in line with the other components, this PR removes the constructor. 

This is a breaking change and should not be backported.

Fixes #1276 

## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] Refactor

## Checklist

### DIFF
--- a/vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java
+++ b/vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java
@@ -116,19 +116,6 @@ public class RichTextEditor
     }
 
     /**
-     * Constructs a {@code RichTextEditor} with the initial value
-     *
-     * @param initialValue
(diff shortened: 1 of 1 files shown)


web-padawan (approved):

web-padawan (changes requested):
Should we also deprecate and remove constructor with a value + value change listener?

ugur-vaadin:
> Should we also deprecate and remove constructor with a value + value change listener?

That was a discussion @yuriy-fix and I had during the day. If that is something that we all agree on, I can proceed with it.

sissbruecker (approved):

PR #4028, opened 2022-11-07
refactor: deprecate constructor with a single string parameter

### LINKED ISSUE #1276: String constructor should be used for label

ugur-vaadin:
## Description

Constructors with a single string parameter should set the label for the component. Currently, RichTextEditor does not support labels but it has a single string parameter constructor which sets the value. For the behaviour to be in line with the other components, this PR deprecates the constructor for v23.3.

This is a temporary change which will be overridden with the [removal](https://github.com/vaadin/flow-components/pull/4027) of the constructor in v24.

Fixes #1276 

## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] Refactor

## Checklist

### DIFF
--- a/vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java
+++ b/vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java
@@ -123,7 +123,8 @@ public class RichTextEditor
      *
      * @see #setValue(Object)
      *
-     * @deprecated Since 23.3, this API is deprecated. Use {@link #setValue(String)} instead.
+     * @deprecated Since 23.3, this API is deprecated in order to maintain API consistency.
+     * Use {@link #setValue(String)} instead.
      */
     @Deprecated
     public RichTextEditor(String initialValue) {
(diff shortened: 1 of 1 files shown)


web-padawan (approved):

vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java
```diff
@@ -122,7 +122,10 @@ public RichTextEditor() {
      *            the initial value in Delta format, not {@code null}
      *
      * @see #setValue(Object)
+     *
+     * @deprecated Since 23.3, this API is deprecated.
```
sissbruecker, line 126:
```suggestion
     * @deprecated Since 23.3, this API is deprecated. Use {@link #setValue(String)} instead.
```
Could add a note what should be used instead. Obvious in this case, but a good practice in general if there is an alternative to what is being deprecated.

sissbruecker (approved):

PR #4034, opened 2022-11-08
fix: properly update selected item key (#4033) (CP: 14.8)

sissbruecker:
Cherry-pick of #4033


web-padawan:
After 14.9.0 stable is out, it became an LTS and there should be no further releases for 14.8 at this point. So, unlike with 23.x we offer no overlap period between minors for Vaadin 14. Therefore I'm closing this PR.

PR #4042, opened 2022-11-08
fix: fix pane freezing in large spreadsheets

tomivirkki:
Addresses the issue described [here](https://docs.google.com/spreadsheets/d/1dpRq-pHZ8DwtEEmmSTVLraix-7AWTzzPp1YutVU0qRg/edit#gid=0&range=33:33): "Freezing panes on the "100_000_rows.xlsx" file leads to browser console error and the sheet breaks and can't be used after that."

The issue was caused by a regression from https://github.com/vaadin/spreadsheet/pull/532


DiegoCardoso (commented):
LGTM

Issue #3258, opened 2022-06-01
[spreadsheet] Locale support does not work as expected


tomivirkki:
The `Spreadsheet.setLocale()` API [exists](https://github.com/vaadin/flow-components/blob/956d4edf8538d93668476b40b10f18a90ae4c434/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/Spreadsheet.java#L1817) but using it doesn't seem to have the desired effect.

In the following example, "Italian" is selected as the active Locale. The formula "=1,1" (comma as a decimal separator) is used but the format does not get recognized.

https://user-images.githubusercontent.com/1222264/171386284-33a5a513-bcc0-49ec-bc7d-3e1e768e335b.mp4

Original test case [here](https://github.com/vaadin/spreadsheet/blob/e37cd0814638298a26851885b5c964beb50ec983/legacy/vaadin-spreadsheet/src/test/java/com/vaadin/addon/spreadsheet/test/FormulaFormatTest.java#L27-L33)

PR #4044, opened 2022-11-09
fix: fix spreadsheet Locale API implementation

tomivirkki:
Fixes https://github.com/vaadin/flow-components/issues/3258

Also, migrate [V8 Spreadsheet ITs](https://github.com/vaadin/spreadsheet/blob/flow-port/legacy/vaadin-spreadsheet/src/test/java/com/vaadin/addon/spreadsheet/test/FormulaFormatTest.java) that depended on the fix.


DiegoCardoso (approved):
LGTM

I wonder if when the location is changed, the formulas/formatting should be reevaluated. On the IT, the value needs to be set again so the formula works. I can see that this was taken as-is from V8, so we can evaluate if we want to make this change at a later moment.

tomivirkki:
> I wonder if when the location is changed, the formulas/formatting should be reevaluated.

This seems to be working already:

https://user-images.githubusercontent.com/1222264/201026591-d07e8bc8-0b66-4ff3-b562-892336b6d2f2.mp4

PR #4025, opened 2022-11-07
refactor!: remove template usage from grid header and footer

sissbruecker:
## Description

Removes the usage of `TemplateRenderer` and `ComponentRenderer` for rendering header and footer content in grid columns, and instead adds connector methods to define header/footer renderer functions on the client grid column that render the respective text or component content. This effectively removes any usage of the `<template>` element for grid headers and footers.

Breaking changes:
- Header and footer do not contain a `<flow-component-renderer>` tag anymore when using component header or footer
- removes protected `AbstractColumn.setHeaderRenderer`
- removes protected `AbstractColumn.getHeaderRenderer`
- removes protected `AbstractColumn.setFooterRenderer`
- removes protected `AbstractColumn.getFooterRenderer`

Closes https://github.com/vaadin/flow-components/issues/3574
Fixes https://github.com/vaadin/web-components/issues/3470

## Type of change

- Refactoring

PR #4055, opened 2022-11-10
fix: create aria label from header component

sissbruecker:
## Description

Fixes the grid column to also create a sorter aria label when using a header component.

Targetting 23.3 as this has already been fixed in 24 as part of this refactoring: https://github.com/vaadin/flow-components/pull/4025

Fixes https://github.com/vaadin/web-components/issues/3470

## Type of change

- Bugfix

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractColumn.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractColumn.java
@@ -47,7 +47,6 @@ abstract class AbstractColumn<T extends AbstractColumn<T>> extends Component
     private boolean headerRenderingScheduled;
     private boolean footerRenderingScheduled;
 
-    private String rawHeaderTemplate;
     private boolean sortingIndicators;
 
     private String headerText;
@@ -137,10 +136,9 @@ abstract class AbstractColumn<T extends AbstractColumn<T>> extends Component
         headerTemplate = rendering.getTemplateElement();
         headerTemplate.setAttribute("class", "header");
 
-        setBaseHeaderTemplate(headerTemplate.getProperty("innerHTML"));
         if (hasSortingIndicators()) {
             headerTemplate.setProperty("innerHTML",
-                    addGridSorter(rawHeaderTemplate));
+                    addGridSorter(headerTemplate.getProperty("innerHTML")));
         }
 
         return rendering;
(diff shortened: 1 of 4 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractColumn.java
```diff
@@ -47,7 +47,6 @@
     private boolean headerRenderingScheduled;
     private boolean footerRenderingScheduled;
 
-    private String rawHeaderTemplate;
```
sissbruecker, line 50: This was assigned, but never used outside of the method where is was assigned.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractRow.java
```diff
@@ -413,8 +413,8 @@ private CELL moveColumnLayerAndJoinCells(Collection<CELL> cellsToJoin,
                 ColumnGroup group = ColumnGroupHelpers.wrapInColumnGroup(grid,
                         col);
                 AbstractColumn<?> oldGroup = leftColumns.next();
-                group.setHeaderRenderer(oldGroup.getHeaderRenderer());
-                group.setFooterRenderer(oldGroup.getFooterRenderer());
+                oldGroup.moveHeaderContent(group);
+                oldGroup.moveFooterContent(group);
```
sissbruecker, line 417: As the new logic for generating the aria-label depends on the header text / component being present, just moving the renderer is not enough. The new methods will assign either the text or the component to the other column / group, which creates a new renderer instead.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractColumn.java
```diff
@@ -311,23 +325,20 @@ protected boolean hasSortingIndicators() {
...
                 .escape(getBottomLevelColumn().getInternalId());
 
-        String textContent = org.jsoup.Jsoup.parse(templateInnerHtml).text();
-        String sortBy = textContent.isBlank() ? ""
+        String textContent = headerComponent != null
+                ? headerComponent.getElement().getTextRecursively()
+                : headerText;
+        if (textContent != null) {
+            textContent = HtmlUtils.escape(textContent);
+        }
```
sissbruecker, line 340: For header components use `getTextRecursively` to get a text for the aria-label.

Also escape the text, which might be something that was done before by parsing and extracting the text through Jsoup.

web-padawan (approved):

PR #4062, opened 2022-11-11
chore: extract building SVG generator into an individual npm script

vursen:
## Description

Similar to https://github.com/vaadin/flow/pull/15099.

The PR extracts building the SVG generator into an individual npm script instead of building it on `postinstall` which may happen to be disabled by `ignore-scripts=true` in `.npmrc` of the developer.

## Type of change

- [x] Bugfix


vursen:
I'm not sure how much value this PR brings. A similar PR was created in Flow because @Artur- happened to have npm scripts disabled in his `.npmrc`. I think it would be reasonable to have that consistent across our Java repos.

sissbruecker (approved):

PR #4074, opened 2022-11-14
fix: clear formula from a cell before updating value

tomivirkki:
## Description

Addresses the DX test finding: "[Invalid value ### can’t be overridden without deleting the cell](https://docs.google.com/document/d/1U6uRAyUsSML_q7U2J6hmAX6q3EmLhmOJJaiJn4LHCrQ/edit#heading=h.g4mjwtu27lqu)"

## Type of change

- [x] Bugfix
- [ ] Feature


DiegoCardoso (approved):
From what I could see, it didn't matter if the formula was invalid or valid for the issue to raise. 

LGTM 🎉

tomivirkki:
> From what I could see, it didn't matter if the formula was invalid or valid for the issue to raise.

The PR includes tests for both cases

Issue #1063, opened 2020-12-07
Switch htmlValue and value


yuriy-fix:
Switch [`setValue`](https://github.com/vaadin/vaadin-flow-components/blob/b34675add5b11b3738aa109bce561360134106a7/vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java#L192), [`getValue`](https://github.com/vaadin/vaadin-flow-components/blob/b34675add5b11b3738aa109bce561360134106a7/vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java#L233) with [`AsHtml#getValue()`](https://github.com/vaadin/vaadin-flow-components/blob/b34675add5b11b3738aa109bce561360134106a7/vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java#L847), [`AsHtml#setValue(String)`](https://github.com/vaadin/vaadin-flow-components/blob/b34675add5b11b3738aa109bce561360134106a7/vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java#L804) logic in order to make it possible to set `htmlValue` by default on flow counter-part instead of operating with Delta.

Desired outcome: Possibility to set html value string with `setValue` and get it with `getValue`.

rolfsmeds:
Should the WC value API be changes correspondingly?

PR #4087, opened 2022-11-15
refactor: deprecate RTE value property in favor of asDelta

sissbruecker:
## Description

Deprecates the `RichTextEditor` value getter and setter in favor of a new `asDelta` wrapper. In 24 we want to make the HTML format the default, so the intention is to temporarily deprecate the value property in order to move existing users of the delta format over to the new wrapper. In 24 we switch the implementation of the value property to use the HTML format, and remove the deprecation.

Part of https://github.com/vaadin/flow-components/issues/1063

## Type of change

- Refactoring

### DIFF
--- a/vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow-integration-tests/src/main/java/com/vaadin/flow/component/richtexteditor/tests/RichTextEditorAsDeltaPage.java
+++ b/vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow-integration-tests/src/main/java/com/vaadin/flow/component/richtexteditor/tests/RichTextEditorAsDeltaPage.java
@@ -1,0 +1,105 @@
+package com.vaadin.flow.component.richtexteditor.tests;
+
+import com.vaadin.flow.component.button.Button;
+import com.vaadin.flow.component.html.Div;
+import com.vaadin.flow.component.orderedlayout.VerticalLayout;
+import com.vaadin.flow.component.richtexteditor.RichTextEditor;
+import com.vaadin.flow.data.binder.Binder;
+import com.vaadin.flow.data.binder.BinderValidationStatus;
+import com.vaadin.flow.data.binder.BindingValidationStatus;
+import com.vaadin.flow.data.value.ValueChangeMode;
+import com.vaadin.flow.function.SerializablePredicate;
+import com.vaadin.flow.router.Route;
+
+import java.io.Serializable;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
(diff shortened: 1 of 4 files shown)


vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow-integration-tests/src/main/java/com/vaadin/flow/component/richtexteditor/tests/RichTextEditorAsDeltaPage.java
```diff
@@ -0,0 +1,105 @@
...
+import com.vaadin.flow.data.value.ValueChangeMode;
+import com.vaadin.flow.function.SerializablePredicate;
+import com.vaadin.flow.router.Route;
+
+import java.io.Serializable;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+@Route(value = "vaadin-rich-text-editor/as-delta")
+public class RichTextEditorAsDeltaPage extends VerticalLayout {
```
sissbruecker, line 19: The IT setup is adapted from the existing ITs testing the value property (`MainView` / `BasicUseIT`) with some cleanup to make them more readable.

vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java
```diff
@@ -987,4 +1013,74 @@ private String generateHtmlValue() {
...
+
+        @Override
+        public Registration addValueChangeListener(
+                ValueChangeListener<? super ValueChangeEvent<String>> valueChangeListener) {
+            return RichTextEditor.this
+                    .addValueChangeListener(valueChangeListener);
+        }
+
+        @Override
+        public void setReadOnly(boolean b) {
```
ugur-vaadin, line 1057: Similar methods in `AsHtml` have JavaDocs. Should we also add them to the methods in `AsDelta` or remove them from the `AsHtml` counterparts?

sissbruecker: I don't think it's necessary, as the methods inherit docs from `HasValue`, and the most interesting methods in this class from a developer perspective are the getter and setter for the value. Most of the other methods are only necessary to get Binder working correctly.

ugur-vaadin (approved):

PR #4095, opened 2022-11-15
fix: update invalid mark on cell refresh

tomivirkki:
## Description

Addresses the issue: "[Updating a cell marked invalid (formula) programmatically does not remove the invalid indicator.](https://docs.google.com/spreadsheets/d/1dpRq-pHZ8DwtEEmmSTVLraix-7AWTzzPp1YutVU0qRg/edit#gid=0)"

Depends on https://github.com/vaadin/flow-components/pull/4089

## Type of change

- [x] Bugfix
- [ ] Feature


DiegoCardoso (approved):

tomivirkki:
Turns out there are multiple places in `CellValueManager` where a formula cell may be marked as invalid. One is `onCellValueChange` which is an RPC intended to be called only from the client. Since fixing the bug could require an excessive amount of refactoring in `CellValueManager` and the reported issue is quite edge-casey, closing this as wontfix.

PR #4115, opened 2022-11-16
refactor: deprecate generated prefix component API in combo box

sissbruecker:
## Description

Deprecates the generated protected methods for adding and removing a prefix component. The API should be replaced with a dedicated prefix component API in the future, similar to the one in `TextField`.

## Type of change

- Refactor


web-padawan:
The label was there by mistake as the PR itself was targeting 23.3 branch, so I removed the label.

PR #4121, opened 2022-11-16
fix: filter all items on select all uncheck

tomivirkki:
## Description

Addresses the issue: "[Clearing the "Select all" checkbox on a filter table does nothing](https://docs.google.com/spreadsheets/d/1dpRq-pHZ8DwtEEmmSTVLraix-7AWTzzPp1YutVU0qRg/edit#gid=0)"

## Type of change

- [x] Bugfix
- [ ] Feature

### DIFF
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/ItemFilter.java
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/ItemFilter.java
@@ -163,10 +163,10 @@ public class ItemFilter extends Div implements SpreadsheetFilter {
                 cancelValueChangeUpdate = true;
                 if (value) {
                     filterCheckbox.setValue(new HashSet<>(allCellValues));
-                    updateFilteredItems(allCellValues);
                 } else {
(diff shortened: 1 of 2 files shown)


DiegoCardoso (commented):
The change looks good, however, it brings a slight difference in behavior depending on whether you click on the `(Select All)` or each checkbox individually.

As seen in the recording below, when there's only one checkbox checked, toggling it on and off, won't be applied to the table. That same behavior is present in the version for FW8. 

https://user-images.githubusercontent.com/262432/202200550-4679d1e2-46ed-476a-afd5-a02c6a02dcf2.mp4

There's a comment here, which makes me believe that this is the intended behavior: 
https://github.com/vaadin/flow-components/blob/e41f55f1e89ff2a15a027f16a33757789e035aaf/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/ItemFilter.java#L189-L191

Checking how Excel works, it looks like the current behavior in Spreadsheet is reproducing how the `(Select All)` works.

https://user-images.githubusercontent.com/262432/202203972-294753b0-10da-47cb-8eef-d2f74dcec1c6.mp4

tomivirkki:
Thanks, the current implementation does look very intentional, and apparently, it's in order to align with how Excel works. Closing.

PR #4122, opened 2022-11-16
fix: prevent filter on popup open

DiegoCardoso:
## Description

This intends to fix ["Table filter hides row when opening"](https://docs.google.com/document/d/1U6uRAyUsSML_q7U2J6hmAX6q3EmLhmOJJaiJn4LHCrQ/edit#heading=h.j44p8x2gb4ch) issue discovered on the DX tests.

This change brings the same behavior from Spreadsheet for V8, where if a filtered row is unhidden from the context menu keeps its state when the popup button is opened. 

## Type of change

- [X] Bugfix
- [ ] Feature

### DIFF
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-integration-tests/src/test/java/com/vaadin/flow/component/spreadsheet/test/SheetFilterTableIT.java
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-integration-tests/src/test/java/com/vaadin/flow/component/spreadsheet/test/SheetFilterTableIT.java
@@ -50,6 +50,22 @@ public class SheetFilterTableIT extends AbstractSpreadsheetIT {
         assertSelectAll(cell);
     }
 
+    @Test
+    public void filteredTable_unhideRowAndOpenPopup_rowIsUnhidden() {
+        loadTestFixture(TestFixtures.SpreadsheetTable);
+        final var cell = getSpreadsheet().getCellAt("B2");
+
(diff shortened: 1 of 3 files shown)


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-integration-tests/src/test/java/com/vaadin/flow/component/spreadsheet/test/SheetFilterTableIT.java
```diff
@@ -50,6 +50,22 @@ public void sheetWithFilterTable_rowIsRemoved_filterOptionsAvailable() {
         assertSelectAll(cell);
     }
 
+    @Test
```
tomivirkki, line 53: Since there's only server-side logic involved, an IT feels like an overkill for the purpose.
The following unit test does the job and takes a fraction of the time to run:

in `FilterTableTest.java`
```java
    @Test
    public void filteredTable_unhideRowAndOpenPopup_rowIsUnhidden() {
        getFilterCheckboxGroup().deselect("4");

        spreadsheet.setRowHidden(3, false);
        getPopupButton().openPopup();

        Assert.assertFalse(spreadsheet.isRowHidden(3));
    }
```

DiegoCardoso: Good catch! Refactored.

PR #4127, opened 2022-11-17
test: use js for lit-renderer-test-component

tomivirkki:


web-padawan (approved):
LGTM. We should have a test in Flow that would check that `@property()` works instead.

tomivirkki:
https://github.com/vaadin/flow/pull/15165 most probably fixes the root issue. If the snapshots builds pass fine, I'll repurpose this PR only to include the log filter cleanup.

web-padawan (approved):

Issue #3861, opened 2022-10-13
GridPro custom edit column value provider


filippo-azserve:
### Describe your motivation

Actually for custom editor columns the editor field value is computed using the `ValueProvider` specified through the GridPro::addEditColumn method:
```
TextField editorField = new TextField();
grid.addEditColumn(Person::getName).custom(
    editorField,
    (item, newValue) -> item.setName(newValue));
```
And there is no way to convert the model value to the presentation one (the editor field value):

```
TextField editorField = new TextField();
// age is an int and should be converted to a String!
grid.addEditColumn(Person::getAge).custom(
    editorField,
    (item, newValue) -> item.setAge(Integer.parseInt(newValue)));
```

### Describe the solution you'd like

Add a new `addEditColumn` overload method for custom editor columns with an additional `valueProvider` parameter:
```
TextField editorField = new TextField();
grid.addEditColumn(Person::getAge).custom(
    editorField,
    item -> "" + item.getAge(), // int to String (valueProvider param)
    (item, newValue) -> item.setAge(Integer.parseInt(newValue)));
```

PR #4208, opened 2022-11-25
feat: GridPro custom editor value provider

sissbruecker:
## Description

Allows setting an editor value provider for columns with custom editors. By default, the editor component would be initialized with the same value provider that is used for rendering the column cells.
Using an editor value provider allows to initialize the editor value with a different value provider, which allows converting the model value into the appropriate type or format used by the editor.

Closes https://github.com/vaadin/flow-components/issues/3861
AC: https://github.com/vaadin/platform/issues/3388

Cherry-picked from https://github.com/vaadin/flow-components/pull/3752 by @filippo-azserve ,  added some JavaDoc and did some cleanup.

## Type of change

- Feature


sissbruecker:
@DiegoCardoso Can you please do a review, as you did already take a look at the previous contribution PR.

DiegoCardoso (approved):

Issue #4059, opened 2022-11-10
Remove internal usages and public API of TemplateRenderer


sissbruecker:
`TemplateRenderer` is scheduled to be removed in 24. Context:
- https://github.com/vaadin/flow-components/issues/857
- https://github.com/vaadin/flow-components/issues/916

First remove all internal usages:

- [x] `Crud` - Used for edit button column
- [x] `TreeGrid` - Used for hierarchy column
- [x] `VirtualList` - Used when setting a value provider as renderer
- [x] Unit tests and integration tests
	- [x] ComboBox
	- [x] Grid
	- [x] TreeGrid
	- [x] Virtual List

Then remove the public API from the `vaadin-renderer-flow` module.

Also consider closing the following issues as obsolete:
- https://github.com/vaadin/flow-components/issues/1303
- https://github.com/vaadin/flow-components/issues/1274

tomivirkki:
Closing as complete

PR #4210, opened 2022-11-25
refactor: remove internal TemplateRenderer use from TreeGrid

tomivirkki:
## Description

Replace `TemplateRenderer` with `LitRenderer` in `TreeGrid` 

Part of https://github.com/vaadin/flow-components/issues/4059

## Type of change

- Refactoring

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/treegrid/it/TreeGridPreloadIT.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/treegrid/it/TreeGridPreloadIT.java
@@ -128,15 +128,13 @@ public class TreeGridPreloadIT extends AbstractTreeGridIT {
     }
 
     @Test
-    public void firstExpanded_reExpand_shouldPreLoadDataForExpandedChildren() {
+    public void firstExpanded_reExpand_shouldUseCachedDataForExpandedChildren() {
         open(Arrays.asList(0), null, null, null, null);
         requestCountReset.click();
 
         getTreeGrid().collapseWithClick(0);
         getTreeGrid().expandWithClick(0);
-        // Expanding a recursively expanded parent doesn't yet trigger a preload
-        // so a second request is made from client.
-        Assert.assertEquals("2", requestCount.getValue());
+        Assert.assertEquals("0", requestCount.getValue());
     }
 
     @Test
@@ -149,13 +147,13 @@ public class TreeGridPreloadIT extends AbstractTreeGridIT {
     }
 
     @Test
-    public void firstExpanded_reExpandChild_shouldPreLoadDataForExpandedChildren() {
+    public void firstExpanded_reExpandChild_shouldUseCachedDataForExpandedChildren() {
         open(Arrays.asList(0), null, null, null, null);
         requestCountReset.click();
 
         getTreeGrid().collapseWithClick(2);
         getTreeGrid().expandWithClick(2);
-        Assert.assertEquals("1", requestCount.getValue());
+        Assert.assertEquals("0", requestCount.getValue());
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/treegrid/it/TreeGridPreloadIT.java
```diff
@@ -128,15 +128,13 @@ public void firstExpanded_scrollByViewportEstimate_shouldPreLoadDataForExpandedC
     }
 
     @Test
-    public void firstExpanded_reExpand_shouldPreLoadDataForExpandedChildren() {
```
tomivirkki, line 131: Seems that there's a difference if the expand item command originates from the server vs the client. 0 additional requests is better than 2 so repurposed these re-expand tests. There are other tests that validate that the content is correct after re-expanding.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -555,14 +554,28 @@ public GridDataView<T> getGenericDataView() {
...
+        Column<T> column = addColumn(LitRenderer.<T> of(
+                "<vaadin-grid-tree-toggle @click=${onClick} .leaf=${!item.children} .expanded=${model.expanded} .level=${model.level}>"
+                        + "${item.name}</vaadin-grid-tree-toggle>")
                 .withProperty("children",
                         item -> getDataCommunicator().hasChildren(item))
                 .withProperty("name",
-                        value -> String.valueOf(valueProvider.apply(value))));
+                        value -> String.valueOf(valueProvider.apply(value)))
+                .withFunction("onClick", item -> {
+                    if (getDataCommunicator().hasChildren(item)) {
```
sissbruecker, line 570: Should we add this condition to `HierarchyComponentRenderer` as well?
https://github.com/vaadin/flow-components/blob/158b50b10a3acd9a5b104d738c4015d284e0be39/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/HierarchyColumnComponentRenderer.java#L27-L33

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -555,14 +554,28 @@ public GridDataView<T> getGenericDataView() {
...
     public Column<T> addHierarchyColumn(ValueProvider<T, ?> valueProvider) {
-        Column<T> column = addColumn(TemplateRenderer
-                .<T> of("<vaadin-grid-tree-toggle "
-                        + "leaf='[[!item.children]]' expanded='{{expanded}}' level='[[level]]'>[[item.name]]"
-                        + "</vaadin-grid-tree-toggle>")
+
+        // The click listener needs to check if the event gets canceled (by
+        // vaadin-grid-tree-toggle) and only invoke the callback if it does.
+        // vaadin-grid-tree-toggle will cancel the event if the user clicks on
+        // a non-focusable element inside the toggle.
```
sissbruecker, line 561: Looks like we are not doing that here though. Compared to `HierarchyColumnRenderer`:
https://github.com/vaadin/flow-components/blob/158b50b10a3acd9a5b104d738c4015d284e0be39/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/HierarchyColumnComponentRenderer.java#L41-L45

From that description it looks like it's not needed here, as we will always only have text node as content. Should we remove the comment then?

tomivirkki: Oops. I first had the same check here but also noticed it's not needed so removed the logic. Apparently forgot to remove the comment. Done.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
tomivirkki: In `HierarchyComponentRenderer`s case, with the conditional click listener, the [check is done client-side](https://github.com/vaadin/web-components/blob/d6f871f2e694aa08748481d32ae9c7f6ba2e6b20/packages/grid/src/vaadin-grid-tree-toggle.js#L192)

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
tomivirkki: That's odd. I pushed the commit but it doesn't show in GH yet

PR #4218, opened 2022-11-25
test: reset UI in teardown for Upload slot tests

web-padawan:
## Description

There is a problem with unit tests introduced in #4207 and now it causes snapshot and other PRs to fail.
Apparently, unit tests for Upload module are not configured correctly adn don't run with ITs. 

Here is the error, it's unclear to me why it happens but at least `@After` block is missing.

```
java.lang.NullPointerException: Cannot invoke "com.vaadin.flow.component.UI.getUIId()" because the return value of "com.vaadin.flow.component.UI.getCurrent()" is null
  at com.vaadin.flow.component.upload.tests.UploadSlotsTest.setDropLabelNull_defaultLabelIsRestored(UploadSlotsTest.java:76)
  ```

## Type of change

- Tests


tomivirkki (approved):
Seems to fix the issue

tomivirkki:
This change fixed the issue when I ran the tests locally (in a dev container). But seems that for some reason the problem still remains on CI https://github.com/vaadin/flow-components/pull/4217

web-padawan:
Yes, the build still fails 😕  Any ideas what could be wrong?

PR #4236, opened 2022-11-28
fix: use correct null check in setTooltipGenerator

web-padawan:
## Description

The null check in `setTooltipGenerator` is incorrectly using `classNameGenerator` 🙈 

## Type of change

- Bugfix


tomivirkki (approved):

tomivirkki:
Could add the following missing test case to `GridTooltipTest`:

```java
    @Test(expected = NullPointerException.class)
    public void setNullTooltipGenerator_throws() {
        grid.addColumn(item -> item).setTooltipGenerator(null);
    }
```

web-padawan:
@tomivirkki Thanks, added.

PR #4245, opened 2022-11-29
refactor: fix detach field test logic

sissbruecker:
Fixes detach buttons in field validation tests to actually remove the component from its container.


vursen (approved):
Haha, good catch, thanks 👍

PR #4250, opened 2022-11-29
fix: remove old license-checker from spreadsheet

tomivirkki:


web-padawan (approved):

Artur-:
How is this "old"?

tomivirkki:
> How is this "old"?

None of the other commercial components have it even though the client-side license-checking was removed from them. But the real problem here seems to be that the check is `if (service.getDeploymentConfiguration().isProductionMode()) {` while it should be `if (!service.getDeploymentConfiguration().isProductionMode()) {`

Artur-:
The web component license checking that has explicit calls is old. This is not in any way old

tomivirkki:
> This is not in any way old

Should the same check then be added to the other Commerical Flow components?

Artur-:
Yes, to all those which do not have a web component on the client side

web-padawan:
> Yes, to all those which do not have a web component on the client side

Spreadsheet is the only one. Other Pro components are covered by `static get cvdlName()` in JS.

tomivirkki:
Closing in favour of https://github.com/vaadin/flow-components/pull/4255

tomivirkki:
Techincally Spreadsheet also has a Web Component. Could we drop this check if we added the following to [`vaadin-spreadsheet.js`](https://github.com/vaadin/flow-components/blob/master/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/resources/META-INF/resources/frontend/vaadin-spreadsheet/vaadin-spreadsheet.js)?

```js
static get cvdlName() {
  return 'vaadin-spreadsheet';
}
```

Artur-:
It needs to be installed as an npm package also

PR #4259, opened 2022-11-29
chore: update license files and headers

tomivirkki:


tomivirkki:
> there's some inconsistency on the license header placement. Some files have it before the package declaration, some after and some even after the imports (like in the AbstractDataLabels.java file)

True, but this was the case also before. 

> there's a mention to CVDL here

Removed the mention

> the URL to the CVDL is used as value to the osgi.bundle.license property present in the POM files. I don't know whether they should be kept as it or updated to the new URL.

Updated the URL

alvarezguille (approved):

PR #4263, opened 2022-11-29
fix!: do not check minlength and pattern when value is an empty string

vursen:
## Description

As I discovered in https://github.com/vaadin/flow-components/pull/4264, the Flow text field components check such constraints as `minlength` and `pattern` even when `value` is an empty string which is different from how their web components handle this case. 

This PR aims to align that behavior to the web components' one, that is, don't check those constraints when `value` is an empty string.

> **Note**
> This PR intentionally doesn't deal with `maxlength` as it already accepts empty strings by its nature.

> **Warning**
> This is an altering-behavior change since someone could be using those constraints as a substitute for `required`.

Part of https://github.com/vaadin/platform/issues/3066.

## Type of change

- [x] Bugfix

## Checklist


vursen:
Let's discuss how we would like this change to go: 
- 24.0 with a feature flag
- 24.0 without a feature flag
- 24.0 and 23.3

vursen:
We agreed internally that this fix will only go to 24.0 for now.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/TextFieldValidationSupport.java
```diff
@@ -77,14 +77,14 @@ ValidationResult checkValidity(String value) {
             return ValidationResult.error("");
         }
 
-        final boolean isMinLengthNotReached = value != null && minLength != null
-                && value.length() < minLength;
+        final boolean isMinLengthNotReached = value != null && !value.isEmpty()
```
sissbruecker, line 80: Nit: Could extract `value != null && !value.isEmpty()` into a variable with a readable name like `hasValue`, and then reuse below.

sissbruecker (approved):

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/TextFieldValidationSupport.java
vursen: If I do so, it starts complaining about potential null pointer access:

image

Do you know how it could be worked around?

vursen: Perhaps using Optional could help.

vursen: Let's probably keep it the way it is

PR #4311, opened 2022-12-07
refactor: extract attach and detach buttons into AbstractValidationPage

sissbruecker:
## Description

Extracts the commonly used attach and detach buttons from validation ITs into `AbstractValidationPage` and adds a convenience method for clicking them to `AbstractValidationIT`.


vaadin-flow-components-shared-parent/vaadin-flow-components-test-util/src/main/java/com/vaadin/tests/validation/AbstractValidationIT.java
```diff
@@ -88,4 +90,11 @@ private boolean shouldSetInvalid(boolean invalid) {
                 "const field = arguments[0]; const invalid = arguments[1]; return field._shouldSetInvalid(invalid)",
                 testField, invalid);
     }
+
+    protected void detachAndReattachField() {
+        $("button").id(DETACH_FIELD_BUTTON).click();
+        $("button").id(ATTACH_FIELD_BUTTON).click();
```
knoobie, line 96: Should there be a check in between both clicks that the component was really removed from the DOM and later added again?

vursen (approved):

vaadin-flow-components-shared-parent/vaadin-flow-components-test-util/src/main/java/com/vaadin/tests/validation/AbstractValidationIT.java
vursen: Good point, we could have that, I think. Especially in light of a mistake that I recently made by using `add()` instead of `remove()`.

sissbruecker: Yeah, can't hurt. Added a check to verify the element gets removed. The following `getTestField` call should also fail if the element doesn't get added again.

vursen (approved):

vaadin-flow-components-shared-parent/vaadin-flow-components-test-util/src/main/java/com/vaadin/tests/validation/AbstractValidationIT.java
knoobie: Thanks! 🙇‍♂️ I had flow's batching of action in mind where I have seen sometimes weird behavior with adding and removing of components in the same server circle

PR #4321, opened 2022-12-08
refactor!: remove deprecated GeneratedVaadinFormLayout

web-padawan:
## Description

Removed deprecated `GeneratedVaadinFormLayout` and `GeneratedVaadinFormItem` classes. 

## Type of change

- Breaking change


vaadin-form-layout-flow-parent/vaadin-form-layout-flow/src/main/java/com/vaadin/flow/component/formlayout/FormLayout.java
```diff
@@ -206,14 +220,56 @@ public FormItem(Component... components) {
...
+         * @param components
+         *            The components to add.
+         * @see <a href=
+         *      "https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot">MDN
+         *      page about slots</a>
+         * @see <a href=
+         *      "https://html.spec.whatwg.org/multipage/scripting.html#the-slot-element">Spec
+         *      website about slots</a>
+         */
+        protected void addToLabel(Component... components) {
```
sissbruecker, line 236: Does this method have any value?

vaadin-form-layout-flow-parent/vaadin-form-layout-flow/src/main/java/com/vaadin/flow/component/formlayout/FormLayout.java
```diff
@@ -206,14 +220,56 @@ public FormItem(Component... components) {
...
+                getElement().appendChild(component.getElement());
+            }
+        }
+
+        /**
+         * Removes all contents from this component, this includes child
+         * components, text content as well as child elements that have been
+         * added directly to this component using the {@link Element} API.
+         */
         public void removeAll() {
```
sissbruecker, line 248: Both `remove` and `removeAll` already exist in `HasComponents`, with a slightly different implementation. The addition here is that they clear the `slot` attribute, which is kind of debatable if the developer has set the attribute manually. Should we keep these?

vaadin-form-layout-flow-parent/vaadin-form-layout-flow/src/main/java/com/vaadin/flow/component/formlayout/FormLayout.java
web-padawan: Yes, it's used here: 
https://github.com/vaadin/flow-components/blob/b64aac94dbbffed372c87683b1c1c973b91690ef/vaadin-form-layout-flow-parent/vaadin-form-layout-flow/src/main/java/com/vaadin/flow/component/formlayout/FormLayout.java#L383-L388

vaadin-form-layout-flow-parent/vaadin-form-layout-flow/src/main/java/com/vaadin/flow/component/formlayout/FormLayout.java
web-padawan: I don't have a strong opinion. Both methods were implemented this way since the initial commit: https://github.com/vaadin/vaadin-form-layout-flow/pull/1. We probably could get rid of them, I just wanted to keep the change in this PR granular.

sissbruecker: OK, makes sense. We can consider doing something about these in a second pass. Also don't feel comfortable making a decision ad-hoc 🙂.

sissbruecker (approved):

PR #4333, opened 2022-12-09
refactor!: remove deprecated GeneratedVaadinSelect

web-padawan:
## Description

1. Removed deprecated `GeneratedVaadinSelect` class,
2. Updated `Select` to implement `HasThemeVariant`.

## Type of change

- Breaking change


vaadin-select-flow-parent/vaadin-select-flow/src/test/java/com/vaadin/flow/component/select/SelectTest.java
```diff
@@ -679,6 +670,7 @@ public void dataViewForFaultyDataProvider_throwsException() {
     }
 
     @Test
+    @Ignore
```
web-padawan, line 673: After updating to use `Select` this test fails with the following error. Any ideas how to fix it?

```
[ERROR] elementHasValue_wrapIntoField_propertyIsNotSetToInitialValue(com.vaadin.flow.component.select.SelectTest)  Time elapsed: 0.277 s  <<< ERROR!
java.lang.NullPointerException: Cannot invoke "com.vaadin.flow.data.provider.KeyMapper.get(String)" because "select.keyMapper" is null
	at com.vaadin.flow.component.select.SelectTest.lambda$elementHasValue_wrapIntoField_propertyIsNotSetToInitialValue$5(SelectTest.java:687)
	at com.vaadin.flow.component.select.SelectTest.elementHasValue_wrapIntoField_propertyIsNotSetToInitialValue(SelectTest.java:688)
```

sissbruecker: The key mapper field isn't initialized yet, as the class initialization only runs after the super constructor. However the super constructor runs the `presentationToValue` method that uses the key mapper.

At least in this case the test scenario doesn't seem realistic. It tests that a value set in a template is retained after creating the component instance. However `Select` uses a `KeyMapper` to generate keys for items, which are then used as component value. So whatever value has been set in the template would need to match whatever keys the KeyMapper generates.

We can fix the isse by adding a null check for `keyMapper` to `Select.presentationToModelValue`. As for the test, we can change it to test that setting a value on the element does not throw an exception when constructing a component from that element.

web-padawan: Thanks, I will update the logic and test accordingly.

web-padawan: When adding a null check for `keyMapper` the original test actually fails as follows:

```
[ERROR] elementHasValue_wrapIntoField_propertyIsNotSetToInitialValue(com.vaadin.flow.component.select.SelectTest)  Time elapsed: 0.277 s  <<< FAILURE!
org.junit.ComparisonFailure: expected:<[foo]> but was:<[]>
	at com.vaadin.flow.component.select.SelectTest.elementHasValue_wrapIntoField_propertyIsNotSetToInitialValue(SelectTest.java:689)
```

web-padawan: Updated the PR as suggested and modified the test.

sissbruecker: Just noticed that the test was using a `TestSelect` class with a custom presentation / model mapper. Then it was truly pointless.

Will take a look at the PR now.

PR #4339, opened 2022-12-12
refactor!: remove deprecated GeneratedVaadinCheckbox

web-padawan:
## Description

Removed deprecated `GeneratedVaadinCheckbox` class and moved some code to `Checkbox`.
Maybe we could consider changing logic to get rid of `setDisabled` / `isDisabledBoolean` 🤔 

Note: I will remove `GeneratedVaadinCheckboxGroup` in a separate PR for release notes.

## Type of change

- Breaking change


vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/Checkbox.java
```diff
@@ -47,8 +63,7 @@ public class Checkbox extends GeneratedVaadinCheckbox<Checkbox, Boolean>
      * Default constructor.
      */
     public Checkbox() {
-        // initial value, default value, accept null value
-        super(false, false, false, true);
+        super("checked", false, false);
```
sissbruecker, line 66: There is a subtle difference here compared to the value initialization from the generated class. Since the initial value is not null (`false`), and in case the `checked` property value is null, the generated class would call effectively call `setPresentationValue(false)`, which would initialize the `checked` property with `false`.

Now after the change, the `checked` property is `null` after initialization. The only difference I can see is an extra sync to the server, probably because the web component initializes the value with `false`:

Bildschirmfoto 2022-12-12 um 11 47 01

If we want to restore the behavior, we would have to add something like:
```java
if (getElement.getProperty("checked") == null) {
    setPresentationValue(false);
}
```
The condition is necessary in order to avoid overwriting any initial value set in a template.

vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/Checkbox.java
```diff
@@ -257,7 +270,27 @@ public void setIndeterminate(boolean indeterminate) {
...
+        return getElement().getProperty("indeterminate", false);
+    }
+
+    /**
+     * If true, the user cannot interact with this element.
+     *
+     * @param disabled
+     *            the boolean value to set
+     */
+    protected void setDisabled(boolean disabled) {
```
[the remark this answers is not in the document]
web-padawan: Changed.

vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/Checkbox.java
web-padawan: Added the check. Should I add some unit test to cover this change? Existing tests didn't catch this.
There is `InjectedCheckboxIT` but it doesn't test `checked` attribute being set in a template.

sissbruecker: One case (checkbox element has initial value) is tested here:
https://github.com/vaadin/flow-components/blob/8bbbb99f5005284b6288bfa1c3bcafe55baf1cf2/vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/test/java/com/vaadin/flow/component/checkbox/tests/CheckboxUnitTest.java#L90-L110

We could add an additional test where we just create a new checkbox instance and verify that the property is initialized with `false` by default.

web-padawan: Thanks, makes sense. I will do that.

vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/Checkbox.java
```diff
@@ -47,12 +63,15 @@ public class Checkbox extends GeneratedVaadinCheckbox<Checkbox, Boolean>
...
      */
     public Checkbox() {
-        // initial value, default value, accept null value
-        super(false, false, false, true);
+        super("checked", false, false);
         getElement().addPropertyChangeListener("indeterminate",
                 "indeterminate-changed", NO_OP);
         getElement().addPropertyChangeListener("checked", "checked-changed",
                 NO_OP);
+        // Avoid overwriting initial checked set in a template.
```
sissbruecker, line 71:
```suggestion
        // Initialize property value unless it has already been set from a template
```

sissbruecker (approved):

PR #4340, opened 2022-12-12
fix: remove invalid state sync (#4325) (CP: 23.2)

sissbruecker:
Cherry-pick of #4325


sissbruecker:
Had to resolve a conflict around imports.

vursen (approved):
Looks good to me. I was a bit concerned that we are introducing a new TestBench public API as part of a fix. I've ignored that fact in earlier PRs for some reason, but it seems too late to change that.

PR #4360, opened 2022-12-14
fix: Fix NPE if component renderer returns null

mstahv:
## Description

Users (including myself couple of times) intuitively return null for e.g. Grid cells that shouldn't contain component and consider this a bug. It is common that only some rows contain e.g. an edit button (for example based on permissions).

Currently we throw a nasty NPE from the internals. With this simple fix we do a null check instead and render nothing.

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

- [ ] I have read the contribution guide: https://vaadin.com/docs/latest/contributing/overview


knoobie:
Thank you 🙇‍♂️ Real quality of life improvement

mstahv:
Replaced wit PR towards master: #4361

PR #4361, opened 2022-12-15
fix: Fix NPE if component renderer returns null

mstahv:
## Description

Users (including myself couple of times) intuitively return null for e.g. Grid cells that shouldn't contain component and consider this a bug. It is common that only some rows contain e.g. an edit button (for example based on permissions).

Currently we throw a nasty NPE from the internals. With this simple fix we do a null check instead and render nothing.

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

- [ ] I have read the contribution guide: https://vaadin.com/docs/latest/contributing/overview


vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/java/com/vaadin/flow/data/renderer/ComponentDataGenerator.java
```diff
@@ -83,6 +83,9 @@ public void generateData(T item, JsonObject jsonObject) {
             nodeId = oldRenderedComponent.getElement().getNode().getId();
         } else {
             Component renderedComponent = createComponent(item);
+            if (renderedComponent == null) {
+                return;
```
tomivirkki, line 87: With virtual scrolling components, that re-use the same container-element, the old content doesn't get removed:

```java
grid.addComponentColumn(item -> {
    if (item.getFirstName().equals("Person 40")) {
        return null;
    }

    return new TextField("", item.getFirstName(), "");

}).setHeader("Test");
```

https://user-images.githubusercontent.com/1222264/207805170-0a6fec81-f481-40ba-a8da-3ec6cb005e38.mp4

vaadin-renderer-flow-parent/vaadin-renderer-flow/src/test/java/com/vaadin/flow/data/renderer/ComponentRendererTest.java
```diff
@@ -132,6 +132,44 @@ public void componentRenderer_childAttachedBeforeParent() {
...
+        TestUIInternals internals = (TestUIInternals) ui.getInternals();
+
+        ComponentRenderer<TestDiv, String> nullRenderer = new ComponentRenderer<>(
+                e -> null);
+
+        Element containerParent = new Element("div");
+        Element container = new Element("div");
+        KeyMapper<String> keyMapper = new KeyMapper<>();
+
+        ComponentDataGenerator<String> rendering = (ComponentDataGenerator<String>) nullRenderer
```
tomivirkki, line 147: When running locally, I'm getting:
```
java.lang.ClassCastException: class com.vaadin.flow.data.renderer.ComponentRenderer$1 cannot be cast to class com.vaadin.flow.data.renderer.ComponentDataGenerator (com.vaadin.flow.data.renderer.ComponentRenderer$1 and com.vaadin.flow.data.renderer.ComponentDataGenerator are in unnamed module of loader 'app')
 at com.vaadin.flow.data.renderer.ComponentRendererTest.allowNullValues([ComponentRendererTest.java:147](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html))
```

tomivirkki (changes requested):

vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/java/com/vaadin/flow/data/renderer/ComponentDataGenerator.java
mstahv: This is now something new to master 🤔

tomivirkki: ComponentRenderer was refactored for V24 to use `Lit` instead of a `<template>` internally. Could be that this part worked differently on V23.

vaadin-renderer-flow-parent/vaadin-renderer-flow/src/test/java/com/vaadin/flow/data/renderer/ComponentRendererTest.java
mstahv: Fixed, this was also issue when moved to master from v23

vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/java/com/vaadin/flow/data/renderer/ComponentDataGenerator.java
mstahv: Do you have good ideas how to test this in the module itself?

tomivirkki: It depends on the fix whether this can be tested with a server-side test case or if it requires an IT. That may be tricky since there isn't an IT setup for ComponentRenderer in the module yet.

mstahv: Damn, I should have done this in the summer 🤯

mstahv: A (probably bit non-optimal) workaround could be to drop in a dummy div component in ComponentDataGenerator if the supplier returns null 🤷‍♂️

mstahv: Like this: #4366 

But I have no idea to test that in a reasonable way...

mstahv: Now using this dummyfix. The good thing (even if this uses slightly more bandwidth) is that it might be easier to backport (didn't check, but just a guess).

tomivirkki (approved):

PR #4377, opened 2022-12-15
refactor!: replace GeneratedVaadinTextField with InternalFieldBase

web-padawan:
## Description

1. Created a new `FieldBase` class that provides common interfaces / public methods that are shared by all the fields,
2. Updated `TextField`, `EmailField`, `PasswordField`, `AbstractNumberField`, `BigDecimalFIeld` to use new class,
3. Removed no longer used `GeneratedVaadinTextField` class and updated corresponding JavaDoc references.

## Type of change

- Breaking change


vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/FieldBase.java
```diff
@@ -0,0 +1,309 @@
...
+     *      "https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot">MDN
+     *      page about slots</a>
+     * @see <a href=
+     *      "https://html.spec.whatwg.org/multipage/scripting.html#the-slot-element">Spec
+     *      website about slots</a>
+     */
+    protected void addToPrefix(Component... components) {
+        for (Component component : components) {
+            component.getElement().setAttribute("slot", "prefix");
+            getElement().appendChild(component.getElement());
```
knoobie, line 238: Would it make sense to add a universal add to slot method to https://github.com/vaadin/flow-components/blob/master/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/SlotUtils.java allowing to re-use it here and on other occurrences?

knoobie: Sounds good 👍

knoobie: Another note: should this be moved to the shared package and more used while you guys are at cleaning the components? https://github.com/vaadin/flow-components/blob/master/vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/HasPrefixAndSuffix.java

web-padawan: Yes, I was thinking about that too. I will create a PR.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/FieldBase.java
```diff
@@ -0,0 +1,389 @@
...
+import com.vaadin.flow.component.shared.HasThemeVariant;
+import com.vaadin.flow.component.shared.HasTooltip;
+import com.vaadin.flow.data.value.HasValueChangeMode;
+import com.vaadin.flow.dom.Element;
+import com.vaadin.flow.function.SerializableBiFunction;
+import com.vaadin.flow.function.SerializableFunction;
+import com.vaadin.flow.shared.Registration;
+
+/**
+ * Provides base functionality for combo box related components, such as
```
knoobie, line 43: Combo Box related?

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/FieldBase.java
```diff
@@ -0,0 +1,389 @@
...
+ * {@link TextField}
+ *
+ * @param <TComponent>
+ *            Type of the component that extends from this class
+ * @param <TValue>
+ *            Type of the value of the extending component
+ */
+public abstract class FieldBase<TComponent extends FieldBase<TComponent, TValue>, TValue>
```
knoobie, line 51: > Created a new FieldBase class that provides common interfaces / public methods that are shared by all the fields

Reading this, I was kinda expecting this in the shared package and ALL input fields using this as base class (which would be awesome, looking back how V8 had a class that all input fields extended as well)

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/FieldBase.java
web-padawan: My bad, copy paste from `ComboBoxBase` - will fix.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/FieldBase.java
web-padawan: That can be an option as well, although for now I would prefer to keep it in this package.
I'm going to update `TextArea` to also use `FieldBase` as it has most of shared logic.

knoobie: I was more referring to all fields including checkbox, combo box and so on 😅 I can still dream about it tho

web-padawan: These could be aligned to some extent, similarly to how we use `@vaadin/field-base` in web components.
However, in some simple cases like `Checkbox` the effort would probably not be worth it.

ugur-vaadin (approved):

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/FieldBase.java
```diff
@@ -0,0 +1,389 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component.textfield;
+
+import com.vaadin.flow.component.AbstractSinglePropertyField;
+import com.vaadin.flow.component.Component;
+import com.vaadin.flow.component.ComponentEvent;
```
tomivirkki, line 20: There are a few unused imports here

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/FieldBase.java
tomivirkki: If we want to leave the room open for further refactorings, such as this, we could consider making the class package private for now so splitting/relocating it later wouldn't cause a breaking change

web-padawan:
> consider making the class package private for now so splitting/relocating it later wouldn't cause a breaking change

@tomivirkki When removing `public` to make the class package private, I'm getting these compilation errors:

```
[ERROR] /Users/serhii/vaadin/flow-components/vaadin-text-field-flow-parent/vaadin-text-field-flow-integration-tests/src/main/java/com/vaadin/flow/component/textfield/tests/ValueChangeModePage.java:[56,18] incompatible types: formal varargs element type com.vaadin.flow.component.textfield.FieldBase<? extends com.vaadin.flow.component.textfield.FieldBase<?,?>&com.vaadin.flow.component.shared.HasThemeVariant<? extends java.lang.Enum<? extends java.lang.Enum<?>&com.vaadin.flow.component.shared.ThemeVariant>&com.vaadin.flow.component.shared.ThemeVariant>,? extends java.lang.Object&java.io.Serializable&java.lang.Comparable<? extends java.lang.Object&java.io.Serializable&java.lang.Comparable<?>>>&com.vaadin.flow.component.shared.HasThemeVariant<? extends java.lang.Enum<? extends java.lang.Enum<?>&com.vaadin.flow.component.shared.ThemeVariant>&com.vaadin.flow.component.shared.ThemeVariant> is not accessible from class com.vaadin.flow.component.textfield.tests.ValueChangeModePage
```

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/FieldBase.java
web-padawan: Removed.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/FieldBase.java
tomivirkki: > When removing public to make the class package private, I'm getting these compilation errors:

So it seems...not an option then

knoobie: If using package private is not possible. How about a strict renaming of the class to `InternalFieldBase` with a JavaDoc comment? At least then you can always say:  it was internal, don't touch it.

web-padawan: @knoobie Good point, thanks. Done.

knoobie: No more getElement dilemma ;)

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/FieldBase.java
web-padawan: Fixed.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/InternalFieldBase.java
```diff
@@ -0,0 +1,423 @@
...
+    }
+
+    /**
+     * String used for the label element.
+     *
+     * @return the {@code label} property from the web component
+     */
+    public String getLabel() {
+        return getElement().getProperty("label");
+    }
```
tomivirkki, line 110: Since the class implements `HasLabel`, these shouldn't be needed?

web-padawan: Checked the `HasLabel` interface implementation and there are two differences:

- [`setLabel(String)`](https://github.com/vaadin/flow/blob/a2fb35292cd0e5328cfd0ba0da32bd20ff378a99/flow-server/src/main/java/com/vaadin/flow/component/HasLabel.java#L38-L40) method doesn't take care of `null`,
- [`getLabel()`](https://github.com/vaadin/flow/blob/a2fb35292cd0e5328cfd0ba0da32bd20ff378a99/flow-server/src/main/java/com/vaadin/flow/component/HasLabel.java#L48-L51) returns `null` as a fallback for the label.

In case of `ComboBoxBase` we don't override these methods though 🤷‍♂️ 
@sissbruecker would you be OK to remove overrides from here too?

sissbruecker: I honestly can't remember. 

In this case, the `getLabel()` implementation is pretty much the same. And the `setLabel()` implementation here is not really consistent:
```
NumberField field = new NumberField();
System.out.println(field.getLabel()); => null

field.setLabel(null);
System.out.println(field.getLabel()); => ""
```

Besides, it would leave an empty `label` attribute on the client-side, not sure if that is necessary. If not, I'd say we can remove the overrides here?

web-padawan: Thanks, good point about inconsistency, I didn't notice it myself. Will remove these overrides then.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/PasswordField.java
```diff
@@ -56,13 +42,8 @@
...
 @NpmPackage(value = "@vaadin/password-field", version = "24.0.0-alpha6")
 @JsModule("@vaadin/password-field/src/vaadin-password-field.js")
-public class PasswordField
-        extends GeneratedVaadinTextField<PasswordField, String> implements
-        CompositionNotifier, Focusable<PasswordField>, HasAllowedCharPattern,
-        HasAutocapitalize, HasAutocomplete, HasAutocorrect, HasClearButton,
-        HasClientValidation, HasHelper, HasLabel, HasPrefixAndSuffix, HasSize,
-        HasStyle, HasThemeVariant<TextFieldVariant>, HasTooltip, HasValidation,
-        HasValidator<String>, HasValueChangeMode, InputNotifier, KeyNotifier {
+public class PasswordField extends InternalFieldBase<PasswordField, String>
```
sissbruecker, line 45: Looks like the value change mode logic can be removed here as well?

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/InternalFieldBase.java
web-padawan: Removed.

tomivirkki (approved):

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/IntegerField.java
```diff
@@ -31,7 +32,8 @@
 @JsModule("@vaadin/polymer-legacy-adapter/style-modules.js")
 @NpmPackage(value = "@vaadin/integer-field", version = "24.0.0-alpha6")
 @JsModule("@vaadin/integer-field/src/vaadin-integer-field.js")
-public class IntegerField extends AbstractNumberField<IntegerField, Integer> {
+public class IntegerField extends AbstractNumberField<IntegerField, Integer>
+        implements HasThemeVariant<TextFieldVariant> {
```
sissbruecker, line 36: `IntegerField` does not implement `HasAllowedCharPattern` anymore. Intentional?

web-padawan: Yes, because it is [hardcoded](https://github.com/vaadin/web-components/blob/ee7af5454f325a64c892d4a7ddffc268faaa7472/packages/integer-field/src/vaadin-integer-field.js#L45) in the web component and not supposed to be changed by the user.
Same applies to `BigDecimalField` which also doesn't implement `HasAllowedCharPattern`.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/FieldBase.java
tomivirkki: One more thing we could do to enforce this being an internal class is to make the constructors package private. That way no-one should be able to make a component that extends `InternalFieldBase`

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/PasswordField.java
web-padawan: Thanks, I missed to remove it. Done.

sissbruecker (approved):

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/FieldBase.java
web-padawan: Missed this comment, sorry about that. Here is a follow up PR: https://github.com/vaadin/flow-components/pull/4410

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/FieldBase.java
web-padawan: Here is a PR that adds `HasPrefix` and `HasSuffix` interfaces: #4458

PR #4379, opened 2022-12-16
chore: make mvn clean remove generated frontend

sissbruecker:
## Description

Currently when switching between version branches, the contents of `frontend/generated` can lead to compilation errors for example because they import modules that do not exist in that version (e.g. tooltip in 23.2). Usually I run `mvn clean` after switching branches, however that doesn't cover the generated frontend files. Let's make `clean` also cover those.

Alternatively it is possible to automatically run `flow:clean-frontend` whenever `clean` is run, however that does not cover all files that we remove with our `clean` configuration. So it's probably better to do everything in our custom clean configuration.

### DIFF
--- a/scripts/templates/pom-integration-tests.xml
+++ b/scripts/templates/pom-integration-tests.xml
@@ -34,6 +34,7 @@
                         </fileset>
                         <fileset>
                             <directory>${project.basedir}/node_modules</directory>
+                            <directory>${project.basedir}/frontend/generated</directory>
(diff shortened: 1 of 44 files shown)


sissbruecker:
I think I'll cherry-pick this manually, as there are probably lots of conflicts due to differences in modules.

web-padawan (approved):

tomivirkki (approved):

PR #4380, opened 2022-12-16
chore: fix live reload for jetty

sissbruecker:
## Description

Restores the live reload for the Jetty server used in IT modules. It seems that for Jetty 11 `scan` needs to be used instead of `scanIntervalSeconds`.


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-integration-tests/pom.xml
```diff
@@ -221,7 +221,7 @@
                         <groupId>org.eclipse.jetty</groupId>
                         <artifactId>jetty-maven-plugin</artifactId>
                         <configuration>
-                            <scanIntervalSeconds>5</scanIntervalSeconds>
+                            <scan>5</scan>
```
tomivirkki, line 224: I'm experiencing some weird side-effects with Spreadsheet IT pages. Didn't try with other components yet:


https://user-images.githubusercontent.com/1222264/208070127-605507e3-eb4d-48ab-9e30-71c4ba73f4da.mp4

tomivirkki: I tried again with other components and it seems to work fine. Let's merge this and if live reload with the spreadsheet turns out to be a problem, it can always be disabled in that module.

tomivirkki (approved):

PR #4404, opened 2022-12-19
feat!: change datepicker default format to ISO8601 for unsupported locale

### LINKED ISSUE #4006: [DatePicker] Use ISO date format as fallback if locale is not supported
### Describe your motivation

Currently, when setting a locale on a date picker that the browser does not support, the date picker will log a warning and then use the `en-US` date pattern as fallback:
https://github.com/vaadin/flow-components/blob/36648d3a48798df315f8dafe80eca39283589f46/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js#L78-L84

This is quite an opinionated choice as a general fallback for any unsupported locale, and a general, locale / region independent locale might be a better a fit.

Use the ISO 8601 format as a fallback instead. This is a neutral format not associated with any country.

This can be considered a breaking change, and should be tackled in a new major.

_No response_

### Additional context

There are plans to refactor locale-based parsing / formatting to use date-fns patterns and introduce some additional checks to verify that the generated pattern is valid. We should make sure that this change covers all places where we use a fallback locale / pattern.

ugur-vaadin:
## Description

With this PR, `DatePicker` fallbacks to ISO 8601 format instead of `en-US` for unsupported locales.

This is a breaking change.

Fixes #4006 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/datepicker/DatePickerLocaleIT.java
```diff
@@ -155,23 +154,23 @@ public void datePicker_setUnsupportedLocale_warningIsShown() {
...
     }
 
     @Test
-    public void datePicker_setUnsupportedLocale_defaultUSLocaleIsUsed() {
+    public void datePicker_setUnsupportedLocale_defaultISO8601FormatIsUsed() {
         picker.setDate(LocalDate.of(2018, Month.MAY, 3));
 
         List<Locale> unsupportedLocales = List.of(new Locale("ar", "SA"),
-                new Locale("th", "TH"), new Locale("fa"), new Locale("ks"));
+                new Locale("th", "TH"), new Locale("fa"));
```
ugur-vaadin, line 165: This omission is due to Kashmiri (`ks`) locale being parsed without a problem by the `toLocaleDateString(locale)` call in the connector.

tomivirkki (approved):
LGTM, this should be marked as a breaking change though. So `feat!: change...`

ugur-vaadin:
> LGTM, this should be marked as a breaking change though. So `feat!: change...`

Edited.

PR #4412, opened 2022-12-20
chore: remove usage of dom-module from charts examples

web-padawan:
## Description

1. Removed HTML files with `<dom-module>` definitions - leftovers from Polymer 2 and HTML Imports era,
2. Moved corresponding CSS to empty `.css` files that are actually included by demos using `@CssImport`,
3. Removed `include = "vaadin-chart-default-theme"` from those `@CssImport` to get rid of warnings.

Note: I didn't bother to check that styles are applied correctly 🙈  We could just drop some of these demos.

## Type of change

- Internal change


web-padawan:
For some reason the contents of CSS files gets removed when running `mvn formatter:format` 🤯  
That's why they are empty on master branch since adding the formatter plugin in https://github.com/vaadin/flow-components/commit/65ccf0b22f24053d922e0db7c25d04752daf8ecb.

sissbruecker (approved):
Verified that styles are still applied 👍

Issue #4395, opened 2022-12-19
DatePicker keeps invalid value when Binder sets empty value


simasch:
### Description

I have a grid with a form that has a DatePicker. The data is bound with a Binder.

The user enters and invalid data like DDDD

Then the user selects another record in the grid. This value of the date of this record is null.

The invalid value DDD stays in the DataPicker.

### Expected outcome

The value of the DatePicker should be empty.

### Minimal reproducible example

none

### Steps to reproduce

as described above

### Environment

Vaadin version(s): 23.3.0

This works fine in 23.2.x

simasch:
This is a blocker! Because it prevents from migrating our application to 23.3.0

vursen:
Just to make sure, is the issue reproducible for you when you have the `enforceFieldValidation` feature flag enabled? Or it doesn't matter?

simasch:
Yes it's the same with that flag enabeld

simasch:
@web-padawan In the meanwhile I've deleted the comment because it also doesn't work in 23.2

simasch:
@web-padawan
To reproduce clone the repo: https://github.com/simasch-scratches/datepicker-with-binder-invalid

- The first two records have no birthday set.
- Enter ddd in the birthday field and switch to the second record.
- The ddd will stay in the birthday field

vursen:
When entering bad input, the value resets to `null`, and that accounts for why a subsequent `setValue(null)` doesn't have any effect. The first solution that comes to mind would be to force the web component to clear the input value each time `setValue(null)` is invoked on the server-side. However, we need to double-check whether it will also work out for the Binder case.

rolfsmeds:
The issue was triaged and currently added to the backlog priority queue for further investigation

vursen:
The issue was fixed. The changes are shipped with the following versions:

| Component | [23.3.2](https://github.com/vaadin/flow-components/releases/tag/23.3.2) | 23.3.3 (not released yet) | [24.0.0.alpha7](https://github.com/vaadin/flow-components/releases/tag/24.0.0.alpha7) |
| ------------ | ------ | ------ | -------------- |
| DatePicker |  ✅ | ✅ | ✅
| TimePicker | | ✅ | ✅
| DateTimePicker | | ✅ | ✅
| NumberField | | ✅ | ✅
| IntegerField | | ✅ | ✅
| BigDecimalField | | ✅ | ✅

PR #4413, opened 2022-12-20
fix: clear possible bad input when setting an empty value

vursen:
## Description

The PR overrides DatePicker's `setValue()` to support clearing the input element from possible bad input when setting an empty value in the case the previous value was already empty. Or, in other words, this PR guarantees that bad input won't remain in the field after you assign an *empty* Binder bean that in fact relates to *another* record.

> **Note**
> The issue isn't only specific to DatePicker but also TimePicker, NumberField, and a few other components. I'm planning to cover them each in their individual PR.

Part of https://github.com/vaadin/flow-components/issues/4395

## Type of change

- [x] Bugfix


vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
```diff
@@ -580,6 +580,21 @@ protected boolean isInputValuePresent() {
...
+        LocalDate oldValue = getValue();
+
+        super.setValue(value);
+
+        if (oldValue == null && value == null && isInputValuePresent()) {
+            // Clear the input element from possible bad input.
+            getElement().executeJs("this.inputElement.value = $0", "");
+            getElement().setProperty("_hasInputValue", false);
+            fireEvent(new ClientValidatedEvent(this, false,
+                    checkValidity(value).isError()));
```
vursen, line 594: **note:** We should probably consider removing this `valid` argument at all. It has become useless since we rely on `isInputValuePresent()` when checking for bad input, instead of an `isClientValid`.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
```diff
@@ -580,6 +580,21 @@ protected boolean isInputValuePresent() {
         return getElement().getProperty("_hasInputValue", false);
     }
 
+    @Override
+    public void setValue(LocalDate value) {
+        LocalDate oldValue = getValue();
+
+        super.setValue(value);
+
+        if (oldValue == null && value == null && isInputValuePresent()) {
```
sissbruecker, line 589: Should this compare against `getEmptyValue`?

sissbruecker (approved):

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
vursen: Good point. Let me update it.

vursen: Done.

sissbruecker (approved):

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
```diff
@@ -579,6 +579,23 @@ protected boolean isInputValuePresent() {
...
+    public void setValue(LocalDate value) {
+        LocalDate oldValue = getValue();
+
+        super.setValue(value);
+
+        if (Objects.equals(oldValue, getEmptyValue())
+                && Objects.equals(value, getEmptyValue())
+                && isInputValuePresent()) {
+            // Clear the input element from possible bad input.
+            getElement().executeJs("this.inputElement.value = $0", "");
```
tomivirkki, line 592: nit: No need to pass the empty value as an argument
```java
getElement().executeJs("this.inputElement.value = ''");
```

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
tomivirkki: At first glance, seems that `checkValidity(value).isError()` would always return `false` here. That would then set the `valid` state of the event to be `false` which sounds kind of sounds backwards. I tried with `!checkValidity(value).isError()` and tests seem to pass.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
vursen: Done.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
vursen: Ah, you are right, it should be `!checkValidity(...` indeed. However, as I mentioned, this argument is unused. I have added it here only because the event is a public API and someone might subscribe to it.

vursen: I guess I'll go open a PR removing it.

vursen: > At first glance, seems that checkValidity(value).isError() would always return false here

By the way, it may return `true` if the component is `required`.

tomivirkki: Tried with this test case and the output is always the same, regardless if the date picker is required or not.

```java
@Test
public void foobar() {
    DatePicker picker = new DatePicker();
    picker.setRequired(true);
    picker.addClientValidatedEventListener(e -> {
        System.out.println("valid? " + e.isValid());
    });
    picker.getElement().setProperty("_hasInputValue", true);
    picker.clear();
}
```

vursen: Created a [ticket](https://github.com/vaadin/flow-components/issues/4425) to deprecate and remove this argument. In the meanwhile, I just pass `true`.

tomivirkki (approved):

PR #4431, opened 2022-12-22
fix(DateTimePicker): clear possible bad input when setting an empty value

vursen:
## Description

The PR overrides DateTimePicker's `setValue()` to support clearing the input element from possible bad input when setting an empty value in the case the previous value was already empty. Or, in other words, this PR guarantees that bad input won't remain in the field after you assign an *empty* Binder bean that in fact relates to *another* record.

Part of https://github.com/vaadin/flow-components/issues/4395

## Type of change

- [x] Bugfix


vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/datetimepicker/validation/BinderValidationIT.java
```diff
@@ -73,48 +74,6 @@ public void required_changeValue_assertValidity() {
...
-    }
-
-    @Test
-    public void required_badInput_setTimeInputValue_blur_assertValidity() {
-        setInputValue(timeInput, "INVALID");
-        timeInput.sendKeys(Keys.TAB);
-        assertServerInvalid();
-        assertClientInvalid();
-        assertErrorMessage("");
-    }
```
vursen, line 93: **note:** Moved these tests to `BasicValidationIT.java` as they aren't really tied to Binder usage.

vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow/src/main/java/com/vaadin/flow/component/datetimepicker/DateTimePicker.java
```diff
@@ -56,6 +54,13 @@ protected void validate() {
 
     void passThroughPresentationValue(LocalDate newPresentationValue) {
         super.setPresentationValue(newPresentationValue);
+
+        if (Objects.equals(newPresentationValue, getEmptyValue())
```
vursen, line 58: **note:** It was hard to find a clean way to get an old value since internal pickers seem to only operate with their presentation value, so I don't check for an old value here. As a consequence, the following JS instructions can be sent even when changing from a non-empty value to an empty value. This shouldn't be a big deal though.

web-padawan (approved):

Issue #1594, opened 2018-10-11
Support for prefix/suffix component


heruan:
It would be useful to set prefix and suffix components in the same way it's done with `TextField`.

In the [advanced demos](https://vaadin.com/components/vaadin-combo-box/html-examples/combo-box-advanced-demos) there is a similar example using `vaadin-combo-box-light`, but there's no Flow component for that.

P.S. Not sure if this is more appropriate in the Web Component repo, if so I'll open an issue there referencing this.

mvysny:
Also the examples at https://vaadin.com/components/vaadin-date-picker/html-examples should mention the possibility to set a component to the prefix slot.

PR #4458, opened 2022-12-30
feat: add HasPrefix and HasSuffix shared interfaces

web-padawan:
## Description

Inspired by https://github.com/vaadin/flow-components/pull/4377#discussion_r1050102495 

Pre-requisite for #1594

Created two separate interfaces: `HasPrefix` and `HasSuffix` based on the `HasPrefixAndSuffix` interface.
I will create a separate PR to update `TextField` and other components to use these new interfaces.

After all the components are updated, we could probably drop `HasPrefixAndSuffix` as no longer needed.

## Type of change

- Internal feature

### DIFF
--- a/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/HasPrefix.java
+++ b/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/HasPrefix.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2000-2022 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component.shared;
+
+import com.vaadin.flow.component.Component;
+import com.vaadin.flow.component.HasElement;
+
+/**
+ * Mixin interface for components that have a prefix slot.
+ *
+ * @author Vaadin Ltd
+ */
+public interface HasPrefix extends HasElement {
+    /**
+     * Adds the given component into this field before the content, replacing
+     * any existing prefix component.
+     * <p>
+     * This is most commonly used to add a simple icon or static text into the
+     * field.
+     *
+     * @param component
+     *            the component to set, can be {@code null} to remove existing
+     *            prefix component
+     */
+    default void setPrefixComponent(Component component) {
+        SlotUtils.setSlot(this, "prefix", component);
+    }
+
+    /**
+     * Gets the component in the prefix slot of this field.
+     *
+     * @return the prefix component of this field, or {@code null} if no prefix
+     *         component has been set
+     * @see #setPrefixComponent(Component)
+     */
+    default Component getPrefixComponent() {
+        return SlotUtils.getChildInSlot(this, "prefix");
+    }
+}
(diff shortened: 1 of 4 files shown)


vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/HasPrefix.java
```diff
@@ -0,0 +1,52 @@
...
+
+import com.vaadin.flow.component.Component;
+import com.vaadin.flow.component.HasElement;
+
+/**
+ * Mixin interface for components that have a prefix slot.
+ *
+ * @author Vaadin Ltd
+ */
+public interface HasPrefix extends HasElement {
```
knoobie, line 26: The chance is probably really low that people have directly used the old interface, but should we add `@Deprecated` in 23.3 for it? Even tho this could again create problems with Eclipse because of the methods.. or they have to be added to all the TextField classes in 23.3 to override the interface's methods..

web-padawan: Unfortunately we can't add `@Deprecated` in 23.3 anymore because it's stable 😕 
I'm afraid we will have to keep `HasPrefixAndSuffix` around if that's a problem.

knoobie: Not a problem per-se, I was just wondering how the new interfaces to be added to the text field based components that currently have the old interface :) The following could work: 

- `HasPrefixAndSuffix extends HasPrefix, HasSuffix`
- Remove all methods from `HasPrefixAndSuffix`
- Add `@Deprecated` to `HasPrefixAndSuffix` in V24 to be removed in V25 

I'm not sure how Eclipse is treating the methods inherited from `HasPrefix`, `HasSuffix` - if they are going to be marked as deprecated or not.

web-padawan: Sounds good to me. I will update `HasPrefixAndSuffix` and keep it as deprecated then.

BTW we might also want to still keep `addToPrefix` methods as deprecated in V24:

- #4459
- #4460

knoobie: Personally, I would argue that you can remove them in V24 - [most of them are protected anyway](https://github.com/vaadin/flow-components/search?p=1&q=addToPrefix).. people had to "deep dive" to access them, forcing them to change this could also improve their code if that was the only reason to sub-class the components and it is already deprecated in 23.3 ;)

web-padawan: Right, most of these methods are protected indeed. Then it's probably fine to remove them 🙈

knoobie: Another alternative could be to add `addToPrefix` as deprecated method into the interface, if you wanna have people gradually change their code (even tho I'm not the biggest fan of this approach)

web-padawan: One subtle difference is that `addToPrefix` accepts multiple components, while `setPrefixComponent` only accepts a single component. Personally, I think the latter should be preferred. We will discuss this with the team.

web-padawan: > how the new interfaces could be added to the text field based components that currently have the old interface 

Created two PRs to update `HasPrefixAndSuffix` and change `InternalFieldBase` to use new interfaces:

- https://github.com/vaadin/flow-components/pull/4464
- https://github.com/vaadin/flow-components/pull/4465

As there are breaking changes in these classes in V24 anyways related to removal of generated classes, I hope this approach would work for most of users. At least the actual methods are exactly the same.

vursen (approved):

PR #4465, opened 2022-12-30
refactor!: use HasPrefix and HasSuffix in field base

web-padawan:
## Description

Based on #4458 

Related to #4464

## Type of change

- Refactor

### DIFF
--- a/vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/InternalFieldBase.java
+++ b/vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/InternalFieldBase.java
@@ -28,6 +28,8 @@ import com.vaadin.flow.component.InputNotifier;
 import com.vaadin.flow.component.KeyNotifier;
 import com.vaadin.flow.component.shared.HasClearButton;
 import com.vaadin.flow.component.shared.HasClientValidation;
+import com.vaadin.flow.component.shared.HasPrefix;
+import com.vaadin.flow.component.shared.HasSuffix;
 import com.vaadin.flow.component.shared.HasTooltip;
 import com.vaadin.flow.component.shared.SlotUtils;
 import com.vaadin.flow.data.binder.HasValidator;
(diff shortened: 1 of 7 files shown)


vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/InternalFieldBase.java
```diff
@@ -49,8 +51,8 @@
         extends AbstractSinglePropertyField<TComponent, TValue>
         implements CompositionNotifier, Focusable<TComponent>,
         HasAutocapitalize, HasAutocomplete, HasAutocorrect, HasClearButton,
-        HasClientValidation, HasHelper, HasLabel, HasPrefixAndSuffix, HasSize,
-        HasStyle, HasTooltip, HasValidation, HasValidator<TValue>,
+        HasClientValidation, HasHelper, HasLabel, HasPrefix, HasSize, HasStyle,
+        HasSuffix, HasTooltip, HasValidation, HasValidator<TValue>,
```
knoobie, line 55: Well, if you remove `HasPrefixAndSuffix` from here - you can completely delete it :D It's even worst now, because people are not going to get compile errors, there get nasty runtime errors and are wondering why `TextField instance HasPrefixAndSuffix` returns false instead of true :) Having `HasPrefixAndSuffix extends HasPrefix, HasSuffix` was meant to let the old interface in this component - thanks to the inheritance, the TextField should now be instance of all three interfaces.

web-padawan: Thanks for clarification. I'll close this PR for now, we can get back to this in Vaadin 25.

web-padawan:
Closing per https://github.com/vaadin/flow-components/pull/4465#discussion_r1059368598

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/InternalFieldBase.java
knoobie: Thanks for listening :)  I should have said parts of the above in my original comment as well, you can't look in my head - that's why I wanted to remove the methods from the interface, so that Eclipse doesn't show them as deprecated, as they are moved from deprecated to the inherited new interfaces

web-padawan: Got it, thanks. That makes a lot of sense 👍

Issue #4425, opened 2022-12-22
Allow creating `ClientValidatedEvent` without the `valid` argument


vursen:
`ClientValidatedEvent` comes with a `valid` argument that was introduced along with the event in `23.2` as part of https://github.com/vaadin/platform/issues/3066:

https://github.com/vaadin/flow-components/blob/d0ffddd50f3e8e6ec6e56d82f13423400ee2e203/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/HasClientValidation.java#L67-L68

We planned to use this argument to identify bad input in the field. However, later we changed our minds and went with the `_hasInputValue` property but didn't remove the argument.

Beginning with https://github.com/vaadin/flow-components/pull/4413, this argument can be computed inaccurately, so it makes even more sense to remove it to avoid possible confusion. Developers are supposed to rely on the component's `invalid` state while the web component's validity state is rather an internal detail.

Let's deprecate it in 24.0 and drop it in 25.0.

vursen:
We decided internally not to deprecate this argument as there is still a chance that things may change during v24 -> v25 and a need for this flag may arise again. Instead, we will:

- Introduce a constructor without the `valid` argument that should be used when creating this event on the server-side
- Emphasize in the `getValid()` method's JSDoc that it always returns `true` if the event originated from the server-side.

PR #4472, opened 2023-01-02
refactor: allow creating ClientValidatedEvent without valid argument

vursen:
## Description

The PR introduces a ClientValidatedEvent constructor overload to allow the event to be created without the `valid` argument. This constructor is supposed to be used when creating the event manually on the server-side.

The PR also mentions in the JSDoc of the related `getValid()` method that it always returns `true` if the event originated from the server-side.

Related to https://github.com/vaadin/flow-components/issues/4425

## Type of change

- [x] Internal


web-padawan (approved):

tomivirkki (approved):
The failing CI tests should get fixed in https://github.com/vaadin/flow-components/pull/4470

PR #4479, opened 2023-01-03
refactor!: relocate Renderer API with custom function name

tomivirkki:
## Description

Follow-up for https://github.com/vaadin/flow-components/pull/4224

Moves the `render(Element container, DataKeyMapper<SOURCE> keyMapper, String rendererName)` API from `LitRenderer` to `Renderer`. Every type of `Renderer` should accept a custom name for the renderer function.

Example:
Before the change, `Grid` [needed to cast the renderer](https://github.com/vaadin/flow-components/blob/cea5c82ea950a600e4a2b65acd4db58ae15c524b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java#L2868-L2869) passed to `setItemDetailsRenderer` as a `LitRenderer` to make it plug into the `rowDetailsRenderer` function on the Web Component. After the change, since the API is defined on the `Renderer`, casting is no longer necessary and any `Renderer` can be passed as an item details renderer.

## Type of change

- Refactoring

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/ColumnPathRenderer.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/ColumnPathRenderer.java
@@ -61,7 +61,7 @@ public class ColumnPathRenderer<SOURCE> extends Renderer<SOURCE> {
 
     @Override
     public Rendering<SOURCE> render(Element container,
-            DataKeyMapper<SOURCE> keyMapper) {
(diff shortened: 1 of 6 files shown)


DiegoCardoso (commented):
`GridPro` has the class `ColumnComponentPathRenderer`, which extends `ColumnPathRenderer`. Should the signature of `render` be updated, so it will override the same method (`render(Element container, DataKeyMapper<SOURCE> keyMapper, String rendererName)`)?

tomivirkki:
> Should the signature of render be updated, so it will override the same method

Probably cleaner that way. Done.

DiegoCardoso (approved):

PR #4483, opened 2023-01-04
refactor!: remove flow-component-renderer

tomivirkki:
WIP: opening a draft PR to run the tests


vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/AbstractComboBoxIT.java
```diff
@@ -105,8 +105,7 @@ protected void assertNotRendered(String innerHTML) {
     }
```
tomivirkki, line 105: Note: Most of the changes in this PR are related to various integration tests that assume there to be a `<flow-component-renderer>` element.

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
```diff
@@ -106,10 +106,11 @@ public Dialog() {
         container.getClassList().add("draggable-leaf-only");
         container.getStyle().set(ElementConstants.STYLE_WIDTH, "100%");
         container.getStyle().set(ElementConstants.STYLE_HEIGHT, "100%");
+        container.getStyle().set("display", "inline-block");
```
tomivirkki, line 109: This fix was originally added in https://github.com/vaadin/vaadin-dialog-flow/pull/195 and then removed in https://github.com/vaadin/vaadin-dialog-flow/pull/232 due to the regression it caused. Since `<flow-component-renderer>` is now removed, I reverted back to the original fix.

Tested the change and both of the previous Dialog scrollbar issues are covered:
- https://github.com/vaadin/vaadin-dialog-flow/issues/185
* https://github.com/vaadin/vaadin-dialog-flow/issues/225

Consider this a temporary fix, since the remaining unnecessary `<div>` wrapper will be completely removed by a separate follow-up PR.

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridDelegateFocusToComponentPage.java
```diff
@@ -45,18 +45,12 @@ public GridDelegateFocusToComponentPage() {
     }
 
     private Div buildComplexComponent(String item) {
-        TextField hiddenTextField = new TextField();
-        hiddenTextField.setValue("hidden");
-        hiddenTextField.getElement().setAttribute("hidden", true);
-        TextField disabledTextField = new TextField();
-        disabledTextField.setValue("disabled");
-        disabledTextField.setEnabled(false);
```
tomivirkki, line 53: When entering from cell navigation mode to interaction mode (by pressing enter), `<vaadin-grid>` Web Component [tries to automatically focus](https://github.com/vaadin/web-components/blob/ed28a0f1f3b2a40ed26c25ffbfe915a1487b9cc3/packages/grid/src/vaadin-grid-keyboard-navigation-mixin.js#L610) the element with "focus-target" attribute or the first focusable element in the cell.

The `<flow-component-renderer>` had some [custom logic](https://github.com/vaadin/flow-components/blob/99682fa0fbe06134cca77c9b7ed7da554f701266/vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/flow-component-renderer.js#L110) to automatically mark an element as "focus-target" _when the node was appended_. This custom logic had additional checks to ignore disabled and hidden elements, which is currently not properly covered by the Web Component's [`isFocusable` check](https://github.com/vaadin/web-components/blob/ed28a0f1f3b2a40ed26c25ffbfe915a1487b9cc3/packages/grid/src/vaadin-grid-active-item-mixin.js#L106).

Since not using `<flow-component-renderer>` falls back to using the Web Component focusing logic, I needed to disabled the checks in this test. The `isFocusable` check in `<vaadin-grid>` should be fixed to properly ignore hidden and disabled nodes instead.

tomivirkki: Submitted a [PR](https://github.com/vaadin/web-components/pull/5263) to fix the issue in `<vaadin-grid>`

vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/flow-component-renderer.js
```diff
@@ -3,6 +3,70 @@ import { html } from '@polymer/polymer/lib/utils/html-tag.js';
...
+    if (referenceNode && referenceNode.parentNode === this) {
+      return originalInsertBefore.call(this, newNode, referenceNode);
+    } else {
+      return originalInsertBefore.call(this, newNode, null);
+    }
+  };
+}
+
+window.Vaadin ||= {};
+window.Vaadin.FlowComponentRenderer ||= { patchVirtualContainer, getNode, renderNodes };
```
sissbruecker, line 69: Could be confusing to use the same name as for the custom element. Since this doesn't really render anything, but rather is used to hosts elements rendered on the server in client-side elements, maybe use `FlowComponentHost`? `renderNodes` might also be a candidate for renaming in this case.

vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/flow-component-renderer.js
```diff
@@ -3,6 +3,70 @@ import { html } from '@polymer/polymer/lib/utils/html-tag.js';
...
 
 class FlowComponentRenderer extends PolymerElement {
```
sissbruecker, line 71: What is the plan for the custom element? I guess we keep this because it is used in custom components / addons? Does it make sense to deprecate this?

tomivirkki: We considered deprecating it but currently, we don't really have a nice public replacement for it so decided to leave it as it is for now. Maybe once the patching workaround is no longer needed, we could consider deprecating `<flow-component-renderer>` in favor of the new helper functions.

vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/flow-component-renderer.js
tomivirkki: Thanks for the suggestion, renamed as `FlowComponentHost`. Also renamed `renderNodes` as `setChildNodes` which better describes the function.

sissbruecker (approved):

jouni:
The "Flow styling issue" note should be removed once v24 is released: https://vaadin.com/docs/latest/components/notification/#recommendations

knoobie:
@tomivirkki Could this change effect the integration with Notification and e.g. TestBench? I've seen quite strange errors regarding Notification since Vaadin 24 and sadly the IT Tests of Notification aren't using the TestBench element, but instead are iterating over the notification-card manually instead of relying on the `_card` property. 

I'm currently seeing the following behavior: 

If I call 

```java
    assertThat($(NotificationElement.class).waitForFirst(10).getText())
      .isEqualTo("abc");
```

the test fails always. But if I do the following, it works:



```java
    // if I call it before.. the assertion afterwards works
    $(NotificationElement.class).waitForFirst(2).getText(); 
    
    assertThat($(NotificationElement.class).waitForFirst(10).getText())
      .isEqualTo("abc");
```

Could this be related to the rendering change on the client side and isn't notified by the current IT Tests because they doesn't use the `_card` property like the NotificationElement?


See: 

https://github.com/vaadin/flow-components/blob/fb3089cd8dbb1b7559fbcad0aaa9a4cec7bd6fc3/vaadin-notification-flow-parent/vaadin-notification-testbench/src/main/java/com/vaadin/flow/component/notification/testbench/NotificationElement.java#L28

https://github.com/vaadin/flow-components/blob/fb3089cd8dbb1b7559fbcad0aaa9a4cec7bd6fc3/vaadin-notification-flow-parent/vaadin-notification-flow-integration-tests/src/test/java/com/vaadin/flow/component/notification/tests/NotificationIT.java#L36

https://github.com/vaadin/flow-components/blob/fb3089cd8dbb1b7559fbcad0aaa9a4cec7bd6fc3/vaadin-notification-flow-parent/vaadin-notification-flow-integration-tests/src/test/java/com/vaadin/flow/component/notification/tests/NotificationIT.java#L73-L90

tomivirkki:
@knoobie I tried running the following test in `NotificationIT.java` and it passes fine:

```java
clickElementWithJs(findElement(By.id("component-notification-button")));

Assert.assertEquals("Hello, I am a notification with components!\nBye",
        $(NotificationElement.class).id("component-notification").getText());
```

If you can make an example that produces the bug, please file a new issue to this repo.

knoobie:
Thanks Tommi! I'm going to try to get a testcase that reproduces it

PR #4488, opened 2023-01-05
chore: update Testbench to 9.0.0.alpha6

ZheSun88:


web-padawan (commented):
LGTM, let's wait for #4487 to be merged first as restarting its build would take a while

ZheSun88:
i guess they both take some time.. as this is changing the root pom  :-)

web-padawan (approved):

PR #4517, opened 2023-01-12
refactor: extract common logic to HasValidationProperties

web-padawan:
## Description

1. Added `HasValidationProperties` interface which is based on [`HasValidation`](https://github.com/vaadin/flow/blob/fd04f545ee3f13b0cf3cdf2b694c6d9ab215c689/flow-server/src/main/java/com/vaadin/flow/component/HasValidation.java#L27) and provides same method signatures. 
The difference is that this interface actually sets / gets corresponding properties: `invalid` and `errorMessage`.
2. Updated all the components that provide these methods to use the new interface instead of `HasValidation`.

## Type of change

- Refactor

### DIFF
--- a/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/HasValidationProperties.java
+++ b/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/HasValidationProperties.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2000-2023 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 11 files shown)


vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/HasValidationProperties.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Gets the error message to show to the when the component is invalid.
+     *
+     * @return the error message or {@code null} if not set
+     */
+    @Override
+    default String getErrorMessage() {
+        return getElement().getProperty("errorMessage", "");
```
vursen, line 48: **note:** returning an empty string by default will be kind of a breaking change.

vursen: Once we agree on the proper return type here, I believe BasicValidation unit tests can be removed in favor of simple `instanceof` checks.

sissbruecker (approved):

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/HasValidationProperties.java
web-padawan: Restored the original behavior for now. We can consider changing this for all the properties later.

PR #4525, opened 2023-01-12
refactor: RichTextEditor value change ITs

sissbruecker:
## Description

Restructures the RTE value change ITs in order to have a cleaner test setup for the change of making the HTML format the default. The current ITs are split over multiple files, are mixed up with other tests, and are incomplete. This change introduces a new IT specifically for testing changes of `value`, `AsHtml` and `AsDelta`, and removes the other ITs.

## Type of change

- Refactoring


vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow-integration-tests/src/main/java/com/vaadin/flow/component/richtexteditor/tests/RichTextEditorSetHtmlValuePage.java
```diff
@@ -1,50 +0,0 @@
...
-import com.vaadin.flow.component.html.Div;
-import com.vaadin.flow.component.html.NativeButton;
-import com.vaadin.flow.component.richtexteditor.RichTextEditor;
-import com.vaadin.flow.data.value.ValueChangeMode;
-import com.vaadin.flow.router.Route;
-
-import java.util.stream.Stream;
-
-@Route(value = "vaadin-rich-text-editor/set-html-value")
-public class RichTextEditorSetHtmlValuePage extends Div {
```
sissbruecker, line 12: I'm not 100% sure about this test. It was introduced by the fix for a BFP (https://github.com/vaadin/vaadin-rich-text-editor/issues/175), where the problem was that setting a value via `AsHtml.setValue` does not trigger a change event via `RichTextEditor.addValueChangeListener`.

However I'm not sure why this is testing every single value change mode, since value change modes should only affect changes coming from the client. Decided to remove this, and instead added a test for setting an initial HTML value to the new `RichTextEditorValuePage`, to cover the issue from the BFP.

vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow-integration-tests/src/test/java/com/vaadin/flow/component/richtexteditor/tests/BasicUseIT.java
```diff
@@ -18,50 +18,6 @@ public void init() {
         String url = getBaseURL().replace(super.getBaseURL(),
                 super.getBaseURL() + "/vaadin-rich-text-editor");
         getDriver().get(url);
-        ButtonElement setValue = getTestButton("setValue");
-        setValue.click();
-    }
-
-    @Test
-    public void rteIsPresent() {
-        Assert.assertTrue($(RichTextEditorElement.class).exists());
```
sissbruecker, line 27: This test is not related to value changes, but it seems quite pointless.

vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow-integration-tests/src/test/java/com/vaadin/flow/component/richtexteditor/tests/RichTextEditorValueIT.java
```diff
@@ -0,0 +1,211 @@
...
+import com.vaadin.flow.testutil.TestPath;
+import com.vaadin.testbench.TestBenchElement;
+import com.vaadin.tests.AbstractComponentIT;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Ignore;
+import org.junit.Test;
+
+@TestPath("vaadin-rich-text-editor/value")
+public class RichTextEditorValueIT extends AbstractComponentIT {
```
sissbruecker, line 13: These ITs have value as there is some JS logic involved in synchronizing HTML value -> Delta value. In the upcoming refactoring, the sync will also work the other way around, which is why I added tests for all scenarios right away.

Note that some assertions and tests are disabled, as the current sync logic, as well as tracking the previous HTML value, is partially broken. I will enable those in the upcoming refactoring.

sissbruecker (commented):
Added some comments, please note that one of them is in a test file that was removed.

tomivirkki (approved):

Issue #4536, opened 2023-01-16
Remove deprecated API and Tools for V24


MarcinVaadin:
### Description

flow-components part of https://github.com/vaadin/flow/issues/15665

- [x] setDataProvider() is now deprecated (in Grid, Select, RadioButtonGroup, CheckboxGroup, ListBox and ComboBox)
  - [x] com.vaadin.flow.component.checkbox.CheckboxGroup.setDataProvider​(DataProvider<T, ?>)
  - [x] com.vaadin.flow.component.combobox.ComboBoxBase.setDataProvider​(DataProvider<TItem, String>)
  - [x]  com.vaadin.flow.component.grid.Grid.setDataProvider​(DataProvider<T, ?>)
  - [x]  com.vaadin.flow.component.listbox.ListBoxBase.setDataProvider​(DataProvider<ITEM, ?>)
  - [x]  com.vaadin.flow.component.radiobutton.RadioButtonGroup.setDataProvider​(DataProvider<T, ?>)
  - [x]  com.vaadin.flow.component.select.Select.setDataProvider​(DataProvider<T, ?>)
- [x] com.vaadin.flow.data.renderer.LitRenderer.getEventHandlers()
- [x]  com.vaadin.flow.data.renderer.LitRenderer.render​(Element, DataKeyMapper, Element)
- [x]  com.vaadin.flow.data.renderer.LitRenderer.withEventHandler​(String, SerializableConsumer)
- [x]  com.vaadin.flow.data.renderer.Renderer.getEventHandlers()
- [x]  com.vaadin.flow.data.renderer.Renderer.getValueProviders()
- [x]  com.vaadin.flow.data.renderer.Renderer.render​(Element, DataKeyMapper, Element)
- [x]  com.vaadin.flow.data.renderer.Renderer.setEventHandler​(String, SerializableConsumer)
- [x]  com.vaadin.flow.data.renderer.Renderer.setProperty​(String, ValueProvider<SOURCE, ?>)
- [x]  com.vaadin.flow.data.renderer.Rendering.getTemplateElement()
- [x] com.vaadin.flow.data.renderer.LocalDateRenderer​(ValueProvider<SOURCE, LocalDate>, DateTimeFormatter)
- [x] com.vaadin.flow.data.renderer.LocalDateTimeRenderer​(ValueProvider<SOURCE, LocalDateTime>, DateTimeFormatter)
- [x] com.vaadin.flow.data.renderer.Renderer​(String)

PR #4544, opened 2023-01-17
refactor!: removed non-serializable constructors from LocalDateRenderer

MarcinVaadin:
chore: removed non-serializable constructors from LocalDateRenderer and LocalDateTimeRenderer

Part of #4536


tomivirkki (approved):

tomivirkki:
Since this is API removal, should it be `refactor!: ...`?

PR #4547, opened 2023-01-17
fix: add dialog and notification children to host element

tomivirkki:
## Description

Fixes the regression discovered in https://github.com/vaadin/flow-components/pull/4538#discussion_r1072032428.

The issue occurred when scheduled JS executions would trigger for a virtual child component before it was actually added to the DOM. The PR fixes the issue by appending the child components as regular elements to the `<vaadin-notification>` and `<vaadin-dialog>` hosts. The script which teleports the child components inside the overlay still works as before.

## Type of change

Bugfix

### DIFF
--- a/vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
+++ b/vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
@@ -87,7 +87,6 @@ public class Dialog extends Component implements HasComponents, HasSize,
 
     private static final String OVERLAY_LOCATOR_JS = "this.$.overlay";
 
-    private Element container;
     private boolean autoAddedToTheUi;
     private int onCloseConfigured;
     private String width;
@@ -105,13 +104,6 @@ public class Dialog extends Component implements HasComponents, HasSize,
      * Creates an empty dialog.
      */
     public Dialog() {
-        container = new Element("div");
-        this.container.addAttachListener(event -> {
-            this.container.executeJs(
-                    "Vaadin.FlowComponentHost.patchVirtualContainer(this)");
-        });
-        getElement().appendVirtualChild(container);
-
         // Needs to be updated on each
         // attach, as element depends on node id which is subject to change if
         // the dialog is transferred to another UI, e.g. due to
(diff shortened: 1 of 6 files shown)


vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
```diff
@@ -1009,12 +1001,14 @@ private void initHeaderFooterRenderer() {
     }
 
     private void setDimension(String dimension, String value) {
-        getElement().executeJs("requestAnimationFrame(e => "
-                + OVERLAY_LOCATOR_JS + ".$.overlay.style[$0]=$1)", dimension,
-                value);
+        getElement().executeJs(OVERLAY_LOCATOR_JS + ".$.overlay.style[$0]=$1",
+                dimension, value);
```
tomivirkki, line 1005: Reverted this not so clean workaround from https://github.com/vaadin/flow-components/pull/4519

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
```diff
@@ -359,7 +351,7 @@ public void add(Component... components) {
         for (Component component : components) {
             Objects.requireNonNull(component,
                     "Component to add cannot be null");
-            container.appendChild(component.getElement());
+            getElement().appendChild(component.getElement());
```
sissbruecker, line 354: Since we now use `getElement()` instead of a separate container, duplicating logic from `HasComponents` seems no longer necessary.

The `remove`, `removeAll` and `getChildren` overrides can be removed.

`add` and `addComponentAtIndex` can be changed to this:
```java
    @Override
    public void add(Component... components) {
        HasComponents.super.add(components);

        updateVirtualChildNodeIds();
    }

    @Override
    public void addComponentAtIndex(int index, Component component) {
        HasComponents.super.addComponentAtIndex(index, component);

        updateVirtualChildNodeIds();
    }
```

The same should work for Notification.

sissbruecker (commented):
Did some manual testing with modifying the child list, could not find any issues 👍

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
tomivirkki: Done. Nice cleanup. Looks like the components have also been overriding the wrong `add` methods. Changed to override `public void add(Collection<Component> components)` and added tests.

sissbruecker (approved):

PR #2923, opened 2022-03-23
feat: Add HasStyle interface to Dialog #2856

krissvaa:
## Description

Partially implement 'HasStyles' interface for Dialog. 
Now class names can be added to dialog overlay.
Adding styles is not supported right now as logic will be a bit convoluted. 

Fixes #2856

## Type of change

- [ ] Bugfix
- [X] Feature

## Checklist

- [X] I have read the contribution guide: https://vaadin.com/docs-beta/latest/guide/contributing/overview/
- [X] I have added a description following the guideline.
- [X] The issue is created in the corresponding repository and I have referenced it.
- [X] I have added tests to ensure my change is effective and works as intended.
- [X] New and existing tests are passing locally with my change.
- [X] I have performed self-review and corrected misspellings.

PR #4549, opened 2023-01-17
refactor: override HasStyle methods in Dialog to use overlayClass

web-padawan:
## Description

Part of https://github.com/vaadin/platform/issues/3593

Related to https://github.com/vaadin/web-components/issues/5121

This PR changes the implementation done in https://github.com/vaadin/flow-components/pull/2923 (released in Vaadin 23.1): replace usage of `MutationObserver` for copying `class` attribute from the host with the newly added `overlayClass` property in the web component.

One notable change is the fact that user's class names set via [`HasStyle.getClassNames().set()`](https://github.com/vaadin/flow/blob/3946b30db5baca340d04d8868dd856e56c89cb38/flow-server/src/main/java/com/vaadin/flow/component/HasStyle.java#L78-L80) are no longer tracked and copied 😕 I'm not sure how to make it possible. Maybe we could mark this method as unsupported.

## Type of change

- Breaking change


sissbruecker:
Regarding getting notified about changes to `ClassList`, we could add a proxy to the original implementation:

<details>
<summary>Lots of code...</summary>

```java

    @Override
    public ClassList getClassNames() {
        ClassList classList = HasStyle.super.getClassNames();
        return new ClassList() {
            @Override
            public int size() {
                return classList.size();
            }

            @Override
            public boolean isEmpty() {
                return classList.isEmpty();
            }

            @Override
            public boolean contains(Object o) {
                return classList.contains(o);
            }

            @Override
            public Iterator<String> iterator() {
                return classList.iterator();
            }

            @Override
            public Object[] toArray() {
                return classList.toArray();
            }

            @Override
            public <T> T[] toArray(T[] a) {
                return classList.toArray(a);
            }

            @Override
            public boolean add(String s) {
                boolean result = classList.add(s);
                Dialog.this.updateOverlayClassName();
                return result;
            }

            @Override
            public boolean remove(Object o) {
                boolean result = classList.remove(o);
                Dialog.this.updateOverlayClassName();
                return result;
            }

            @Override
            public boolean containsAll(Collection<?> c) {
                return classList.containsAll(c);
            }

            @Override
            public boolean addAll(Collection<? extends String> c) {
                boolean result = classList.addAll(c);
                Dialog.this.updateOverlayClassName();
                return result;
            }

            @Override
            public boolean retainAll(Collection<?> c) {
                boolean result = classList.retainAll(c);
                Dialog.this.updateOverlayClassName();
                return result;
            }

            @Override
            public boolean removeAll(Collection<?> c) {
                boolean result = classList.removeAll(c);
                Dialog.this.updateOverlayClassName();
                return result;
            }

            @Override
            public void clear() {
                classList.clear();
                Dialog.this.updateOverlayClassName();
            }
        };
    }
```

</details>

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
```diff
@@ -1029,6 +1022,104 @@ private void attachComponentRenderer() {
...
+     *            the CSS class name to remove, not <code>null</code>
+     * @return <code>true</code> if the class name was removed,
+     *         <code>false</code> if the class list didn't contain the class
+     *         name
+     */
+    @Override
+    public boolean removeClassName(String className) {
+        boolean result = HasStyle.super.removeClassName(className);
+
+        if (result == true) {
```
sissbruecker, line 1055:
```suggestion
        if (result) {
```

Legioth:
There will still be a leaking abstraction even if `dialog.getClassNames()` is patched since the developer might just as well end up doing something along the lines of `dialog.getElement().setAttribute("class", "something")`. Doing that is still less likely than something like `dialog.getClassNames().clear()`.

From what I can see, there are two different "pure" solutions available:
* Deprecate everything related to `HasStyles` on `Dialog` and instead provide API that is explicitly targeting `overlayClass`. This would also address the other inconsistency in how the "class" part of `HasStyles` works different than its "style" part.
* Introduce a framework feature to fire events whenever the class list of a specific element is changed.

Both of those solutions are on the other hand challenging in other ways so there is no clear cut winner.

Implementing a custom `ClassList` proxy might be less code if it's based on `AbstractSet` instead of directly implementing all abstract methods from the `Set` interface. You can see an example of this in  `NodeList.SetView` that backs the regular `ClassList` implementation.

web-padawan:
> Implementing a custom ClassList proxy might be less code if it's based on AbstractSet instead of directly implementing all abstract methods from the Set interface. You can see an example of this in NodeList.SetView 

Does it mean we need to have something like `ClassListView` that would only implement methods that are supposed to call `updateOverlayClassName`? Sounds reasonable to me, I'll try to get it working based on that example.

web-padawan:
@Legioth Updated the PR with `OverlayClassListProxy` logic implemented by @sissbruecker based on your suggestion.
We might want to move this class to the `com.vaadin.flow.component.shared` package and add `Internal` to its name.

tomivirkki (approved):

mshabarov:
Possibility for developers to call `dialog.getElement().setAttribute("class", "something")` for a simple styling reminds me a recent discussion about recommending to developers always use component's high level API and think that they do something wrong if they use `getElement()`...
Just for making it clear: if someone sets the "class" attribute now when we have "overlayClass" in web component, would the styles just not apply to the Dialog's overlay? Or it will break the component? Does the "class" attribute even make any sense to the latest Dialog version? If we strictly don't want developers to set the "class" attribute, can we intercept it with also proxying the Element of Dialog and throw an exception?

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
```diff
@@ -981,6 +978,109 @@ private void attachComponentRenderer() {
...
+     * Adds a CSS class name to the dialog overlay element.
+     *
+     * @param className
+     *            the CSS class name to add, not <code>null</code>
+     */
+    @Override
+    public void addClassName(String className) {
+        HasStyle.super.addClassName(className);
+
+        updateOverlayClassName();
```
Legioth, line 995: Probably redundant to call these now when the backing collection already does the same

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
```diff
@@ -991,4 +1091,77 @@ public Style getStyle() {
...
+        }
+
+        private void updateOverlayClass() {
+            hasStyle.getElement().setProperty("overlayClass",
+                    hasStyle.getClassName());
+        }
+
+        @Override
+        public Iterator<String> iterator() {
+            return classList.iterator();
```
Legioth, line 1111: Removal through the iterator doesn't trigger `updateOverlayClass()` which might or might not be an acceptable compromise.

At the same time, implementing a custom iterator that handles removals in the "right" way would make it redundant to override most of the other methods in this class for any other purpose than to slightly optimize performance.

Legioth:
> If we strictly don't want developers to set the "class" attribute, can we intercept it with also proxying the Element of Dialog and throw an exception?

I don't think we have a good mechanism for doing that. Overriding `Dialog.getElement()` would probably be ugly but it would work to some extent. It would still not cover all potential cases since you can also get an element reference through other means.

web-padawan:
> Just for making it clear: if someone sets the "class" attribute now when we have "overlayClass" in web component, would the styles just not apply to the Dialog's overlay? Or it will break the component? Does the "class" attribute even make any sense to the latest Dialog version?

@mshabarov Thanks for good questions. The main idea behind the new approach is to remove `MutationObserver`, so obviously the `class` attribute changes done via `getElement()` will no longer apply to `vaadin-dialog-overlay`.

IMO we can accept this as a limitation - as you mentioned, there is a PR in Flow to mark `getElement()` as deprecated, because every time people use it, this means that there is a lack of proper Java API in the Flow component.

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
sissbruecker: Removed all unnecessary overrides.

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
```diff
@@ -991,4 +1006,60 @@ public Style getStyle() {
...
+            return new IteratorProxy(classList.iterator());
+        }
+
+        @Override
+        public int size() {
+            return classList.size();
+        }
+
+        @Override
+        public boolean add(String s) {
```
mshabarov, line 1035: Maybe I missed something, but don't we need `remove` and `clear` methods here still to update the overlay class? Or how the overlay class is updated if I call `dialog.getClassNames().remove(myStyleClass)`?

sissbruecker: This works due to how `AbstractSet` and `AbstractCollection` are implemented - they always use the iterator and its `remove` method for any removal operation. So just updating the overlay class name in the iterator proxy's `remove` method covers all these cases.

mshabarov: Thanks, now I see.

mshabarov (approved):

PR #4555, opened 2023-01-18
fix: select custom editor text after value is set

### LINKED ISSUE #4305: GridPro custom editor fields input selection behavior is inconsistent
### Description

Default editor fields in GridPro automatically select the contents of the editor input when you navigate between the cells (see https://vaadin.com/docs/latest/components/grid-pro). This makes sense, it's easy to start overriding the contents, you don't need to start erasing the previous values first. If you use a `.custom` editor (even with stock Vaadin components), the input selection occurs inconsistently - sometimes the input content is selected, other times it's not, even when navigating within a single GridPro.

### Expected outcome

The input selection should be consistent, and preferably it should happen always if the editor component supports it. Alternatively, there could be some kind of workaround that would ensure the contents of the input are selected when a cell is navigated into.

gridpro

```java
    public static class Bean {

        public Bean(double doubleProp, LocalDate localDateProp, BigDecimal bigDecimalProp) {
            this.doubleProp = doubleProp;
            this.localDateProp = localDateProp;
            this.bigDecimalProp = bigDecimalProp;
        }

        public double getDoubleProp() {
            return doubleProp;
        }

        public void setDoubleProp(double doubleProp) {
            this.doubleProp = doubleProp;
        }

        public LocalDate getLocalDateProp() {
            return localDateProp;
        }

        public void setLocalDateProp(LocalDate localDateProp) {
            this.localDateProp = localDateProp;
        }

        public BigDecimal getBigDecimalProp() {
            return bigDecimalProp;
        }

        public void setBigDecimalProp(BigDecimal bigDecimalProp) {
            this.bigDecimalProp = bigDecimalProp;
        }

        double doubleProp;
        LocalDate localDateProp;
        BigDecimal bigDecimalProp;
    }

      protected void initView() {
        GridPro<Bean> grid = new GridPro<>();
        grid.setEnterNextRow(true);
        List<Bean> beans = new ArrayList<>();
        for (int i = 0; i < 10; i++) {
            beans.add(new Bean(i, LocalDate.ofYearDay(2022, 1+i*20), new BigDecimal(i * 3)));
        }
        grid.addEditColumn(bean -> bean.getDoubleProp()).custom(new NumberField(),
                (bean, s) -> bean.setDoubleProp(s)).setHeader("prop1");
        grid.addEditColumn(bean -> bean.getLocalDateProp())
                .custom(new DatePicker(),
                (bean, s) -> bean.setLocalDateProp(s)).setHeader("prop2");
        Grid.Column<Bean> prop3 = grid.addEditColumn(
                bean -> bean.getBigDecimalProp()).custom(new BigDecimalField(),
                (bean, s) -> bean.setBigDecimalProp(s)).setHeader("prop3");
        grid.setItems(beans);
        add(grid);
 
    }
```

### Steps to reproduce

1. Add the snippet into a view
2. Call the initView method
3. Open the browser
4. Navigate between GridPro cells with tab and enter

### Environment

Vaadin version(s): 23.2.10
OS: Windows 10

### Browsers

Chrome

DiegoCardoso:
## Description

When the user enters a cell with a custom editor, the GridPro calls the `setValue()` method, which makes the field loses the selected text that is performed by the web component. 

This change adds a method to `gridProConnector` that does the same logic present on the web component to select the text on the opened editor _only for the columns with a custom editor_.

Fixes #4305

## Type of change

- [X] Bugfix
- [ ] Feature

### DIFF
--- a/vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow-integration-tests/src/main/java/com/vaadin/flow/component/gridpro/tests/MainView.java
+++ b/vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow-integration-tests/src/main/java/com/vaadin/flow/component/gridpro/tests/MainView.java
@@ -7,6 +7,7 @@ import java.util.Random;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
+import com.vaadin.flow.component.UI;
 import com.vaadin.flow.component.button.Button;
 import com.vaadin.flow.component.combobox.ComboBox;
 import com.vaadin.flow.component.gridpro.GridPro;
@@ -34,6 +35,13 @@ public class MainView extends VerticalLayout {
 
         Div eventsPanel = new Div();
         eventsPanel.setId("events-panel");
+
+        Div selectionPanel = new Div();
+        selectionPanel.setId("selection-panel");
+
+        UI.getCurrent().getElement().executeJs(
+                "document.addEventListener('selectionchange', () => $0.textContent = document.getSelection().toString())",
+                selectionPanel.getElement());
 
         GridPro<Person> grid = new GridPro<>();
         Button disableGrid = new Button("Disable Grid");
@@ -112,7 +120,7 @@ public class MainView extends VerticalLayout {
         disableGrid.addClickListener(click -> grid.setEnabled(false));
 
         add(grid, itemDisplayPanel, subPropertyDisplayPanel, eventsPanel,
-                disableGrid);
+                selectionPanel, disableGrid);
     }
 
     protected void createBeanGridWithEditColumns() {
--- a/vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow-integration-tests/src/test/java/com/vaadin/flow/component/gridpro/tests/BasicIT.java
+++ b/vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow-integration-tests/src/test/java/com/vaadin/flow/component/gridpro/tests/BasicIT.java
@@ -9,6 +9,7 @@ import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.openqa.selenium.Keys;
+import org.w3c.dom.Document;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -256,6 +257,24 @@ public class BasicIT extends AbstractParallelTest {
         TestBenchElement textField = cell.$("vaadin-text-field").first();
         // should have converted integer model value into string editor value
         Assert.assertEquals("2019", textField.getProperty("value"));
+    }
+
+    @Test
+    public void gridWithCustomEditors_navigateToWithTabKey_textIsSelected() {
+        var cell_0_4 = grid.getCell(0, 4);
+        assertCellEnterEditModeOnDoubleClick(0, 4, "vaadin-combo-box");
+
+        waitUntil(e -> cell_0_4.$("vaadin-combo-box").exists());
+        var input = cell_0_4.$("vaadin-combo-box").first();
+        input.sendKeys(Keys.TAB);
+
+        Assert.assertEquals("person1@vaadin.com",
+                getPanelText("selection-panel"));
+
+        var cell_0_5 = grid.getCell(0, 5);
+        input = cell_0_5.$("input").first();
+        input.sendKeys(Keys.TAB);
+        Assert.assertEquals("2019", getPanelText("selection-panel"));
     }
 
     private void assertCellEnterEditModeOnDoubleClick(Integer rowIndex,
--- a/vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/resources/META-INF/resources/frontend/gridProConnector.js
+++ b/vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/resources/META-INF/resources/frontend/gridProConnector.js
@@ -17,6 +17,13 @@
   }
 
   window.Vaadin.Flow.gridProConnector = {
+    focusCustomEditor: (editor) => {
+      if (editor instanceof HTMLInputElement) {
+        editor.select();
+      } else if (editor.focusElement && editor.focusElement instanceof HTMLInputElement) {
+        editor.focusElement.select();
+      }
+    },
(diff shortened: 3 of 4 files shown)


vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow-integration-tests/src/main/java/com/vaadin/flow/component/gridpro/tests/MainView.java
```diff
@@ -35,6 +36,13 @@ protected void createEditorColumns() {
         Div eventsPanel = new Div();
         eventsPanel.setId("events-panel");
 
+        Div selectionPanel = new Div();
+        selectionPanel.setId("selection-panel");
+
+        UI.getCurrent().getElement().executeJs(
+                "document.addEventListener('selectionchange', () => $0.textContent = document.getSelection().toString())",
+                selectionPanel.getElement());
```
DiegoCardoso, line 44: Tried to check the selection from `document.getSelection().toString()` on the IT with the `executeScript` method, but it was always returning `null`, so I came up with this solution instead.

vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/resources/META-INF/resources/frontend/gridProConnector.js
```diff
@@ -17,6 +17,13 @@
   }
 
   window.Vaadin.Flow.gridProConnector = {
+    focusCustomEditor: (editor) => {
+      if (editor instanceof HTMLInputElement) {
+        editor.select();
+      } else if (editor.focusElement && editor.focusElement instanceof HTMLInputElement) {
+        editor.focusElement.select();
+      }
+    },
```
DiegoCardoso, line 26: I tried to add a `value-changed` listener to the component so that it would perform this task, but that won't work for native fields. Adding a `change` listener doesn't work either, since the value is changed programmatically and the event is not fired. So, instead, the solution was to explicitly call the method from GridPro after the value is set.

sissbruecker: I think doing this explicitly after setting the value makes a lot of sense. Using the `value-changed` event also sounds like that could interfere with typing.

vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow-integration-tests/src/main/java/com/vaadin/flow/component/gridpro/tests/MainView.java
sissbruecker: Did you maybe forget to add a return?

The following IT works fine for me:
```java
    @Test
    public void gridWithCustomEditors_navigateToWithTabKey_textIsSelected() {
        var cell_0_4 = grid.getCell(0, 4);
        assertCellEnterEditModeOnDoubleClick(0, 4, "vaadin-combo-box");

        waitUntil(e -> cell_0_4.$("vaadin-combo-box").exists());
        var input = cell_0_4.$("vaadin-combo-box").first();
        input.sendKeys(Keys.TAB);

        String selection = (String) getCommandExecutor().executeScript("return document.getSelection().toString()");
        Assert.assertEquals("person1@vaadin.com", selection);

        var cell_0_5 = grid.getCell(0, 5);
        input = cell_0_5.$("input").first();
        input.sendKeys(Keys.TAB);

        selection = (String) getCommandExecutor().executeScript("return document.getSelection().toString()");
        Assert.assertEquals("2019", selection);
    }
```

vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow-integration-tests/src/test/java/com/vaadin/flow/component/gridpro/tests/BasicIT.java
```diff
@@ -9,6 +9,7 @@
 import org.junit.Before;
 import org.junit.Test;
 import org.openqa.selenium.Keys;
+import org.w3c.dom.Document;
```
tomivirkki, line 12: nit: unused import

DiegoCardoso: Thanks. Removed.

vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/resources/META-INF/resources/frontend/gridProConnector.js
```diff
@@ -17,6 +17,13 @@
   }
 
   window.Vaadin.Flow.gridProConnector = {
+    focusCustomEditor: (editor) => {
```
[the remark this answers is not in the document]
DiegoCardoso: I wasn't sure about the name, so I got "inspired" by the name of the [method this behavior was taken from](https://github.com/vaadin/web-components/blob/2a51598e6db4bd4fa441fbe3b26030f82f2c3242/packages/grid-pro/src/vaadin-grid-pro-edit-column.js#L198-L207):

```js
  _focusEditor(editor) {
    editor.focus();
    if (this.editorType === 'checkbox') {
      editor.setAttribute('focus-ring', '');
    } else if (editor instanceof HTMLInputElement) {
      editor.select();
    } else if (editor.focusElement && editor.focusElement instanceof HTMLInputElement) {
      editor.focusElement.select();
    }
  }
```

Changed to `selectAll(editor)` since it makes more sense.

vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow-integration-tests/src/main/java/com/vaadin/flow/component/gridpro/tests/MainView.java
DiegoCardoso: Oh, yeah. Forgot to add `return` on the JS expression. Now it works, thanks!

sissbruecker (approved):

PR #4560, opened 2023-01-19
chore: remove legacy template adapter imports

tomivirkki:
## Description

Since the Flow components no longer use a `<template>` element for rendering, remove the unnecessary import for `@vaadin/polymer-legacy-adapter/template-renderer.js` and the `suppress-template-warning` attribute.

## Type of change

- Cleanup


web-padawan (approved):
LGTM. Consider also removing `style-modules.js` in a separate PR.

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/frontend/src/grid-on-client-and-slot.js
```diff
@@ -17,13 +17,23 @@ class GridOnClientAndSlot extends LitElement {
 
   render() {
     return html`
-      <vaadin-grid id="tree">
-        <vaadin-grid-column>
-          <template>
-            <vaadin-grid-tree-toggle leaf="[[item.leaf]]" expanded="{{expanded}}" level="[[level]]">
-              <div class="tree-cell">[[item.name]]</div>
-            </vaadin-grid-tree-toggle>
-          </template>
```
tomivirkki, line 26: @web-padawan I needed to change a couple of tests that were still using the `<template>` API to use `renderer` instead

PR #4569, opened 2023-01-20
fix: Make it possible to remove all features from a VectorSource

Artur-:


vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/source/VectorSource.java
```diff
@@ -55,6 +55,12 @@ public void removeFeature(Feature feature) {
         removeChild(feature);
     }
 
+    public void removeAllFeatures() {
```
sissbruecker, line 58: As you usually interact with the `FeatureLayer` of the map through `map.getFeatureLayer()` to add / remove features, should we add this method there as well?

Artur-: Yes, added there too

sissbruecker (approved):

PR #4576, opened 2023-01-20
refactor: override HasStyle implementation in ConfirmDialog

web-padawan:
## Description

Same as #4549 but for `ConfirmDialog`.

Based on #4573

## Type of change

- Refactor


vaadin-confirm-dialog-flow-parent/vaadin-confirm-dialog-flow-integration-tests/src/main/java/com/vaadin/flow/component/confirmdialog/tests/StylingPage.java
```diff
@@ -21,8 +21,7 @@ public StylingPage() {
         addClassNameFoo.setId("add-foo");
 
         Button setClassNameBar = new Button("Set class bar", e -> {
-            dialog.setClassName("foo bar");
-            dialog.getClassNames().set("foo", false);
+            dialog.setClassName("bar");
```
web-padawan, line 24: Looks like the case in this IT isn't covered by the logic of `InternalOverlayClassListProxy` 😕 
However, this approach looks like an edge case to me, so we can probably accept it.

tomivirkki: This could be fixed by implementing `setClassName` to support multiple class names split by whitespace. Something like:

```java
public void setClassName(String className) {
    getClassNames().clear();
    if (className != null) {
        getClassNames().addAll(Arrays.asList(className.split(" ")));
    }
}
```

web-padawan: Thanks, done.

tomivirkki: Makes sense to add the same change to `Dialog` also.

vaadin-confirm-dialog-flow-parent/vaadin-confirm-dialog-flow/src/main/java/com/vaadin/flow/component/confirmdialog/ConfirmDialog.java
```diff
@@ -29,7 +28,9 @@
 import com.vaadin.flow.component.UI;
 import com.vaadin.flow.component.dependency.JsModule;
 import com.vaadin.flow.component.dependency.NpmPackage;
+import com.vaadin.flow.component.shared.InternalOverlayClassListProxy;
```
knoobie, line 31: Would it make sense to use the same package semantic flow is using? "shared.internal.ClassName" so that don't even see this class / interace when they navigate to the package? This could also allow to drop internal from the name

web-padawan: Thanks, makes sense 👍  I will do that in a separate PR after this one lands.

vaadin-confirm-dialog-flow-parent/vaadin-confirm-dialog-flow-integration-tests/src/main/java/com/vaadin/flow/component/confirmdialog/tests/StylingPage.java
web-padawan: Thanks, done.

tomivirkki (approved):

vaadin-confirm-dialog-flow-parent/vaadin-confirm-dialog-flow/src/main/java/com/vaadin/flow/component/confirmdialog/ConfirmDialog.java
web-padawan: Created #4581 to move the class as you suggested.

PR #4598, opened 2023-01-25
refactor!: rename InternalFieldBase to TextFieldBase

vursen:
## Description

It was decided internally to rename `InternalFieldBase` to `TextFieldBase` in light of the plans to introduce a top-level 
 `FieldBase` class that will contain common functionality to field components in general, not only text fields.

## Type of change

- [x] Breaking change


web-padawan (commented):
Consider moving the class to `com.vaadin.textfield.internal` for now to indicate that it isn't public API, similarly to what we did for `OverlayClassListProxy`.

web-padawan (changes requested):
Looks like you included a Git submodule for ComboBox by accident?

vursen:
> Looks like you included a Git submodule for ComboBox by accident?

Yeah, that was an accident.

> Consider moving the class to com.vaadin.textfield.internal for now to indicate that it isn't public API, similarly to what we did for OverlayClassListProxy.

Good point, done.

web-padawan:
> Consider moving the class to com.vaadin.textfield.internal for now

As discussed in the daily, let's not make this change and keep the class where it previously was.
The name `TextFieldBase` is quite well aligned with `ComboBoxBase` and is not problematic.

tomivirkki (approved):

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/TextFieldBase.java
```diff
@@ -45,7 +45,7 @@
  * @param <TValue>
  *            Type of the value of the extending component
  */
-public abstract class InternalFieldBase<TComponent extends InternalFieldBase<TComponent, TValue>, TValue>
+public abstract class TextFieldBase<TComponent extends TextFieldBase<TComponent, TValue>, TValue>
```
sissbruecker, line 48: `TextFieldBase` could now be made package-private if we don't intend to expose it.

sissbruecker: Well that might be a problem if you want to take a collection of fields, and do something generic with them, like generating placeholders. You would not be able to cast to `TextFieldBase` then.

vursen: Well, turns out this class isn't really internal considering the use-case above.

sissbruecker: Yes, but it's probably fine as is. This is definitely public API, because developers are going to cast to `TextFieldBase`, and the methods in it also need to be stable. However making the constructor package-private is good, that gives us some guarantee that the implementation isn't used in custom classes.

web-padawan (approved):

PR #4602, opened 2023-01-26
refactor: add check for delta values

sissbruecker:
## Description

Adds a check to `RichTextEditor.setValue` that prevents developers from setting values in the Delta format. If the check detects a possible Delta value, an `IllegalArgumentException` is thrown. This is intended to make sure developers don't accidentally pass a Delta value after migrating to v24, which could lead to data corruption. This also un-deprecates the `AsHtml` wrapper in order to keep the legacy behavior, which allows setting HTML values starting with either `[` or `{`.

## Type of change

- Refactoring

### DIFF
--- a/vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java
+++ b/vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java
@@ -169,6 +169,18 @@ public class RichTextEditor
      * <p>
      * Note: {@link Binder} will take care of the {@code null} conversion when
      * integrates with the editor, as long as no new converter is defined.
+     * <p>
+     * Since v24, this method only accepts values in the HTML format, whereas in
+     * v23 and earlier this method would accept values in the Delta format. In
+     * order to prevent data corruption, passing a value that starts with either
+     * <code>[</code> or <code>{</code> will now throw an
+     * {@link IllegalArgumentException}, as it might indicate that the value is
+     * in the Delta format. In order to keep using the Delta format, use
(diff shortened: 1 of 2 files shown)


vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java
```diff
@@ -169,6 +169,18 @@ public void setValueChangeMode(ValueChangeMode valueChangeMode) {
      * <p>
      * Note: {@link Binder} will take care of the {@code null} conversion when
      * integrates with the editor, as long as no new converter is defined.
+     * <p>
+     * Since v24, this method only accepts values in the HTML format, whereas in
+     * v23 and earlier this method would accept values in the Delta format. In
+     * order to prevent data corruption, passing a value that starts with either
+     * <code>[</code> or <code>{</code> will now throw an
+     * {@link IllegalArgumentException}, as it might indicate that the value is
```
tomivirkki, line 177: I wonder if this is too sneaky. If someone doesn't notice this comment, the exception might occur only until in production.

tomivirkki (approved):

vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java
sissbruecker: There was a long discussion about this solution in this [internal Slack thread](https://vaadin.slack.com/archives/C3TGRP4HY/p1674637523802949). Some alternatives, like changing the API so that it wouldn't compile anymore, were considered, but we ultimately decided to go with the exception.

Issue #4278, opened 2022-12-01
[Map] Investigate stylable marker labels implementation


sissbruecker:
### Describe your motivation

In order to introduce stylable labels for map markers we should do some initial research:
- Verify API design proposed in the [acceptance criteria](https://github.com/vaadin/platform/issues/3533) and a [contribution PR](https://github.com/vaadin/flow-components/pull/3450). Should we adopt the low-level OpenLayers API wholesale, or can we add any convenience APIs?
- Should the [`text` instance](https://openlayers.org/en/latest/apidoc/module-ol_style_Style-Style.html#setText) of a marker's style be initialized by default, or only when the developer sets a label? Are there any downsides of initializing by default, is there a performance impact?
- How can we allow customization of the default styles of the labels?
	- Does OpenLayers provide a mechanic that can be used here, or do we need to implement this on our own?
	- Prototype possible approaches on how the `text` instance of a marker's style can be initialized with custom settings.
- What the default styles should be, so that they are readable with various map types. Should involve getting feedback from a designer.
- Verify performance does not degrade when not using labels, and is adequate when applying a reasonable amount of labels.

#### Related

- [Acceptance criteria](https://github.com/vaadin/platform/issues/3533)
- [Contribution PR](https://github.com/vaadin/flow-components/pull/3450)

sissbruecker:
> Verify proposed API design

- Adopting the low-level OpenLayers API (adding Java equivalents for [ol/style/Text](https://openlayers.org/en/v6.15.1/apidoc/module-ol_style_Text-Text.html) and [ol/style/Style#setText](https://openlayers.org/en/v6.15.1/apidoc/module-ol_style_Style-Style.html#setText)) makes sense to align with the existing component implementation, and provide low-level access if we decide to not add convenience APIs immediately
- The proposed `MarkerFeature.setLabel(String)` convenience API to set the label string makes sense
- Storing the label string together with the label styles, as OpenLayers does, is problematic:
	- it doesn't allow for reuse of label style instances, which increases memory usage and the synchronization payload
	- it's likely that other features, such as default label styles per map, would be more difficult to implement
- I would suggest to deviate from the OpenLayers design here, and decouple the label string from the label styles. Store the label string on the map `Feature` instead, and make sure the label is applied to the respective label style before rendering a marker.
- As discoverability of methods like `marker.getStyle.getText().setFont(String)` is bad, additionally `MarkerFeature` should have convenience methods for setting common label style properties
[...]

Issue #2898, opened 2022-03-18
[map] add styleable label to MarkerFeature


jgoderbauer:
### Describe your motivation

When there are several markers at the same time visible, it is hard for the end user to see what a marker is actually representing in the real world. 
While this can be improved with custom icons, its not enough if there are several markers closeby.

### Describe the solution you'd like

It should be possible to add a label to a marker.

Here is screenshot of how this could look alike:
![image](https://user-images.githubusercontent.com/54622999/160557770-b9f98e8b-8386-4f50-95ea-e1a6532507d8.png)
In this example there is a marker (blue dot) and above a label that provides information for that given marker.

The label also should be styleable via the flow api (provide information how to to style the label regards to the font, color, size, hover-color and so on and the relativ position to the marker like top, bottom, etc.). This is necessery to keep a consitant look of the marker and it's label.

What may be a challenge is the position (x,y, and z) of the label, especially when several labels are overlapping.

### Describe alternatives you've considered

Maybe it would be better if the solution is: create and add a Label to _any_ Feature (via the flow api).

### Additional context

Example of this may could be done in OpenLayers (but maybe there is a better way?)

```javascript
var iconFeature = new Feature({
  geometry: new Point([0, 0]),
  name: 'Water',
});

var vectorSource = new VectorSource({
  features: [iconFeature],
});
 vectorSource.on('addfeature', function() {
  map.getView().fit(vectorSource.getExtent());
}); 

var iconStyle = new ol.style.Style({
        image:  new ol.style.Circle({
            radius: 6,
            stroke: new ol.style.Stroke({
                color: '#fff'
            }),
            fill: new ol.style.Fill({
                color: '#3399CC'
            })
        })
    });
var labelStyle = new ol.style.Style({
  text: new ol.style.Text({
    font: '12px Calibri,sans-serif',
    overflow: true,
    fill: new ol.style.Fill({
      color: '#000'
    }),
    stroke: new ol.style.Stroke({
      color: '#fff',
      width: 3
    }),
    offsetY: -12
  })
});
var style = [iconStyle, labelStyle];
var vectorLayer = new VectorLayer({
  source: vectorSource,
  style: function(feature) {
        labelStyle.getText().setText(feature.get('name'));
        return style;
      }
});

var map = new Map( {
        target: 'map',
        layers: [
          new ol.layer.Tile({
            source: new ol.source.OSM()
          }),
          vectorLayer
        ],
        view: new ol.View({
          center: [0, 0],
[...]

PR #4632, opened 2023-02-01
feat: set labels for map markers

sissbruecker:
## Description

Allows to set and remove text labels for map markers. Labels are rendered with a default style for now, I'll add customization of the marker's label style in a follow-up PR.

Part of: https://github.com/vaadin/flow-components/issues/2898
AC: https://github.com/vaadin/platform/issues/3533
Research: https://github.com/vaadin/flow-components/issues/4278#issuecomment-1411810550

## Type of change

- Feature

### DIFF
--- a/vaadin-map-flow-parent/vaadin-map-flow/src/main/resources/META-INF/resources/frontend/vaadin-map/synchronization/index.js
+++ b/vaadin-map-flow-parent/vaadin-map-flow/src/main/resources/META-INF/resources/frontend/vaadin-map/synchronization/index.js
@@ -9,6 +9,9 @@
  */
 import Feature from 'ol/Feature';
 import Point from 'ol/geom/Point';
+import Fill from 'ol/style/Fill';
+import Stroke from 'ol/style/Stroke';
+import Text from 'ol/style/Text';
 import View from 'ol/View';
 import { synchronizeImageLayer, synchronizeTileLayer, synchronizeVectorLayer } from './layers.js';
 import {
@@ -20,6 +23,16 @@ import {
 } from './sources.js';
 import { synchronizeIcon, synchronizeFill, synchronizeStroke, synchronizeStyle } from './styles.js';
 import { convertToCoordinateArray, synchronizeCollection } from './util.js';
+
+/**
+ * Fallback text style to use for features that don't have a custom one
+ */
+const fallbackTextStyle = new Text({
+  font: '13px sans-serif',
+  offsetY: 10,
+  fill: new Fill({ color: '#333' }),
+  stroke: new Stroke({ color: '#fff', width: 3 })
+});
 
 function synchronizeMap(target, source, context) {
   if (!target) {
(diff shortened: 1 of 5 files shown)


vaadin-map-flow-parent/vaadin-map-flow/src/main/resources/META-INF/resources/frontend/vaadin-map/synchronization/index.js
```diff
@@ -62,7 +75,31 @@ function synchronizeFeature(target, source, context) {
...
+      style.setText(fallbackTextStyle);
+    }
+    // Set the feature's label on the text style. This is safe even when using
+    // the default text style instance, as for each feature using the default
+    // text style, this function will be called again before rendering.
+    if (style.getText()) {
+      style.getText().setText(source.label);
+    }
+    return style;
+  });
```
sissbruecker, line 102: This is more involved than it needs to be at the moment, but it is a starting point for adding customization of the label's styles.

vaadin-map-flow-parent/vaadin-map-testbench/src/main/java/com/vaadin/flow/component/map/testbench/MapElement.java
```diff
@@ -429,7 +429,7 @@ public GeometryReference getGeometry() {
         }
 
         public StyleReference getStyle() {
-            return new StyleReference(executor, path("getStyle()"));
+            return new StyleReference(executor, path("getStyle()()"));
```
sissbruecker, line 432: `getStyle` returns a style function now.

DiegoCardoso (commented):
The change looks good, but I shared one clarification/change request

sissbruecker:
@DiegoCardoso 
> The change looks good, but I shared one clarification/change request

Where? 😅

vaadin-map-flow-parent/vaadin-map-flow/src/main/resources/META-INF/resources/frontend/vaadin-map/synchronization/index.js
```diff
@@ -21,6 +24,16 @@ import {
...
 
+/**
+ * Fallback text style to use for features that don't have a custom one
+ */
+const fallbackTextStyle = new Text({
+  font: '13px sans-serif',
+  offsetY: 10,
+  fill: new Fill({ color: '#333' }),
+  stroke: new Stroke({ color: '#fff', width: 3 })
+});
```
DiegoCardoso, line 35: I got a bit stuck here trying to understand one thing about this shared `Text` object and would like to share my understanding to check whether I got it right.

For all `Feature` objects that don't have a custom text style, we are setting `fallbackTextStyle`. So, all objects will share the same instance, which means that when we put the actual text on the line `#99`, we are changing the same object. From what I could test, it works with multiple labels because:
1. When a `Feature` have its label changed, OL will call `target.getStyle()` for all the `Feature` objects
2. Immediately after getting the `Style` object, OL will proceed to render the feature and its label and, at some point, it copies the text string to another object it creates.

So, although we at the end are changing the text on the same object, that doesn't seem to be an issue here.

----

If I got everything correctly, I would propose to improve the IT a little bit by adding more than one `Feature` and making changes on theirs label in different orders, just to make sure that no undesired behavior happens.

DiegoCardoso (commented):
I don't know what happened with my review... I had to write it again, but here you go!

vaadin-map-flow-parent/vaadin-map-flow/src/main/resources/META-INF/resources/frontend/vaadin-map/synchronization/index.js
sissbruecker: Yep, that is correct. It's a common pattern in OpenLayers, for example in this demo: https://openlayers.org/en/latest/examples/street-labels.html. My assumption is that for each feature, OL calls the style function, uses the returned style to immediately render that feature on the canvas, and then discard it.

Adding an IT for multiple labels make senses. Changing labels in different orders probably doesn't make a difference due to how the sync works, which always collects children in the same order, so I would skip that.

sissbruecker: Changed the IT to cover multiple features / labels.

DiegoCardoso (approved):

PR #4634, opened 2023-02-02
fix: use the same cell reference constructor in order to ensure consistency

### LINKED ISSUE #4588: Spreadsheet does not trigger addCellValueChangeListener for anything else than text input
### Description

In order to sync a spreadsheet cell value with an external component, I do
```
        spreadsheet.addCellValueChangeListener(event -> {
            CellReference a2 = new CellReference(spreadsheet.getCell("A2"));
            CellReference d4 = new CellReference(spreadsheet.getCell("D4"));
            if (event.getChangedCells().contains(a2) || event.getChangedCells().contains(d4)) {
                updateInvoiceNumberAndSource();
            }

        });
```

This works fine when typing text but when doing undo, redo or pasting a value, the event is not triggered

### Expected outcome

A CellValueChangeListener is triggered whenever the Cell Value Changes

Above

### Steps to reproduce

1. Paste or do undo/reado

### Environment

Vaadin version(s): 24.0.0.alpha9
OS: mac

### Browsers

Chrome

ugur-vaadin:
## Description

In Apache POI, `CellReference` has multiple constructors. In the version that Vaadin V8 used, the constructor `CellReference(Cell cell)` did not use the sheet name from the `Cell` object. However, while having the same signature, this constructor started using the sheet name in the new versions. This creates inconsistencies when checking equality for `CellReference`s since we use different constructors in various locations. 

While this check can be easily be done in other ways, if we want the equality check to hold regardless, we can use the constructors that use the sheet name everywhere. This PR makes this change and adds the sheet names to the constructors wherever necessary. 

One drawback to this is adding sheet name in cases where we do not really need them in the sake of consistency. Another is having to add two booleans repeatedly just to be able to use the constructors with the sheet name. Also, this might possibly break some implementations if users are depending on some exact check on the references.

Fixes #4588

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/SpreadsheetHandlerImpl.java
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/SpreadsheetHandlerImpl.java
@@ -334,7 +334,9 @@ public class SpreadsheetHandlerImpl implements SpreadsheetServerRpc {
         for (int x = region.getFirstColumn(); x <= region
                 .getLastColumn(); x++) {
             for (int y = region.getFirstRow(); y <= region.getLastRow(); y++) {
-                cells.add(new CellReference(y, x));
+                cells.add(new CellReference(
+                        spreadsheet.getActiveSheet().getSheetName(), y, x,
+                        false, false));
             }
         }
         fireCellValueChangeEvent(cells);
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/command/CellValueCommand.java
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/command/CellValueCommand.java
@@ -148,7 +148,8 @@ public class CellValueCommand extends SpreadsheetCommand
 
     @Override
     public CellReference getSelectedCellReference() {
-        return new CellReference(selectedCellRow, selectedcellCol);
+        return new CellReference(spreadsheet.getActiveSheet().getSheetName(),
+                selectedCellRow, selectedcellCol, false, false);
     }
 
     @Override
@@ -339,13 +340,15 @@ public class CellValueCommand extends SpreadsheetCommand
         for (Object o : values) {
             if (o instanceof CellValue) {
                 CellValue cellValue = (CellValue) o;
-                changedCells
-                        .add(new CellReference(cellValue.row, cellValue.col));
+                changedCells.add(new CellReference(
+                        spreadsheet.getActiveSheet().getSheetName(),
+                        cellValue.row, cellValue.col, false, false));
             } else {
                 CellRangeValue cellRangeValue = (CellRangeValue) o;
                 for (int r = cellRangeValue.row1; r <= cellRangeValue.row2; r++) {
                     for (int c = cellRangeValue.col1; c <= cellRangeValue.col2; c++) {
-                        changedCells.add(new CellReference(r, c));
+                        changedCells.add(new CellReference(
+                                getSheet().getSheetName(), r, c, false, false));
                     }
                 }
             }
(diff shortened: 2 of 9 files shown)


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/command/CellValueCommand.java
```diff
@@ -148,7 +148,8 @@ public void captureCellRangeValues(CellRangeAddress... cellRanges) {
 
     @Override
     public CellReference getSelectedCellReference() {
-        return new CellReference(selectedCellRow, selectedcellCol);
+        return new CellReference(spreadsheet.getActiveSheet().getSheetName(),
```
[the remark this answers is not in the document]
ugur-vaadin: Addressed in the latest version

vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/SpreadsheetHandlerImpl.java
```diff
@@ -334,7 +334,9 @@ private void fireCellValueChangeEvent(CellRangeAddress region) {
         for (int x = region.getFirstColumn(); x <= region
                 .getLastColumn(); x++) {
             for (int y = region.getFirstRow(); y <= region.getLastRow(); y++) {
-                cells.add(new CellReference(y, x));
+                cells.add(new CellReference(
```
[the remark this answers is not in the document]
ugur-vaadin: Addressed in the latest version

rolfsmeds:
As we are not aware of any users having problems due to this issue, perhaps it would be better to wait until the next major as it will introduce some breaking changes?

yuriy-fix:
Added to the v25 breaking changes list to be revisited before v25.

vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/CellSet.java
DiegoCardoso, line 1: Would it make sense to make `CellSet` extend `HashSet` and override the `contains` method to apply the custom logic done here? This would make it a bit less of a breaking change, I suppose, as the returning types could still be left as `Set`. If `CellSet` become unused externally, it could even be made package private.

The `contains` method would become something like this:
```java
    @Override
    public boolean contains(Object object) {
        if (this.isEmpty()) {
            return false;
        }
        if (object instanceof CellReference cellReference) {
            if (cellReference.getSheetName() == null) {
                CellReference cellWithSheetName = new CellReference(
                        this.iterator().next().getSheetName(),
                            cellReference.getRow(), cellReference.getCol(),
                    cellReference.isRowAbsolute(),
                    cellReference.isColAbsolute());
                return super.contains(cellWithSheetName);
            }
        }
        return super.contains(object);
```

ugur-vaadin: Updated the PR so that the `CellSet` extends `HashSet`. Even though it is still `public`, I also added it in the JavaDoc that it is internal.

tomivirkki: Messing with the expected behavior of `Set.contains` feels dangeorus.

Screenshot 2025-11-10 at 15 46 29

Also other `Set` API isn't aligned:
```java
var a2Cell = new CellReference(1, 0);
var changedCells = event.getChangedCells();
if (changedCells.contains(a2Cell)) {
    // Doesn't actually get removed
    changedCells.remove(a2Cell);
}
```

ugur-vaadin: Updated the implementation. We now throw an `IllegalArgumentException` if contains is called with a `CellReference` that does not contain a sheet name. Otherwise, the regular `HashSet` behaviour is now intact. I also added this the the JavaDocs for `CellSet` and the events that use `CellSet`.

tomivirkki (approved):

PR #4647, opened 2023-02-07
feat: customize marker label styles

sissbruecker:
## Description

Allow customization of marker label styles by adding support for setting and synchronizing a custom `Text` style instance ([OL API docs](https://openlayers.org/en/v6.15.1/apidoc/module-ol_style_Text-Text.html)) for a feature's`Style`. 

This PR only includes basic support for synchronizing `Text` style instances, and there are  currently no convenience methods for setting the label styles. I plan to add convienence methods to `Feature` in a separate PR that would allow to customize label styles without having to go through `Feature -> Style -> Text`, and would also create a style / text instances on demand.

This PR is based off of the [PR that introduces marker labels](https://github.com/vaadin/flow-components/pull/4632). As that PR can only be merged after a v24 has been branched out, I've set this one to merge against the existing PR to get a clean diff for the style customization.

Part of: https://github.com/vaadin/flow-components/issues/2898
AC: https://github.com/vaadin/platform/issues/3533
Research: https://github.com/vaadin/flow-components/issues/4278#issuecomment-1411810550

## Type of change

- Feature


vaadin-map-flow-parent/vaadin-map-flow-integration-tests/src/test/java/com/vaadin/flow/components/map/FeatureLabelIT.java
```diff
@@ -58,13 +68,64 @@ public void updateLabels() {
...
+        Assert.assertEquals("left", text.getTextAlign());
+        Assert.assertEquals("bottom", text.getTextBaseline());
+        Assert.assertEquals("#fff", text.getFill().getColor());
+        Assert.assertEquals("#000", text.getStroke().getColor());
+        Assert.assertEquals(5, text.getStroke().getWidth());
+        Assert.assertEquals("#1F6B75", text.getBackgroundFill().getColor());
+        Assert.assertEquals("#fff", text.getBackgroundStroke().getColor());
+        Assert.assertEquals(2, text.getBackgroundStroke().getWidth());
+
+        waitUntil(driver -> getRenderCount() == 1);
```
sissbruecker, line 96: All the style assertions can only check that the style function returns correct values, but not that markers have actually been re-rendered using the new styles. 

Adding a check for the number of renders should improve this a bit, especially since changing any style in OpenLayers requires manually forcing a re-render.

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/AbstractConfigurationObject.java
```diff
@@ -161,6 +162,8 @@ protected void deepMarkAsDirty() {
      * resolved on the client-side.
      */
     protected void addChild(AbstractConfigurationObject configurationObject) {
+        Objects.requireNonNull(configurationObject,
+                "Child configuration object must not be null");
```
sissbruecker, line 166: Added an explicit check here. This is not a breaking change, the method would still have failed with a null pointer later on.

Also added a method that allows passing null, which is more convenient to use for objects that have nullable properties.

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/style/Text.java
```diff
@@ -0,0 +1,413 @@
...
+import com.fasterxml.jackson.annotation.ObjectIdGenerators;
+import com.vaadin.flow.component.map.configuration.AbstractConfigurationObject;
+import com.vaadin.flow.component.map.configuration.Constants;
+
+import java.io.Serializable;
+
+/**
+ * Text style that defines how to render texts, such as labels, on the map.
+ */
+public class Text extends AbstractConfigurationObject {
```
sissbruecker, line 14: Not sure about the naming. For one this conflicts with the existing Flow `Text` class, and then the name is misleading since this class (in contrast to the OpenLayers class) doesn't contain actual text data, only styles. Maybe rename it `TextStyle`?

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/style/Text.java
```diff
@@ -0,0 +1,413 @@
...
+        font = "13px sans-serif";
+        offset = new TextOffset(0, 10);
+        scale = 1;
+        rotation = 0;
+        rotateWithView = false;
+        textAlign = TextAlign.CENTER;
+        fill = new Fill("#333");
+        stroke = new Stroke("#fff", 3);
+        padding = 0;
+    }
```
sissbruecker, line 38: These are duplicated from the defaults in the connector for now. There are more default property values specified here, as we don't get the automatic fallback values that OpenLayers provides on the client-side.

The duplication could be removed if we would configure the client-side fallback text style from the server-side. I think it's OK as is for now. If / when add the feature to allow changing the default label style per map, then this would solve itself.

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/style/Text.java
DiegoCardoso: Yeah, even for the OpenLayers class I find it a bit confusing that you `style.getText().setText()` to actually set the text content. `TextStyle` seems to be more precise with the purpose of this class.

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/style/Style.java
```diff
@@ -64,4 +65,16 @@ public void setStroke(Stroke stroke) {
...
+    @JsonIdentityReference(alwaysAsId = true)
+    public Text getText() {
+        return text;
+    }
+
+    public void setText(Text text) {
+        removeChild(this.text);
+        this.text = text;
+        addNullableChild(text);
+    }
```
DiegoCardoso, line 79: Missing javadoc.

DiegoCardoso (commented):
The changes overall look good to me.

One thing I noticed is the lack of documentation on the constructors and some methods like in `Fill`, `Stroke`, and `Text` classes. Looking elsewhere, I could see that most constructors are lacking documentation, even in the `Map` class.

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/style/Text.java
sissbruecker: Renamed the class to `TextStyle`. I also renamed the property in `Style` to `textStyle`, and added a Jackson annotation to map it to `text` on serialization.

vaadin-map-flow-parent/vaadin-map-flow-integration-tests/src/main/java/com/vaadin/flow/component/map/FeatureLabelPage.java
```diff
@@ -43,7 +44,41 @@ public FeatureLabelPage() {
...
 
+        NativeButton setLabelStyle = new NativeButton("Set custom label style",
+                e -> {
+                    marker1.getStyle().setText(createCustomLabelStyle());
+                });
+        setLabelStyle.setId("set-label-style");
+
+        NativeButton updateLabelStyle = new NativeButton(
+                "Update custom label style", e -> {
+                    marker1.getStyle().getText().setFont("15px sans-serif");
```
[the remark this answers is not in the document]
sissbruecker: Added a null check.

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/style/Style.java
sissbruecker: Thanks, added.

Good point about the other missing documentation. We can try to do this as a follow-up maybe. Though I feel constructor docs are not that urgent, and Fill, Stroke are somewhat self-explanatory.

sissbruecker:
@DiegoCardoso I'm questioning my choice to model some values as `double`. `TextOffset` and `padding` are basically pixel values and I can't really image a purpose for specifying sub-pixel values. Objections to using `int` instead?

DiegoCardoso:
> @DiegoCardoso I'm questioning my choice to model some values as `double`. `TextOffset` and `padding` are basically pixel values and I can't really imagine a purpose for specifying sub-pixel values. Objections to using `int` instead?

It makes sense to go with `int` instead of `double` for this property. 👍🏻

sissbruecker:
@DiegoCardoso Updated to use `int` instead.

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/style/TextStyle.java
```diff
@@ -0,0 +1,413 @@
...
+    }
+
+    /**
+     * Sets the padding for the text's background. The padding is applied to all
+     * sides of the background. Default value is {@code 0}.
+     *
+     * @param padding
+     *            the new padding
+     */
+    public void setPadding(int padding) {
```
DiegoCardoso, line 383: Should we support setting the padding for each side or, at least, for each pair of sides (top/bottom and left/right) from the start? My reason is that usually, one wants at least to have bigger horizontal padding than the vertical one.

sissbruecker: My plan was to keep this as simple as possible and wait and see if there is sufficient interest in this 🙂

DiegoCardoso: That was my first assumption, indeed. But, nevertheless, it still made me wonder whether it makes sense to add this API from the start. Anyway, we can add an enhancement ticket to address this behavior and wait for any reactions/requests.

DiegoCardoso (approved):

PR #4657, opened 2023-02-08
fix: update effectiveSize at the end of clear

### LINKED ISSUE #4611: TreeGrid: refreshAll shows empty rows if number of items exceeds page size
### Description

When calling refreshAll on a TreeGrid empty rows are shown instead of items, as soon as the number of items exceeds the page size value.

### Expected outcome

See all the my TreeGrid items after calling refreshAll.

Example project: 
[v23-tree-grid-refresh-issue.zip](https://github.com/vaadin/flow-components/files/10522440/v23-tree-grid-refresh-issue.zip)

There're 2 examples in it, the first one has a page size set to 10 an has 11 items. The second one, has 60 items and it's using the default page size (50). 

### Steps to reproduce

1 - Run attached example project.
2 - In first example (left) select an item and then click "refresh all" button.
3 - Scroll to the bottom of the tree and see last item is missing and an empty row is shown instead. 

tree-grid-page-size-10

4 - In second example (right) select an item and then click "refresh all" button.
5 - Scroll to the bottom of the tree and see last items are missing and empty rows are shown instead.

tree-grid-default-page-size

Also, notice that in both examples, when clicking on the empty rows, a item from the tree is selected.

### Environment

Vaadin version(s): 23.3.5
OS: Windows 10

### Browsers

Chrome, Firefox

ugur-vaadin:
## Description

The effective size was not correctly updated after `grid.$connector.clear`. This led to miscalculation of indexes, requests, and page numbers. This PR adds the update to the end of `clear`.

Fixes #4611 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -397,20 +397,29 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
                 parentUniqueKey
               );
             }
-          } else if (cache[root] && cache[root][page]) {
-            callback(cache[root][page]);
           } else {
-            rootPageCallbacks[page] = callback;
+            // workaround: sometimes grid-element gives page index that overflows
+            page = Math.min(page, Math.floor(grid.size / grid.pageSize));
 
-            rootRequestDebouncer = Debouncer.debounce(
-              rootRequestDebouncer,
-              timeOut.after(grid._hasData ? rootRequestDelay : 0),
-              () => {
-                grid.$connector.fetchPage(
-                  (firstIndex, size) => grid.$server.setRequestedRange(firstIndex, size),
-                  page,
-                  root
-                );
+            if (cache[root] && cache[root][page]) {
+              callback(cache[root][page]);
+            } else {
+              rootPageCallbacks[page] = callback;
+
+              rootRequestDebouncer = Debouncer.debounce(
+                rootRequestDebouncer,
+                timeOut.after(grid._hasData ? rootRequestDelay : 0),
+                () => {
+                  grid.$connector.fetchPage(
+                    (firstIndex, size) => grid.$server.setRequestedRange(firstIndex, size),
+                    page,
+                    root
+                  );
+                }
+              );
+            }
+          }
+        });
               }
             );
           }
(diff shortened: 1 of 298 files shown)


tomivirkki:
Seems that the root problem might be is caused by the connector making content update requests synchronously in `grid.$connector.set`. The `set` function populates the content for the requested rows iteratively, but during each iteration, selections and calls to `updateGridItemsInDomBasedOnCache` are being made.

Tried the following and it seems to fix the issue on the IT page, but it's just a quick tryout and may still need some more thinking:

Screenshot 2023-02-14 at 13 07 35

Since `updateGridItemsInDomBasedOnCache` is pretty much the same as calling `grid.requestContentUpdate()`, we could consider using that instead.

There seems to be some other places where `updateGridItemsInDomBasedOnCache` is called on each iteration while resolving callbacks. May want to check those out as well.

ugur-vaadin:
I have considered this solution too, both works. Also, the change in the PR might be overkill regarding the scope of the issue. So it is probably a better solution for this issue alone. 

What do you think about the logic in the PR outside the scope? The cache currently does not take the page numbering inconsistency into account. The web component does actually prevent such calls when there is already a pending request. In this case too, there is already a pending request effectively for the same page, but under a different index. Do you think that there might be more cases where this is applicable and might cause a bug?

tomivirkki:
There probably shouldn't be any page numbering inconsistency between the connector and the web component. That sounds accidental rather than expected behavior. If that's the case, building a fix on a false premise could be dangerous.

The fact that the non-existent root level page 2 is requested in the IT page scenario happens because the connector manipulates the grid cache externally on `clear()` but doesn't update the grid's effective size -> grid ends up with more rows than it should -> content gets requested for the rows that shouldn't be there.

We should probably add `grid._effectiveSize = grid._cache.effectiveSize;` at the end of `grid.$connector.clear` to make sure the grid has the correct size after cache has been tinkered with. The connector function `updateAllGridRowsInDomBasedOnCache` seems to already have this, but `clear` doesn't.

ugur-vaadin:
You are right, these calls were never valid. Another question would be the role of the page adjustments in the `grid.$connector.dataProvider` and whether to keep them.

Currently, there are 2 lines of code (one for root pages, one for children) with the explanation "workaround: sometimes grid-element gives page index that overflows". I suspect that these are not required anymore. It might be even worse that they are there, since they might be masking other bugs like size etc not properly synced. Those lines can be traced back to one of [commit](https://github.com/vaadin/vaadin-grid-flow/pull/241/commits/2d9ced410d34befe44e648a3d15ad5c8f9e183a4)s in the initial TreeGrid merge [PR](https://github.com/vaadin/vaadin-grid-flow/pull/241/). Removing them do not cause any tests to fail, but the coverage is not reliable at this point. What do you think?

tomivirkki:
As you mentioned, the coverage is not reliable at this point so it's probably best to leave those lines untouched for now.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -667,6 +667,9 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
           const firstPage = index / grid.pageSize;
           const updatedPageCount = Math.ceil(items.length / grid.pageSize);
 
+          const selections = [];
```
tomivirkki, line 670: Are these changes still needed? I assumed the addition of `grid._effectiveSize = grid._cache.effectiveSize;` fixes the issue.

ugur-vaadin: Removed the other changes

PR #4665, opened 2023-02-10
fix: use property change listener for opened change event

### LINKED ISSUE #4654: vaadin-details: OpenedChangeEvent fires upon initialization of the page
### Description

The OpenedChangeEvent fires as soon as the page with the Details Component loads. This seems odd because the Component at this point has only just "come into existence" and nothing has changed yet.
	    
See reproducer view in attached zip file

### Expected outcome

I would expect that no OpenedChangeEvent is triggered upon initialization of the page.

[reproducerView.zip](https://github.com/vaadin/flow-components/files/10683236/reproducerView.zip)

### Steps to reproduce

1. Replace the MainView from the vaadin demo project with the view from the attached reproducer zip file
2. Add a breakpoint in the constructor of com.vaadin.flow.component.details.Details.OpenedChangeEvent
3. Start a debug session

### Environment

Vaadin version(s): 14.9.2 / 23.3.5
OS: Windows 10

### Browsers
reproducible with Firefox 109.0.1

### LINKED ISSUE #4659: vaadin-details: There is no reliable way to tell if an OpenedChangeEvent is triggered by user interaction or not
### Description

We need to distinguish events that come from an actual user versus events that were triggered programmatically. The reason for that is that we record QA test scripts based on user interaction with our app. If we are unable to make this distinction, we do not know the relevant events to record in these test scripts. This prevents us from running our automated QA tests for this component.

In some instances, we can use the property `isFromClient` of the event to make that distinction. But for vaadin-details this is unreliable because programmatically triggered events can come from the client, see https://github.com/vaadin/flow-components/issues/4654

### Expected outcome

We need a way to tell if an event is triggered by user interaction. 

On some events in vaadin 8 there used to be a property `userOriginated` for example in com.vaadin.client.widget.grid.sort.SortEvent

We would need something like that for this component.

[reproducerView.zip](https://github.com/vaadin/flow-components/files/10695024/reproducerView.zip)

### Steps to reproduce

1. Replace the MainView from the vaadin demo project with the view from the attached reproducer zip file
2. Add a breakpoint in the constructor of com.vaadin.flow.component.details.Details.OpenedChangeEvent
3. Start a debug session

### Environment

Vaadin version(s): 14.9.2 / 23.3.5
OS: Windows 10 

### Browsers

Issue is not browser related

sissbruecker:
## Description

Currently the opened change listener for `Details` listens for the `opened-changed` event from the client, which means that even when changing the opened state on the server, the event will always be from the client.

This change fixes that by using a property change listener instead, so that the event will correctly report whether the state was changed on the client or on the server. This also properly initializes the opened state on the server, so that the initial opened change event from initializing the Polymer property is discarded.

Fixes #4659
Fixes #4654

## Type of change

- Bugfix

### DIFF
--- a/vaadin-details-flow-parent/vaadin-details-flow-integration-tests/src/main/java/com/vaadin/flow/component/details/tests/DetailsEventsPage.java
+++ b/vaadin-details-flow-parent/vaadin-details-flow-integration-tests/src/main/java/com/vaadin/flow/component/details/tests/DetailsEventsPage.java
@@ -1,0 +1,29 @@
+package com.vaadin.flow.component.details.tests;
+
+import com.vaadin.flow.component.details.Details;
+import com.vaadin.flow.component.html.Div;
+import com.vaadin.flow.component.html.NativeButton;
+import com.vaadin.flow.router.Route;
+
+@Route("vaadin-details/events")
+public class DetailsEventsPage extends Div {
+    public DetailsEventsPage() {
+        Details details = new Details();
+
+        Div output = new Div();
+        output.setId("output");
+        output.getStyle().set("white-space", "pre");
+        details.addOpenedChangeListener(e -> {
+            output.setText(output.getText() + String.format(
+                    "Opened changed: opened=%s, isFromClient=%s\n",
+                    e.isOpened(), e.isFromClient()));
+        });
+
+        NativeButton toggle = new NativeButton("Toggle", e -> {
+            details.setOpened(!details.isOpened());
+        });
+        toggle.setId("toggle");
+
+        add(details, new Div(toggle), output);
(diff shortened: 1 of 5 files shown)


web-padawan (approved):
LGTM. We should have noticed this before, but Details has never been very popular component 🤷‍♂️

sissbruecker:
Thanks for the prompt review 👍.

I'll merge / cherry-pick this on Monday.

knoobie:
Just a quick note: I've implemented `NativeDetails` in `flow` two years ago, with the exact same behavior that clientside is always true. Not sure if you would/should/must tackle it as well to align them again.

PR #4666, opened 2023-02-12
refactor: remove an unused connector method

vursen:
## Description

Nobody seems to use the `ensureHierarchy` method of the grid connector. If so, let's remove it.

## Type of change

- [x] Refactor


sissbruecker:
`HierarchicalDataCommunicator` does use it 🙈:
https://github.com/vaadin/flow/blob/402b2b621a4298da4bf06c80a0b5049f26e47ea5/flow-data/src/main/java/com/vaadin/flow/data/provider/hierarchy/HierarchicalDataCommunicator.java#L141

vursen:
> HierarchicalDataCommunicator does use it 🙈:

No way! And same do `$connector.expandItems` and `$connector.collapseItems`. That sucks.

Issue #4500, opened 2023-01-10
DateTimePicker: i18n not applied when used within a grid component column


rappenze:
### Description

When I create a component column in grid and add a DateTimePicker, it does not use the provided i18n information.
Be aware that the same just using a DatePicker works as expected.

Also DateTimePicker added in a Layout works as expected. 

Potentially there is a timing problem during loading when the DateTimePicker is attached within a flow-component-renderer?

![image](https://user-images.githubusercontent.com/29733488/211553487-f32c33bd-f795-45ad-91e4-2f89a8c2ffba.png)

### Expected outcome

The DateTimePicker sohuld work correctly when rendered within a grid component column.
Currently that is not the case, it ignores all things applied by the i18n information.

### Minimal reproducible example

```java
package com.sowatec.arregulo.web.ui.field;

import com.vaadin.flow.component.datepicker.DatePicker.DatePickerI18n;
import com.vaadin.flow.component.datetimepicker.DateTimePicker;
import com.vaadin.flow.component.grid.Grid;
import com.vaadin.flow.component.html.Div;
import com.vaadin.flow.router.Route;

@Route("test")
public class TestView extends Div {
  public TestView() {
    Grid<String> grid = new Grid<>();
    grid.addComponentColumn(row -> {
      DateTimePicker dateTime = new DateTimePicker("Date and time");
      DatePickerI18n i18n = new DatePickerI18n();
      i18n.setCancel("Test cancel text");
      i18n.setDateFormat("dd.MM.yyyy");
      dateTime.setDatePickerI18n(i18n);
      add(dateTime);
      return dateTime;
    });

    grid.setItems("Row 1", "Row 2");
    add(grid);
  }
}

```

### Steps to reproduce

1. Open the view
2. Open the date dropdown, it does not show the custom cancel text set by i18n
3. Enter a date and click outside the editor, you see that the custom format is not used to format the date

### Environment

Vaadin version(s): 23.3.2
OS: Windows

### Browsers

Issue is not browser related

rolfsmeds:
This might have been accidentally fixed with the removal of the `FlowComponentRenderer` element in V24. Let's check, and fix in V23 regardless.

tomivirkki:
Confirmed it works as expected in V24:

https://user-images.githubusercontent.com/1222264/212266516-8ea791cc-6f0f-4d78-8a57-236adb0b3b2b.mp4

tomivirkki:
[...]

PR #4706, opened 2023-02-21
refactor: revert i18n workaround

### LINKED ISSUE #4667: DateTimePicker: i18n not applied when used within a TreeGrid component column
### Description

When I create a component column in TreeGrid and add a DateTimePicker, it does not use the provided i18n information.
Be aware that the same just using a DatePicker works as expected and the same issue for a Grid was fixed in #4500.

Even in the TreeGrid the i18n information is only ignored when you have a child row and do initial expand from the server side.

Potentially there is a timing problem during loading when the DateTimePicker is attached within a flow-component-renderer?

image

### Expected outcome

The DateTimePicker sohuld work correctly when rendered within a TreeGrid component column.
Currently that is not the case, it ignores all things applied by the i18n information in child rows when I set them initially to expanded.

```java
package com.sowatec.arregulo.web.ui.field;


import com.vaadin.flow.component.datepicker.DatePicker.DatePickerI18n;
import com.vaadin.flow.component.datetimepicker.DateTimePicker;
import com.vaadin.flow.component.html.Div;
import com.vaadin.flow.component.treegrid.TreeGrid;
import com.vaadin.flow.data.provider.hierarchy.TreeData;


@Route("test")
public class TestView extends Div {
  public TestView() {
    TreeGrid<String> grid = new TreeGrid<>();
    grid.addHierarchyColumn(row -> row);
    grid.addComponentColumn(row -> {
      DateTimePicker dateTime = new DateTimePicker("Date and time");
      DatePickerI18n i18n = new DatePickerI18n();
      i18n.setCancel("Test cancel text");
      i18n.setDateFormat("dd.MM.yyyy");
      dateTime.setDatePickerI18n(i18n);
      add(dateTime);
      return dateTime;
    });


    TreeData<String> data = new TreeData<>();
    data.addRootItems("Row 1", "Row 2");
    data.addItem("Row 1", "Child 1");
    data.addItem("Row 1", "Child 2");
    data.addItem("Row 2", "Child 3");
    data.addItem("Row 2", "Child 4");
    grid.setTreeData(data);
    grid.expand("Row 1", "Row 2");
    add(grid);
  }
}
```

### Steps to reproduce

1. Open the view
2. Open the date dropdown in a child row, it does not show the custom cancel text set by i18n
3. Enter a date and click outside the editor, you see that the custom format is not used to format the date

### Environment

Vaadin version(s): 23.3.6
OS: Windows

### Browsers

Issue is not browser related

ugur-vaadin:
## Description

There is a known timing [issue](https://github.com/vaadin/flow-components/issues/4500) with updating `DateTimePicker` i18n. A [workaround](https://github.com/vaadin/flow-components/pull/4593) was applied for the regular `Grid`, but the problem persisted for the children items of `TreeGrid`.

This PR reverts the previous workaround, while a related web component [PR](https://github.com/vaadin/web-components/pull/5604) makes sure that the i18n is only updated after the component is ready.

This fix needs the web component [PR](https://github.com/vaadin/web-components/pull/5604) to be merged first.

Partially fixes #4667 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
@@ -2,6 +2,7 @@ import dateFnsFormat from 'date-fns/format';
 import dateFnsParse from 'date-fns/parse';
 import dateFnsIsValid from 'date-fns/isValid';
 import { extractDateParts, parseDate as _parseDate } from '@vaadin/date-picker/src/vaadin-date-picker-helper.js';
+import { DatePicker } from '@vaadin/date-picker/src/vaadin-date-picker.js';
 
 (function () {
   const tryCatchWrapper = function (callback) {
@@ -142,6 +143,17 @@ import { extractDateParts, parseDate as _parseDate } from '@vaadin/date-picker/s
           return referenceDate ? new Date(referenceDate.year, referenceDate.month, referenceDate.day) : new Date();
         }
 
+        function _isDateTimePickerDatePicker() {
+          return datepicker.constructor.name === 'DateTimePickerDatePicker';
+        }
+
+        datepicker.ready = tryCatchWrapper(function () {
+          DatePicker.prototype.ready.call(datepicker);
+          if (_isDateTimePickerDatePicker()) {
+            datepicker.dispatchEvent(new Event("date-time-picker-date-picker-ready"));
+          }
+        });
+
         datepicker.$connector.updateI18n = tryCatchWrapper(function (locale, i18n) {
           // Either use custom formats specified in I18N, or create format from locale
           const hasCustomFormats = i18n && i18n.dateFormats && i18n.dateFormats.length > 0;
@@ -151,25 +163,16 @@ import { extractDateParts, parseDate as _parseDate } from '@vaadin/date-picker/s
           const usedFormats = hasCustomFormats ? i18n.dateFormats : [createLocaleBasedDateFormat(locale)];
           const formatterAndParser = createFormatterAndParser(usedFormats);
 
-          const updatedI18n = Object.assign({}, datepicker.i18n, i18n, formatterAndParser);
-
           // Merge current web component I18N settings with new I18N settings and the formatting and parsing functions
-          if (datepicker.constructor.name === 'DateTimePickerDatePicker') {
-            // If the date picker is a part of a date time picker, defer setting I18N property into microtask as a
-            // workaround for: https://github.com/vaadin/flow-components/issues/4500
-            // An additional layer of deferring with a MutationObserver was added as a workaround for:
-            // https://github.com/vaadin/flow-components/issues/4667
-            // These workarounds are only necessary for v23, and will not be removed in v24.
-            queueMicrotask(() => {
-              datepicker.i18n = updatedI18n;
-              const observer = new MutationObserver(() => {
-                observer.disconnect();
-                datepicker.i18n = updatedI18n;
-              });
-              observer.observe(datepicker.parentNode, { attributes: true, subtree: true });
-            });
-          } else {
-            datepicker.i18n = updatedI18n;
+          const updatedI18n = Object.assign({}, datepicker.i18n, i18n, formatterAndParser);
+          datepicker.i18n = updatedI18n;
+          // If the date picker is a part of a date time picker, defer setting I18N property after the element
+          // is ready as a workaround for both:
+          // https://github.com/vaadin/flow-components/issues/4500
+          // https://github.com/vaadin/flow-components/issues/4667
+          // This workaround is only necessary for v23, and will be removed in v24.
+          if (_isDateTimePickerDatePicker()) {
+            datepicker.addEventListener("date-time-picker-date-picker-ready", () => datepicker.i18n = updatedI18n);
           }
         });
       })(datepicker)
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
@@ -143,18 +143,9 @@ import { DatePicker } from '@vaadin/date-picker/src/vaadin-date-picker.js';
           return referenceDate ? new Date(referenceDate.year, referenceDate.month, referenceDate.day) : new Date();
         }
 
-        function _isDateTimePickerDatePicker() {
-          return datepicker.constructor.name === 'DateTimePickerDatePicker';
-        }
-
         datepicker.ready = tryCatchWrapper(function () {
-          if (_isDateTimePickerDatePicker()) {
-            datepicker.dispatchEvent(new Event("date-time-picker-date-picker-ready-call"));
-          }
           DatePicker.prototype.ready.call(datepicker);
-          if (_isDateTimePickerDatePicker()) {
-            datepicker.dispatchEvent(new Event("date-time-picker-date-picker-ready"));
-          }
+          datepicker.dispatchEvent(new Event("date-picker-ready"));
         });
 
         datepicker.$connector.updateI18n = tryCatchWrapper(function (locale, i18n) {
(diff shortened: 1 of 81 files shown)


vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
```diff
@@ -142,6 +143,11 @@ import { extractDateParts, parseDate as _parseDate } from '@vaadin/date-picker/s
           return referenceDate ? new Date(referenceDate.year, referenceDate.month, referenceDate.day) : new Date();
         }
 
+        datepicker.ready = tryCatchWrapper(function () {
+          DatePicker.prototype.ready.call(datepicker);
+          datepicker.dispatchEvent(new Event("date-picker-ready"));
```
vursen, line 148:
```suggestion
          datepicker.dispatchEvent(new CustomEvent("date-picker-ready"));
```

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
```diff
@@ -152,9 +158,14 @@ import { extractDateParts, parseDate as _parseDate } from '@vaadin/date-picker/s
           const formatterAndParser = createFormatterAndParser(usedFormats);
 
           // Merge current web component I18N settings with new I18N settings and the formatting and parsing functions
-          // Defer setting I18N property into microtask as a workaround for: https://github.com/vaadin/flow-components/issues/4500
-          queueMicrotask(() => {
-            datepicker.i18n = Object.assign({}, datepicker.i18n, i18n, formatterAndParser);
+          const updatedI18n = Object.assign({}, datepicker.i18n, i18n, formatterAndParser);
```
vursen, line 161: It is probably better to compute the updated i18n in the animation frame. There is a chance that `datepicker.i18n` can change in the ready callback.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
```diff
@@ -152,9 +158,14 @@ import { extractDateParts, parseDate as _parseDate } from '@vaadin/date-picker/s
...
+          datepicker.i18n = updatedI18n;
+          // Update I18N property again after the element is ready as a workaround for both:
+          // https://github.com/vaadin/flow-components/issues/4500
+          // https://github.com/vaadin/flow-components/issues/4667
+          // This workaround is only necessary for v23, and will be removed in v24.
+          datepicker.addEventListener("date-picker-ready", () => {
```
vursen, line 167: What will happen if you call `DatePicker.setLocale(...)` after the ready callback has been already invoked? I'm afraid the i18n update will never happen in that case.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
```diff
@@ -152,9 +158,14 @@ import { extractDateParts, parseDate as _parseDate } from '@vaadin/date-picker/s
...
+            requestAnimationFrame(() => datepicker.i18n = updatedI18n);
```
tomivirkki, line 168: Not sure if related to this PR, but if the root items are initially collapsed and then expanded, the time pickers have the wrong locale:

https://user-images.githubusercontent.com/1222264/220933193-7ee74951-50fe-43a4-bcce-315027ef2f8a.mp4

tomivirkki, line 168: Prefer microtask timing. With animation frame, you can see the wrong locale flasing really briefly on page load.

vursen: Can be related to https://github.com/vaadin/flow-components/pull/4706/files#r1115759943

ugur-vaadin: I rolled back the change and it is still reproducible.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
ugur-vaadin: I think the first update call still handles those cases. The workaround is just the extra call that updates the i18n again after the initial erroneous override.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
ugur-vaadin: It actually works locally without the animation frame. But I could not get the TeamCity builds to pass. That is the reason I added that part. Microtask did not work as well. Do you have any suggestions?

tomivirkki: Ok, I see. Maybe we could then just add a synchronous call `datepicker.i18n = updatedI18n;` before the rAF line to avoid the flash.

Another option could be to use (instead of the custom event) a custom property change observer for `i18n` which always reverts to the desired value. [Example](https://github.com/vaadin/flow-components/blob/2d60208657c8c4353455d740cbe1bf3cad5c5e90/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js#L1002) in grid connector.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
vursen: Aha, I see. Then my only concern would be that it adds more and more listeners that are never invoked. I don't know how frequently updatei18N is usually called – it probably depends on the application. I think it would make sense to check whether the component is ready before adding a listener.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
ugur-vaadin: I think this last option really makes sense. I updated the PR based on that approach.

ugur-vaadin: That part is still reproducible with the latest updates. Maybe we can separate that issue. What do you think?

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
ugur-vaadin: This was a very valid concern. However, I recently updated the PR based on Tomi's [suggestion](https://github.com/vaadin/flow-components/pull/4706#discussion_r1115859796), and now there are no extra listeners involved.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
vursen: Is it reproducible on 23.3 without these changes?

ugur-vaadin: I checked again, and it is reproducible. I also checked how it works on v24, and it shows "12:00 PM" for both root rows and the expanded rows. I will verify if that is the expected behavior. If so, the time pickers in the root row are broken in v23.3, otherwise the ones in the children rows are broken along with the time pickers in v24.

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/treegrid/it/TreeGridDateTimePickerIT.java
```diff
@@ -0,0 +1,44 @@
...
+    @Before
+    public void before() {
+        open();
+        setupTreeGrid();
+    }
+
+    @Test
+    public void shouldHaveI18nAppliedToRoot() {
+        Assert.assertEquals("13.06.2000", $(DateTimePickerElement.class)
+                .id("id-Row 1").getDatePresentation());
```
vursen, line 36:
```suggestion
                .id("id-Row-1").getDatePresentation());
```

vursen:
I'm not sure if we need these tests now that we have fixed it in the web components. I guess it should be enough to remove the workaround and all the related tests. @tomivirkki What's your opinion?

vursen (approved):
I think it can be considered a `refactor:` since it is more like a clean-up.

PR #4739, opened 2023-02-24
refactor: debounce update in grid connector confirm parent

tomivirkki:
## Description

When multiple hierarchy updates in `TreeGrid` take place at once, the `confirmParent` function in `gridConnector` gets called multiple times. Currently, the `confirmParent` function requests `<vaadin-grid>` to update its rows synchronously after each invocation.

This PR debounces the update request which reduces the overhead on hierarchy updates.

The change has a 40% impact on the `expandtime` metric in the [Grid benchmark tests](https://bender.vaadin.com/buildConfiguration/Flow_Components_BenchmarkTests_Grid?mode=builds).

## Type of change

Performance enhancement

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -1,5 +1,5 @@
 import { Debouncer } from '@polymer/polymer/lib/utils/debounce.js';
-import { timeOut, animationFrame } from '@polymer/polymer/lib/utils/async.js';
+import { timeOut, animationFrame, microTask } from '@polymer/polymer/lib/utils/async.js';
 import { Grid } from '@vaadin/grid/src/vaadin-grid.js';
 import { ItemCache } from '@vaadin/grid/src/vaadin-grid-data-provider-mixin.js';
 import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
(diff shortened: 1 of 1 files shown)


web-padawan (approved):

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -906,7 +906,11 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
           grid.$server.confirmParentUpdate(id, parentKey);
 
           if (!grid.loading) {
-            grid.__updateVisibleRows();
+            grid.__confirmParentUpdateDebouncer = Debouncer.debounce(
+              grid.__confirmParentUpdateDebouncer,
+              microTask,
+              () => grid.__updateVisibleRows()
```
vursen, line 912: Why do we need to update visible rows? Isn't it already covered by _debouncerApplyCachedData in the web component?

tomivirkki: I considered the option of removing the call altogether. But in order to keep focus on performance enhancements within the ongoing project, I'll just assume that the call was originally added here for a reason. Once we have the connector fully covered with tests, it's going to be less risky to remove things we assume to be no longer necessary.

vursen: OK, makes sense.

vursen (approved):

PR #4763, opened 2023-03-03
refactor: avoid unnecessary update in grid connector

tomivirkki:
## Description

Replace a call to the heavy `updateAllGridRowsInDomBasedOnCache()` in the grid connector's data provider with a manual size update to avoid an unnecessary content update.

The change has a 21% impact on the `expandedrendertime` and 55% impact on the `expandtime` metrics in the [Grid benchmark tests](https://bender.vaadin.com/buildConfiguration/Flow_Components_BenchmarkTests_Grid?mode=builds).

## Type of change

Performance enhancement

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -1,5 +1,5 @@
 import { Debouncer } from '@polymer/polymer/lib/utils/debounce.js';
-import { timeOut, microTask } from '@polymer/polymer/lib/utils/async.js';
+import { timeOut, animationFrame, microTask } from '@polymer/polymer/lib/utils/async.js';
 import { Grid } from '@vaadin/grid/src/vaadin-grid.js';
 import { ItemCache } from '@vaadin/grid/src/vaadin-grid-data-provider-mixin.js';
 import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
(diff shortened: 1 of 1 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -145,7 +145,7 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
             level: targetCache.getLevel()
           });
 
-          ensureSubCacheDebouncer = Debouncer.debounce(ensureSubCacheDebouncer, animationFrame, () => {
```
tomivirkki, line 148: Animationframe seems unnecessary here. Replaced with microtask timing.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -381,12 +381,13 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
               ensureSubCacheQueue = [];
               callback(cache[parentUniqueKey][page], cache[parentUniqueKey].size);
 
-              // Flush after the callback to have the grid rows up-to-date
-              updateAllGridRowsInDomBasedOnCache();
+              // Update effective size
+              grid._cache.updateSize();
+              grid._effectiveSize = grid._cache.effectiveSize;
```
tomivirkki, line 386: `updateAllGridRowsInDomBasedOnCache` consists of the lines added here and an explicit call to update all the grid's rendered rows. Since we only require the size update here, we can skip the unnecessary (and heavy) content update. A manual update request is made at the end of the block.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
tomivirkki: Decided to revert this after all and have this PR concern only about one thing.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
vursen: Ideally, these size updates shouldn't be needed because there is `_debouncerApplyCachedData` in the web component that does exactly the same with a 0ms delay.

vursen (approved):

PR #4768, opened 2023-03-04
refactor: do not remove the sub-cache on item collapse

vursen:
## Description

The PR drops grid connector logic that removes the sub-cache when its associated item collapses. This logic only worked when the toggle action would originate from the web component but not from the server. For example, it didn't work in the most common use case, when you are toggling a row with <kbd>Space</kbd> or a click.

## Type of change

- [x] Refactor


vursen:
Note, I temporarily removed the entire `grid.expandItem` and `grid.collapseItem` overrides to check if they are covered with tests.

tomivirkki (approved):

PR #4775, opened 2023-03-07
refactor: avoid unnecessary re-renders in grid connector

tomivirkki:
## Description

Remove the unnecessary `ensureSubCacheQueue = [];` from gridConnector's dataProvider. This will enable `isLoading` to return `true` when the state is [checked](https://github.com/vaadin/web-components/blob/6eb1e181763c67abb71bd406f2ded33495d63c02/packages/grid/src/vaadin-grid-data-provider-mixin.js#L433) in the vaadin-grid-data-provider-mixin, which will prevent a synchronous re-render.

The change has a 20% impact on the `expandedsorttime` and 14% impact on the `expandedrendertime` metrics in the [Grid benchmark tests](https://bender.vaadin.com/buildConfiguration/Flow_Components_BenchmarkTests_Grid?mode=builds).

## Type of change

Performance enhancement


tomivirkki:
After discussing with @vursen we came up with a cleaner solution, which still has the same performance impact.

tomivirkki:
Closing in favor of https://github.com/vaadin/flow-components/pull/4792

PR #4790, opened 2023-03-09
Revert "refactor: reduce grid eager fetch viewport size estimate"

tomivirkki:
## Description

This PR reverts https://github.com/vaadin/flow-components/pull/4748

Even though the first contentful paint is a bit faster with a lower viewport estimate, the Grid immediately needs to start fetching more items for the rows that are in the DOM but still outside the viewport. This has an overall negative impact on how performant the component feels especially on a slower network connection.

## Type of change

Refactor


vursen:
How many rows does the grid usually render outside the viewport?

tomivirkki:
> For my own information, how many rows does the grid usually render outside the viewport?

With the default grid height `400px` and the default row height, seems like 10 inside the viewport and 8 outside of it. But typically in real-life apps, a larger height is used for the grid. If I double the grid's height to `800px`, then there are 21 items inside the viewport and 20 outside.

vursen (approved):

vursen:
> If I double the grid's height to 800px, then there are 21 items inside the viewport and 20 outside.

I wonder why the grid needs this estimation considering that the connector requests a range, so it should ideally be immediately known how many rows the grid needs to render... am I mixing up different concepts?

tomivirkki:
> I wonder why the grid needs to make such estimations in general. 

The estimate is used for preloading the items, even before the Grid is first rendered.

PR #4792, opened 2023-03-09
refactor: avoid unnecessary render cycles in the data provider callback

tomivirkki:
## Description

This change improves how a data provider callback is processed in `gridConnector`. When the callback resolves synchronously from the cache, the updated data provider checks if there is a pending request for another parent item which could be resolved synchronously as well.

The change has a 73% impact on the `expandedsorttime` and 62% impact on the `expandedrendertime` metrics in the [Grid benchmark tests](https://bender.vaadin.com/buildConfiguration/Flow_Components_BenchmarkTests_Grid?mode=builds).

Depends on https://github.com/vaadin/web-components/pull/5647

## Type of change

Performance enhancement


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -910,6 +914,7 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
         });
 
         grid.$connector.confirm = tryCatchWrapper(function (id) {
+          grid.$connector.confirmingParentItems = true;
```
tomivirkki, line 917: I'll probably extract this to a separate PR since it's somewhat unrelated. An ideal fix would be to make sure `expandedItems` are assigned _before_ `confirm` executes.

vursen (approved):

PR #4813, opened 2023-03-16
fix: add jakartaee-web-api test dependency

tomivirkki:
Fix the `ClassesSerializableTest` test failures

### DIFF
--- a/vaadin-accordion-flow-parent/vaadin-accordion-flow/pom.xml
+++ b/vaadin-accordion-flow-parent/vaadin-accordion-flow/pom.xml
@@ -44,6 +44,12 @@
             <artifactId>slf4j-simple</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>jakarta.platform</groupId>
+            <artifactId>jakarta.jakartaee-web-api</artifactId>
+            <version>10.0.0</version>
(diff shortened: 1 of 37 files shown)


vaadin-accordion-flow-parent/vaadin-accordion-flow/pom.xml
```diff
@@ -44,6 +44,12 @@
             <artifactId>slf4j-simple</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>jakarta.platform</groupId>
+            <artifactId>jakarta.jakartaee-web-api</artifactId>
+            <version>10.0.0</version>
```
knoobie, line 50: Wanna add this versioning (dependencyManagement) to the root pom like done in flow to ease migrating later to 11?

tomivirkki: Done

vursen (approved):

knoobie:
The same is needed for v24 as well or?

vursen:
> The same is needed for v24 as well or?

The [related Flow fix](https://github.com/vaadin/flow/pull/16223) was back-ported to 24.0, so good catch. Let me add a 24.0 label.

Issue #4787, opened 2023-02-14
Bound IntegerField does not show value when initially invisible


roelnoten:
### Description of the bug

When an IntegerField is bound with a binder and initially invisible, then a value which is set via field.setValue() is not shown in the integerfield.
See code example below.  This does not display the 123 value in the integerfield when clicking the button.
Once you remove integerField.setVisible(false) or binder.readBean(this) then it works.

Note that the value is in fact available in the browser: a document.getElementsByTagName("vaadin-integer-field")[0].value should '123'.

This is a recent regression. It works in Vaadin 23.3.1 and 23.3.2 and fails in 23.3.5 and 23.3.6.

### Expected behavior

values set in java should be displayed in browser.

### Minimal reproducible example
```
@Route("valueissue")
public class ValueIssue extends VerticalLayout {

    private Integer value;

    public ValueIssue() {
        IntegerField integerField = new IntegerField("integerfield");
        integerField.setVisible(false);
        add(integerField);

        Binder binder = new Binder<>(ValueIssue.class);
        binder.forField(integerField).bind("value");
        binder.readBean(this);

        add(new Button("click", (event) -> {
            integerField.setVisible(true);
            integerField.setValue(123);
        }));
    }

    public Integer getValue() {
        return value;
    }

    public void setValue(Integer value) {
        this.value = value;
    }
}
```

### Versions

- Vaadin / Flow version: 23.3.5
- Java version: 11
- OS version:
- Browser version (if applicable):
- Application Server (if applicable):
- IDE (if applicable):

vursen:
The issue has been fixed for several components. The fix is available in the following versions:

| Component   | 23.3.9 (not released yet) | [24.0.2](https://github.com/vaadin/flow-components/releases/tag/24.0.2) | 24.0.3 (not released yet) |
|:------------- | -------------------- | --------------------------- | --------- |
| DatePicker    | ✅ | ✅ | ✅ 
| TimePicker   | ✅ |      | ✅
| IntegerField  | ✅ | ✅ | ✅
| NumberField | ✅ | ✅ | ✅
| BigDecimalField | ✅ | | ✅

PR #4815, opened 2023-03-16
fix: do not attempt to clear bad input when value is present

vursen:
## Description

There is an optimization in Flow to only send the final value for an element property in the case the property is updated multiple times using `setProperty()` within one round-trip. In terms of a set of commands that Flow sends to the client, the first value set command will be replaced with the final value set command. However, the optimization doesn't apply to `executeJs()`. It is sent as many times as you call it. 

It is therefore possible that

```java
integerField.setValue(null);
integerField.setValue(1234);
```

```js
{ command: 'setProperty', name: 'value', value: '1234' }
```

```js
{ command: 'setProperty', name: 'value', value: '1234' }
```

The PR fixes that by adding a check to prevent the execution of `executeJs()` clearing the input element's value if the component's value is no longer empty.

> **Note**
> I'll do the same fix for DatePicker, TimePicker, BigDecimalField separately.

Fixes https://github.com/vaadin/flow-components/issues/4787

## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/AbstractNumberField.java
+++ b/vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/AbstractNumberField.java
@@ -138,11 +138,19 @@ public abstract class AbstractNumberField<C extends AbstractNumberField<C, T>, T
 
         super.setValue(value);
 
+        // Clear the input element from possible bad input.
         if (Objects.equals(oldValue, getEmptyValue())
                 && Objects.equals(value, getEmptyValue())
                 && isInputValuePresent()) {
-            // Clear the input element from possible bad input.
-            getElement().executeJs("this.inputElement.value = ''");
+            // The check for value presence before clearing the input element's
+            // value is necessary to ensure the last non-empty value won't get
+            // cleared when clear() and setValue(...) are subsequently called
+            // within one round-trip.
+            // Note, Flow is optimized to only send the last component's value
+            // when changing it several times during a round-trip.
+            // The last set value is sent in place of the first set one.
+            getElement()
+                    .executeJs("if (!this.value) this._inputElementValue = ''");
             getElement().setProperty("_hasInputValue", false);
             fireEvent(new ClientValidatedEvent(this, false));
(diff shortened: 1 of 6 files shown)


vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/AbstractNumberField.java
```diff
@@ -138,11 +138,19 @@ public void setValue(T value) {
...
+            // The check for value presence guarantees that a non-empty value
+            // won't get cleared when setValue(null) and setValue(...) are
+            // subsequently called within one round-trip.
+            // That may happen otherwise due to the Flow optimization:
+            // Flow only sends the final component value to the client
+            // when you update the value multiple times during a round-trip
+            // and the final value is sent in place of the first one.
+            getElement()
+                    .executeJs("if (!this.value) this._inputElementValue = ''");
             getElement().setProperty("_hasInputValue", false);
```
bwajtr, line 154: This PR looks Ok to me, I only wonder if ` getElement().setProperty("_hasInputValue", false);` shouldn't get a similar treatment. However, I'm not sure if I understand what `_hasInputValue` should express: I thought simply that _input has some value in it_, but when I set a value to a field using flow's `integerField.setValue(1234)` it is still `false`. It changes to `_hasInputValue==true` only when I set the input manually by typing text in the field... is that correct?

bwajtr (approved):

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/AbstractNumberField.java
vursen: Yeah, it is a known inconsistency that `_hasInputValue` can indicate `false` when the value is set programmatically. This currently doesn't cause real issues because Flow relies on this flag when it needs to determine whether _the user_ has entered something in the field. The inconsistency is described in: https://github.com/vaadin/web-components/issues/5639.

The `_hasInputValue` property isn't prone to this regression because we are using `setProperty`. Only the final value is sent to the client.

PR #4824, opened 2023-03-17
chore: stop enforcing LF line endings by formatter

bwajtr:
## Description

When you install `git` on Winwows, you'll automatically have `core.autocrlf` set to `true`, which will basically change all `LF` line endings to `CRLF` line endings when you clone the repository. Git then ensures that `CRLF` is converted back to `LF` on commit. However, because you have `CRLF` locally, running `mvn formatter:format` with actually change **_all_** of your Java files -> changing only the line endings. This is very inconvenient for Windows developers. The solution to this is to **not** enforce line endings by formatter, but actually let `git` handle it automatically.
  - I've set the formatter to ignore line endings
  - I've introduced .gitattributes file, which enables git to handle `CRLF` to `LF` conversion for Windows developers automatically, regardless of their local settings of `core.autocrlf`.  Git makes sure, thanks to this file, that remote will always contain only `LF`. The topic is explained [here](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings?platform=windows).

## Type of change

 [x] Chore


Artur-:
So like this https://markentier.tech/posts/2021/10/autocrlf-true-considered-harmful/ ? :D

bwajtr:
> So like this https://markentier.tech/posts/2021/10/autocrlf-true-considered-harmful/ ? :D

Not sure what this comment means, but the purpose of this PR is actually to introduce the `.gitattributes`, which is the recommended solution to this line-ending issue (even by the article). We can discuss the content of .gitattributes -> whether or not to enforce LF, even for Windows users. However, I don't see a reason to enforce such a thing. With `.gitattributes`, there shouldn't be any problems with this anymore. And enforcing LF on Windows is really annoying: The linked article argues that modern editors can handle LF even on Windows. That is true, but in that case, you have to change the defaults of _all your other_ toolchain to use LF as well. If you work on just a single project, that's probably not that big deal, but once you start mixing projects, it's starts to be annoying.

The same silly argument would be to enforce CRLF even for Mac/Linux users, because _decent editor can handle any line ending_, right?

bwajtr:
> In general, everything is much easier when all developers have the same environment. Now, I don't see a problem if Windows users convert and convert back some files although it does not seem very fruitful.
> 
> The thing that could be a problem is that you are not only removing the LF requirement for your machine and Windows, you are removing it from the build the validates what goes into the repository and that these files have the correct line ending

Yes, exactly. With `.gitattributes.` there is no need for the build to check for line endings because that's already being handled by git. Details about this [topic also here](https://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/).

vursen (approved):

Issue #4786, opened 2023-02-12
BigDecimal field using binder accept values but won't display them


mrmimo:
### Description of the bug

I am creating BigDecimalField using the following code:

//property is String = 'price', binder is set with a bean containing price property with BigDecimal non null value

BigDecimalField x = new BigDecimal(property)
binder.bind(x, property)
x.setPlaceHolder("placehd")
form.add(x)

I use exactly same code for TextField with String property of the above bean.

### Expected behavior

TextField works as expected. But BigDecimalField doesn't display any value, if used with a placeholder as above it displays the 'placehd' text, not the value from the bean. However if I examine the HTML produced for the field it looks like:

```
<input slot="input" value="0,23" type="text" id="input-vaadin-big-decimal-field-23" placeholder="placehd" aria-labelledby="label-vaadin-big-decimal-field-15" inputmode="decimal">
```

It evidently contains the value from the bean. I can edit it, save the form and it accepts values, but it won't display them. I added debug log in AbstractField setValue to make sure the calling patterns for TextField and BigDecimal are the same and it goes like this:

```
setValue in com.vaadin.flow.component.textfield.TextField value = 
setValue in com.vaadin.flow.component.textfield.TextField value = 
setValue in com.vaadin.flow.component.textfield.BigDecimalField value = null
setValue in com.vaadin.flow.component.textfield.TextField value = 
setValue in com.vaadin.flow.component.textfield.BigDecimalField value = null
setValue in com.vaadin.flow.component.textfield.TextField value = subject
setValue in com.vaadin.flow.component.textfield.TextField value = description
setValue in com.vaadin.flow.component.textfield.TextField value = GTU_12
setValue in com.vaadin.flow.component.textfield.BigDecimalField value = 112.03
setValue in com.vaadin.flow.component.textfield.TextField value = szt
setValue in com.vaadin.flow.component.textfield.BigDecimalField value = 0.23
```

If I don't use binder, but instead use setValue on BigDecimalField - it display the value correctly. My working theory is that while using binder there are two setXXX cycles - apparently first values are set to null, then they change to non-null values, and it does not generate proper calls in javascript counterpart.  To prove this concept I created the following code:

```
BigDecimalField f = new BigDecimalField("xxxx");
f.setValue(null);
f.setValue(new BigDecimal("123.00"));
...
```

In such case BigDecimal won't display any value, even though the html value parameter is set with 123,00 - it displays empty placeholder.

### Minimal reproducible example

Create a form with:
[...]

PR #4828, opened 2023-03-20
fix(TimePicker): do not attempt to clear bad input when value is present 

vursen:
## Description

There is an optimization in Flow to only send the final value for an element property in the case the property is updated multiple times using `setProperty()` within one round-trip. In terms of commands sent to the client, the final value set command will replace the first value set command. However, the optimization doesn't apply to `executeJs()`. It is sent as many times as you call it. 

It is therefore possible that

```java
timePicker.setValue(null);
timePicker.setValue(LocalTime.of(12, 0));
```

```js
{ command: 'setProperty', name: 'value', value: '12:00' }
```

```js
{ command: 'setProperty', name: 'value', value: '12:00' }
```

The PR fixes that by adding a check to prevent the execution of `executeJs()` clearing the input element's value if the component's value is no longer empty.

Depends on
- https://github.com/vaadin/web-components/issues/5695.

Related to https://github.com/vaadin/flow-components/issues/4786#issuecomment-1460173710

## Type of change

- [x] Bugfix


vursen:
> **Warning**
> The PR is currently blocked by https://github.com/vaadin/web-components/issues/5695.

ugur-vaadin (approved):

PR #4834, opened 2023-03-20
fix(BigDecimalField): restore input value when setting a non-empty value

vursen:
## Description

There is an optimization in Flow to only send the final value for an element property in the case the property is updated multiple times using `setProperty()` within one round-trip. In terms of commands sent to the client, the first value set command will be replaced with the final value set command. However, the optimization doesn't apply to `executeJs()`. It is sent as many times as you call it. 

It is therefore possible that

```java
bigDecimalField.setValue(null);
bigDecimalField.setValue(BigDecimal.valueOf(1234));
```

```js
{ command: 'setProperty', name: 'value', value: '1234' }
```

```js
{ command: 'setProperty', name: 'value', value: '1234' }
```

The PR fixes that by restoring the input element's value when setting a non-empty component value, in case it was cleared due to bad input in the above case.

Fixes https://github.com/vaadin/flow-components/issues/4786
## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/BigDecimalField.java
+++ b/vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/BigDecimalField.java
@@ -228,9 +228,10 @@ public class BigDecimalField extends TextFieldBase<BigDecimalField, BigDecimal>
             getElement().setProperty("_hasInputValue", false);
             fireEvent(new ClientValidatedEvent(this, false));
         } else {
-            // Restore the input element's value to prevent a non-empty value
-            // from getting cleared when setValue(null) and setValue(...) are
-            // subsequently called within one round-trip.
+            // Restore the input element's value in case it was cleared
+            // in the previous branch which can happen when setValue(null)
+            // and setValue(...) are subsequently called within one round-trip
+            // and there was bad input.
             getElement().executeJs("this._inputElementValue = $0",
                     FORMATTER.apply(this, value));
         }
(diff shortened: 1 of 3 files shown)


vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/BigDecimalField.java
```diff
@@ -224,9 +224,15 @@ public void setValue(BigDecimal value) {
...
-            getElement().executeJs("this.inputElement.value = ''");
+            getElement().executeJs("this._inputElementValue = ''");
             getElement().setProperty("_hasInputValue", false);
             fireEvent(new ClientValidatedEvent(this, false));
+        } else {
+            // Restore the input element's value in case it was cleared
+            // in the above branch. That can happen when setValue(null)
+            // and setValue(...) are subsequently called within one round-trip
+            // and there was bad input.
+            getElement().executeJs("this._inputElementValue = this.value");
```
vursen, line 235: The solution could actually be much simpler if we had a `_hasBadInput` property instead of `_hasInputValue` (see a proposal in https://github.com/vaadin/web-components/issues/5639).

Example:

```java
if (Objects.equals(oldValue, getEmptyValue())
    && Objects.equals(value, getEmptyValue())
    && isBadInputPresent()) {
    getElement().executeJs("if (this._hasBadInput) this._inputElementValue = ''");
    getElement().setProperty("_hasBadInput", false);
    fireEvent(new ClientValidatedEvent(this, false));
}
```

vursen: BigDecimalField is different from such components as TimePicker, IntegerField, etc, in the way that it doesn't clear its value when the user enters bad input. For that reason, I had to use a different approach than in [other components](https://github.com/vaadin/flow-components/pull/4828/files#diff-82671d2386a473925418e897ecccbb7c6ef2ab89bb7fe9df89b41809958244f4R260-R268). I rather restore the input element's value than prevent `executeJs` from clearing it.

vursen: If we had a `_hasBadInput` property ([proposal](https://github.com/vaadin/web-components/issues/5639)), we could share a single approach between components.

Example:

```java
if (Objects.equals(oldValue, getEmptyValue())
    && Objects.equals(value, getEmptyValue())
    && isBadInputPresent()) {
    getElement().executeJs("if (this._hasBadInput) this._inputElementValue = ''");
    getElement().setProperty("_hasBadInput", false);
    fireEvent(new ClientValidatedEvent(this, false));
}
```

ugur-vaadin (approved):

PR #4848, opened 2023-03-22
chore: set javaparser version used by spreadsheet

MarcinVaadin:
## Description

Set javaparser dependency to be inline with vaadin-dev-server.


pom.xml
```diff
@@ -595,6 +596,11 @@
                 <artifactId>poi-ooxml-full</artifactId>
                 <version>${poi.version}</version>
             </dependency>
+            <dependency>
+                <groupId>com.github.javaparser</groupId>
```
Artur-, line 600: To be sure that this version is used, you should add an exclusion for javaparser for whereever it originally comes from

Artur- (changes requested):

pom.xml
MarcinVaadin: Excluded from poi-ooxml and defined explicitly

sissbruecker (approved):

PR #4879, opened 2023-03-27
fix: Lazy init of disable on click

caalador:
## Description

Only init disable on click
if actually enabled for button.
Move function to buttonFunctions.js
so each button only needs to call the function.

Moving function makes 5000 buttons request
0.5 MB smaller and lazy init to only calll
method for disbleOnClick buttons drops 1MB
from the request.

### DIFF
--- a/vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java
+++ b/vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java
@@ -361,8 +361,8 @@ public class Button extends Component
      * if server-side handling takes some time.
      */
     private void initDisableOnClick() {
-        getElement().executeJs(
-                "window.Vaadin.Flow.button.initDisableOnClick($0)");
+        getElement()
+                .executeJs("window.Vaadin.Flow.button.initDisableOnClick($0)");
     }
 
     private void updateIconSlot() {
--- a/vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java
+++ b/vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java
@@ -20,7 +20,6 @@ import com.vaadin.flow.component.ClickEvent;
 import com.vaadin.flow.component.ClickNotifier;
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.ComponentEventListener;
-import com.vaadin.flow.component.DetachEvent;
 import com.vaadin.flow.component.Focusable;
 import com.vaadin.flow.component.HasAriaLabel;
 import com.vaadin.flow.component.HasEnabled;
--- a/vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java
+++ b/vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java
@@ -57,6 +57,7 @@ import java.util.stream.Stream;
 @JsModule("@vaadin/polymer-legacy-adapter/style-modules.js")
 @NpmPackage(value = "@vaadin/button", version = "24.1.0-alpha5")
 @JsModule("@vaadin/button/src/vaadin-button.js")
+@JsModule("./buttonFunctions.js")
 public class Button extends Component
         implements ClickNotifier<Button>, Focusable<Button>, HasAriaLabel,
         HasEnabled, HasPrefix, HasSize, HasStyle, HasSuffix, HasText,
@@ -65,6 +66,7 @@ public class Button extends Component
     private Component iconComponent;
     private boolean iconAfterText;
     private boolean disableOnClick = false;
+    private boolean disableInited = false;
 
     // Register immediately as first listener
     private Registration disableListener = addClickListener(
@@ -340,6 +342,7 @@ public class Button extends Component
         this.disableOnClick = disableOnClick;
         if (disableOnClick) {
             getElement().setAttribute("disableOnClick", "true");
+            initDisableOnClick();
         } else {
             getElement().removeAttribute("disableOnClick");
         }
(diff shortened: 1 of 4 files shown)


vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java
```diff
@@ -359,10 +362,15 @@ public boolean isDisableOnClick() {
      * if server-side handling takes some time.
      */
     private void initDisableOnClick() {
-        getElement().executeJs("var disableEvent = function () {"
-                + "if($0.getAttribute('disableOnClick')){"
-                + " $0.setAttribute('disabled', 'true');" + "}" + "};"
-                + "$0.addEventListener('click', disableEvent)");
+        if(!disableInited) {
```
tomivirkki, line 365: Let's not add the `disableInited` flag since it will make the feature not work on detach -> attach / preserve on refresh.

Better to do the check in the JS function:

```js
if (!button.__hasDisableOnClickListener) {
  button.addEventListener('click', disableOnClickListener);
  button.__hasDisableOnClickListener = true;
}
```

vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java
```diff
@@ -340,6 +342,7 @@ public void setDisableOnClick(boolean disableOnClick) {
         this.disableOnClick = disableOnClick;
         if (disableOnClick) {
             getElement().setAttribute("disableOnClick", "true");
```
tomivirkki, line 344: Could change this undefined attribute to a property.

vaadin-button-flow-parent/vaadin-button-flow/src/main/resources/META-INF/resources/frontend/buttonFunctions.js
```diff
@@ -0,0 +1,10 @@
+window.Vaadin.Flow.button = {
+  initDisableOnClick: (target) => {
+    const disableEvent = function (target) {
+      if (target.getAttribute('disableOnClick')) {
+        target.setAttribute('disabled', 'true');
+      }
+    };
```
tomivirkki, line 7: Move this function outside the `window.Vaadin.Flow.button`-object so it doesn't get re-created on each call:

```js
function disableOnClickListener({ currentTarget: button }) {
  button.disabled = !!button.disableOnClick;
}
```

vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java
```diff
@@ -359,10 +362,15 @@ public boolean isDisableOnClick() {
...
-                + "$0.addEventListener('click', disableEvent)");
+        if(!disableInited) {
+            getElement().executeJs(
+                    "window.Vaadin.Flow.button.initDisableOnClick($0)");
+            disableInited = true;
+        }
+//        var disableEvent = function () {"
+//                + "if($0.getAttribute('disableOnClick')){"
+//                + " $0.setAttribute('disabled', 'true');" + "}" + "};"
+//                + "$0.addEventListener('click', disableEvent)");
```
tomivirkki, line 373: Remove the commented code

vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java
knoobie: Please don't ;) I'm relying on this attribute for styling

tomivirkki: Oh, ok, and you're probably not the only one :) Let's keep it as an attribute then.

...maybe we can align it with our other component attributes and change it to `<vaadin-button disable-on-click>` in one of the upcoming majors.

vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java
caalador: done

vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java
caalador: done

vaadin-button-flow-parent/vaadin-button-flow/src/main/resources/META-INF/resources/frontend/buttonFunctions.js
```diff
@@ -0,0 +1,10 @@
+window.Vaadin.Flow.button = {
+  initDisableOnClick: (target) => {
```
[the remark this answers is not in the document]
caalador: done

vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java
```diff
@@ -20,6 +20,7 @@
 import com.vaadin.flow.component.ClickNotifier;
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.ComponentEventListener;
+import com.vaadin.flow.component.DetachEvent;
```
tomivirkki, line 23: Unused import

caalador: removed

vaadin-button-flow-parent/vaadin-button-flow/src/main/resources/META-INF/resources/frontend/buttonFunctions.js
```diff
@@ -0,0 +1,12 @@
+window.Vaadin.Flow.button = {
+  initDisableOnClick: (button) => {
+    if (!button.__hasDisableOnClickListener) {
+      button.addEventListener('click', disableOnClickListener(button));
```
tomivirkki, line 4: Shouldn't run the function here but pass the function as the click listener

```suggestion
      button.addEventListener('click', disableOnClickListener);
```

vaadin-button-flow-parent/vaadin-button-flow/src/main/resources/META-INF/resources/frontend/buttonFunctions.js
```diff
@@ -0,0 +1,12 @@
...
+      button.__hasDisableOnClickListener = true;
+    }
+  }
+}
+
+function disableOnClickListener({currentTarget: button}) {
+  button.disabled = button.hasAttribute('disableOnClick');
+}
```
tomivirkki, line 12: nit: it's [recommended](https://typescript-eslint.io/rules/no-use-before-define/) to define the function above from where it's used so I'd suggest moving this function above the `window.Vaadin.Flow.button`.

tomivirkki (approved):

PR #4909, opened 2023-03-31
fix: allow unregistering event handlers through event

### LINKED ISSUE #4802: Details OpenedChangeListener: unregisterListener can only be called inside the event listener
### Description

I register a listener for `OpenedChangeEvent`s on a Vaadin `Details` component and inside the listener, I call `event.unregisterListener`.
Since Vaadin 23.3.7 this results in an exception:
> java.lang.IllegalStateException: unregisterListener can only be called inside the event listener
>		at com.vaadin.flow.component.ComponentEvent.unregisterListener(ComponentEvent.java:90)

### Expected outcome

The listener should be unregistered and no exception should be thrown.

```
AccordionPanel panel = new AccordionPanel();
// Add panel content
panel.addThemeVariants(DetailsVariant.FILLED, DetailsVariant.REVERSE);
panel.addOpenedChangeListener(e -> {
  if (e.isOpened())
    e.unregisterListener();
});
```

### Steps to reproduce

1. Add an AccordionPanel to a Vaadin  application using the above code
2. Add a few items to the panel
3. Open the webpage and open one of the Details

### Environment

Vaadin version(s):  23.3.7

### Browsers

Issue is not browser related

### Notes
The issue does not occur on 23.3.6 and is related to https://github.com/vaadin/flow-components/pull/4665

ugur-vaadin:
## Description

Fixes #4802 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


vaadin-details-flow-parent/vaadin-details-flow/src/main/java/com/vaadin/flow/component/details/Details.java
```diff
@@ -321,8 +322,17 @@ public boolean isOpened() {
...
+                @Override
+                protected void fireComponentEvent(
+                        PropertyChangeEvent propertyChangeEvent) {
+                    Details.this.fireEvent(new OpenedChangeEvent(Details.this,
+                            propertyChangeEvent.isUserOriginated()));
+                }
+            };
+        }
+        return openedPropertyChangeEventHandler.addListener(listener);
     }
```
sissbruecker, line 337: Looking at how `ComponentEvent.unregisterListener` works, it depends on `ComponentEventBus` to set the unregister listener:
https://github.com/vaadin/flow/blob/3aaaeea838d6b120dd2e717bb1891479c0a2980e/flow-server/src/main/java/com/vaadin/flow/component/ComponentEventBus.java#L227-L235

So rather than implementing our custom solution here, maybe we should reuse `ComponentEventBus`. I was able to get the tests to pass by making two changes. First, define the `opened` property change listener once in the constructor, and make it fire the `OpenedChangeEvent`:
```java
getElement().addPropertyChangeListener("opened",
        event -> fireEvent(new OpenedChangeEvent(this, event.isUserOriginated())));
```
Then `addOpenedChangeListener` just becomes:
```java
public Registration addOpenedChangeListener(
        ComponentEventListener<OpenedChangeEvent> listener) {
    return addListener(OpenedChangeEvent.class, listener);
}
```
Both `fireEvent` and `addListener` delegate to the component event bus, which makes `unregisterListener` work automatically.

sissbruecker:
@ugur-vaadin Was there anything else you wanted to add to the PR? Any other components to cover?

ugur-vaadin:
> @ugur-vaadin Was there anything else you wanted to add to the PR? Any other components to cover?

There is another example of this in `ContextMenuBase.addOpenedChangeListener`, which is only different in terms of using generics. Controlling the behaviour in the constructor makes sense and seems clean. IIRC, the tests used to pass for this draft as well, but I might be wrong. There was also a discussion about potentially adding a Flow helper for this event conversion like `Element.forwardPropertyChange(propertyName, ComponentEventListener listener),` but your suggested approach makes it unnecessary.

sissbruecker (approved):

vursen:
The fix should probably be backported to 23.3 as the issue was reported there.

sissbruecker:
@vursen Yep, this should be ported all the way to v14, at least for `Details`, where this is a regression. But I'll have to do the cherry-picks manually, the events have been moved around when removing the generated classes in v24.

PR #4933, opened 2023-04-14
chore: Use global imports for Material

Artur-:
Uses the global JS files for Material (added in https://github.com/vaadin/web-components/pull/5756) so the theme can be loaded at any point in time and styles always end up in the beginning of `<head>`

This is the same as #4808 done for Lumo


Artur-:
Likely needs a new web component release and update here for tests to pass

vursen:
I rebased the branch from `main`. Let's see if the tests pass.

vursen:
Besides, could you @Artur- please mention in the description what this PR follows up on?

Artur-:
Updated the description. Still needs a new web component release containing https://github.com/vaadin/web-components/pull/5756

vursen (approved):

web-padawan (approved):

Artur-:
Weird. This should now pass and locally it works

Artur-:
Ah ok so it's some other problem as #5034 fails in the same way

PR #4946, opened 2023-04-18
feat: introduce SideNav component Flow API

bwajtr:
## Description

The initial version of Flow API for `side-nav` and `side-nav-item` web components. Based on [vcf-nav](https://github.com/vaadin/vcf-nav), which is already in use in Vaadin Start.

This component is considered experimental and should be explicitly enabled via the feature flag.

Includes flow-component parts for:

https://github.com/vaadin/vcf-nav/issues/21
https://github.com/vaadin/vcf-nav/issues/19
https://github.com/vaadin/vcf-nav/issues/20
Fixes # (issue reference to be added)

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


pom.xml
```diff
@@ -87,6 +87,7 @@
         <module>vaadin-rich-text-editor-flow-parent</module>
         <module>vaadin-spreadsheet-flow-parent</module>
         <module>vaadin-field-highlighter-flow-parent</module>
+        <module>vaadin-side-nav-flow-parent</module>
```
web-padawan, line 90: BTW, we might want to sort these alphabetically at some point. I will create a PR.

vaadin-side-nav-flow-parent/vaadin-side-nav-flow-integration-tests/src/test/java/com/vaadin/flow/component/sidenav/tests/SideNavHierarchyIT.java
sissbruecker, line 1: Integration tests are a bit excessive at the moment. While having more tests is always good, there is a trade-off in that more tests also significantly slow down CI. For the most part we want to use integration tests to verify any custom client-side integration that the Flow component does, for example when running custom JS, listening for events, or synchronizing properties.

What we don't want to do is a full test coverage of every case, especially if those are implemented in the web component, or can be covered by unit tests.

I think for this component we want to cover some basic use-cases (render menu with items, clicking navigates, toggle expanded), and that the synchronization of the properties from client to server works (those annotated with `@Synchronize`).

On the other hand some things that we don't need integration tests for, or can be covered with unit tests:
- updating properties
- adding / removing items
- different constructor variants
- creating items from view components
- specific cases like clicking item that has no path - should be covered by web component

vaadin-side-nav-flow-parent/vaadin-side-nav-flow/src/main/java/com/vaadin/flow/component/sidenav/SideNavItem.java
```diff
@@ -0,0 +1,294 @@
...
+            StateTree tree = (StateTree) getElement().getNode().getOwner();
+            router = tree.getUI().getInternals().getRouter();
+        }
+        if (router == null) {
+            router = VaadinService.getCurrent().getRouter();
+        }
+        if (router == null) {
+            throw new IllegalStateException(
+                    "Implicit router instance is not available. "
+                            + "Use overloaded method with explicit router parameter.");
```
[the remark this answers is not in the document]
bwajtr: Changed wording of the exception

vaadin-side-nav-flow-parent/vaadin-side-nav-flow-integration-tests/src/test/java/com/vaadin/flow/component/sidenav/tests/SideNavHierarchyIT.java
bwajtr: Ok, I got the point. I've dramatically simplified the IT tests

sissbruecker: Thanks, looks great 👍

One thing that could be added to the unit tests is creating an item / setting a path from a view class. That's a bit tricky because it requires setting up some Flow internals and a router configuration, but it should be doable. Still, we can do this in a follow up PR.

vaadin-side-nav-flow-parent/vaadin-side-nav-flow/src/main/java/com/vaadin/flow/component/sidenav/SideNav.java
```diff
@@ -0,0 +1,183 @@
...
+    /**
+     * Set a textual label for this side navigation menu.
+     * <p>
+     * This label can help the end user to distinguish groups of navigation
+     * items. The label is also available for screen reader users.
+     *
+     * @param label
+     *            the label to set
+     */
+    public void setLabel(String label) {
```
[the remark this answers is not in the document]
bwajtr: It makes sense to me to remove the label when set to null, so the developer can "return to initial state" after setLabel() is used. However, I'd keep the possibility to set the empty string as it may be desired for some reason.

vaadin-side-nav-flow-parent/vaadin-side-nav-flow/src/main/java/com/vaadin/flow/component/sidenav/SideNav.java
```diff
@@ -0,0 +1,183 @@
...
+     * Creates a new menu with the given label.
+     *
+     * @param label
+     *            the label to use
+     */
+    public SideNav(String label) {
+        setLabel(label);
+    }
+
+    private boolean isLabelElement(Element element) {
```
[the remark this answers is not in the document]
bwajtr: Ok, I did that. Setting null removes the created element. 
I made also the same change for SideNavItem, cause label handling was almost the same there...

sissbruecker (approved):

PR #5002, opened 2023-04-28
feat: add setTooltipGenerator method for the entire grid

### LINKED ISSUE #4998: Add setTooltipGenerator to Grid (Row level)
### Describe your motivation

A tooltip with setTooltipGenerator can be defined at Column level in a Grid. In V8 you were able to define a tooltip with setDescriptionGenerator at Grid (Row) and Column levels. So it would be great to replicate this option and the behavior that is:

- If a tooltip is defined at Grid level is the default tooltip
- If a tooltip is defined at Column level this tooltip is used for this column and the rest continues to have the default one (if it exist) no matter the order of definition (first Grid and second Column or viceversa) 

Just replicate the V8 behavior as described. Internally setTooltipGenerator at Grid level would define for every Column that doesn't have a predefined generator the provided one. If later on a generator is defined for a Column it would override the default one if it exists. 

Seems that V8 approach is convenient so no alternatives considered

### Additional context

_No response_

sosa-vaadin:
## Description

The grid class will now hold a tooltip generator of its own. The columns will now check if their generator produces some content, and if not, it will try with the grid's generator. The `<vaadin-tooltip>` element will be added to the grid when either the grid generator is set, or any column generator is set.
Some refactoring was needed so that the content of the tooltip that each column would generate could be extracted with a single method call from the tests; as well as to add the `<vaadin-tooltip>` element to the grid when calling the grid's version of the method. 

Fixes #4998 

## Type of change

- [x] Feature

## Checklist


sosa-vaadin:
Might be a good moment to ask why 3 unit tests are marked to be ignored.

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridTooltipIT.java
```diff
@@ -74,6 +74,58 @@ public void dynamicallyAddGenerator_hoverOverTooltipColumnCell_showTooltip() {
...
+                            getActiveTooltipText());
+                } else {
+                    // check that previously set column (5) tooltip was not
+                    // replaced
+                    Assert.assertEquals(
+                            "First name of the person is " + firstName,
+                            getActiveTooltipText());
+                }
+            }
+        }
```
sissbruecker, line 98: Let's simplify this to check the tooltip of two cells, one using the grid's tooltip, the other using the column's tooltip.

I'd also prefer accessing cells by their contents as done in the other tests. Please also adapt the other added tests to do so.

sissbruecker (commented):
Some nits here and there, otherwise the change looks good

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -3895,7 +3883,7 @@ private void generateTooltipTextData(T item, JsonObject jsonObject) {
         JsonObject tooltips = Json.createObject();
 
         idToColumnMap.forEach((id, column) -> {
-            String cellTooltip = column.getTooltipGenerator().apply(item);
+            String cellTooltip = generateTooltipContent(item, column);
             if (cellTooltip != null) {
                 tooltips.put(id, cellTooltip);
             }
```
sissbruecker, line 3889: Looking at this again, I think we should optimize this. Currently, if you have like 50 columns, we would send the same tooltip 50 times, which can significantly increase the payload. We would also generate the tooltip 50 times, which could be a performance issue if there is database access involved.

I think we can put the grid tooltip into `tooltips` with a special key like `row`, and then in the Javascript tooltip generator below use this as a fallback if the object doesn't contain a tooltip for the specific column node id.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -3906,6 +3899,28 @@ private void generateTooltipTextData(T item, JsonObject jsonObject) {
...
+     *
+     * @param item
+     *            the item of the row where the tooltip would show
+     * @param column
+     *            the column where the tooltip should show
+     * @return the content of the tooltip this column would show with the
+     *         specified item, or {@code null} if no tooltip generator is
+     *         available at neither the column nor grid level
+     */
+    String generateTooltipContent(T item, Column<T> column) {
```
sissbruecker, line 3916: This method is not used anymore in the component, and the respective tests for it are obsolete.

sissbruecker (approved):

PR #4934, opened 2023-04-17
feat: add hierarchical scrollToIndex and scrollToEnd support

ugur-vaadin:
## Description

- Adds hierarchical `scrollToIndex` support to `TreeGrid`. 
- Updates `TreeGrid.scrollToEnd` to scroll to the last item instead of the last root item.
- Adds hierarchical `scrollToIndex` support to `TreeGridElement`. 
- Takes the parent items of the visible items into account when fetching a page. Tries to match the page number of the visible items, even if their parent is not visible but has the same depth as of the request. Previously, the the page number for the visible item was only considered if they were of the same depth as the request. This change prevents clearing the cache for the visible items, which can create a loop in some cases.
- Removes ignore annotations for related tests.
- Adds tests.

Fixes #3505

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [x] I have read the contribution guide: https://vaadin.com/docs/latest/contributing/overview
- [x] I have added a description following the guideline.
- [x] The issue is created in the corresponding repository and I have referenced it.
- [x] I have added tests to ensure my change is effective and works as intended.
- [x] New and existing tests are passing locally with my change.
- [x] I have performed self-review and corrected misspellings.

#### Additional for `Feature` type of change

- [x] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.

Issue #3505, opened 2021-07-08
Override scrollToIndex(number) in TreeGrid


sosa-vaadin:
### Describe your motivation

`TreeGrid` inherits the `scrollToIndex(int)` method from `Grid` but some complications arise from this. The complete structure of the Grid is needed to properly calculate how to scroll to an item using its index. One example is a node at index 3 with 5 children is expanded, the following node (previously at index 4) would move to index 9. When collapsed, the node at index 9 would move back to index 4.

### Describe the solution you'd like

We need to provide a reliable method that can consistently scroll to any item using its index. The tree should be traversed in a pre-order traversal manner.

### Describe alternatives you've considered

Making it possible to only scroll to root nodes. I'm not sure if this solution would work if any node is expanded since we would probably face the previous issue of having a node's index change when previous nodes are expanded or collapsed.

sirbris:
Any updates on this?

yuriy-fix:
We need to implement the Flow counterpart of the logic introduced in the `web-components` with https://github.com/vaadin/web-components/pull/5428. We would like to keep the existing API, so we should have two methods available:

```
public void scrollToIndex(int rowIndex)
public void scrollToIndex(int... indexes)
```

tomivirkki:
Also
- Update the implementation of [`scrollToEnd`](https://github.com/vaadin/flow-components/blob/fb3089cd8dbb1b7559fbcad0aaa9a4cec7bd6fc3/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java#L4390) as advised [here](https://github.com/vaadin/web-components/pull/5428#issuecomment-1415656675)
- Add API for scrolling to a hierarchical index to the `TreeGridElement` (TB)
- Re-enable the tests annotated with `@Ignore("Fix as part of https://github.com/vaadin/flow-components/issues/3505")`

PR #5003, opened 2023-04-28
test: extract tree grid test data generator

ugur-vaadin:
## Description

This PR extracts `TreeGrid` test data generation logic to a new class as suggested in a PR [comment](https://github.com/vaadin/flow-components/pull/4934#discussion_r1176570889).

Related to #3505

## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] Refactor

## Checklist

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/treegrid/it/TreeGridStringDataGenerator.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/treegrid/it/TreeGridStringDataGenerator.java
@@ -1,0 +1,62 @@
+package com.vaadin.flow.component.treegrid.it;
+
+import com.vaadin.flow.data.provider.hierarchy.TreeData;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+public class TreeGridStringDataGenerator {
+
+    private final TreeData<String> data = new TreeData<>();
+
+    private final Map<String, String> parentPathMap = new HashMap<>();
+
+    private Map<Integer, String> levelToNameMap;
+
+    public TreeGridStringDataGenerator() {
+        this(Map.of(0, "Granddad", 1, "Dad", 2, "Son"));
+    }
+
+    public TreeGridStringDataGenerator(Map<Integer, String> levelToNameMap) {
+        this.levelToNameMap = levelToNameMap;
+    }
+
+    public TreeData<String> generate(int... countPerLevel) {
+        List<String> itemsForLevel = null;
+        for (int level = 0; level < countPerLevel.length; level++) {
+            itemsForLevel = addItemsForLevel(levelToNameMap.get(level),
+                    countPerLevel[level], itemsForLevel);
+        }
+        return data;
+    }
+
+    private List<String> addItemsForLevel(String name, int numberOfItems,
+            List<String> parentItems) {
+        if (parentItems == null) {
+            return addItemsToParent(name, numberOfItems, null);
+        }
+        return parentItems.stream()
+                .map(parent -> addItemsToParent(name, numberOfItems, parent))
+                .flatMap(List::stream).collect(Collectors.toList());
+    }
+
+    private List<String> addItemsToParent(String name, int numberOfItems,
+            String parent) {
+        return IntStream.range(0, numberOfItems)
+                .mapToObj(index -> createItem(name, parent, index))
+                .collect(Collectors.toList());
+    }
+
+    private String createItem(String name, String parent, int index) {
+        String thisPath = Optional.ofNullable(parentPathMap.get(parent))
+                .map(path -> path + "/" + index).orElse("" + index);
+        String item = (name + " " + thisPath).trim();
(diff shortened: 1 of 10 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/treegrid/it/TreeGridStringDataGenerator.java
```diff
@@ -0,0 +1,62 @@
...
+
+public class TreeGridStringDataGenerator {
+
+    private final TreeData<String> data = new TreeData<>();
+
+    private final Map<String, String> parentPathMap = new HashMap<>();
+
+    private Map<Integer, String> levelToNameMap;
+
+    public TreeGridStringDataGenerator() {
```
vursen, line 20: The current API, which is based on a Map, may be difficult to understand, particularly when encountering `TreeGridStringDataGenerator` for the first time somewhere in the code. I wonder if using the builder pattern would improve it.

Example:
```java
TreeGridStringDataBuilder
  .addLevel(50)
  .addLevel(100)
  .addLevel(300)
  .build()
```
```java
TreeGridStringDataBuilder
  .addLevel(50, "Graddad")
  .addLevel(100, "Dad")
  .addLevel(300, "Son")
  .build()
```

What do you think?

ugur-vaadin: That was actually my initial approach. The reason I used the default names was the generator being used with the default names in all the cases but one. The depth was also never more than 3. This reduces the code to something like `new TreeGridStringDataGenerator().generate(3, 3, 3)`, which seemed brief and understandable. However, it is true that the builder approach is much cleaner in general, especially if it is to be used in new test pages. So, if you think that it is the better option here, I will update the PR.

vursen: I wonder if there is a real need for custom item names in the tests. Couldn't they be always named "Item 1", "Item 1/1", etc?

ugur-vaadin: We don't really need different names since we provide the tree path anyway. However, in the [case](https://github.com/vaadin/flow-components/pull/5003/files#diff-efe17d4338c514600b81bb1ac969051a60782d3e27abc13de6e7634bdafbb3c2L79-R69) where we use the name map, it is used to provide a longer name.

vursen: Agreed, let's leave it be 👍

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/treegrid/it/TreeGridStringDataGenerator.java
```diff
@@ -0,0 +1,62 @@
...
+        this(Map.of(0, "Granddad", 1, "Dad", 2, "Son"));
+    }
+
+    public TreeGridStringDataGenerator(Map<Integer, String> levelToNameMap) {
```
vursen, line 24: Could we use an array instead of a Map here?

```java
new TreeGridStringDataGenerator("Graddad", "Dad", "Son").generate(3, 3, 3)
```

ugur-vaadin: Yes, using a map is definitely unnecessary. I will change that. For names, if they are required in the constructor and don't have a default, I think then builder approach would make more sense. I don't have a strong opinion on which option to choose. It can also be modified somewhat easily afterwards. If you think that enforcing setting the names explicitly would make it clearer and more future-proof, I will convert it to builder.

vursen: Well, now I'm thinking that the Builder approach might be a better option after all. I checked and the item names were explicit before but now we are hiding them.

ugur-vaadin: I updated the branch

vursen (approved):

vursen:
Since this PR only affects tests, I would use the `test:` prefix rather than `refactor:`.

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/treegrid/it/TreeGridStringDataGenerator.java
vursen: Thanks, looks good now. 👍

Issue #4967, opened 2023-04-21
Grid selection working unexpectedly


tomivirkki:
### Description

The select all -checkbox no longer toggles all items if you've done manual item selections. This seems to be a regression from #4902.

Current behavior:

https://user-images.githubusercontent.com/1222264/233565497-f52e1aee-382d-4898-818e-d4c01b20a7e4.mp4

Expected behavior:

https://user-images.githubusercontent.com/1222264/233565487-1c60a64e-8188-44ff-9ab4-a3e10e4b9771.mp4

### Expected outcome

See above

### Minimal reproducible example

Open the "in-memory grid" example in the IT pages

### Steps to reproduce

1. select all items
2. deselect some items
3. toggle "select all" checkbox

### Environment

Vaadin version(s): 24.1-SNAPSHOT

### Browsers

Issue is not browser related

vursen:
Confirmed that the regression is indeed caused by https://github.com/vaadin/flow-components/pull/4902. Before that, deselection would deselect all items as expected.

PR #4902, opened 2023-03-30
fix: do not refresh items on select

ugur-vaadin:
## Description

When an item is selected via `Grid.select`, the provided item is cached in the selection model. However, that item is not necessarily the same item as in the grid. The ID match is enough for that item to be considered for selection. Then, the item is refreshed using the object in this cache. This can lead to incorrect data being sent to the client with `updateFlatData` for the regular `Grid`. 

This PR removes the refresh calls on both single and multi selection models. For the client-side `selected` property state of an item to be consistent, the property is now updated for the server-side selection/deselection calls as well.

Fixes #4861 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

- [x] I have read the contribution guide: https://vaadin.com/docs/latest/contributing/overview
- [x] I have added a description following the guideline.
- [x] The issue is created in the corresponding repository and I have referenced it.
- [x] I have added tests to ensure my change is effective and works as intended.
- [x] New and existing tests are passing locally with my change.
- [x] I have performed self-review and corrected misspellings.

PR #5018, opened 2023-05-02
fix: prevent adding duplicate items to selectedItems on selection

vursen:
## Description

The PR updates `$connector.doSelection` to prevent adding duplicate items to `selectedItems` which can happen when selecting the same item first through the client and then through the server. Previously, possible duplicate items were removed by `$connector.updateFlatItems` but as of https://github.com/vaadin/flow-components/pull/4902 that method is no longer called on selection. Duplicate items are problematic because `$connector.doDeselection` is currently designed so that it only removes the first occurrence of an item, which can lead to UI inconsistencies as was observed in the issue.

Fixes https://github.com/vaadin/flow-components/issues/4967

## Type of change

- [x] Bugfix


sissbruecker (approved):
LGTM. Please add target labels, looks like this should target 24.0 and 23.3.

vursen:
> looks like this should target 24.0 and 23.3.

Thanks, added but the PR anyway requires a manual cherry-pick because both `24.0` and `23.3` don't have web-test-runner tests set up.

PR #5023, opened 2023-05-05
chore: improve NpmPackage detection in version script

sissbruecker:
## Description

The script currently looks for any occurrence of `@NpmPackage` in Java files, and then tries to detect a version from that. This is a problem is `@NpmPackage` occurs somewhere in a comment like here: 
https://github.com/vaadin/flow-components/blob/3445a4fe6b08bd61988f2fbf4a43c6e8f74a2cf8/vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/LumoUtility.java#L33-L36

The script for bumping NPM package versions currently fails due to that.

This PR improves the script to also look for an opening parenthesis.


scripts/lib/versions.js
```diff
@@ -84,7 +84,7 @@ async function getVersionsJson() {
 }
 
 async function getAnnotations(){
-  const cmd = 'grep -r @NpmPackage ./vaadin*parent/*/src/*/java';
+  const cmd = 'grep -r @NpmPackage\\( ./vaadin*parent/*/src/*/java';
```
Artur-, line 87: Maybe even more reliable would be to match start-of-line -> possibly whitespace -> `@NpmPackage`. Then you avoid everything inside comments, even if it says `Use @NpmPackage("foo")`

sissbruecker: Thought about that, but then a multiline comment could also match.

Artur-: Yeah, I was thinking about javadoc only where you tend to have ` * ` on each comment line

Artur- (approved):

PR #5048, opened 2023-05-11
fix: add default for marker text baseline

sissbruecker:
## Description

Currently `textBaseline` in `TextStyle` has no default value, which causes synchronization into the OL `Text` class to fail if the developer does not provide a custom value. 

This change adds a default, and also adapts the sync logic to handle missing values.

## Type of change

- Bugfix


DiegoCardoso (changes requested):
I noticed that when you click on the "set default custom text style" button. the label disappears:

https://github.com/vaadin/flow-components/assets/262432/8a88a855-b057-4884-9042-36b74b1cb671

When I logged the `context.lookup.get(source.stroke)` statement in the `syncrhonizeText` function, it returned `undefined`, although there's a stroke defined for this text style.

sissbruecker:
Good catch, the stroke and fill needed to be explicitly added as children. Added a commit to fix it.

DiegoCardoso (approved):
LGTM

PR #5076, opened 2023-05-23
refactor: improve testbench api

ugur-vaadin:
## Description

This PR addresses some shortcomings and issues discovered during the Start integration of `SideNav`. 

Improvements regarding item access through the component:
- `includeChildren` option for `SideNavElement.getItems`
- querying items with the new `SideNavElement.getItemByPath` and `SideNavElement.getItemByLabel`
- getting the active item with the new `SideNavElement.getSelectedItem`

Improvements regarding the individual items:
- `includeChildren` option for `SideNavItemElement.getItems`
- a separate way of clicking the anchor via `SideNavItemElement.clickAnchor`
- `SideNavItemElement.navigate` was renamed to `SideNavItemElement.click` (overrides the regular `click`)
- getting the path via `SideNavItemElement.getPath`
- bug fix in `SideNavItemElement.getLabel`

Note: Should be back-ported to v24.1.

No related issue.

## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] Refactor

## Checklist

### DIFF
--- a/vaadin-side-nav-flow-parent/vaadin-side-nav-testbench/src/main/java/com/vaadin/flow/component/sidenav/testbench/SideNavElement.java
+++ b/vaadin-side-nav-flow-parent/vaadin-side-nav-testbench/src/main/java/com/vaadin/flow/component/sidenav/testbench/SideNavElement.java
@@ -41,7 +41,7 @@ public class SideNavElement extends TestBenchElement {
         return hasAttribute("collapsible");
     }
 
-    public void clickExpandButton() {
+    public void toggle() {
         final WebElement element = getWrappedElement().getShadowRoot()
                 .findElement(By.cssSelector("summary[part='label']"));
         executeScript("arguments[0].click();", element);
--- a/vaadin-side-nav-flow-parent/vaadin-side-nav-testbench/src/main/java/com/vaadin/flow/component/sidenav/testbench/SideNavElement.java
+++ b/vaadin-side-nav-flow-parent/vaadin-side-nav-testbench/src/main/java/com/vaadin/flow/component/sidenav/testbench/SideNavElement.java
@@ -51,6 +51,7 @@ public class SideNavElement extends TestBenchElement {
             throw new NoSuchElementException(
                     "Nav does not contain a toggle button", e);
         }
+        // click() on elements in shadow DOM does not work with Chrome driver
         executeScript("arguments[0].click();", element);
     }
 
--- a/vaadin-side-nav-flow-parent/vaadin-side-nav-testbench/src/main/java/com/vaadin/flow/component/sidenav/testbench/SideNavElement.java
+++ b/vaadin-side-nav-flow-parent/vaadin-side-nav-testbench/src/main/java/com/vaadin/flow/component/sidenav/testbench/SideNavElement.java
@@ -17,6 +17,7 @@ package com.vaadin.flow.component.sidenav.testbench;
 
 import java.util.List;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
@@ -32,16 +33,6 @@ import com.vaadin.testbench.elementsbase.Element;
 @Element("vaadin-side-nav")
 public class SideNavElement extends TestBenchElement {
 
-    public List<SideNavItemElement> getItems() {
-        // get only the direct vaadin-side-nav-item of this vaadin-side-nav
-        return wrapElements(findElements(By.xpath("vaadin-side-nav-item")),
-                getCommandExecutor())
-                .stream()
-                .map(testBenchElement -> testBenchElement
-                        .wrap(SideNavItemElement.class))
-                .collect(Collectors.toList());
-    }
-
     public String getLabel() {
         return $("span").attributeContains("slot", "label").first().getText();
     }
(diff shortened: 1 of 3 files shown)


vaadin-side-nav-flow-parent/vaadin-side-nav-testbench/src/main/java/com/vaadin/flow/component/sidenav/testbench/SideNavElement.java
```diff
@@ -56,4 +47,32 @@ public void clickExpandButton() {
         executeScript("arguments[0].click();", element);
     }
 
+    public List<SideNavItemElement> getItems(boolean includeChildren) {
```
sissbruecker, line 50: I think we should keep an overload without the boolean param that returns items without nested children.

vaadin-side-nav-flow-parent/vaadin-side-nav-testbench/src/main/java/com/vaadin/flow/component/sidenav/testbench/SideNavItemElement.java
```diff
@@ -53,17 +56,28 @@ public boolean isActive() {
...
     }
 
-    public String getLabel() {
-        final WebElement unnamedSlot = getWrappedElement().getShadowRoot()
-                .findElement(By.cssSelector("slot:not([name])"));
-        return (String) executeScript(
-                "return arguments[0].assignedNodes()[0].textContent;",
-                unnamedSlot);
+    @Override
+    public void click() {
```
sissbruecker, line 60: `navigate` would at least give you an idea what it does, with `click` you won't know if it triggers navigation or if it toggles the expanded state (or both?).

Maybe we should have `navigate` and `toggleExpanded`. The implementations could also be improved to check if the respective elements (anchor, toggle button) actually exist, and throw a meaningful exception otherwise. It looks like the methods for clicking the anchor / toggle button would currently just result in a script error.

vaadin-side-nav-flow-parent/vaadin-side-nav-testbench/src/main/java/com/vaadin/flow/component/sidenav/testbench/SideNavItemElement.java
```diff
@@ -53,17 +56,28 @@ public boolean isActive() {
...
+        click(1, 1);
+    }
+
+    public void clickAnchor() {
+        executeScript("arguments[0].click();", getWrappedElement()
+                .getShadowRoot().findElement((By.cssSelector("a"))));
     }
```
sissbruecker, line 67: If `click` does the same as `navigate` did before, why do we need this?

The click could probably be triggered through the Selenium API rather than running a script.

vaadin-side-nav-flow-parent/vaadin-side-nav-testbench/src/main/java/com/vaadin/flow/component/sidenav/testbench/SideNavElement.java
ugur-vaadin: I could not be sure about the naming. At first, I was thinking of going with something like `getDirectChildItems` and `getItemsRecursively`. Do you have a suggestion?

sissbruecker: I think the current name is fine, just add an overload and maybe tune the param name to use `nested` and `items`:
```java
List<SideNavItemElement> getItems();
List<SideNavItemElement> getItems(boolean includeNestedItems);
```

vaadin-side-nav-flow-parent/vaadin-side-nav-testbench/src/main/java/com/vaadin/flow/component/sidenav/testbench/SideNavItemElement.java
ugur-vaadin: During the Start integration, we discovered that `navigate` will not work with `click(1,1)` implementation when the drawer is hidden. However, the `click` does not trigger navigation at all, if not overridden. I don't have a quick solution to do them both as I haven't debugged it to see why it is the case. But in the unofficial `AppNav` component in Start had its separate TestBench classes and it used the logic of `getAnchor().click()` to navigate. The `click` not working by itself can be addressed separately I guess. Also, it might be a bug in the component itself.

So, I can 
- revert the name to `navigate`
- use the logic in `clickAnchor` to solve the hidden drawer issue 
- remove the `click` override
- use only `navigate` on tests

How does it sound?

vaadin-side-nav-flow-parent/vaadin-side-nav-testbench/src/main/java/com/vaadin/flow/component/sidenav/testbench/SideNavItemElement.java
ugur-vaadin: The thing is, `click` should do both. If an item is not navigable and has child items, a click should result in expanding the item. However, like I mentioned in the other thread, how the clicks are handled in the component needs to be debugged to be certain. It might be a bug. There is a test `clickNonNavigableParent_childNodesDisplayed` which checks this double behaviour. The reason I included both is to still have an API that covers the test cases, and also provide a direct way to handle navigation through the anchor.

Throwing a meaningful exception makes sense. There are enough lines that result in script errors already.

vaadin-side-nav-flow-parent/vaadin-side-nav-testbench/src/main/java/com/vaadin/flow/component/sidenav/testbench/SideNavElement.java
ugur-vaadin: Done.

PR #5120, opened 2023-06-05
Revert "refactor: avoid an extra render on grid connector confirm parent (#4894)"

### LINKED ISSUE #5045: Expanded TreeGrid with small page size does not display the correct items
### Description

I used a TreeGrid with page size 2 and expanded the items recursively. Expanding and scrolling causes unexpected behaviour which can only be corrected after scrolling far down and back up to clear the cache.

https://github.com/vaadin/flow-components/assets/106508695/959c0862-2832-45af-b730-bb0dda7fdc12

`TreeGridScrollToPage` can be used to reproduce the issues. Setting page size small like 2 will produce such results. Using different numbers of items and different combinations of expanding/scrolling will result in various unexpected behaviours. It is likely that the root cause is the same.

### Expected outcome

I would expect the grid to be expanded/scrolled correctly regardless of the page size.

```java
@Route("vaadin-grid/treegrid-regression")
public class TreeGridRegressionPage extends Div {

    public TreeGridRegressionPage() {
        TreeGrid<String> grid = new TreeGrid<>();
        grid.setPageSize(2);
        grid.addHierarchyColumn(String::toString).setHeader("Item");

        TreeData<String> data = new TreeGridStringDataBuilder()
                .addLevel("Granddad", 50).addLevel("Dad", 20)
                .addLevel("Son", 20).build();
        grid.setDataProvider(new TreeDataProvider<>(data));

        add(grid);
    }
}
```

### Steps to reproduce

1. Add the above page to `vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests`
2. Open it in the browser
3. Expand `Granddad 0`
4. Try to scroll back and forth
5. See that the grid displays incorrect items and some rows even appear blank.

### Environment

Vaadin version(s): 24.1
OS: MacOS

### Browsers

Issue is not browser related

DiegoCardoso:
## Description

This reverts commit b6ff3ffada160ad5ab90b3b95de8cd37e39fbf75 that causes the regression described in #5045.

I am planning to add a test with the same use case in a separate PR.

Fixes #5045

## Type of change

- [x] Bugfix
- [ ] Feature

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -889,6 +889,14 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
           }
           // Let server know we're done
           grid.$server.confirmParentUpdate(id, parentKey);
+
+          if (!grid.loading) {
+            grid.__confirmParentUpdateDebouncer = Debouncer.debounce(
+              grid.__confirmParentUpdateDebouncer,
+              microTask,
+              () => grid.__updateVisibleRows()
+            );
+          }
         });
 
         grid.$connector.confirm = tryCatchWrapper(function (id) {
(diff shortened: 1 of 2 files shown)


sissbruecker (approved):

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -889,6 +889,14 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
           }
           // Let server know we're done
           grid.$server.confirmParentUpdate(id, parentKey);
+
+          if (!grid.loading) {
+            grid.__confirmParentUpdateDebouncer = Debouncer.debounce(
+              grid.__confirmParentUpdateDebouncer,
+              animationFrame,
```
DiegoCardoso, line 896: For some reason, the `microTask` timing function was breaking one of the tests. Changing it to `animationFrame` fixes the issue.

tomivirkki:
> I am planning to add a test with the same use case in a separate PR.

Please don't forget to add the test case. Otherwise, the [bug](https://github.com/vaadin/flow-components/issues/5045) is very likely to reoccur during future connector refactorings

Issue #5064, opened 2023-05-16
GridProElement getLastVisibleRowIndex() throws exception when Grid is empty


l33the:
How to reproduce:

Have GridProElement with no content, and call getLastVisibleRowIndex().

Returns this:

```
java.lang.NullPointerException: Cannot invoke "java.lang.Long.intValue()" because the return value of "com.vaadin.flow.component.gridpro.testbench.GridProElement.executeScript(String, Object[])" is null

	at com.vaadin.flow.component.gridpro.testbench.GridProElement.getLastVisibleRowIndex(GridProElement.java:93)
org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
	at org.junit.jupiter.engine.extension.TimeoutExtension.intercept(TimeoutExtension.java:156)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestableMethod(TimeoutExtension.java:147)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestMethod(TimeoutExtension.java:86)
	at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker$ReflectiveInterceptorCall.lambda$ofVoidMethod$0(InterceptingExecutableInvoker.java:103)
	at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker.lambda$invoke$0(InterceptingExecutableInvoker.java:93)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$InterceptedInvocation.proceed(InvocationInterceptorChain.java:106)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.proceed(InvocationInterceptorChain.java:64)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.chainAndInvoke(InvocationInterceptorChain.java:45)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.invoke(InvocationInterceptorChain.java:37)
	at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker.invoke(InterceptingExecutableInvoker.java:92)
	at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker.invoke(InterceptingExecutableInvoker.java:86)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$invokeTestMethod$7(TestMethodTestDescriptor.java:217)
[...]

PR #5156, opened 2023-06-16
fix: Improve robustness of GridElement getLastVisibleRowIndex/getFirstVisibleRowIndex

TatuLund:
It is better to return -1 than throw NPE here

- Remove also outdated comment
- Make isRowInView public

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-testbench/src/main/java/com/vaadin/flow/component/grid/testbench/GridElement.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-testbench/src/main/java/com/vaadin/flow/component/grid/testbench/GridElement.java
@@ -77,11 +77,16 @@ public class GridElement extends TestBenchElement {
     /**
      * Gets the index of the first row which is at least partially visible.
      *
-     * @return the index of the first visible row
+     * @return the index of the first visible row, -1 if Grid is empty
      */
     public int getFirstVisibleRowIndex() {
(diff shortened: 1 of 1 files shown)


vursen:
https://github.com/vaadin/flow-components/issues/5064 seems related

vaadin-grid-flow-parent/vaadin-grid-testbench/src/main/java/com/vaadin/flow/component/grid/testbench/GridElement.java
```diff
@@ -181,8 +190,7 @@ public int getLastVisibleRowIndex() {
      * @return <code>true</code> if the row is at least partially in view,
      *         <code>false</code> otherwise
      */
-    private boolean isRowInView(int rowIndex) {
-        // Private for now because this seems to be slightly incorrect
+    public boolean isRowInView(int rowIndex) {
```
vursen, line 193: Could you please clarify what use-case you are solving by this change?

vursen (changes requested):
Here are changes that we request based on the internal discussion regarding this PR:

- fix NPE by returning -1 for both `GridElement` and `GridProElement` as they don't seem to share the respective methods (see the [issue](https://github.com/vaadin/flow-components/issues/5064) for GridPro).
- address the visibility and naming of the `isRowInView` method in a separate PR.

TatuLund:
@vursen done

vursen (approved):

PR #5246, opened 2023-07-11
refactor: reuse GridSelectionColumnBaseMixin in Flow Grid selection column

sissbruecker:
## Description

Reuse the `GridSelectionColumnBaseMixin` introduced with https://github.com/vaadin/web-components/pull/6143 in the `<vaadin-grid-flow-selection-column>` element.

Closes https://github.com/vaadin/web-components/issues/6103

## Type of change

- Refactoring

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/vaadin-grid-flow-selection-column.js
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/vaadin-grid-flow-selection-column.js
@@ -1,156 +1,97 @@
 import '@vaadin/grid/vaadin-grid-column.js';
 import { GridColumn } from '@vaadin/grid/src/vaadin-grid-column.js';
-{
-  class GridFlowSelectionColumnElement extends GridColumn {
+import { GridSelectionColumnBaseMixin } from '@vaadin/grid/src/vaadin-grid-selection-column-base-mixin.js';
 
-    static get is() {
-      return 'vaadin-grid-flow-selection-column';
-    }
+class GridFlowSelectionColumn extends GridSelectionColumnBaseMixin(GridColumn) {
 
-    static get properties() {
-      return {
+  static get is() {
+    return 'vaadin-grid-flow-selection-column';
+  }
 
-        /**
-         * Automatically sets the width of the column based on the column contents when this is set to `true`.
-         */
-        autoWidth: {
-          type: Boolean,
-          value: true
-        },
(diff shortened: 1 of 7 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/vaadin-grid-flow-selection-column.js
```diff
@@ -1,156 +1,97 @@
...
+   *
+   * @override
+   */
+  _defaultHeaderRenderer(root, _column) {
+    super._defaultHeaderRenderer(root, _column);
+    const checkbox = root.firstElementChild;
+    if (checkbox) {
+      checkbox.id = 'selectAllCheckbox';
     }
+  }
```
sissbruecker, line 43: This should be cleaned up as well, right now every grid defines this ID, which means there will be elements with duplicate IDs in the DOM.

Could use a class name that is set by the header renderer as alternative. However there is no TestBench method for getting the checkbox, so this would also be a breaking change as developers are probably relying on the ID as well.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/vaadin-grid-flow-selection-column.js
```diff
@@ -1,156 +1,97 @@
...
+  _selectAll() {
+    this.selectAll = true;
+    this.$server.selectAll();
+  }
 
-    _onDeselectEvent(e) {
-      if (e.detail.userOriginated) {
-        this.selectAll = false;
-      }
-    }
```
vursen, line 150: Did this work at all? I can't find neither `select` nor `deselect` events in the web component.

vursen: As of this PR, we preemptively update `this.selectAll` on the client-side once the user clicks Select All / Deselect All. I wonder if this change can affect anything.

sissbruecker: No those didn't work. Maybe it was part of the connector at some point 🤔.

Anyway, all it would have done would be to optimistically deselect the select all checkbox, so I adopted that into `deselectItem`. Doesn't seem to cause any harm with the existing tests.

vursen (approved):

Issue #5192, opened 2023-06-27
[TreeGrid] Client-side errors: e.item is undefined


sissbruecker:
### Description

Just opening the [drag&drop with filters example](https://vaadin.com/docs/latest/example?embed=grid-drag-drop-filters-wc.js&import=component/grid/grid-drag-drop-filters.ts) in the Grid documentation throws multiple errors in the browser console. Apart from that the example seems to work properly.

<details>
<summary>Error log</summary
```
There seems to be an error in Vaadin Grid:e.item is undefinedPlease submit an issue to https://github.com/vaadin/flow-components/issues/new/choose [example:1:1509](https://vaadin.com/docs/latest/example?embed=grid-drag-drop-filters-wc.js&import=component/grid/grid-drag-drop-filters.ts%20line%206%20%3E%20injectedScript)
    tryCatchWrapper https://vaadin.com/docs/latest/example?embed=grid-drag-drop-filters-wc.js&import=component/grid/grid-drag-drop-filters.ts line 6 > injectedScript:1
    _filterDragAndDrop https://vaadin.com/docs/latest/55985-e46cbe3b233f7c4ff68e.js:353
    filterDragAndDrop https://vaadin.com/docs/latest/55985-e46cbe3b233f7c4ff68e.js:353
    o https://vaadin.com/docs/latest/55985-e46cbe3b233f7c4ff68e.js:2
    filterDragAndDrop https://vaadin.com/docs/latest/55985-e46cbe3b233f7c4ff68e.js:353
    _dragDropAccessChanged https://vaadin.com/docs/latest/55985-e46cbe3b233f7c4ff68e.js:353
    K https://vaadin.com/docs/latest/229-7a9d543be452da2c1707.js:2
    L https://vaadin.com/docs/latest/229-7a9d543be452da2c1707.js:2
    _propertiesChanged https://vaadin.com/docs/latest/229-7a9d543be452da2c1707.js:2
    _propertiesChanged https://vaadin.com/docs/latest/vaadin/VAADIN/build/FlowClient-d5d5e377.js:3
    _flushProperties https://vaadin.com/docs/latest/229-7a9d543be452da2c1707.js:2
    _invalidateProperties https://vaadin.com/docs/latest/229-7a9d543be452da2c1707.js:2
    _setProperty https://vaadin.com/docs/latest/229-7a9d543be452da2c1707.js:2
    <anonym> https://vaadin.com/docs/latest/229-7a9d543be452da2c1707.js:2
[...]

PR #5040, opened 2023-05-09
fix: make empty grid clear loading state

sissbruecker:
## Description

WIP

## Type of change

- Bugfix

Issue #5249, opened 2023-07-11
Tree Grid loses data after moving item between branches


thevaadinman:
### Description

In the attached demo application, grab an item from one branch and move it to the other. Then move it back. The item vanishes from view.

This issue was sent to us through Expert Chat.

### Expected outcome

The item is not supposed to vanish when dragged and dropped.

### Minimal reproducible example

```java
package com.example.application;

import java.util.ArrayList;
import java.util.List;
import java.util.Objects;
import java.util.Optional;
import java.util.UUID;
import java.util.stream.Stream;

import com.vaadin.flow.component.grid.dnd.GridDropEvent;
import com.vaadin.flow.component.grid.dnd.GridDropMode;
import com.vaadin.flow.component.orderedlayout.VerticalLayout;
import com.vaadin.flow.component.treegrid.TreeGrid;
import com.vaadin.flow.data.provider.hierarchy.AbstractBackEndHierarchicalDataProvider;
import com.vaadin.flow.data.provider.hierarchy.HierarchicalQuery;
import com.vaadin.flow.function.SerializablePredicate;
import com.vaadin.flow.router.Route;

@Route("")
public class MainView extends VerticalLayout {

	private TreeGrid<TreeNode> treeGrid;
	private TreeDataProvider treeDataProvider;
	private List<TreeNode> draggedNodes = new ArrayList<>();

	public MainView() {
		setup();
	}

	private void setup() {
		treeGrid = new TreeGrid<>();
		treeGrid.addHierarchyColumn(TreeNode::getName);
		treeGrid.setRowsDraggable(true);
		treeGrid.setDropMode(GridDropMode.ON_TOP);

		treeGrid.addDragStartListener(event -> {
			draggedNodes.clear();
			draggedNodes.addAll(event.getDraggedItems());
		});

		treeGrid.addDragEndListener(event -> draggedNodes.clear());

		treeGrid.addDropListener(this::drop);

		TreeNode root = new TreeNode("root", null);
		populateTreeGrid(root);

		treeDataProvider = new TreeDataProvider(root);
		treeGrid.setDataProvider(treeDataProvider);

		addAndExpand(treeGrid);
	}

	private void populateTreeGrid(TreeNode parent) {
		buildSubTree(parent, "child", 2);
		parent.getChildren().forEach(child -> buildSubTree(child, "grandchild", 2));
	}

	private void buildSubTree(TreeNode parent, String prefix, int numberOfChildren) {
[...]

PR #5250, opened 2023-07-12
fix: check for null item in drag and drop filters

### LINKED ISSUE #5177: [Grid] Cannot read properties of undefined (reading 'dragDisabled')
### Description

I am making use of the Vaadin Grid Component. In some cases it only has 4 columns - but inside these columns I tend to have components and a few other controls.

I frequently encounter the error :
```
There seems to be an error in Vaadin Grid:
Cannot read properties of undefined (reading 'dragDisabled')
Please submit an issue to https://github.com/vaadin/flow-components/issues/new/choose 
```
But there is no associated error on the server side. Also I am using the @Push mechanism to update it sometimes (using the appropriate UI.access() method etc).

In some cases the application simply carries on. In other cases, the whole session for that user is 'broken'.

Unclear as to how to debug further and give more information here.

### Expected outcome

This error should not occur. 
Or at least if it does - should specify what can be done to remedy.
Certainly should not hang the application.

Wish I could provide the application code. But it is quite complex.
Happy to add additional logging as appropriate. Please advise.

### Steps to reproduce

Wish I could provide the application code. But it is quite complex.
Happy to add additional logging as appropriate. Please advise.

### Environment

Vaadin version(s): 23.3.14
OS: Windows and Linux. Chrome Browser primarily. But also Edge.

### Browsers

Chrome, Firefox, Edge

DiegoCardoso:
## Description

Fix the regression introduced by https://github.com/vaadin/flow-components/pull/5040 by adding a check for the `item` property in `row` in `dragFilter` and `dropFilter` methods. Sometimes the method can be invoked before `item` is defined for the row.

Fixes #5177

## Type of change

- [X] Bugfix
- [ ] Feature

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/DragAndDropGridIT.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/DragAndDropGridIT.java
@@ -15,6 +15,7 @@
  */
 package com.vaadin.flow.component.grid.it;
 
+import org.hamcrest.CoreMatchers;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
@@ -24,6 +25,12 @@ import org.openqa.selenium.WebElement;
 import com.vaadin.flow.component.grid.testbench.GridElement;
 import com.vaadin.tests.AbstractComponentIT;
 import com.vaadin.flow.testutil.TestPath;
+import org.openqa.selenium.logging.LogEntry;
+import org.openqa.selenium.logging.LogType;
+
+import java.util.List;
+import java.util.logging.Level;
+import java.util.stream.Collectors;
 
 @TestPath("vaadin-grid/drag-and-drop")
 public class DragAndDropGridIT extends AbstractComponentIT {
@@ -34,6 +41,12 @@ public class DragAndDropGridIT extends AbstractComponentIT {
     public void init() {
         open();
         grid = $(GridElement.class).first();
+    }
+
+    @Test
+    public void dragAndDropDefined_gridLoaded_noErrors() {
+        List<LogEntry> logs = getLogEntries(Level.SEVERE);
+        Assert.assertTrue(logs.isEmpty());
     }
 
     @Test
@@ -57,6 +70,7 @@ public class DragAndDropGridIT extends AbstractComponentIT {
 
     @Test
     public void noDropMode_dropOnRow_dropEventNotFired() {
+        click("no-drop-mode");
         fireDrop(3, "on-top");
         assertMessages("", "", "");
     }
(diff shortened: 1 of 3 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/DragAndDropGridIT.java
```diff
@@ -36,6 +43,12 @@ public void init() {
         grid = $(GridElement.class).first();
     }
 
+    @Test
+    public void dragAndDropDefined_gridLoaded_noErrors() {
+        List<LogEntry> logs = getLogEntries(Level.SEVERE);
+        Assert.assertTrue(logs.isEmpty());
```
sissbruecker, line 49:
```suggestion
        checkLogsForErrors();
```
We usually use this to check for client-side errors.

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/DragAndDropGridIT.java
```diff
@@ -24,6 +25,12 @@
 import com.vaadin.flow.component.grid.testbench.GridElement;
 import com.vaadin.tests.AbstractComponentIT;
 import com.vaadin.flow.testutil.TestPath;
+import org.openqa.selenium.logging.LogEntry;
+import org.openqa.selenium.logging.LogType;
+
+import java.util.List;
+import java.util.logging.Level;
+import java.util.stream.Collectors;
```
sissbruecker, line 33: A few unused imports here, will probably not require any new imports if the other suggestion is applied.

sissbruecker:
Let's also set target labels to match the previous PR that introduced the regression.

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/DragAndDropGridIT.java
DiegoCardoso: Removed.

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/DragAndDropGridIT.java
DiegoCardoso: Ah, wasn't aware of this helper method. Done.

vursen:
I'm currently checking if [this issue](https://github.com/vaadin/flow-components/issues/5249) is also caused by this regression. If that's the case, then the solution should probably be different.

vursen:
> I'm currently checking if https://github.com/vaadin/flow-components/issues/5249 is also caused by this regression. If that's the case, then the solution should probably be different.

No, it's not. 👍

DiegoCardoso:
This seems to also fix https://github.com/vaadin/flow-components/issues/5192

sissbruecker (approved):

vursen (approved):

Issue #4997, opened 2023-04-27
Dialog server side modality can block opened-changed of previously closed Dialog


probert94:
### Description

Consider having a `Dialog` with a `TextField`. The Dialog should close on outside click.
Now a `ValueChangeListener` that opens a new modal `Dialog` (for example a validation message) is added to the `TextField`.
If you now enter a value into the TextField and click outside the Dialog, a `ValueChangeEvent` is triggered (cause of the focus loss) at (almost) the same time as the dialog is closed cause of the outside click.
As the `ValueChangeEvent` occur a little earlier and causes a new modal to open, the `OpenedChangeEvent` is ignored (inert).
This causes the `Dialog` to stay "open" on the server side and if it is modal itself, this causes the whole `UI` to freeze, as all nodes are marked as inert.

The same issue has been reported several times (#4026, #4743) but those issues are closed as solved.
Unfortunately the fix does not seem to solve the issue (Tested with Vaadin 24.0.5 which should include the fix).

### Expected outcome

The `Dialog` should be closed on the server, not only on the client.

### Minimal reproducible example

```java
@PageTitle("Main")
@Route(value = "")
public class MainView extends HorizontalLayout {
    public MainView() {
    Button btnOpenDlg = new Button("Open", e -> openDlg1());
    add(btnOpenDlg);
  }
  private static void openDlg1() {
    Dialog dlg = new Dialog();
    dlg.setModal(true);
    dlg.setCloseOnOutsideClick(true);
    HorizontalLayout toolBar = new HorizontalLayout();
    Span title = new Span("Title");
    toolBar.addAndExpand(title);
    dlg.add(toolBar);
    VerticalLayout content = new VerticalLayout();
    TextField tfValue = new TextField("Value");
    tfValue.addValueChangeListener(e -> openDlg2(e.getValue()));
    content.add(tfValue);
    dlg.add(toolBar);
    dlg.add(content);
    dlg.addOpenedChangeListener(e -> System.out.println(e.isOpened() ? "Opened" : "Closed"));
    dlg.open();
  }
  private static void openDlg2(String value) {
    try {
      Thread.sleep(100);
    }
    catch (InterruptedException e) {
      throw new RuntimeException(e);
    }
    Dialog dlg = new Dialog();
    dlg.setModal(true);
    Div content = new Div();
    content.setText(value);
    dlg.add(content);
    dlg.open();
  }
}
```

Sidenote:
[...]

PR #5258, opened 2023-07-14
fix: keep dialog open when inert

sissbruecker:
## Description

Currently a dialog's opened state can desync between the client and the server, for example:
- on mouse down, a focus out is triggered, which opens a second modal dialog, which sets the first dialog to be inert
- on the subsequent click event, the dialog closes on the client, but the server will never process the opened-changed event from the client because the dialog is inert, so the server-side dialog stays open

This fix involves:
- prevent the client dialog overlay from closing itself on outside click and escape presses (should be the only ways to close a dialog on the client)
- instead, make a call to the server component to close itself - if the dialog has become inert in the meantime this call will never be processed and the dialog will stay open on both the client and the server
- close the dialog from the server, while still firing the opened changed event as coming from the client

Fixes https://github.com/vaadin/flow-components/issues/4997

## Type of change

- Bugfix

### DIFF
--- a/vaadin-dialog-flow-parent/vaadin-dialog-flow-integration-tests/src/test/java/com/vaadin/flow/component/dialog/tests/DialogTestPageIT.java
+++ b/vaadin-dialog-flow-parent/vaadin-dialog-flow-integration-tests/src/test/java/com/vaadin/flow/component/dialog/tests/DialogTestPageIT.java
@@ -63,9 +63,8 @@ public class DialogTestPageIT extends AbstractComponentIT {
 
         Assert.assertEquals("The open state of the dialog is true",
                 message.getText());
-        Assert.assertEquals(
-                "There should not be initial events before opening the dialog",
-                "Number of event is 0", eventCounterMessage.getText());
+        Assert.assertEquals("There should one event from opening the dialog",
+                "Number of event is 1", eventCounterMessage.getText());
         Assert.assertEquals("The event came from server",
                 eventSourceMessage.getText());
 
@@ -76,7 +75,7 @@ public class DialogTestPageIT extends AbstractComponentIT {
         checkDialogIsClosed();
         Assert.assertEquals("The open state of the dialog is false",
                 message.getText());
-        Assert.assertEquals("Number of event is 1",
+        Assert.assertEquals("Number of event is 2",
                 eventCounterMessage.getText());
         Assert.assertEquals("The event came from server",
                 eventSourceMessage.getText());
--- a/vaadin-dialog-flow-parent/vaadin-dialog-flow-integration-tests/src/test/java/com/vaadin/flow/component/dialog/tests/DialogTestPageIT.java
+++ b/vaadin-dialog-flow-parent/vaadin-dialog-flow-integration-tests/src/test/java/com/vaadin/flow/component/dialog/tests/DialogTestPageIT.java
@@ -64,7 +64,7 @@ public class DialogTestPageIT extends AbstractComponentIT {
         Assert.assertEquals("The open state of the dialog is true",
                 message.getText());
         Assert.assertEquals("There should one event from opening the dialog",
-                "Number of event is 1", eventCounterMessage.getText());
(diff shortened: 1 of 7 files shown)


vaadin-dialog-flow-parent/vaadin-dialog-flow-integration-tests/src/test/java/com/vaadin/flow/component/dialog/tests/DialogTestPageIT.java
```diff
@@ -63,9 +63,8 @@ public void dialogWithOpenedChangeListener() {
 
         Assert.assertEquals("The open state of the dialog is true",
                 message.getText());
-        Assert.assertEquals(
-                "There should not be initial events before opening the dialog",
-                "Number of event is 0", eventCounterMessage.getText());
+        Assert.assertEquals("There should one event from opening the dialog",
+                "Number of event is 1", eventCounterMessage.getText());
```
sissbruecker, line 67: Not sure if I'm missing something, but opening the dialog should fire an opened change event?

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/test/java/com/vaadin/flow/component/dialog/DialogOpenedChangeListenerTest.java
```diff
@@ -0,0 +1,97 @@
...
+import com.vaadin.flow.server.VaadinSession;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+public class DialogOpenedChangeListenerTest {
```
sissbruecker, line 14: Since the new implementation uses custom logic to dispatch the opened change event, added a new set of tests to verify that they are properly fired, that the is from client flag is correct, and that no duplicate events are fired.

vaadin-dialog-flow-parent/vaadin-dialog-flow-integration-tests/src/test/java/com/vaadin/flow/component/dialog/tests/DialogTestPageIT.java
sissbruecker: Yeah OK, there actually was a bug in that just adding a closed dialog to the DOM causes an opened change event. Added a fix here: https://github.com/vaadin/flow-components/pull/5258/commits/538ba4b3e37c0d49f5f6f978e45d102056e0d0cf

Still kept the updated assertions, so that the first actual change event changes the counter to `1`, rather then kepp it at `0`.

vaadin-dialog-flow-parent/vaadin-dialog-flow-integration-tests/src/test/java/com/vaadin/flow/component/dialog/tests/DialogTestPageIT.java
```diff
@@ -76,7 +75,7 @@ public void dialogWithOpenedChangeListener() {
         checkDialogIsClosed();
         Assert.assertEquals("The open state of the dialog is false",
                 message.getText());
-        Assert.assertEquals("Number of event is 1",
+        Assert.assertEquals("Number of event is 2",
```
ugur-vaadin, line 78: Nit: typo
```suggestion
        Assert.assertEquals("Number of events is 2",
```

ugur-vaadin (approved):

vaadin-dialog-flow-parent/vaadin-dialog-flow-integration-tests/src/test/java/com/vaadin/flow/component/dialog/tests/DialogTestPageIT.java
sissbruecker: Done

PR #5300, opened 2023-07-28
refactor: simplify getting cache by key

sissbruecker:
The `ItemCache.prototype.getCacheAndIndexByKey` and `grid.$connector.getCacheByKey` do not actually return the cache associated with that key, but its parent cache, which is rather confusing. In all usages, the code then extracts the child cache by key / index. 

We can simplify that by just using a single method that actually returns the cache for the key. That also makes `getCacheAndIndexByKey` obsolete, as the index from the result was only ever used to access the child cache again.

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -64,19 +64,18 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
             }
           });
 
-          ItemCache.prototype.getCacheAndIndexByKey = tryCatchWrapper(function (key) {
+          ItemCache.prototype.getCacheByKey = tryCatchWrapper(function (key) {
+            // Start looking in this cache
             for (let index in this.items) {
               if (this.grid.getItemId(this.items[index]) === key) {
-                return { cache: this, scaledIndex: index };
+                return this.itemCaches[index];
               }
             }
-            const keys = Object.keys(this.itemkeyCaches);
-            for (let i = 0; i < keys.length; i++) {
-              const expandedKey = keys[i];
-              const subCache = this.itemkeyCaches[expandedKey];
-              let cacheAndIndex = subCache.getCacheAndIndexByKey(key);
-              if (cacheAndIndex) {
-                return cacheAndIndex;
+            // Look through sub-caches
+            for (let index of Object.keys(this.itemCaches)) {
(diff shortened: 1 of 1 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -64,19 +64,18 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
             }
           });
 
-          ItemCache.prototype.getCacheAndIndexByKey = tryCatchWrapper(function (key) {
+          ItemCache.prototype.getCacheByKey = tryCatchWrapper(function (key) {
+            // Start looking in this cache
             for (let index in this.items) {
```
vursen, line 69: **idea:** This could be further optimized by iterating through `itemkeyCaches`, which contains only the items that are expanded.

sissbruecker: I was actually planning to remove `itemkeyCaches` after this PR, because they don't have any purpose at the moment.

sissbruecker: Your idea would indeed make them relevant, it's a trade-off between optimizing this loop and reducing complexity / confusion in the connector. I kind of prefer the second option 🙂.

vursen: Ah, that would be great.

vursen (approved):

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -773,8 +761,7 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
           }
           let cacheToClear = grid._cache;
           if (parentKey) {
-            const cacheAndIndex = grid._cache.getCacheAndIndexByKey(pkey);
-            cacheToClear = cacheAndIndex.cache.itemCaches[cacheAndIndex.scaledIndex];
+            cacheToClear = grid._cache.getCacheByKey(pkey);
```
tomivirkki, line 764: Would be ideal to have wtr test coverage for all refactored connector code just to be on the safe side. Many places in the connector, such as this, haven't yet been covered. This particular change is not too complex though so it's not an issue. Much of the code is also covered by ITs but it's impossible to tell exactly which parts.

PR #5301, opened 2023-07-28
refactor: remove item key caches

sissbruecker:
The item key caches are currently only created and then again removed in `clear` - there is no other usage. They can be removed.

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -23,8 +23,6 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
         // a server grid is initialized
         if (!isItemCacheInitialized) {
           isItemCacheInitialized = true;
-          // Storing original implementation of the method to be used for client
-          // side only grids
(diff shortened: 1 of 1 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -54,12 +54,7 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
           ItemCache.prototype.doEnsureSubCacheForScaledIndex = tryCatchWrapper(function (scaledIndex) {
```
vursen, line 54: This whole method override is now unnecessary since it 100% repeats the [original method](https://github.com/vaadin/web-components/blob/f5ff5170fbf8453095e6a711239dc7e6881699e0/packages/grid/src/vaadin-grid-data-provider-mixin.js#L74-L80) in the web component.

sissbruecker: I knew there was something I wanted to check... Perfect, thanks for the heads up. Removed.

Looks like Github is having issues with showing the commit though.

vursen: There are currently methods that invoke `doEnsureSubCacheForScaledIndex `. They need to be updated to invoke `ensureSubCacheForScaledIndexOriginal` instead.

vursen (approved):

PR #5304, opened 2023-07-31
refactor: remove unnecessary data communicator reset in Grid

### LINKED ISSUE #5303: DataChangeEvent is not fired when GridListDataView#setFilter is called
### Description

DataChangeEvent is not fired when GridListDataView#setFilter is called. Note, the event is fired when filter is set via DataProvider instead. This should be analogous.

### Expected outcome

Event is fired.

```java
    @Test
    public void dataView_setFilter_methodsUseFilteredData() {
        AtomicInteger refreshed = new AtomicInteger(0);
        String[] items = new String[] { "item1", "item2", "item3", "item4" };
        GridListDataView<String> dataView = grid.setItems(items);
        grid.getDataProvider().addDataProviderListener(e -> {
            refreshed.incrementAndGet();
        });

        dataView.setFilter(s -> s.endsWith("4"));
        Assert.assertEquals("Filter change did not fire DataChangeEvent", 1,
                refreshed.get());

        Assert.assertEquals("Filter was not applied to data size", 1,
                dataView.getItemCount());

        Assert.assertTrue("Expected item is missing from filtered data",
                dataView.contains(items[3]));
        Assert.assertFalse(
                "Item that should be filtered out is available in the data",
                dataView.contains(items[1]));

        Assert.assertEquals("Wrong item on row for filtered data.", items[3],
                dataView.getItem(0));
    }
```

### Steps to reproduce

Run the unit test

### Environment

Vaadin version(s): Vaadin 24.1.4
OS: 

### Browsers

_No response_

TatuLund:
Fixes #5303


sissbruecker:
I think this is a rather opinionated change, and personally I would not agree with it. The purpose of the `DataChangeEvent` is for notifying the component / data communicator about changes to the data, like added, removed, or updated items, which allows the component to update its presentation. Changing the filter affects the **view** over that data, similar to changing sort order, and as such should not trigger that event. This would also introduce new side-effects, such as closing the current editor when a filter changes.

As such, I'll close this PR and related issue. As for alternatives, for an application, which sets the filter in the first place, it should also be possible to do some other thing as part of the same step rather than adding some indirection via an event of the grid. Other than that, maybe the data view should fire an event if the viewed data changes.

TatuLund:
I re-opened the PR based @mshabarov conclusion.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -4663,7 +4663,8 @@ private void onInMemoryFilterOrSortingChange(
         updateInMemorySorting(sortComparator);
         updateInMemoryFiltering(filter);
 
-        dataCommunicator.reset();
+        // Call refreshAll as data may have changed due filter change
+        getDataProvider().refreshAll();
```
mshabarov, line 4667: I propose to move `refreshAll()` from here to the data view.
Then the `setFilter` will be like on the Flow side:
```
    @Override
    public AbstractListDataView<T> setFilter(SerializablePredicate<T> filter) {
        DataViewUtils.setComponentFilter(component, filter);
        fireFilteringOrSortingChangeEvent(filter,
                (SerializableComparator<T>) DataViewUtils
                        .getComponentSortComparator(component).orElse(null));
        refreshAll();
        return this;
    }
```
This would fix all similar issues in data view, not only the Grid.

mshabarov (changes requested):

mshabarov:
Same issue is raised here https://github.com/vaadin/flow-components/issues/5465. 
This confirms my theory for making the refresh all on the DataView side to fix all the components implementing DataView API.
I will make a pull request on the Flow side.

mshabarov:
But this PR is valid. We need to get rid of extra reset call in the Grid in the end.

mshabarov:
I've made the PR eventually https://github.com/vaadin/flow/pull/20263, but this PR needs a removal of `dataCommunicator.reset` and probably not only in Grid, but in other list components.

mshabarov (approved):

sissbruecker:
Looks like ComboBox is the only other component that uses DataCommunicator. There, setting a filter on a list data view results in 5 calls to `DataCommunicator.reset`, three of which come from `DataCommunicator` itself. While we could optimize away the additional calls from ComboBox itself, that would require untangling a lot of the reset logic that is called in various cases other than the filter changing. For now I'll assume that `DataCommunicator.reset` does not incur a major performance penalty, given that it's called multiple times by that class itself for a single action, and avoid refactoring internals of the component to optimize these calls.

sissbruecker (approved):

PR #5327, opened 2023-08-04
fix: reject emails with double dot between domain and tld

vursen:
## Description

As indicated in issue https://github.com/vaadin/flow/issues/17333, this PR changes the validation pattern of an email address by invalidating the dot-dot in the domain name.

In PR https://github.com/vaadin/flow/pull/17353, I made EmailValidator.PATTERN public to use it in EmailField class. So the pattern is not longer duplicates between EmailValidator and EmailField

Fixes vaadin/flow#17333

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/EmailField.java
+++ b/vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/EmailField.java
@@ -27,6 +27,7 @@ import com.vaadin.flow.data.binder.Binder;
 import com.vaadin.flow.data.binder.ValidationStatusChangeEvent;
 import com.vaadin.flow.data.binder.ValidationStatusChangeListener;
 import com.vaadin.flow.data.binder.Validator;
+import com.vaadin.flow.data.validator.EmailValidator;
 import com.vaadin.flow.data.value.ValueChangeMode;
 import com.vaadin.flow.shared.Registration;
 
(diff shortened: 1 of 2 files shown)


vursen:
@gronono It seems that the web component's [pattern](https://github.com/vaadin/web-components/blob/2c024e8fd462d178430418f76a61f498fb549998/packages/email-field/src/vaadin-email-field.js#L42) is prone to the same issue, so it needs a fix too.

gronono:
> @gronono It seems that the web component's [pattern](https://github.com/vaadin/web-components/blob/2c024e8fd462d178430418f76a61f498fb549998/packages/email-field/src/vaadin-email-field.js#L42) is prone to the same issue, so it needs a fix too.

Ok. I will do it early next week.

vursen:
It seems my assumption about the web component was wrong. It already contains and passes the [corresponding test](https://github.com/vaadin/web-components/blob/2c024e8fd462d178430418f76a61f498fb549998/packages/email-field/test/validation.common.js#L29).

gronono:
> It appears that my concern about the web component's pattern was unfounded. The web component already includes and passes the [corresponding test](https://github.com/vaadin/web-components/blob/2c024e8fd462d178430418f76a61f498fb549998/packages/email-field/test/validation.common.js#L29).

There is something wrong with the test.

With the current pattern, 'email@example..com' should be valid:
```javascript
const pattern = '^([a-zA-Z0-9_\\.\\-+])+@[a-zA-Z0-9\\-.]+\\.[a-zA-Z0-9\\-]{2,}$'
const regex = new RegExp(pattern);
console.log('email@example.com', regex.test('email@example.com'))
console.log('email@example..com', regex.test('email@example..com'))
```
Both tests show true. So the regex accepts them.

gronono:
It seems that the browser adds a default behavior on top of the pattern.

In this example, we should accept any values. But the field becomes invalid for an incorrect email address.
```html
<label for="emailAddress">Email</label><br />
<input
  id="emailAddress"
  type="email"
  pattern=".*" />
```
```css
input:invalid {
  border: 1px solid red
}

input:valid {
  border: 1px solid green
}
```

vursen:
Oh, you are right. The browser appears to match the entered emails against its [own regular expression](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email#basic_validation) which is more extensive. This makes me wonder if we should consider using that regular expression for the Flow component as well and if a predefined pattern is needed at all in the web component.

vursen:
After further consideration, I'm inclined to believe that the potential difference in client-side email validation due to the browser using a different regular expression does not hinder this PR from merging. When using the Flow counterpart, the web component can only trigger server-side validation, while its own validation results are ignored. Thus, removing the pattern from the web component is a separate topic, which doesn't affect the Flow component.

sissbruecker (approved):

PR #5329, opened 2023-08-04
feat: add event listeners for Grid DataProvider change

### LINKED ISSUE #5324: Grid: allow registering event listeners to notify them when Grid DataProvider change
### Describe your motivation

Currently there's no way to perform any actions when Grid DataProvider change, other than overriding Grid.setDataProvider() method. If that's not feasible, the alternative would be triggering an event when DataProvider changes.

In order to perform some computations when the DataProvider for Grid is set or changed, it'd be useful to have the possibility to register listeners and notify them when this change occurs.

A use case could be performing some actions on the column header content when Grid DataProvider is set. For example, a column which values are rendered as checkboxes, and has a Select All checkbox on its header.

Consider adding an API to register listeners so they get notified when Grid's DataProvider is set:

```
public Registration addDataProviderChangeListener(
            ComponentEventListener<DataProviderChangeEvent<T>> listener) {
   return addListener(DataProviderChangeEvent.class, (ComponentEventListener) listener);
}

public void setDataProvider(DataProvider<T, ?> dataProvider) {
   ...
   fireEvent(new DataProviderChangeEvent<>(this));
}

```

_No response_

### Additional context

Grid web component has already a "data-provider-changed" event which is only triggered the first time DataProvider is set from Java.

flang:
## Description

Fire a `DataProviderChangeEvent` event to notify listeners when Grid DataProvider changes.

Closes #5324

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [] I have not completed some of the steps above and my pull request can be closed immediately.

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridDataProviderChange.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridDataProviderChange.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2000-2023 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


sissbruecker:
This feels like moving application concerns (How can I get notified about changes to the data provider in one part of my application, after setting it in a different part of the application) into the component. In general, this should be doable without doing changes to the core component, or overriding it. Other than that, implementing such an event might be more useful in Flow's data communicator rather than in a specific component. As such I'll close the PR.

PR #5339, opened 2023-08-09
docs: state that getColumn() returns null for multi-selection column

### LINKED ISSUE #5337: ItemClickEvent.getColumn() returns null when a selction column of a grid is clicked.
### Description

An ItemClickListener was added to a grid with multi selection mode enabled . To reduce falsely triggering the click event, when a selection checkbox is slightly missed  the click action shouldn't fire if the column is the selection column. ItemClickEvent.getColumn() returns null for the selection column even though the documentation states otherwise. 

### Expected outcome

ItemClickEvent.getColumn() should either state that it can return null for internal created columns as the multi selection column or the selection column should be reachable and clearly identifiable with functions like grid.getColumns() etc.

```java

Grid<Person> grid = new Grid<>(Person.class, false);
grid.addColumn(Person::getFirstName).setHeader("First name");
grid.addColumn(Person::getLastName).setHeader("Last name");
grid.addColumn(Person::getEmail).setHeader("Email");

List<Person> people = DataService.getPeople();
grid.setItems(people);

dataGrid.addItemClickListener(event -> System.out.println(event.getColumn());

```

### Steps to reproduce

1. Add Multi Selection mode and ItemClickListener to grid.
2. In the Clickevent Action call ItemClickEvent.getColumn
3. Click th selection column beside its checkbox.

### Environment

Vaadin version(s): 23.3.2
OS: Windows

### Browsers

Edge

TatuLund:
Fixes #5337

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/ItemClickEvent.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/ItemClickEvent.java
@@ -111,7 +111,8 @@ public class ItemClickEvent<T> extends ClickEvent<Grid<T>> {
     /**
      * Gets the column that was clicked.
      *
-     * @return the clicked column, not {@code null}
+     * @return the clicked column, {@code null} if clicked column was the multi
+     *         select column.
      */
     public Grid.Column<T> getColumn() {
         return column;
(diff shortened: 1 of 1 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/ItemClickEvent.java
```diff
@@ -111,7 +111,8 @@ public T getItem() {
     /**
      * Gets the column that was clicked.
      *
-     * @return the clicked column, not {@code null}
+     * @return the clicked column, {@code null} if clicked column was the multi
+     *         select column.
```
vursen, line 115:
```suggestion
     * @return the clicked column or {@code null} if it was the multi selection column.
```

vursen (approved):

Issue #1027, opened 2021-02-18
refreshItem(startNode, refreshChildren) does NOT refresh children!


mmoser18:
Calling the method` tree.getDataProvider().refreshItem(node, true);` will NOT refresh the children. The second parameters seems ignored. 

Maybe the cautious phrasing in the JavaDocs "_in a hierarchical DataProvider it is supposed to refresh all of the children as well in case 'refreshChildren' is true._" should have been a warning. It might "be supposed to" but it **does not**. 

Or then the default `com.vaadin.flow.data.provider.hierarchy.TreeDataProvider` is not a hierachical DataProvider as one would expect although it explicitly `extends AbstractHierarchicalDataProvider`. So, what *is* then a hierarchical data provider?

stefanuebe:
This sample shows a simple tree with a 3 level hierarchy. Every items has a name and a "counter" variable. 

There are 2 buttons, which increase the counter by 1 for either all items or the selected items plus its children. 

The "+ All" works fine. It increases all counters and calls a refreshAll. 
The "+ Selected" does not. It also increases all counters (of the selected item and its children) and calls a refreshItem with children. But the refreshing only occurs for the selected item and its direct children. Any grandchildren are not refreshed.

```java
@Route("refresh-issue")
public class TreeGridRefreshExpandedIssueView extends VerticalLayout {

    public TreeGridRefreshExpandedIssueView() {

        int counter = 1;

        List<Model> items = new LinkedList<>();

        for (int i = 0; i < 10; i++) {
            Model modelI = new Model("Item " + i, counter++);
            items.add(modelI);
            for (int j = 0; j < 3; j++) {
                Model modelJ = new Model("Item " + i + "." + j, counter++);
                modelI.getChildren().add(modelJ);
                for (int k = 0; k < 3; k++) {
                    Model modelK = new Model("Item " + i + "." + j + "." + k, counter++);
                    modelJ.getChildren().add(modelK);
                }
            }
        }

        TreeGrid<Model> grid = new TreeGrid<>();
        grid.addHierarchyColumn(Model::getName).setResizable(true).setHeader("Name");
        grid.addColumn(Model::getCounter).setResizable(true).setHeader("Counter");
        grid.setDataProvider(new MyDataProvider(items));

        Button refrAll = new Button("All", VaadinIcon.PLUS.create(), event -> {
            for (Model item : items) {
                increase(item);
            }
            grid.getDataProvider().refreshAll();
        });
[...]

PR #5340, opened 2023-08-09
fix: handle added and removed nodes when refreshing tree nodes

sissbruecker:
## Description

Refreshing a tree node from the server while using the refresh children option currently leads to the grid not displaying the newly added child nodes. For updates initiated from the server, the current connector logic only handles updating existing child items / rows, but does not properly handle changes to the size of the parent item.

This fixes the connector to update the respective grid sub-cache size and effective grid size if it has changed, which causes grid to render the newly added items.

Fixes https://github.com/vaadin/flow-components/issues/5249
Fixes https://github.com/vaadin/flow-components/issues/1027

## Type of change

- Bugfix


sissbruecker:
Also tagged https://github.com/vaadin/flow-components/issues/1027 as fixed. The OP doesn't provide a reproduction, so it's not clear what the original issue was. From testing I can confirm that updating, adding and removing direct children works. Nested hierarchical updates don't seem to work, but the `refreshItem` API also doesn't promise that.

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/treegrid/it/TreeGridChangingHierarchyIT.java
```diff
@@ -138,6 +138,23 @@ public void remove_item_from_root() {
         Assert.assertEquals("b", grid.getCell(0, 0).getText());
     }
 
+    @Test
+    public void addChildrenToExpandedNode_childrenDisplayed() {
```
tomivirkki, line 142: The issue is fixed on the root level, but it still remains in the sub-levels:

https://github.com/vaadin/flow-components/assets/1222264/ce79ed3f-9f24-4622-8158-9f93a729011e

tomivirkki (approved):

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/treegrid/it/TreeGridChangingHierarchyIT.java
sissbruecker: Sure, done.

PR #5394, opened 2023-08-29
feat: add FontIcon component

tomivirkki:
## Description

Adds `FontIcon` component, which can be used to render glyphs from an icon font collection.

Related to https://github.com/vaadin/web-components/issues/6212

## Type of change

- [ ] Bugfix
- [x] Feature

### DIFF
--- a/vaadin-icons-flow-parent/vaadin-icons-flow-integration-tests/src/test/java/com/vaadin/flow/component/icon/tests/FontIconIT.java
+++ b/vaadin-icons-flow-parent/vaadin-icons-flow-integration-tests/src/test/java/com/vaadin/flow/component/icon/tests/FontIconIT.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2000-2023 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component.icon.tests;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.openqa.selenium.By;
+
+import com.vaadin.flow.testutil.TestPath;
+import com.vaadin.tests.AbstractComponentIT;
+
+@TestPath("vaadin-icons/font-icons")
+public class FontIconIT extends AbstractComponentIT {
+
+    @Before
+    public void init() {
+        open();
+    }
+
+    @Test
+    public void fontAwesomeIcon() {
+        var icon = findElement(By.tagName("vaadin-icon"));
+        // Get the pseudo element's width and height separated by a space
+        var beforeBounds = executeScript(
+                """
+                        const { width, height } = window.getComputedStyle(arguments[0], '::before');
+                        return `${width} ${height}`;
+                        """,
+                icon);
+
+        Assert.assertEquals("21px 24px", beforeBounds);
+    }
+}
--- a/vaadin-icons-flow-parent/vaadin-icons-flow-integration-tests/src/test/java/com/vaadin/flow/component/icon/tests/FontIconIT.java
+++ b/vaadin-icons-flow-parent/vaadin-icons-flow-integration-tests/src/test/java/com/vaadin/flow/component/icon/tests/FontIconIT.java
@@ -32,9 +32,9 @@ public class FontIconIT extends AbstractComponentIT {
     }
 
     @Test
-    public void fontAwesomeIcon() {
-        var icon = findElement(By.tagName("vaadin-icon"));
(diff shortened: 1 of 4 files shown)


vaadin-icons-flow-parent/vaadin-icons-flow/src/test/java/com/vaadin/flow/component/icon/tests/FontIconTest.java
```diff
@@ -0,0 +1,111 @@
...
+import com.vaadin.flow.component.icon.FontIcon;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class FontIconTest {
+
+    @Test
+    public void constructor_hasFont() {
+        var icon = new FontIcon("fa-solid", "fa-user");
```
knoobie, line 27: Would it be worth to create a IT test with font awesome integration to ensure that the whole chain with flow -> web components -> flow components work? Testbench could even be used to create and compare screenshots later on to ensure that not the DOM but also the visual appearance is correct.

tomivirkki: I did verify this manually in a test application but an IT would be a good addition here as well.

tomivirkki: Note that such ITs can't be added before the Flow Components are using the next alpha version of the web components (I needed to manually hack the test app to use the updated vaadin-icon).

vaadin-icons-flow-parent/vaadin-icons-flow/src/test/java/com/vaadin/flow/component/icon/tests/FontIconTest.java
```diff
@@ -0,0 +1,111 @@
...
+        Assert.assertNull(icon.getElement().getProperty("char"));
+    }
+
+    @Test
+    public void setColor_hasColor() {
+        var icon = new FontIcon();
+        icon.setColor("red");
+        Assert.assertEquals("red", icon.getColor());
+        Assert.assertEquals("red", icon.getElement().getStyle().get("color"));
+    }
```
DiegoCardoso, line 110: Should there also be a test that checks that calling `icon.setColor(null)` removes the color?

tomivirkki: Added

vaadin-icons-flow-parent/vaadin-icons-flow/src/test/java/com/vaadin/flow/component/icon/tests/FontIconTest.java
tomivirkki: Seems that the current alpha already had the `font` API for icon available, so I was able to add the IT

vaadin-icons-flow-parent/vaadin-icons-flow-integration-tests/src/test/java/com/vaadin/flow/component/icon/tests/FontIconIT.java
```diff
@@ -0,0 +1,48 @@
...
+        // Get the ::before element's width and height separated by a space
+        var beforeBounds = executeScript(
+                """
+                        const { width, height } = window.getComputedStyle(arguments[0], '::before');
+                        return `${width} ${height}`;
+                        """,
+                icon);
+
+        // Expect the icon size to match FontAwesome "user" icon's size
+        Assert.assertEquals("21px 24px", beforeBounds);
```
knoobie, line 46: Just a question / note: Ensures the Vaadin Icon that all icon's have the same width or has the Developer apply the fa-fw class name? (e.g. fa-exclamation is really narrow as example) 

How does some apply other font awesome related styles like fa-2xl or fa-rotate? Can those be directly applied to the Vaadin Icon?

tomivirkki: The `FontIcon` is sized the same as the regular `Icon` component, so by default, it is 24px/24px (`--lumo-icon-size-m`) and it's size can be changed with `setSize()`.

A font icon inside the `<vaadin-icon>` element will by default take 100% of the host's height, but its width is only determined by the icon's own aspect ratio (in the `fa-user`'s case, the width inside a 24px high icon becomes 21px etc). The icon is, by default, horizontally centered inside the host.

I added an outline around the `<vaadin-icon>` elements in the following examples:

Screenshot 2023-08-30 at 9 45 27

The plan is not to support all of FontAwesome's features specifically, but many of the utility classes still work fine, and they can be passed to the `FontIcon`'s constructor, or to `setFont()` method.

Screenshot 2023-08-30 at 9 46 49

Screenshot 2023-08-30 at 9 47 23

Some features, such as the sizing classes (`fa-xs`, `fa-4x`) don't work, see https://github.com/vaadin/web-components/pull/6418. You can instead size the `FontIcon` the same as the `Icon` component.

Screenshot 2023-08-30 at 10 03 35

knoobie: Thanks for the comment! This are all really valuable information for the docs later on :)

DiegoCardoso (approved):

Issue #5229, opened 2023-07-06
[Grid] Restore scrollToIndex performance enhancement


sissbruecker:
We reverted a performance enhancement for `scrollToIndex` as it introduced a regression when using it after adding items:
- https://github.com/vaadin/flow-components/pull/4854
- https://github.com/vaadin/flow-components/issues/5187

The root cause for the regression might actually be in the data communicator as mentioned here: https://github.com/vaadin/flow-components/issues/5187#issuecomment-1607524557. After the Flow team had a chance to take a look at the data communicator issue, we should look into restoring the performance enhancement.

sissbruecker:
Depends on https://github.com/vaadin/flow/issues/17288

sissbruecker:
After further investigation, it turns out this is not related to data communicator, and instead caused by two issues related to the performance enhancement:

**Calling `DataCommunicator.setRequestedRange` with a start index that does not align with the grid's page size**

This one is straightforward, as there is a check in the grid's connector that throws an error if the start index of a data range does not align with the page size:
https://github.com/vaadin/flow-components/blob/b002fbad41ec0207b69ed9d6fc4ef20b9c90e6a6/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js#L642-L644

This was the original cause of the error observed in the bug issue. The fix would be to calculate a proper start index when requesting a range.

**Calling `DataCommunicator.setRequestedRange` on the server does not update the cached requested ranges in the grid's connector**

Even when requesting data from a valid page start index, the grid only displays rows from the page that was scrolled to, but rows from the surrounding pages are empty, and are also not loaded when scrolling up or down. The problem is that calling `setRequestedRange` on the server desynchronizes the range configured in the data communicator with the range configured in the grid connectors `lastRequestedRanges`. So the data communicator will only return data from that range, and the grid connector is still convinced that it has configured the data communicator to return a range that includes additional pages around the one that was scrolled to, so it doesn't trigger another load even though the grid is missing data.

To properly fix this, we would somehow need to make sure that both client and server request the same range for a scroll index. The connector's implementation for that is here: https://github.com/vaadin/flow-components/blob/b002fbad41ec0207b69ed9d6fc4ef20b9c90e6a6/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js#L295-L342

[...]

PR #5228, opened 2023-07-06
Revert "refactor: preload items on grid scroll to index (#4854)"

sissbruecker:
## Description

Reverts #4854 which introduced a regression when using `scrollToIndex` and adds a respective test.

Fixes https://github.com/vaadin/flow-components/issues/5187

PR #5459, opened 2023-09-12
refactor: preload items on grid scroll to index

tomivirkki:
## Description

This PR re-implements preloading of items on scrollToIndex. The [previous implementation](https://github.com/vaadin/flow-components/pull/4854) was [reverted](https://github.com/vaadin/flow-components/pull/5228) due to the regressions it caused.

The new implementation:
- Makes sure that the preloaded page ranges are valid on the root level based on the current page size. For example, if you scroll to index 25 and the page size is 50, 25-75 is not a valid page range to preload because the page doesn't start at index 25. In this case, the grid should preload a range from 0-100 (pages 0-50 and 50-100).
- Makes sure the `lastRequestedRanges` record for the root level is cleared if it ends up in an invalid state where the last requested range is only partially covered (For example, grid web component expects 4 pages but only 2 are preloaded. If the property was not cleared, the connector would be blocked from making another data request to the server for the remaining 2 pages).

Fixes https://github.com/vaadin/flow-components/issues/5229

## Type of change

Performance enhancement

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridScrollToIT.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridScrollToIT.java
@@ -93,15 +93,19 @@ public class GridScrollToIT extends AbstractComponentIT {
     public void grid_addItem_scrollToIndex_twice() {
         $("button").id("add-row-and-scroll-to-index").click();
         // Wait until finished loading
-        waitUntil(e -> !grid.getRow(grid.getFirstVisibleRowIndex()).hasAttribute("loading"));
+        waitUntil(e -> !grid.getRow(grid.getFirstVisibleRowIndex())
+                .hasAttribute("loading"));
 
         $("button").id("add-row-and-scroll-to-index").click();
-        waitUntil(e -> !grid.getRow(grid.getFirstVisibleRowIndex()).hasAttribute("loading"));
+        waitUntil(e -> !grid.getRow(grid.getFirstVisibleRowIndex())
+                .hasAttribute("loading"));
 
         // Find the content element of the first visible row cell
-        var slot = grid.getCell(grid.getFirstVisibleRowIndex(), 0).findElement(By.tagName("slot"));
-        var content = grid.findElement(By.cssSelector(
-                "vaadin-grid-cell-content[slot='" + slot.getPropertyString("name") + "']"));
+        var slot = grid.getCell(grid.getFirstVisibleRowIndex(), 0)
+                .findElement(By.tagName("slot"));
+        var content = grid
+                .findElement(By.cssSelector("vaadin-grid-cell-content[slot='"
+                        + slot.getPropertyString("name") + "']"));
         // Expect the content element to be displayed
         Assert.assertTrue(content.isDisplayed());
     }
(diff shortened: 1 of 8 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -4538,6 +4538,31 @@ private void fireColumnReorderEvent(List<Column<T>> columns) {
      *            zero based index of the item to scroll to in the current view.
      */
     public void scrollToIndex(int rowIndex) {
```
tomivirkki, line 4540: The following is covered by the unit tests in `GridScrollTest.java`

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -919,6 +919,18 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
...
+          if (Object.keys(rootPageCallbacks).length) {
+            // There are still unresolved callbacks waiting for data to the root level,
+            // which means that the range grid requested items for was only partially filled.
+            //
+            // This can happen for example if you preload some items without knowing exactly
+            // how many items the grid web component is going to request.
+            //
+            // Clear the last requested range for the root level to unblock
+            // any possible data requests for the same range in fetchPage.
+            delete lastRequestedRanges[root];
```
tomivirkki, line 931: This change is covered by the new WTR tests in `grid-connector.test.ts` (and the ITs)

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridScrollToIT.java
```diff
@@ -89,4 +89,40 @@ public void grid_addItem_scrollToIndex() {
...
+                .hasAttribute("loading"));
+
+        // Find the content element of the first visible row cell
+        var slot = grid.getCell(grid.getFirstVisibleRowIndex(), 0)
+                .findElement(By.tagName("slot"));
+        var content = grid
+                .findElement(By.cssSelector("vaadin-grid-cell-content[slot='"
+                        + slot.getPropertyString("name") + "']"));
+        // Expect the content element to be displayed
+        Assert.assertTrue(content.isDisplayed());
```
vursen, line 110: I'm a bit confused what this test actually does. If the cell content is displayed (I assume it means no display: none), does it gurantee that the content is actually there? Would it make sense to check for the cell text instead?

tomivirkki: This also checks `visibility: hidden` which is applied to the content in the case a row `<tr>` is in a loading state:

Screenshot 2023-09-18 at 13 16 47

Checking for the text would not be meaningful in the case of the error (missing changes to the connector) because the text content would actually be correct. The problem is that the content is invisible due to the row being stuck in a loading-state.

vursen: Ah, I see. Would it be enough to just wait for the row to exit the loading state then, similarly to the test below?

tomivirkki: Technically, it would be enough, but I kind of considered checking the `loading` attribute of an element inside the grid's shadow root as an internal detail, whereas the real user-facing symptom of the issue is the content not being displayed.

vursen (approved):

PR #5484, opened 2023-09-20
fix: ignore removed nodes in flow component directive

tomivirkki:
## Description

There's currently an issue in flow-component-directive that can cause an infinite loop in case a re-render is requested for a node that has already been removed from the Flow client registry.

The bug can be reproduced on the `/vaadin-grid/treegrid-component-renderer` IT page by sorting the first column. The issue is sneaky because the user doesn't necessarily notice there's an infinite loop running behind the scene, but you can catch it with a debugger.

This PR fixes the issue by storing the id of a node render retry timeout, and then clearing the timeout whenever the directive is requested to update (and also on disconnect).

## Type of change

Bugfix

### DIFF
--- a/vaadin-renderer-flow-parent/vaadin-renderer-flow-integration-tests/test/flow-component-renderer.test.ts
+++ b/vaadin-renderer-flow-parent/vaadin-renderer-flow-integration-tests/test/flow-component-renderer.test.ts
@@ -138,7 +138,7 @@ describe('flow-component-renderer', () => {
     component.nodeId = 1;
     await nextFrame();
 
-    // getByNodeId should only have been called for node id 1 only
+    // getByNodeId should only have been called for node id 1
     getByNodeId.getCalls().forEach((call) => expect(call.args[0]).to.equal(1));
     expect(component.firstElementChild).to.equal(elements[1]);
   });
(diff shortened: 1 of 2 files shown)


vaadin-renderer-flow-parent/vaadin-renderer-flow-integration-tests/test/flow-component-renderer.test.ts
```diff
@@ -67,4 +89,94 @@ describe('flow-component-renderer', () => {
...
+    const { getByNodeId } = window.Vaadin.Flow.clients.ROOT;
+    const component = fixtureSync<TestComponent>(`<test-component></test-component>`);
+
+    // Try to render a lazily added node
+    component.nodeId = 0;
+    await nextFrame();
+
+    // Disconnect
+    component.remove();
+    getByNodeId.resetHistory();
```
vursen, line 177: Should it be before `component.remove()`?

tomivirkki: The order doesn't seem to make a difference. Maybe it could be changed for clarity.

vursen: Yeah, I'd suggest changing it to avoid confusion.

tomivirkki: Done

vursen (approved):

PR #5491, opened 2023-09-21
fix: properly clear pending requests on grid item collapse

### LINKED ISSUE #5378: TreeGrid with filter spams server
### Describe the bug

I have a TreeGrid with a TextField filter.
When input is non-empty, I filter and expand all folders, and when input is empty I clear filter and collapse all folders.

When I clear with this, I get a never-ending request/response loop:
```
treeDataProvider.clearFilters();
treeGrid.collapse(allFolders);
```

The application still works though. If I enter some filter text again, it stops spamming the server. If I clear it, the spamming starts again.
When I switch the order, it works just fine.

### Expected-behavior

Ideally, this should just be default behavior. As long as it isn't, the order in which I do clear+collapse shouldn't matter

### Reproduction

- treeDataProvider.clearFilters();
- treeGrid.collapse(allFolders);

Have uploaded an example project to github:
https://github.com/guttormvik2/treegrid-filter-bug

- Write "Child 8" and you should see all those children
- Clear filter. All parents are collapsed and everything is ok
- Tick "Clear then collapse" and do same as above -> Browser goes into a request/response loop

### System Info

Windows 10, Vaadin 24.1.6, Firefox 116.0.3

tomivirkki:
## Description

Fixes #5378 

The original issue can be reproduced with the following simplified example:

```java
var treeData = new TreeData<String>();
var parentRow = "Parent";
treeData.addItem(null, parentRow);
treeData.addItem(parentRow, "Child");

var treeDataProvider = new TreeDataProvider<>(treeData);
var treeGrid = new TreeGrid<>(treeDataProvider);
treeGrid.addHierarchyColumn(s -> s);

add(treeGrid);
add(new Button("Expand and collapse", e -> {
	treeGrid.expand("Parent");
	treeGrid.collapse("Parent");
}));
```

The problematic case is when an item is programmatically expanded and collapsed immediately.

The loop is caused by an invalid data request that originates due to an issue in the grid connector's removeFromArray helper function which never removes the last item in the given array. This PR fixes the issue.

The PR also addresses a closely related issue where the grid could end up with pending requests in the cache when an item is expanded and collapsed soon after.

## Type of change

Bugfix


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -803,19 +803,13 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
 
         grid.$connector.removeFromQueue = tryCatchWrapper(function (item) {
           let itemId = grid.getItemId(item);
+          // The treePageCallbacks for the itemId are about to be discarded ->
+          // Resolve the callbacks with an empty array to not leave grid in loading state
+          Object.values(treePageCallbacks[itemId] || {}).forEach((callback) => callback([]));
```
vursen, line 808: It seems, if we don't provide the size for nested level requests, this [logic](https://github.com/vaadin/web-components/blob/e5c179c5fc67c374b52f24455b026fb242da6ce1/packages/grid/src/vaadin-grid-data-provider-mixin.js#L394-L399) will set the size to the length of the provided items array, which is 0 in this case. I wonder if it can cause any issues. Probably not, considering that `ensureSubCacheForScaledIndex` will anyway request the first page and set the correct size back when the item is expanded again. Just wanted to shine a spotlight on this.

ugur-vaadin (approved):

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
tomivirkki: I did consider this but concluded it shouldn't be an issue because, as you mentioned, the page will get requested again anyway and the correct size will be provided after the response.

vursen (approved):

Issue #5469, opened 2023-09-14
TabSheet inside of TabSheet is disabled


javier-godoy:
### Description

Consider the attached example. After I click "Tab 1_1" the SelectedChangeListener executes (reverting back to "Tab 1_0") where "Tab 2_1" is now disabled.

![image](https://github.com/vaadin/flow-components/assets/11554739/130bd20a-2d12-42aa-bd06-2fdf8660f6f4)

### Expected outcome

I would expect that "Tab 2_1" remains enabled.

### Minimal reproducible example

```java
TabSheet tabs1 = new TabSheet();
    TabSheet tabs2 = new TabSheet();

    tabs1.add("Tab 1_0", tabs2);
    tabs1.add("Tab 1_1", new Div());
    tabs2.add("Tab 2_0", new Button("Click", ev -> Notification.show("Clicked!")));
    tabs2.add("Tab 2_1", new Button("Click", ev -> Notification.show("Clicked!")));
    add(tabs1);

    tabs1.addSelectedChangeListener(ev -> {
      if (ev.getSelectedTab() == tabs1.getTabAt(1)) {
        tabs2.setSelectedIndex(1);
        tabs1.setSelectedIndex(0);
      }
    });
```

### Steps to reproduce

1. Add the snippet above to a view
2. Click "Tab 1_1"
3. Notice how "Tab 2_1" is disabled.

### Environment

Vaadin version(s):  23.3.26

PR #5506, opened 2023-09-25
fix: ignore ancestor enabled state in tabs internal logic

tomivirkki:
## Description

Certain [internal logic](https://github.com/vaadin/flow-components/blob/51aefa5f2512be6644b031da1aadcb55e6725a52/vaadin-tabs-flow-parent/vaadin-tabs-flow/src/main/java/com/vaadin/flow/component/tabs/Tabs.java#L674) of `Tabs` currently uses `isEnabled()` to determine if the currently selected tab is enabled or not. Using `isEnabled()` is problematic for this purpose because it also [traverses the node's ancestors](https://github.com/vaadin/flow/blob/cb63006b56777d46c881d41bc416d8d7189cd47d/flow-server/src/main/java/com/vaadin/flow/internal/StateNode.java#L1158) and assumes a tab is disabled even if the tab itself was actually enabled but one of its ancestor containers wasn't.

This PR addresses the issue by changing the logic to use [`isEnabledSelf()`](https://github.com/vaadin/flow/blob/cb63006b56777d46c881d41bc416d8d7189cd47d/flow-server/src/main/java/com/vaadin/flow/internal/StateNode.java#L1175) instead.

Fixes https://github.com/vaadin/flow-components/issues/5469

## Type of change

Bugfix

### DIFF
--- a/vaadin-tabs-flow-parent/vaadin-tabs-flow/src/main/java/com/vaadin/flow/component/tabs/Tabs.java
+++ b/vaadin-tabs-flow-parent/vaadin-tabs-flow/src/main/java/com/vaadin/flow/component/tabs/Tabs.java
@@ -671,7 +671,8 @@ public class Tabs extends Component
             return;
         }
 
-        if (currentlySelected == null || currentlySelected.isEnabled()) {
+        if (currentlySelected == null
+                || currentlySelected.getElement().getNode().isEnabledSelf()) {
             selectedTab = currentlySelected;
             getChildren().filter(Tab.class::isInstance).map(Tab.class::cast)
                     .forEach(tab -> tab.setSelected(false));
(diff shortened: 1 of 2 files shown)


DiegoCardoso (approved):
LGTM
Although I couldn't figure out why on the original example, some parent would report itself as not enabled.

tomivirkki:
> Although I couldn't figure out why on the original example, some parent would report itself as not enabled.

`TabSheet` intentionally [disables](https://github.com/vaadin/flow-components/blob/2f78468bf362be808b30c5ffcb992caa7800f108/vaadin-tabs-flow-parent/vaadin-tabs-flow/src/main/java/com/vaadin/flow/component/tabs/TabSheet.java#L346) the content of all the hidden tabs because hidden UI shouldn't be accessible even with dev tools hacks. So in the case where the reported issue occurred, the second `TabSheet` was temporarily inside disabled content while its tab was changed, and that triggered the bug in `Tabs`.

ugur-vaadin (approved):

Issue #4984, opened 2023-04-24
Change of Visibility of MenuItems in MenuBar fails if initially set to false


fschoning:
### Description of the bug

MenuBar with MenuItem visibility initialised to false cannot be changed to being visible later on.

### Expected behavior

MenuItem should be made visible

### Minimal reproducible example

```
package com.example.application.views.helloworld;

import com.example.application.views.MainLayout;
import com.vaadin.flow.component.button.Button;
import com.vaadin.flow.component.contextmenu.MenuItem;
import com.vaadin.flow.component.menubar.MenuBar;
import com.vaadin.flow.component.orderedlayout.HorizontalLayout;
import com.vaadin.flow.router.PageTitle;
import com.vaadin.flow.router.Route;
import com.vaadin.flow.router.RouteAlias;

@PageTitle("Hello World")
@Route(value = "hello", layout = MainLayout.class)
@RouteAlias(value = "", layout = MainLayout.class)
public class HelloWorldView extends HorizontalLayout {

    private MenuBar menuBar;
    private MenuItem menuItemOne;
    private MenuItem menuItemTwo;

    public HelloWorldView() {
        Button changeVisbibility = new Button("Change visibility");
        changeVisbibility.addClickListener(e -> {
            menuItemOne.setVisible(!menuItemOne.isVisible());
        });

        add(createToolbar(), changeVisbibility);
    }

    private HorizontalLayout createToolbar() {
        menuBar = new MenuBar();

        var toolbarLayout = new HorizontalLayout();
        toolbarLayout.setWidthFull();
        toolbarLayout.setDefaultVerticalComponentAlignment(Alignment.CENTER);

        menuItemOne = menuBar.addItem("New");
        menuItemTwo = menuBar.addItem("Two");



        // ************************************************************************************************************
        menuItemOne.setVisible(false);
        //menuItemOne.setVisible(true);  // If the visibility here is set to true, then it all works fine
        // ************************************************************************************************************



        HorizontalLayout rightButtonLayout = new HorizontalLayout();
        rightButtonLayout.getElement().getStyle().set("margin-left", "auto");       // align to the far right
        rightButtonLayout.setDefaultVerticalComponentAlignment(Alignment.CENTER);

        toolbarLayout.add(menuBar, rightButtonLayout);
        return toolbarLayout;
    }

}

```

### Versions

- Vaadin / Flow version: 24.0.4 as well as v23
- Java version: 17
- OS version: Win11
- Browser version (if applicable):
[...]

PR #5539, opened 2023-10-04
fix: attach observer before filtering hidden items

yuriy-fix:
Fixes https://github.com/vaadin/flow-components/issues/4984


web-padawan (approved):
The connector fix looks good to me. Added a test to verify that it works.

Issue #5533, opened 2023-10-02
RichTextEditor: extra white space is removed when content is loaded


mihannus:
### Description

We're using the RichTextEditor in Delta format (JSON) mode, and the value is bound to a String field that is stored in the database. When the component is loaded with existing Delta format data then extra whitespace is removed (tabs, spaces). 

![rich_text_editor_1](https://github.com/vaadin/flow-components/assets/7468249/bb2ac80e-bc23-4805-a9ed-14a6a71261fc)
![rich_text_editor_2](https://github.com/vaadin/flow-components/assets/7468249/aac26868-3180-4f24-babf-1f094b9bc514)

### Expected outcome

We'd expect the text to appear as it was when it was saved, with extra whitespaces preserved.

For example when populating the component with this data:
`[{"attributes":{"bold":true},"insert":"Line 1"},{"insert":"\n\tIndent 1\n\t\tIndent 2\nLast line with\textra         spaces and a \ttab\n\n"}]
`
then when expect to get the same data back, but now we get:
`[{"attributes":{"bold":true},"insert":"Line 1"},{"insert":"\nIndent 1\nIndent 2\nLast line with extra spaces and a tab\n"}]
`

### Minimal reproducible example

```
getBinder().forField(ingredientListEditor.asDelta())
	.bind(product -> product.getIngredientList(), (product, s) -> {
		product.setIngredientList(s);
	});
```

### Steps to reproduce

1. Create RichTextEditor component
2. Write text that contains tabs and extra spaces into the editor component
3. Retrieve the JSON Delta format data from the component (should be correct at this point)
4. Re-populate the component with the retrieved data
5. Observe that the extra whitespace has been removed

### Environment

Vaadin version(s): 24.1.10
OS: Windows 10

### Browsers

Chrome

TatuLund:
I do not know if this is related, but when html value is cleaned, Jsoup will strip white spaces

https://github.com/vaadin/flow-components/blob/main/vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java#L273

mihannus:
@TatuLund Yes I think you're right that it's Jsoup that strips the white spaces. I ran the application in the debugger and when `RichTextEditor.sanitize` was called with the following input:
```
[...]

PR #5540, opened 2023-10-04
fix: Do not strip whitespaces when sanitizing html

TatuLund:
Fixes https://github.com/vaadin/flow-components/issues/5533

### DIFF
--- a/vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java
+++ b/vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java
@@ -270,13 +270,17 @@ public class RichTextEditor
     }
 
     static String sanitize(String html) {
-        return org.jsoup.Jsoup.clean(html,
+        org.jsoup.nodes.Document.OutputSettings settings = new org.jsoup.nodes.Document.OutputSettings();
+        settings.prettyPrint(false);
+        String safeHtml = org.jsoup.Jsoup.clean(html, "",
                 org.jsoup.safety.Safelist.basic()
                         .addTags("img", "h1", "h2", "h3", "s")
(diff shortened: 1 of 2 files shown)


vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java
```diff
@@ -270,13 +270,17 @@ private String getDeltaValue() {
     }
 
     static String sanitize(String html) {
-        return org.jsoup.Jsoup.clean(html,
+        org.jsoup.nodes.Document.OutputSettings settings = new org.jsoup.nodes.Document.OutputSettings();
+        settings.prettyPrint(false);
```
tomivirkki, line 274: This doesn't seem to fix the [original issue](https://github.com/vaadin/flow-components/issues/5533). You can try it with the following:

```java
editor.asDelta().setValue(
                "[{\"attributes\":{\"bold\":true},\"insert\":\"Line 1\"},{\"insert\":\"\\n" + //
                                "\\tIndent 1\\n" + //
                                "\\t\\tIndent 2\\n" + //
                                "Last line with\\textra         spaces and a \\ttab\\n" + //
                                "\\n" + //
                                "\"}] ");
```

The above will still produce:
Screenshot 2023-10-04 at 11 55 41

In fact, the result is the same even if you return the given `html` as such from the `sanitize` method.

TatuLund: I need to dig this deeper. It may be that the fix needs changes in web component as well. Or if the white spaces are stripped also in Quill. If the later, we are stuck. If it is not the trouble maker, i.e. issue in our code, this Java side change is needed anyway, but if it is Quill related, this change is meaningless.

yuriy-fix:
We will proceed with PR once WC alpha6 is available.

tomivirkki (approved):

PR #5543, opened 2023-10-05
refactor!: make applayout afternavigation package private

### LINKED ISSUE #5449: AppLayout does not provide UI in afterNavigation method
### Description

The design of the afterNavigation method in the `AppLayout` is contradictory to the one of the `AfterNavigationObserver`. While in the method of the `ANO` interface the UI can be accessed via `getUI()`, the method of the `AppLayout` does not yet provide it.
This may lead to confusion for devs, as they might expect the same behavior based on the same method naming and similar documentation. 

### Expected outcome

The UI should be available when calling getUI inside `afterNavigation`.

Possible solutions: 
1. Override `getUI` in the `AppLayout` to return `UI.getCurrent()` as an alternative, if the component has not yet been attached to the UI. Not sure if this could lead to a false UI in any normal case?
2. Let `AppLayout` implement the `AfterNavigationObserver` interface and move the call of `afterNavigation` there. This could be classified as a breaking change as it might interfere with any manual implementation of the interface. It would however align more with our base interfaces and the intended usage.

```
public class MainLayout extends AppLayout {
    ...

    @Override
    protected void afterNavigation() {
        getUI().orElseThrow(IllegalStateException::new);
    }
    ...
}
```

### Steps to reproduce

1. Create a starter app from our website.
2. Go to the MainLayout.
3. Call `getUI()` inside the pre-implemented `afterNavigation` method. The returned value is an empty optional. 

### Environment

Vaadin version(s): 24.1.8
OS: Windows

### Browsers

Issue is not browser related

ugur-vaadin:
## Description

`AppLayout.afterNavigation` does not serve the same purpose as `AfterNavigationObserver.afterNavigation`, therefore causes confusion. It does not need to be public. This PR makes the API package-private. 

Marked as breaking since it might have been used in place of `AfterNavigationObserver.afterNavigation` in some cases.

Fixes #5449 

## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] Refactor

## Checklist


vursen (approved):

yuriy-fix:
Should be revisited before 25.0. Added to the breaking changes list for 25.0

DiegoCardoso (approved):

Issue #5537, opened 2023-10-03
Validation refactor to fix issues related to the validated event


vursen:
## Motivation

There are several problems with the current approach based on the validated event:

1. Server-side validation is triggered on blur even if nothing has been typed ([#17429](https://github.com/vaadin/flow/issues/17429), [#6146](https://github.com/vaadin/flow-components/issues/5712))
2. Server-side validation is triggered too frequently, resulting in performance degradation ([#4390](https://github.com/vaadin/flow-components/issues/4390)).
3. Server-side validation is triggered on client property changes, for example resulting in unexpected Binder revalidation after a bean reset ([#5399](https://github.com/vaadin/flow-components/issues/5399), [#5319](https://github.com/vaadin/flow-components/issues/5319), [#5330](https://github.com/vaadin/flow-components/issues/5330), [#5328](https://github.com/vaadin/flow-components/issues/5328)).
4. Server-side validation is triggered inconsistently for text fields: fields validate on change when valid but on input when invalid ([#4986](https://github.com/vaadin/flow-components/issues/4986)). 

All these problems have emerged after the introduction of the validated event, so we need to find an alternative. We should remember that the underlying purpose of the validated event was to trigger server-side validation when the user enters bad input, as there is no change event in that case.

## Solution

For components that allow entering bad input, let’s replace the validated event with a new event `unparsable-change`:
- This event should be fired by the web component whenever the user attempts to commit or clear bad input. These actions usually happen on Enter, outside click or blur. However, the specific list of actions may vary depending on the component.
- This event should be mutually exclusive with the change event to prevent sending two events to the server in one round-trip and thus rule out the possibility of duplicate server-side validation.
- The Flow component is supposed to subscribe to this event (in addition to the change event) and trigger validation whenever it occurs.

However, switching to the new event will address (4) only partially. The validation will be still inconsistent in eager mode: text fields will validate on input when entering valid values but on change when entering bad input. To improve this, let’s additionally subscribe the text field Flow components to `hasInputValue` property change and trigger validation when it occurs in eager mode.

For the rest components, let's remove the validated event so that they will only validate on change event.

## Affected components

- [x] DatePicker
- [x] TimePicker
- [x] NumberField
- [x] IntegerField
- [x] BigDecimalField
- [x] ComboBox
- [x] MultiSelectComboBox
- [x] Select
- [x] RadioGroup
- [x] CheckboxGroup

> **Note** 
[...]

PR #5581, opened 2023-10-13
fix!(BigDecimalField): use custom logic instead of validated event

vursen:
## Description

The PR refactors BigDecimalField to make it trigger validation whenever a `value-changed` event is received from the client while the value still remains empty on the server, instead of relying on the `validated` event.

Depends on
- https://github.com/vaadin/flow-components/pull/5590

> **Warning**
> It's a behavior altering change because it removes validation on blur.

Part of #5537 

## Type of change

- [x] Bugfix


vursen:
I added tests for eager mode and they uncovered that there is an unexpected initial validation that is caused by the `value-changed` event that Polymer fires during the web component's initialization. We should ensure that the client-side value property is initialized with an empty string to prevent initial `value-changed` events. DatePicker had this kind of problem long time ago and it was fixed by https://github.com/vaadin/flow-components/pull/2878.

web-padawan (approved):

PR #5590, opened 2023-10-16
fix: ensure client-side value property is initialized

vursen:
## Description

Ensures that TextField initializes its client-side value property with an empty string to prevent Polymer from generating `value-changed` events when converting `null` to `""` on its own. Refer to https://github.com/vaadin/flow-components/pull/5581#issuecomment-1764670968 for more details on why this is necessary.

Note, the same fix was applied to DatePicker a while ago: https://github.com/vaadin/flow-components/pull/2878.

Part of #5537 

## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-text-field-flow-parent/vaadin-text-field-flow/src/test/java/com/vaadin/flow/component/textfield/tests/BigDecimalFieldTest.java
+++ b/vaadin-text-field-flow-parent/vaadin-text-field-flow/src/test/java/com/vaadin/flow/component/textfield/tests/BigDecimalFieldTest.java
@@ -16,15 +16,23 @@
 package com.vaadin.flow.component.textfield.tests;
 
 import com.vaadin.flow.component.AbstractField;
+import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.HasAriaLabel;
+import com.vaadin.flow.component.UI;
 import com.vaadin.flow.component.shared.HasTooltip;
 import com.vaadin.flow.component.shared.InputField;
 import com.vaadin.flow.component.textfield.BigDecimalField;
 import com.vaadin.flow.component.textfield.TextFieldVariant;
+import com.vaadin.flow.di.Instantiator;
+import com.vaadin.flow.dom.Element;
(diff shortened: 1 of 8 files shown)


vaadin-text-field-flow-parent/vaadin-text-field-flow/src/test/java/com/vaadin/flow/component/textfield/tests/BigDecimalFieldTest.java
```diff
@@ -52,9 +60,41 @@ public void setValueNull() {
...
+    public void initialValueIsNotSpecified_valuePropertyHasEmptyString() {
+        BigDecimalField bigDecimalField = new BigDecimalField();
+        Assert.assertNull(bigDecimalField.getValue());
+        Assert.assertEquals("",
+                bigDecimalField.getElement().getProperty("value"));
+    }
+
+    @Override
+    @Test
+    public void initialValueIsNull_valuePropertyHasEmptyString() {
```
vursen, line 72: **note:** BigDecimalField's constructor doesn't have an overload that accepts an initial value.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/test/java/com/vaadin/flow/component/textfield/tests/IntegerFieldTest.java
```diff
@@ -59,9 +66,39 @@ public void assertStepGreaterThanZero() {
...
-                field.getElement().getProperty("value"));
+    public void initialValueIsNotSpecified_valuePropertyHasEmptyString() {
+        IntegerField integerField = new IntegerField();
+        Assert.assertNull(integerField.getValue());
+        Assert.assertEquals("", integerField.getElement().getProperty("value"));
+    }
+
+    @Override
+    @Test
+    public void initialValueIsNull_valuePropertyHasEmptyString() {
```
vursen, line 77: **note:** NumberField's constructor doesn't have overloads that accept an initial value.

PR #5629, opened 2023-10-26
fix!(TextField): remove validated event subscription

vursen:
## Description

The PR removes the subscription to the `validated` event from the remaining text field components.

> **Warning**
> It's a behavior altering change because it removes validation on blur.

Part of #5537 

## Type of change

- [x] Bugfix


vaadin-flow-components-shared-parent/vaadin-flow-components-test-util/src/main/java/com/vaadin/tests/validation/AbstractValidationIT.java
```diff
@@ -46,6 +46,7 @@ protected void assertValidationCount(int expected) {
                 .parseInt($("div").id(SERVER_VALIDATION_COUNTER).getText());
         Assert.assertEquals("The field should have validated " + expected
                 + " times on the server-side", expected, actual);
+        resetValidationCount();
```
vursen, line 49: **note:** Implicitly resetting the validation count after each assertion simplifies tests. I haven't come up with a use-case where it would be useful to preserve the count after the assertion.

tomivirkki (approved):

Issue #5608, opened 2023-10-19
Spreadsheet table filter popups still displayed in the UI even if setVisible(false) and setHeaderHidden(true) are called


WoozyG:
### Description

there is no way to hide/remove the Spreadsheet component overlays for table column filter menus.

As the title says, even if setVisible(false) and setHeaderHidden(true) are called in an overridden registerTable() method, the UI still renders the overlays.

There is no way to remove the overlays entirely, and we can't inject a subclass of SpreadsheetFilterTable or replace with SpreadsheetTable without possibly creating side effects, as there is no preventing the caller from doing further work with the instance they passed, even if the implementation ignores it.

### Expected outcome

If the overlay has no content, it shouldn't be displayed at all.

Ideally, this would not be needed if #5607 were fixed.

### Minimal reproducible example

https://github.com/WoozyG/sheet-tables

### Steps to reproduce

Run the given sample project, and note the tables still have overlay buttons even though they have no content.

### Environment

Vaadin version(s): 24.2
OS: N/A

### Browsers

Issue is not browser related

WoozyG:
This is still an issue in 24.4.7.  The problem stems partly from `SpreadsheetFactory.loadSheetTables()` which doesn't check if the XLSX table has autoFilter set or not, and always creates a filter table.  Thus, if we don't want filters, there is neither a way to turn them off or to define the file to not use them, and have it honored.

We need tables, but filters are implemented very incompletely, and completely mess up the display if the table is dynamically updated.  We need a way to turn them off.

PR #5657, opened 2023-11-02
fix: register spreadsheet filter tables just once

### LINKED ISSUE #5605: SpreadsheetFilterTable instances are created and registered every time a sheet is activated, but never deleted (object leak)
### Description

Render a spreadsheet with 2 tabs, each with a table defined.
Click back and forth between the tabs.
new SpreadsheetFilterTable instances are registered every single activation, but never removed or dereferenced.

This creates an object leak, obviously.

It also means one can't extend Spreadsheet to do any meaningful work with tables, as the instances used change every time a tab is displayed again.

See this sample project:

and note the log messages as you switch worksheets in the browser.

### Expected outcome

table objects should be created once, 1:1 with their POI object counterparts, and registered once.

FIlter overlays should probably not be coupled with this object, but be an "active sheet" UI scoped item instead. 

Co-mingling data objects and UI state is never a good idea.

Vaadin starter with SpringBoot and a slightly modified default workbook and added logging to show the issue.

### Steps to reproduce

Run the linked project, switch between workbook tabs, and watch the log messages.

### Environment

Vaadin version(s): 24.2
OS: N/A

### Browsers

Issue is not browser related

bwajtr:
## Description

There was a problem with spreadsheet table registration, where a table was loaded (and registered) from the underlying XSSFSheet model every time the active sheet was changed. So, for example, having a table on Sheet1 meant that every time the user switched to Sheet1 from another sheet, the table was loaded and stored in memory (without the previous instance to be released). This caused a memory leak (with the old instances being effectively ignored, so it went unnoticed).

There was also one connected bug: when you deleted a table on Sheet1, went to Sheet2, and then again returned to Sheet1, the table that was previously deleted reappeared.

The solution to both problems is to load and register all the tables just once when the whole workbook is loaded. 
However, change of the table initialization code was also needed, because the previous table creation code assumed, that the table being initialized is on active sheet, which is not always the case after this change -> this resulted for example in creation of the `WorksheetAutoFilterTable`. 

Also, I've removed `PopupButton` initialization from the constructor of the `SpreadsheetTable` - having that initialization there prevented seamless extending of the `SpreadsheetTable` class. The initialization of the popup button is now delayed and always done on a first `reload()` call.

Furthermore, the Spreadsheet was made a little bit more extensible by offering two `protected` methods that are responsible for the creation and registration of the `SpreadsheetTable` instances: `buildWorksheetAutoFilterTable()` and `buildFilterTable()` can now be used to affect how the tables are created in the Spreadsheet.

Fixes #5605

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


DiegoCardoso (changes requested):
There's a change in the behavior introduced here (don't know exactly where). In the project attached to the original issue, the author overrides the `registerTable` method and, essentially, do:

```java
    @Override
	public void registerTable(SpreadsheetTable table) {
		table.getPopupButtons().forEach(pb -> {
			pb.setVisible(false);
			pb.setHeaderHidden(true);
		});
		super.registerTable(table);
	}
```
In the current Spreadsheet, when you click on the popup buttons, nothing appears in the overlay (you can barely see the empty overlay). Running against this branch, the popup button's content appears in the overlay.

We can argue that the buttons shouldn't be added at all when `PopupButton#setVisible` is called with `false` (as described in https://github.com/vaadin/flow-components/issues/5608), but probably calling `PopupButton#setHeaderHiddern` with `true` should hide the header inside the overlay.

vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/SpreadsheetTable.java
```diff
@@ -68,10 +68,6 @@ public SpreadsheetTable(Spreadsheet spreadsheet, Sheet sheet,
         this.sheet = sheet;
         this.fullTableRegion = fullTableRegion;
         popupButtons = new HashMap<CellReference, PopupButton>();
-
-        if (isTableSheetCurrentlyActive()) {
-            initPopupButtons();
-        }
```
DiegoCardoso, line 74: I understand now what caused the difference in behavior. 

Before, by the time the `registerTable` is called, all `PopupButton` instances of the table are available (for the active sheet), so it makes sense to perform some actions on them (like calling `PopupButton#setHeaderHidden`) and then call `super.registerTable(table)` afterward. 
 
By removing this code and calling `reload()` in the `registerTable` method, we delay creating the instances so that they won't be available by any subclass that overrides the method.

Could you explain why this code has been removed?

DiegoCardoso:
I was also wondering whether creating all tables upfront should be done or not. It's not uncommon to have large files with multiple sheets each of them having one or more tables. It's safe to assume that users won't always open all sheets, so we might be creating objects that will never be used.

bwajtr:
I had a call about this PR with Diego and decided to take a completely different approach for this bug. 
The changes are so different that I've decided to close this PR and open another one: https://github.com/vaadin/flow-components/pull/5685

bwajtr:
I had a call about this PR with Diego and decided to take a completely different approach for this bug. 
The changes are so different that I've decided to close this PR and open another one: https://github.com/vaadin/flow-components/pull/5685

Issue #5607, opened 2023-10-19
SpreadsheetFilterTable doesn't respect Excel enable/disable table filters setting


WoozyG:
### Description

XLSX file format has a property defining whether a table should show filter popups or not.  POI doesn't respect it, and always creates the popups, whether the developer or workbook author wants them or not.

### Expected outcome

table objects created from XSSFTable should only initialize the filter popup if the CTTable has an appropriate CTAutoFilter member, and then only for columns defined in that AutoFilter.

### Minimal reproducible example

https://github.com/WoozyG/sheet-tables

The first sheet has no autofilter for the table, but the second sheet does for its table.  Both display the filter popups.

### Steps to reproduce

Run the project.  Note the tables both have filter popups (empty, due to other code for a different issue).

Open the spreadsheet file directly in Excel.  Only the second sheet has filter drop-downs on its table.

### Environment

Vaadin version(s): 24.2
OS: N/A

### Browsers

Issue is not browser related

WoozyG:
Note in Excel table filters can be toggled via 

ctrl-shift-L 

or the Filter button in "Sort & Filter" group on Data ribbon tab

WoozyG:
As we can't allow filters in our app, as hiding full rows can hide charts anchored to those rows next to tables, I consider this more a bug than an enhancement - Vaadin currently has no way to fully disable/hide the filter dropdowns in v24.2.  I was able to do so in V8 without hassle.

WoozyG:
Noting for the record this not a low impact for us - I can't currently let customers see a beta build because of this, and can't move forward with acceptance testing of the V24 migration, as we can't find a way to get rid of the client side overlays.

PR #5685, opened 2023-11-09
fix: prevent tables to be loaded from XSSFSheet multiple times

### LINKED ISSUE #5605: SpreadsheetFilterTable instances are created and registered every time a sheet is activated, but never deleted (object leak)

### LINKED ISSUE #5606: Spreadsheet[Filter]Table doesn't reference the POI Table object it is built from
### Description

If we override Spreadsheet.registerTable() to capture table processing and add functionality, we have to scan the workbook and find the backing Excel table object rather than just referencing it from the passed in SpreadsheetTable or SpreadsheetFilterTable (the only class actually used currently, and from a static method we can't touch).

### Expected outcome

table.getTable() should exist and return the XSSFTable object it was built from.

code for SpreadsheetTable - should have XSSFTable as a constructor parameter instead of its cell references

### Steps to reproduce

see source code for SpreadsheetTable and SpreadsheetFactory.loadSheetTables()

### Environment

Vaadin version(s): 24.2
OS: N/A

### Browsers

Issue is not browser related

bwajtr:
## Description

There was a problem with spreadsheet table registration, where a table was loaded (and registered) from the underlying XSSFSheet model every time the active sheet was changed. So, for example, having a table on Sheet1 meant that every time the user switched to Sheet1 from another sheet, the table was loaded and stored in memory (without the previous instance to be released). This caused a memory leak (with the old instances being effectively ignored, so it went unnoticed).

The solution is to load and register the tables just once - when the table is registered, we now also remember the origin of the table (why was the table loaded and registered): either because of the presence of `XSSFSheet.getCTWorksheet().getAutoFilter()` or because there are instances of `XSSFSheet.getTables()`.
When the active sheet is changed, we now check the existing tables and we do not register a new table if the table for the given AutoFilter or XSSFTable already exists.

The upside of this PR is that it also fixed one additional feature request (#5606). Having references to `XSSFTable` in `SpreadsheetTable` will help us in the future to work with the tables properly: you'll need such reference when you want to modify or delete the underlying XSSFTable (https://github.com/vaadin/flow-components/issues/5693)

Fixes #5605
Fixes #5606

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/SpreadsheetTable.java
```diff
@@ -201,4 +205,43 @@ public Collection<PopupButton> getPopupButtons() {
...
     }
 
+    /**
+     * Sets the CTAutoFilter object that represents this table in the underlying
+     * POI model.
+     *
+     * @param ctWorksheetAutoFilter
+     *            Referenced autofilter.
+     */
+    public void setCtWorksheetAutoFilter(CTAutoFilter ctWorksheetAutoFilter) {
```
DiegoCardoso, line 215: I am thinking about making these setters either package private or protected. I can't see the reason for having these as public API. WDYT?

bwajtr: Yeah... I think that having getter public is useful, but setters can be protected...

bwajtr: --> added a commit which changes the visiblity

DiegoCardoso (approved):
LGTM

WoozyG:
I would very much prefer any new code NOT be in the static methods in SpreadsheetFactory - in fact, that class is a huge problem in general.  

I want to customize nearly every step of spreadsheet initialization, either to hook in additional functionality or fix issues ahead of Vaadin.  Having so much of the construction done via `static` methods makes this tricky, and sometimes I have to wait for Vaadin to do something only to undo it in a later step because I can't touch the static methods or set up anything earlier in the constructor flow.

I see that the proposed changes here add even more static code to that class.  If it is possible to refactor to let that code be part of the Spreadsheet class, or even an instantiable object we could extend to wrap base calls, that would be helpful.  I may have mentioned these statics in another issue (which I'm not seeing) or perhaps just in conversation at Vaadin Create.  Can't remember for sure if I talked or typed about it.

WoozyG:
#5608 or #5607 may related to part of the reason this issue was created to replace the previous one.  I don't have a strong preference on how it's done, but it may be worth checking if it is easier to incorporate fixing those now rather than separately later.

bwajtr:
@WoozyG Thanks for review. We have discussed internally the same thing. I agree that `SpreadsheetFactory` is written in a way that prevents easy extensibility of the `Spreadsheet` component. However, in this PR, we decided to focus purely on a bugfix of the memory leak issue, and we decided not to mix in any refactoring of the `SpreadsheetFactory`. However, the static nature of the `SpreadsheetFactory` is still an issue that should be addressed somehow - therefore, I've created a new task to handle such bigger refactoring; please comment on it here: https://github.com/vaadin/flow-components/issues/5706

Issue #5597, opened 2023-10-18
Dialog and Notification cannot be used with Session Replication


MichaelPluessErni:
When trying to use Session Replication with Vaadin, the Deserialization fails when a Notification or a nested Dialog is present on the webpage. The error is the exact situation described here:
https://vaadin.com/docs/latest/tools/kubernetes/session-replication#serializedlambda-classcastexception

This is a problem as these components are attached to the view and thus the VaadinSession and thus cannot be excluded from serialization.

Example 1: Inside a Dialog, a button is clicked that opens a ConfirmDialog, without closing the Dialog
Output from Kubernetes Kit - Session Replication Debug Tool:
```
ERRORS DURING SERIALIZATION/DESERIALIZATION PROCESS:
====================================================
DESERIALIZATION_FAILED: cannot assign instance of java.lang.invoke.SerializedLambda to field com.vaadin.flow.component.dialog.Dialog.childDetachListener of type com.vaadin.flow.dom.ElementDetachListener in instance of com.vaadin.flow.component.dialog.Dialog

SERIALIZED LAMBDA CLASS CAST EXCEPTION BEST CANDIDATES:
=======================================================
	SerializedLambda[capturingClass=class com.vaadin.flow.component.dialog.Dialog, functionalInterfaceMethod=com/vaadin/flow/dom/ElementDetachListener.onDetach:(Lcom/vaadin/flow/dom/ElementDetachEvent;)V, implementation=invokeVirtual com/vaadin/flow/component/dialog/Dialog.lambda$new$76f1ff5c$1:(Lcom/vaadin/flow/dom/ElementDetachEvent;)V, instantiatedMethodType=(Lcom/vaadin/flow/dom/ElementDetachEvent;)V, numCaptured=1]
	- field (class "com.vaadin.flow.dom.Element$2", name: "val$detachListener", type: "interface com.vaadin.flow.dom.ElementDetachListener")
[...]
```

End of associated StackTrace:
```
java.lang.ClassCastException: cannot assign instance of java.lang.invoke.SerializedLambda to field com.vaadin.flow.component.dialog.Dialog.childDetachListener of type com.vaadin.flow.dom.ElementDetachListener in instance of com.vaadin.flow.component.dialog.Dialog
	at java.base/java.io.ObjectStreamClass$FieldReflector.setObjFieldValues(ObjectStreamClass.java:2227)
[...]

PR #5698, opened 2023-11-13
fix: replace lambda in fields to fix notification and dialog serialization

tomivirkki:
## Description

Fixes https://github.com/vaadin/flow-components/issues/5597

I [wasn't able to write unit tests](https://github.com/vaadin/flow-components/issues/5597#issuecomment-1794754772) for this case. The issue was, however, reproducible in a local Kubernetes cluster and I verified that this change fixes the issue in the same environment.

## Type of change

Bugfix


DiegoCardoso (approved):
I was trying to understand why these lambdas were not being serialized and came across [this article](https://www.baeldung.com/java-serialize-lambda). I couldn't run the same analysis locally, but it feels strange to me since `ElementDetachListener` implements `Serializable`, so I guess it _should_ be serializable. 

Just wanted to mention this, but anyway, LGTM

PR #5724, opened 2023-11-15
fix: defer setting defaults until tooltip.ts is loaded

vursen:
## Description

It turns out that `TooltipConfiguration` didn't work correctly in production mode. In this mode, `tooltip.ts` can be loaded later than `executeJs` from the `UIInitListener` are executed. This in turn resulted in "Cannot read properties of undefined" JS errors, which prevented the setting of tooltip defaults on a page refresh. However, we didn't see these errors because Flow didn't log JS errors in production until 24.3. Also, the tests didn't catch that the defaults didn't get updated because they always set the defaults again after a page refresh, which left no chance for the issue to manifest.

In this PR, I updated `TooltipConfiguration` to defer setting defaults for tooltips until `tooltip.ts` is loaded in the browser.

A follow-up to https://github.com/vaadin/flow/issues/18016

## Type of change

- [x] Bugfix


vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/resources/META-INF/resources/frontend/tooltip.ts
```diff
@@ -3,9 +3,19 @@ import { Tooltip } from '@vaadin/tooltip/src/vaadin-tooltip.js';
...
 _window.Vaadin.Flow = _window.Vaadin.Flow || {};
+_window.Vaadin.Flow.tooltip = _window.Vaadin.Flow.tooltip || {};
 
-_window.Vaadin.Flow.tooltip = {
-  setDefaultHideDelay: (hideDelay: number) => Tooltip.setDefaultHideDelay(hideDelay),
-  setDefaultFocusDelay: (focusDelay: number) => Tooltip.setDefaultFocusDelay(focusDelay),
-  setDefaultHoverDelay: (hoverDelay: number) => Tooltip.setDefaultHoverDelay(hoverDelay),
+_window.Vaadin.Flow.tooltip.setDefaultHideDelay = (hideDelay: number) => Tooltip.setDefaultHideDelay(hideDelay);
+_window.Vaadin.Flow.tooltip.setDefaultFocusDelay = (focusDelay: number) => Tooltip.setDefaultFocusDelay(focusDelay);
+_window.Vaadin.Flow.tooltip.setDefaultHoverDelay = (hoverDelay: number) => Tooltip.setDefaultHoverDelay(hoverDelay);
```
tomivirkki, line 10: This could be shortened to:
```js
const tooltip = (((window as any).Vaadin ||= {}).Flow ||= {}).tooltip ||= {};

Object.assign(tooltip, {
  setDefaultHideDelay: (hideDelay: number) => Tooltip.setDefaultHideDelay(hideDelay),
  setDefaultFocusDelay: (focusDelay: number) => Tooltip.setDefaultFocusDelay(focusDelay),
  setDefaultHoverDelay: (hoverDelay: number) => Tooltip.setDefaultHoverDelay(hoverDelay),  
});
```

tomivirkki (approved):

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/resources/META-INF/resources/frontend/tooltip.ts
vursen: Thanks, updated `tooltip.ts`. Also updated `executeJS` to use `||=` and `?` .

tomivirkki (approved):

PR #5725, opened 2023-11-15
fix: avoid using formatRawCellContents() because it ignores conditional formatting

### LINKED ISSUE #5651: Flow Spreadsheet ignores numeric conditional formatting - regression from v8
### Description

CustomDataFormatter has some incorrect logic for numbers, that can ignore conditional formatting, while conditional formatting is correctly applied in some cases, primarily where the base cell format is "General".

In particular, the code and behavior do not match the JavaDoc for `CustomDataFormatter.formatCellValue(Cell, FormulaEvaluator, ConditionalFormattingEvaluator)`

The JavaDoc indicates the default behavior is only changed if the formatting contains >=3 segments, but that's not how it behaves.

See the included sample project, and compare the display in Excel vs. Vaadin.

the logic calls `super.formatCellValue()` if the cell's base format is `General`, but that isn't correct.  It should find any conditional format applied, and prefer its format.

Then later, if the cell type is NUMERIC, it should only call `formatNumericValueUsingFormatPart` if the resolved format string actually has enough parts, otherwise just use the default `super` behavior.

I think for now I may have to just eliminate use of `CustomDataFormatter`.  We don't have users with multiple locales, so for now that would be OK, but I'm sure other customers would not have that option. 

### Expected outcome

number formatting in Vaadin would match number formatting in Excel for conditionally formatted cells with a basic base number format that isn't "General"

https://github.com/WoozyG/sheet-tables latest commit

### Steps to reproduce

check out and run the specified test project based on Vaadin Start.  Compare the 2nd sheet displayed values with the project XLSX file opened in Excel, and notice that the last column in the table should be formatted with one decimal place, but is not in Vaadin.

### Environment

Vaadin version(s):  tested 24.2.1, probably applies 24+
OS: N/A

### Browsers

Issue is not browser related

bwajtr:
## Description

Some background first: The problem is in `CustomDataFormatter` class. This class was introduced six years ago to workaround issues in the Apache POI `DataFormatter` class. The problems it fixes are described in the class JavaDoc. Some of the issues this custom implementation addressed were already fixed in Apache POI. However, the code and Javadoc in this class haven't been updated accordingly. I've investigated on whether it would be possible to remove significant parts of the `CustomDataFormatter` code, but I've unfortunately found that despite fixes in POI, some of the issues are still reproducible, and therefore, the majority of the code has to stay.

After this [POI bug was resolved](https://bz.apache.org/bugzilla/show_bug.cgi?id=60422), the `DataFormatter` class got much better in formatting using defined Locale. Therefore, I did two things:

1. When formatting numbers, I've replaced a call to `super.formatRawCellContents()` with `super.formatCellValue()`. The former can format the number correctly but ignores any cell Conditional Formatting. The latter also formats numbers correctly (probably because of POI fixes, which happened in the past six years), but also considers cell's Conditional Formatting. All the tests are still passing after this change + this change fixes the issue reported in #5651 

2. I've modified the JavaDoc of the class and the `formatCellValue()` method to reflect the reality better. For example, It is not true anymore that CellFormat does not use Locale info... also the JavaDoc of `formatCellValue()` wasn't describing what the method does.

Fixes #5651 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


DiegoCardoso (approved):

bwajtr:
Note that I wanted to cherry-pick this change also to `23.3` and `23.4` branches, but I don't think it makes sense: during cherry-picking, I found out that a [vital change related to locale handling](https://github.com/vaadin/flow-components/commit/8d5b7df5f09d9056d911ff2c2e1ab520c8eb3118) in the `CustomDataFormatter` was not backported to `23.3`. The solution in this PR kind of relies on this change...

PR #5736, opened 2023-11-16
fix: init all CellValueManager locale specific fields properly

bwajtr:
## Description

As noted in [this comment](https://github.com/vaadin/flow-components/issues/5651#issuecomment-1814222309): a problem in the `Spreadsheet` component originally identified by @WoozyG: When a `CellValueManager` is created, a `Locale` from the current UI is used to create new instance of `CellValueManager.formatter`. However, the `CellValueManager.formatter` is not the only one in this class that is locale-specific. As you can see in the `CellValueManager.updateLocale()` method, there are also `localeDecimalSymbols`, `originalValueDecimalFormat`, and `cellValueFormatter` that should all take into account a change in the Locale.

Creating just `formatter` in the constructor means, that part of the class uses different Locale than `formatter` for some time... until the `CellValueManager.updateLocale()` is called... That normally happens when Spreadsheet is attached.

The solution is to call `CellValueManager.updateLocale()` in the constructor instead of just creating only a new `formatter`.
It was also possible to write a test for this.

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/CellValueManager.java
```diff
@@ -91,7 +91,7 @@ public class CellValueManager implements Serializable {
     private CellValueHandler customCellValueHandler;
     private CellDeletionHandler customCellDeletionHandler;
 
-    private DataFormatter formatter;
+    private DataFormatter formatter = new CustomDataFormatter();
```
DiegoCardoso, line 94: Is it needed to create a new instance here?
I see that `updateLocale` creates a new instance of `CustomDataFormatter` using the current locale.

bwajtr: It is needed, by my opinion, to keep backwards compatibility: `new CustomDataFormatter()` was called in the constructor in the previous version when `UI.getCurrent() == null`

Newly, we call  `updateLocale` only when `UI.getCurrent() != null`

DiegoCardoso (approved):

PR #5749, opened 2023-11-21
feat: add AbstractLogin.showErrorMessage()

### LINKED ISSUE #1525: New API for setting error title and message directly

mvysny:
## Description

Adds `AbstractLogin.showErrorMessage()` which allows you to set the login error message programmatically, when implementing custom authentication flow.

Fixes #1525 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/AbstractLogin.java
```diff
@@ -222,4 +224,46 @@ public ForgotPasswordEvent(AbstractLogin source, boolean fromClient) {
...
+        if (loginI18n == null) {
+            loginI18n = LoginI18n.createDefault();
+        }
+        if (loginI18n.getErrorMessage() == null) {
+            loginI18n.setErrorMessage(new LoginI18n.ErrorMessage());
+        }
+        loginI18n.getErrorMessage().setTitle(title);
+        loginI18n.getErrorMessage().setMessage(message);
+        setI18n(loginI18n);
+        setError(true);
```
web-padawan, line 267: What is the motivation for calling `setError(true);`? In other components e.g. `TextField`, the `setErrorMessage` method only sets error text, but doesn't force it to be shown. IMO this method should behave the same.

mvysny: Good point! The motivation is that I'm calling this function overwhelmingly after the login failed:
```
try {
  login(username, password);
  navigateTo("");
} catch (LoginException ex) {
  setErrorMessage("Login failed", translate(ex.failureKey));
}
```
Therefore, I always want also the error to appear, and therefore I'd like the `setErrorMessage()` to call `setError(true)` since I assume this kind of omission will be the number 1 cause of problems.

That being said, `setErrorMessage()` indeed sounds similar to other setters, and calling `setError(true)` sounds like an unintended side-effect.

I'd still like to keep the call to `setError(true)`; how about if we rename the function to `showError()` or `showErrorMessage()` or such?

mvysny: Renamed to `showErrorMessage()`

yuriy-fix:
We will proceed with the PR after branching out for new minor. (~ couple of weeks)

sissbruecker (approved):

PR #5752, opened 2023-11-22
fix: use last parsed date and value to correct century of short year

### LINKED ISSUE #4636: [DatePicker] Selecting a value in the overlay results in wrong date, when using short year formats (works in V23)
### Description

This is an V22 issue (and maybe 14.9.x, haven't it tested there).

The setup is simple. Have a date picker with a short year format, e. g. "dd.MM.yy". When using the overlay to select a year, the date picker misinterpretes those years.

For instance, when selecting a date in 1922, the year will be changed to 2022. Selecting 2070 will result in 1970. Anything outside of 19xx or 20xx will result in those centuries (e.g. 1880 will be changed to 1980).

Setting additional, long year formats has no effect on the outcome. It still changes the years.

The date to select
a

The resulting date
b
c

### Expected outcome

When  using the overlay, the selected year should be considered and not be replaced.

        DatePicker picker = new DatePicker();
        DatePickerI18n i18n = new DatePickerI18n();
        i18n.setDateFormats("dd.MM.yy"); // we can add additional long year formats, does not change the outcome
        picker.setI18n(i18n);
        add(picker);

        picker.addValueChangeListener(event -> Notification.show(String.valueOf(event.getValue())));

### Steps to reproduce

1. Create a V22 project
2. Add the sample code
3. Open the overlay of the date picker
4. Select for instance 1.1.1920 or 1.1.2070.
5. Check the resulting year. Should be 2020 / 1970.

### Environment

Vaadin version(s): 22.0.28
OS: Windows

### Browsers

Issue is not browser related

ugur-vaadin:
## Description

Date picker does a year correction for short year formats when calculating the full year. This correction does not get applied properly when the selection is done through the overlay. The reason is that there is no previously parsed date. 

This PR:
- Refactors the parsing logic to reduce duplication and increase clarity
- Uses the date picker value to correct the century in the first parse after overlay open
- Uses last parsed date to correct the century if last parse attempt is successful and the dates match except the century
- Does not correct the century if last parse attempt was a failure

Fixes #4636 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
@@ -79,8 +79,11 @@ import { extractDateParts, parseDate as _parseDate } from '@vaadin/date-picker/s
             // The current year check handles the case where a four-digit year is parsed, then formatted
             // as a two-digit year, and then parsed again. In this case we want to keep the century of the
             // originally parsed year, instead of using the century of the reference date.
+            if (!datepicker.value) {
+              return;
+            }
             const currentDate = _parseDate(datepicker.value);
-            if (
+            if (dateFnsIsValid(currentDate) &&
               currentDate.getDate() === date.getDate() &&
               currentDate.getMonth() === date.getMonth() &&
               currentDate.getFullYear() % 100 === date.getFullYear() % 100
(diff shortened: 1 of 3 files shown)


vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
```diff
@@ -2,6 +2,7 @@ import dateFnsFormat from 'date-fns/format';
 import dateFnsParse from 'date-fns/parse';
 import dateFnsIsValid from 'date-fns/isValid';
 import { extractDateParts, parseDate as _parseDate } from '@vaadin/date-picker/src/vaadin-date-picker-helper.js';
+import { DatePicker } from '@vaadin/date-picker/src/vaadin-date-picker.js';
```
DiegoCardoso, line 5: Unused import.
```suggestion
```

ugur-vaadin: Removed

DiegoCardoso (approved):

PR #5768, opened 2023-11-24
fix: call updateSelectionModeOnClient on attach

### LINKED ISSUE #5645: Reattaching Grid breaks Selection Mode
### Description

When adding a Grid a second time (either adding it, even if it was already in the view, or adding it, while it was previously removed) the SelectionMode SINGLE will break. (for this Bug to work the Grid must at the same time be shown with setVisible(true).

What do I mean with breaking? Users on the client are now able to select items in the grid, and every time that happens an exception is thrown on the server:
java.lang.IllegalStateException: Client tried to update selection even though selection mode is currently set to NONE.

The right Grid itself also gives its rows too little vertical space.
Reattaching a Grid will break other components as well. For example the Content of other Grids will be removed.

### Expected outcome

I know that the Grid could just be hidden with setVisible(false) and be shown again with setVisible(true) but removing and attaching the Grid should still not lead to this behaviour.

For a minimal reproducible Example please clone this repo:
https://github.com/DarioViva42/selectionmode-bug/tree/master

### Steps to reproduce

Start the Application
You will see a SplitLayout. On the left side (PrimaryComponent) there is a Grid with two elements.
Select an element in this Grid.
On the right side (SecondaryComponent) another Grid will show up.
At the same time you will notice that the content from the first grid disappeared.
You can also select Elements in the right grid, even though the SelectionMode was set to NONE.
Now change the selection on the left Grid.
What you will notice is that the elements on the right Grid take up too little vertical space.
When scrolling in the right grid, or moving the slider from the SplitLayout the Items will grow to show all content.
When you select an element while the elements are still too small, then the selected element will take more space, hiding the rows below it.

### Environment

Vaadin version(s): 24.2.1
OS: Windows 10

### Browsers

Issue is not browser related

DiegoCardoso:
- fix: call updateSelectionModeOnClient on attach
- test: add test checking no server log error is thrown

## Description

Add a call to `updateSelectionModeOnClient` on the `onAttach` method to ensure that the
selection mode is correctly set in case a Grid is first attached not visible and detached
and reattached again in the same roundtrip.

The cause of this issue is that Flow does not send JS commands to the client when the component
is not visible. If the component is then detached, Flow clears any pending JS command, which makes
the selection mode to be not synched between server and client.

Fixes #5645

## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/DetachReattachPage.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/DetachReattachPage.java
@@ -22,6 +22,7 @@ import com.vaadin.flow.component.html.NativeButton;
 import com.vaadin.flow.component.html.Span;
 import com.vaadin.flow.data.renderer.ComponentRenderer;
 import com.vaadin.flow.router.Route;
+import com.vaadin.flow.server.VaadinSession;
 
 @Route(value = "vaadin-grid/detach-reattach-page")
 public class DetachReattachPage extends Div {
@@ -37,6 +38,26 @@ public class DetachReattachPage extends Div {
 
         NativeButton btnDetach = new NativeButton("Detach", e -> remove(grid));
         btnDetach.setId("detach-button");
+
+        NativeButton btnSelectionModeNone = new NativeButton(
+                "Change to selection none",
+                e -> grid.setSelectionMode(Grid.SelectionMode.NONE));
+        btnSelectionModeNone.setId("selection-mode-none-button");
+
+        NativeButton btnHideGrid = new NativeButton("Hide grid",
+                e -> grid.setVisible(false));
+        btnHideGrid.setId("hide-grid-button");
+
+        NativeButton btnShowGrid = new NativeButton("Show grid",
+                e -> grid.setVisible(true));
+        btnShowGrid.setId("show-grid-button");
+
+        NativeButton btnDetachAndReattach = new NativeButton(
+                "Detach and reattach", e -> {
+                    remove(grid);
+                    add(grid);
+                });
+        btnDetachAndReattach.setId("detach-and-reattach-button");
 
         NativeButton btnDisallowDeselect = new NativeButton("Disallow deselect",
                 e -> {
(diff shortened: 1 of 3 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/DetachReattachPage.java
```diff
@@ -68,7 +89,22 @@ public DetachReattachPage() {
...
+
+        NativeButton btnSetCustomErrorHandler = new NativeButton(
+                "Set custom error handler", e -> {
+                    // Set error handler to show errors in the UI
+                    VaadinSession.getCurrent().setErrorHandler(event -> {
+                        errorMessage.setText(
+                                "Error: " + event.getThrowable().getMessage());
+                    });
+                });
+        btnSetCustomErrorHandler.setId("set-custom-error-handler-button");
```
tomivirkki, line 103: nit: Why the custom error handler instead of using `checkLogsForErrors();` in the test case?

tomivirkki (approved):

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/DetachReattachPage.java
DiegoCardoso: AFAIK `checkLogsForErrors()` only checks the client side logs. In this case, the error can only be verified from the server side and it doesn't show anything in the browser. That's why I had to add the handler there.

DiegoCardoso: For some reason, I wasn't seeing the errors before, but I can confirm that `checkForLogs` works to assert that the error is gone.

tomivirkki: Actually, I wonder if this only works when running in dev mode? Could it give false positives in production mode if the server-side error wouldn't get reflected on the client?

DiegoCardoso: I tried running it with `-Dvaadin.productionMode` as it seems to be used in some of the commands in the `build.sh` script and the test failed if I remove the change in Grid.

PR #5791, opened 2023-11-29
fix: don't escape text in Notification#setText

### LINKED ISSUE #5790: Notification#setText incorrectly formats text
### Description

Running `Notification.show("Java > JavaScript");` currently displays `Java &gt; JavaScript` on the screen.

### Expected outcome

I would expect that `Notification.show("Java > Javascript");` would show a notification with the text `Java > Javascript`.

```java
add(new Button("Click me", e -> CustomNotification.show(">"));
```

This code should display the `>` character as the text to a notification when the button is pressed, but currently displays `&gt;`.

### Steps to reproduce

1. Add the snippet above to a view
2. Click the button
3. Notice how the Notification displays `&gt;` instead of `>`

### Environment

Vaadin version(s): Any (unfixed in current version: 24.2.4)
OS: Any

### Browsers

Issue is not browser related

maxcai314:
fix: don't escape text in Notification#setText

`Notification.show("Java > JavaScript");` currently displays `Java &gt; JavaScript` on the screen.

This bug causes any symbol that HTML escapes (`" ' & < >`) to display incorrectly (e.g. as `&gt;`).
This can be fixed by directly running `this.getElement().setProperty("text", text);` in Notification#setText.

This change will fix the functionality of Notification#setText with special characters.

Fixes #5790

### DIFF
--- a/vaadin-notification-flow-parent/vaadin-notification-flow/src/main/java/com/vaadin/flow/component/notification/Notification.java
+++ b/vaadin-notification-flow-parent/vaadin-notification-flow/src/main/java/com/vaadin/flow/component/notification/Notification.java
@@ -282,8 +282,7 @@ public class Notification extends Component implements HasComponents, HasStyle,
      */
     public void setText(String text) {
         removeAll();
-        this.getElement().setProperty("text",
-                text != null ? HtmlUtils.escape(text) : null);
+        this.getElement().setProperty("text", text);
         this.getElement().callJsFunction("requestContentUpdate");
     }
 


maxcai314:
This bug most likely exists in other setText methods in the code as well.

vaadin-notification-flow-parent/vaadin-notification-flow/src/main/java/com/vaadin/flow/component/notification/Notification.java
```diff
@@ -282,8 +282,7 @@ public static Notification show(String text) {
      */
     public void setText(String text) {
         removeAll();
-        this.getElement().setProperty("text",
-                text != null ? HtmlUtils.escape(text) : null);
+        this.getElement().setProperty("text", text);
```
tomivirkki, line 285: Please also add the following test case to `NotificationTest.java`:

```java
@Test
public void setText_notificationHasUnmodifiedText() {
    Notification notification = new Notification();
    notification.setText("foo > bar");

    Assert.assertEquals("foo > bar",
            notification.getElement().getProperty("text"));
}
```

tomivirkki (approved):

PR #5797, opened 2023-11-30
fix: respect column filter enabled/disabled in the table

### LINKED ISSUE #5607: SpreadsheetFilterTable doesn't respect Excel enable/disable table filters setting

bwajtr:
## Description

When creating sheet content, the Spreadsheet component reads data about tables in the underlying POI data model and creates a `SpreadsheetTable` instance for each found table. When the table is created, a `PopupButton` containing the filtering options for the given column is created in every cell of the table header. In Excel you can create a "table", but either disable the filtering of the table columns completely, or keep the filtering enabled, but hide the popup buttons. In both cases the popup buttons are not visible in the excel table - this prevents the users to filter/sort the table content. You can check examples of such tables here:
[table_with_disabled_autofilter.xlsx](https://github.com/vaadin/flow-components/files/13515751/table_with_disabled_autofilter.xlsx)

The problem was that the table filtering enable/disable settings were not respected in the `Spreadsheet` component at all and the popup buttons were always created.

This PR adds code to the `SpreadsheetTable` popup button initialization which checks whether the popup buttons should be created or not. The code attempts to cover all cases where the popup buttons should be hidden.
I had to add several new constructors to the related code because I needed references to the backing `xssfTable` or `ctWorksheetAutoFilter` right in the constructors - because the popup buttons are initialized in the constructor of `SpreadsheetTable`.

Added automatic test to check all these cases.

Fixes #5607 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/SpreadsheetTable.java
```diff
@@ -67,10 +72,13 @@ public SpreadsheetTable(Spreadsheet spreadsheet,
      *            Cell range to build the table in
      */
     public SpreadsheetTable(Spreadsheet spreadsheet, Sheet sheet,
-            CellRangeAddress fullTableRegion) {
+            CellRangeAddress fullTableRegion,
+            CTAutoFilter ctWorksheetAutoFilter, XSSFTable xssfTable) {
```
DiegoCardoso, line 76: Changing this constructor without maintaining the old one would cause breaking changes if anyone else uses it. The class is public, so we should be careful with this type of change. In general, ideally, we should try not to add any new API as part of a fix.

bwajtr: Yes, you are correct. I've re-added the constructor in a new commit.
I understand that adding a new API as part of the fix is not nice, however, the other options on how to fix this issue were even worse in my opinion. Because initPopupButtons() is called in the constructor, I need to set the `xssfTable`/`ctWorksheetAutoFilter` before this call - and that is doable only via a new constructor.

DiegoCardoso (approved):

PR #5798, opened 2023-11-30
fix: call to setSplitterPoisition on splitter-dragend event

### LINKED ISSUE #2124: SplitLayout position - set manually + programatically - regression ?
### Description

In v21 if one sets the splitter position programatically and then slides it manually, it can't be re-set again to the same value.
This behaves as expected in v20.

### Expected outcome

By manually sliding the splitter and then clicking 'set to 0', the slider should go back to the position 0.

### Actual outcome

The splitter position doesn't change

```java
SplitLayout splitLayout = new SplitLayout();
splitLayout.addToPrimary(new Div(new Span("PRIMARY")));
splitLayout.addToSecondary(new Div(new Span("SECONDARY")));
add(splitLayout);

splitLayout.addToSecondary(new Button("SET TO 0", buttonClickEvent -> splitLayout.setSplitterPosition(0)));
splitLayout.addToSecondary(new Button("SET TO 90", buttonClickEvent -> splitLayout.setSplitterPosition(90)));

splitLayout.setSplitterPosition(0);
```

### Steps to reproduce

1. Manually slide the splitter to a random position
2. Click 'set to 0'

Or:

1. Click 'set to 90'
2. Manually slide the splitter to a random position
3. Click 'set to 90'

DiegoCardoso:
## Description

Add a call to `setSplitterPosition` on the default `splitter-dragend` listener to force synching the value from the client-side to the server-side. Previously, although we saved the split position in the server when the user changes it in the client, we can't reset it to the previous value from the server because Flow still thinks that the split position hasn't been changed.

This effectively will send a command to the client on every split change, but as it's part of the same roundtrip that the event is triggered, it won't cause any extra communication.

Fixes #2124

## Type of change

- [X] Bugfix
- [ ] Feature

### DIFF
--- a/vaadin-split-layout-flow-parent/vaadin-split-layout-flow-integration-tests/src/test/java/com/vaadin/flow/component/splitlayout/tests/SplitterPositionIT.java
+++ b/vaadin-split-layout-flow-parent/vaadin-split-layout-flow-integration-tests/src/test/java/com/vaadin/flow/component/splitlayout/tests/SplitterPositionIT.java
@@ -102,6 +102,31 @@ public class SplitterPositionIT extends AbstractComponentIT {
         Assert.assertTrue(splitPosition < 81);
     }
 
+    @Test
+    public void setSplitterPositionFromServer_moveOnClient_resetToOriginal() {
+        // Add split layout with 30% splitter position
+        $(NativeButtonElement.class).id("createLayoutJavaApi").click();
+        $(NativeButtonElement.class).id("setSplitPositionJavaApi").click();
+
+        var split = $(SplitLayoutElement.class).id("splitLayoutJavaApi");
+        var primaryComponent = split.getPrimaryComponent();
+
+        var flexBasisInitial = primaryComponent.getCssValue("flex-basis");
+
+        // Move splitter by 150px
+        var splitter = split.getSplitter();
+        Actions resizeAction = new Actions(getDriver());
+        resizeAction.dragAndDropBy(splitter, 150, 0);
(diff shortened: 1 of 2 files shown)


vaadin-split-layout-flow-parent/vaadin-split-layout-flow-integration-tests/src/test/java/com/vaadin/flow/component/splitlayout/tests/SplitterPositionIT.java
```diff
@@ -102,6 +102,31 @@ public void nestedSplitLayouts_moveParentSplitter_getSplitterPosition() {
...
+        var primaryComponent = split.getPrimaryComponent();
+
+        var flexBasisInitial = primaryComponent.getCssValue("flex-basis");
+
+        // Move splitter by 150px
+        var splitter = split.getSplitter();
+        Actions resizeAction = new Actions(getDriver());
+        resizeAction.dragAndDropBy(splitter, 150, 0);
+        resizeAction.perform();
+
```
ugur-vaadin, line 121: Does it make sense to also check whether `flex-basis` has been properly updated? I looked at the other tests and there doesn't seem to be any for this. We can maybe add something like `Assert.assertNotEquals(flexBasisInitial, primaryComponent.getCssValue("flex-basis"));` or add another test to be on the safe side. It is not a blocker though. What do you think?

ugur-vaadin (approved):
LGTM. Added a suggestion comment regarding the test but it's not a blocker.

PR #5806, opened 2023-12-01
fix: use executeJS to set flex value in SplitLayout

DiegoCardoso:
## Description

>[!WARNING]
>A cherry-pick from #5798 could not be performed for V24.1 and below because the API used as part of the solution for it, is not available in the earlier versions.

This PR uses an alternative approach that replaces the use of the `getElement().getStyle().set(...)` API to use `getElement().executeJS(...)`, to ensure that the flex value is always sent to the client, to work around the default Flow behavior that avoids sending values from server to the client in case the value in the server didn't change.

Some notes:
- I am unsure whether this approach is better than the one that landed in V24.2+. In V24.2+ the value from the client is sent back from the server after the `dragend` event, but it will send any data from the server when two equal split position values are set in a row. On the other hand, the new approach will not send any data back to the client on `dragend` but will always update the split position, no matter if the split position value hasn't changed.
- Maybe there could be some logic checking whether the new split position is the same as the current and only use the `getElement().executeJS()` API for this case, but I didn't want to make this any more complex, as I think the downside here might be not big for us to worry too much about.

Part of #2124

## Type of change

- [X] Bugfix


DiegoCardoso:
> LGTM if we should back-port the fix to those branches.

I guess so since the issue has been reported in V21.

Issue #5513, opened 2023-09-26
When sorting a grid with multisort from the backend and hiding a column, the sort order indicator is not updated correctly.


jencal-essers:
### Description

When sorting a grid with multisort from the backend and hiding a column, the sort order indicator is not updated correctly.
The indicator shows that the only applied sort, has order 2.

### Expected outcome

The sort order indicator shows order number 1 or no number.

### Minimal reproducible example

In the bookstore example 
https://github.com/jencal-essers/bookstore-example-sort-order-bug/commit/985b5b1225584732c698439273f246b4efa640dc

### Steps to reproduce

* run the bookstore example (mvn)
* login as admin
* navigate to inventory
* Click the sort 1 button
* Click the sort 2 button 

### Environment

Vaadin version(s): 24.1.12
OS: Windows in wsl ubuntu

### Browsers

Issue is not browser related

vursen:
Confirmed that the issue is reproducible in the latest 24.3. 

Note, the example project is missing `grid.setMultiSort(true, MultiSortPriority.APPEND)`, and the "Sort 2" button isn't added to the page.

vursen:
Here is a more streamlined example for reproduction:

```java
@Route("grid")
public class GridView extends Div {

    public static class Product {

        private final String product;
        private final double price;

        public Product(String product, double price) {
            this.product = product;
            this.price = price;
        }

        public String getProduct() {
            return product;
        }

        public double getPrice() {
            return price;
        }
    }

    public GridView() {
        Grid<Product> grid = new Grid<Product>();
        Grid.Column<Product> product = grid.addColumn(Product::getProduct).setHeader("Product").setSortable(true);
        Grid.Column<Product> price = grid.addColumn(Product::getPrice).setHeader("Price").setSortable(true);

        grid.setMultiSort(true);
        grid.setItems(
                new Product("iPhone", 599.99),
                new Product("iPad", 899.99),
                new Product("MacBook", 1299.99));

        Button sort1 = new Button("sort1");
        sort1.addClickListener(event -> {
            grid.sort(GridSortOrder.asc(product)
                    .thenAsc(price).build());
            price.setVisible(true);
        });

        Button sort2 = new Button("sort2");
        sort2.addClickListener(event -> {
            grid.sort(GridSortOrder.asc(product).build());
            price.setVisible(false);
        });

        add(grid, sort1, sort2);
    }
}
```

vursen:
[...]

PR #5817, opened 2023-12-07
fix: re-render sorter priority indicators when setting new directions

vursen:
## Description

The PR updates the grid connector to ensure the all the sorters, including hidden ones, re-render their priority indicator when new sort directions are applied from the server via `$connector.setSorterDirections`. Additionally, it disables the automatic exclusion of hidden columns from sorting by the web component. In Flow, it's still the developer's responsibility to remove hidden columns from the backend sort order when they get hidden.

Depends on 
- https://github.com/vaadin/web-components/pull/6927

Fixes https://github.com/vaadin/flow-components/issues/5513

## Type of change

- [x] Bugfix


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -390,6 +390,10 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
                     }
                   });
                 });
+
+                // Manually recalculate sorter order indicators in case some of the sorters
+                // were hidden and therefore didn't notify the grid about their direction change.
```
vursen, line 395:
```suggestion
                // Manually update sorter order indicators in case some of the sorters
                // were hidden and therefore didn't notify the grid about their direction change.
```

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
tomivirkki, line 1: All the tests pass for me locally even without these changes in the connector

vursen: Hmm, let me check

tomivirkki (approved):

PR #5826, opened 2023-12-11
fix: don't invoke LitRenderer.withFunction handler when item resolves to null

### LINKED ISSUE #5778: NPE in TreeGrid (race condition) while refreshing and clicking hierachical column
### Description

While refreshing the Grid and simultaneous clicking rapidly the hierachical column it's possible that the application throws an unexpected error: 

```java
2023-11-27T11:06:22.916+01:00 ERROR 979 --- [nio-7009-exec-3] c.v.flow.server.DefaultErrorHandler      :
java.lang.NullPointerException: null
  at java.base/java.util.Objects.requireNonNull(Unknown Source)
  at java.base/java.util.ImmutableCollections$List12.<init>(Unknown Source)
  at java.base/java.util.List.of(Unknown Source)
  at com.vaadin.flow.component.treegrid.TreeGrid.lambda$addHierarchyColumn$63241ece$1(TreeGrid.java:583)
  at com.vaadin.flow.data.renderer.LitRenderer.lambda$withFunction$71d1d723$1(LitRenderer.java:330)
  at com.vaadin.flow.data.renderer.LitRenderer.lambda$createJsRendererFunction$f31f4741$1(LitRenderer.java:205)
  at com.vaadin.flow.internal.nodefeature.ReturnChannelMap.lambda$registerChannel$2a20409b$1(ReturnChannelMap.java:124)
  at com.vaadin.flow.internal.nodefeature.ReturnChannelMap$ChannelImpl.invoke(ReturnChannelMap.java:74)
  at com.vaadin.flow.server.communication.ReturnChannelHandler.handleNode(ReturnChannelHandler.java:78)
  at  com.vaadin.flow.server.communication.rpc.AbstractRpcInvocationHandler.handle(AbstractRpcInvocationHandler.java:74)
  at com.vaadin.flow.server.communication.ServerRpcHandler.handleInvocationData(ServerRpcHandler.java:459)
  at com.vaadin.flow.server.communication.ServerRpcHandler.lambda$handleInvocations$2(ServerRpcHandler.java:440)
  at java.base/java.util.ArrayList.forEach(Unknown Source)
  at com.vaadin.flow.server.communication.ServerRpcHandler.handleInvocations(ServerRpcHandler.java:440)
  at com.vaadin.flow.server.communication.ServerRpcHandler.handleRpc(ServerRpcHandler.java:323)
  at com.vaadin.flow.server.communication.UidlRequestHandler.synchronizedHandleRequest(UidlRequestHandler.java:114)
  at com.vaadin.flow.server.SynchronizedRequestHandler.handleRequest(SynchronizedRequestHandler.java:40)
  at com.vaadin.flow.server.VaadinService.handleRequest(VaadinService.java:1529)
  at com.vaadin.flow.server.VaadinServlet.service(VaadinServlet.java:398)
  at com.vaadin.flow.spring.SpringServlet.service(SpringServlet.java:106)
  at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:658)


```

Solution: add a nullcheck before passing the value from the client to the data communicator 

https://github.com/vaadin/flow-components/blob/e39becb6cc49a96297a9e2a4a5aa039a68754992/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java#L580-L588

```java

if(null == item) {
  return;
}

```

### Expected outcome

No Error

See comments below

### Environment

Vaadin version(s): 24.2.4

bwajtr:
## Description

When defining the LitRenderer (to be used in Grids, ComboBoxes, VirtualLists etc.) you can also attach an event handler to it, which is invoked when the appropriate event is generated in JavaScript. A `click` handler could be defined for example as follows:
```
        return LitRenderer
                .<Item> of("<div @click=${onClick}>${item.name}</div>")
                .withProperty("name", Item::key)
                .withFunction("onClick",
                        item -> System.out.println("Item clicked: " + item));
```
In the above usecase the "onClick" function handler on the server side is invoked when the item is clicked in the grid, and the clicked `item` is passed to the handler.

The problem is, that it may happen that the `item` parameter in the "onClick" function handler can be null: Imagine having a Grid with LitRenderer as above and a "Refresh" button next to it, that sets a completely new set of items to the Grid. If the Refresh action takes a long time, then you have time to click on the item in the Grid. The click is not performed immediately, but it is rather "scheduled" to be sent to the server after the Refresh action finishes. After the refresh, the click action is sent to the server, but the "onClick" event handler is invoked with a `null` item, because the item key for which the click was done is not valid anymore (is not in the grid data set after the refresh).

This is exactly what happened in #5778: You have such click handler in the TreeGrid for Hierarchy columns: https://github.com/vaadin/flow-components/blob/8b25219a4424fb9d0854837f6b6cd61b66b1a8a3/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java#L580
So when you click on the expand button in TreeGrid during a long refresh, then the above handler is invoked with `null` item, which leads to NPE, because the clicked item is not valid after the refresh anymore.

The issue could be fixed directly in TreeGrid onClick handler as @knoobie suggests in the issue, but I think this is a general problem with `LitRenderer`. I don't think it makes sense to invoke the function handler when the item key resolves to a `null` item (=> function handlers are actions that should be typically (not exclusively) done for a specific item => if you don't know the item, then the function handler makes a little sense). There is no indication in JavaDoc if the `item` parameter in the handler is nullable or not. [The documentation](https://vaadin.com/docs/latest/components/grid/flow#handling-events) does not elaborate on it either, however, it somewhat indicates that the item is always not-null because of missing null checks in the example code.

This PR introduces a null check that prevents calling a `LitRenderer` function handler when the item key resolves to null (no item with such key exists in the current grid data set).

There are several places across the code where a similar check is done and which indicates that it's a normal situation that `keyMapper.get(key)` might return null:
https://github.com/vaadin/flow-components/blob/4842332634138bd3de458d16201acd9aead96206/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java#L286
this one even explains when such situation can happen:
https://github.com/vaadin/flow-components/blob/4842332634138bd3de458d16201acd9aead96206/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java#L3361

I think that `LitRenderer` misses such a check too. Sorry for a long description ;).

Fixes #5778

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/java/com/vaadin/flow/data/renderer/LitRenderer.java
```diff
@@ -201,8 +201,9 @@ private Registration createJsRendererFunction(Element container,
                     SerializableBiConsumer<SOURCE, JsonArray> handler = clientCallables
                             .get(handlerName);
                     SOURCE item = keyMapper.get(itemKey);
-
-                    handler.accept(item, args);
+                    if (item != null) {
+                        handler.accept(item, args);
+                    }
```
knoobie, line 206: I like this more general approach 👍

vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/java/com/vaadin/flow/data/renderer/LitRenderer.java
```diff
@@ -201,8 +201,9 @@ private Registration createJsRendererFunction(Element container,
                     SerializableBiConsumer<SOURCE, JsonArray> handler = clientCallables
                             .get(handlerName);
                     SOURCE item = keyMapper.get(itemKey);
-
-                    handler.accept(item, args);
+                    if (item != null) {
```
vursen, line 204: Is there a possibility that a new item will coincidentally reuse the key of one of the previous items?

bwajtr: By default the `KeyMapper` uses Java object identity (hashCode) for equality comparisons. So, if you override equals/hashCode then yes: the keyMapper.get() resolves to the "new" item even when "old" item was clicked in the grid.

Also, for Grid and TreeGrid you can specify your own Key value provider, like so:
```
treeGrid.setUniqueKeyDataGenerator("key", Item::key);
```
Meaning that you can use whatever you want to specify what is the key of an object - and the keyMapper will use that.

From my point of view, two item instances with the same key (same identity) represent one entity - so it should not (theoretically) matter if you run the function handler over the "old" one or the "new" one. It might become a problem when the identity function is not defined correctly and two "not identical" objects share the same key. 

In any case, if you think this is an issue worth pursuing, I suggest creating a new bug for it - because I think this concept is not related to this bug that much...

vursen: I agree with your reasoning 👍 Thanks for the thorough comment.

vursen (approved):

PR #5847, opened 2023-12-15
fix: refresh also combobox value json-data when renderer is set

### LINKED ISSUE #5270: Lit renderer shows empty overlay with readonly multiselect combobox
### Description

I have a multiselect combobox with a custom lit renderer. This renderer works as expected if the MSCB is not read only. I used the lit renderer to render user name on a first line and email and phone just below. However when I set the MSCB to be read only and click the MSCB, I only get a blank overlay.
Screenshot 2023-07-19 at 22 40 20
Screenshot 2023-07-19 at 22 52 02

### Expected outcome

I would expect to see whatever is selected, which works normally if I don't use a custom renderer.

```
    public static Renderer<UserDto> userRenderer() {
        final String template = """
                <div style="display: flex;">
                  <div>
                    <div>${item.fullName}</div>
                    <div style="font-size: var(--lumo-font-size-s); color: var(--lumo-secondary-text-color);">${item.email}</div>
                  </div>
                </div>
                        """;

        return LitRenderer.<UserDto>of(template)
                .withProperty("email", UserDto::getEmailWithPhone)
                .withProperty("fullName", UserDto::getFullName);
    }
```

### Steps to reproduce

1. create a multiselect combobox
2. add elements to it
3. select some of the elements
4. set the MSCB to be readonly
5. click the MSCB

### Environment

Vaadin version(s): 24.1.2
OS: Ventura

### Browsers

Chrome

bwajtr:
## Description

The `MutliSelectComboBox` (MSCB) overlay seems to be empty because the configured `LitRenderer` misses the data configured by `.withProperty()` methods and is therefore unable to render the item properly. An example of such `LitRenderer` is this:
```
return LitRenderer.<Person> of("<div>Name: ${item.fullName}</div>")
                .withProperty("fullName", Person::getFullName);
```
If the renderer does not have the `fullName` data, it will render only "Name: " in the dropdown.

When the MSCB is in read-only mode, you can still open the dropdown overlay (in contrast to normal ComboBox) to review all the selected items. However, when in `read-only` mode, the backend data provider is **not** invoked to fetch the dropdown data, but rather the `selectedItems` property on the web component is used as a data feed for the `LitRenderer`.

As Serhii pointed out in [his comment](https://github.com/vaadin/flow-components/issues/5270#issuecomment-1822797167), the problem is that the `selectedItems` property doesn't contain keys using `propertyNamespace` expected by `LitRenderer` (so in addition to `label` it should contain `lr_0_fullName` etc) - simply it does not contain the data configured with `.withProperty()` method.

I tracked the issue to a point that the `DataGenerator` which is responsible for generating those data for the `LitRenderer` is not invoked when you set MSCB value using `select()` method. The order of actions is the problem here. Consider the following code:

```
...
multiSelect.setReadOnly(true);
multiSelect.setRenderer(createLitRenderer());
multiSelect.select(people.get(0), people.get(1));
...
```
1. When you set the renderer at line 2, the renderer is registered, but the first render is scheduled only as `BeforeClientResponse` action
2. `.select()` is called on line 3 which ultimately causes a call of `MultiSelectComboBox.refreshValue()` that generates and sets the `selectedItems` on the webcomponent - but it generates only `label` data (using `ItemLabelGenerator`)
3. `BeforeClientResponse` processing happens at last - calling the `render` method of the `LitRenderer` - now finally the data generator which is responsible for generating data for the `LitRenderer` is registered to the combobox ([here](https://github.com/vaadin/flow-components/blob/abe9a5085edb505fcf1d9dea289f6af5533499bb/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxRenderManager.java#L76)).

The problem is that at this point the `selectedItems` value is not regenerated, although the set of `DataGenerator`s has changed.

So, this PR refreshes also the combobox value (selected items) data before a client response after Renderer is "rendered".

I think comparable situation [is here](https://github.com/vaadin/flow-components/blob/e39becb6cc49a96297a9e2a4a5aa039a68754992/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxBase.java#L393-L396) - when you set new `ItemLabelGenerator` you also have to do both `dataController.reset()` and ` this.refreshValue();`.

However, I'm not sure if the place where I chose to call `comboBox.refreshValue();` is the correct one... I'm opened to other suggestions.

Fixes #5270 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxRenderManager.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxRenderManager.java
@@ -81,6 +81,8 @@ class ComboBoxRenderManager<TItem> implements Serializable {
 
         comboBox.getDataController().reset();
         if (comboBox.getValue() != null) {
+            // renderer might added new dataGenerator -> refresh getValue() JSON
+            // data so it also contains the new dataGenerator data
             comboBox.refreshValue();
         }
     }
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxRenderManager.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxRenderManager.java
@@ -15,13 +15,13 @@
  */
 package com.vaadin.flow.component.combobox;
 
-import com.vaadin.flow.data.renderer.Renderer;
-import com.vaadin.flow.shared.Registration;
-
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
+
+import com.vaadin.flow.data.renderer.Renderer;
+import com.vaadin.flow.shared.Registration;
 
 /**
  * Internal class for managing rendering related logic for combo box components
@@ -80,5 +80,8 @@ class ComboBoxRenderManager<TItem> implements Serializable {
         renderingRegistrations.add(rendering.getRegistration());
 
         comboBox.getDataController().reset();
+        if (comboBox.getValue() != null) {
+            comboBox.refreshValue();
+        }
     }
 }
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/main/java/com/vaadin/flow/component/combobox/test/MultiSelectComboBoxLitRendererPage.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/main/java/com/vaadin/flow/component/combobox/test/MultiSelectComboBoxLitRendererPage.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2000-2023 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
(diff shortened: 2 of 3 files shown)


vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxRenderManager.java
```diff
@@ -80,5 +80,8 @@ private void render() {
         renderingRegistrations.add(rendering.getRegistration());
 
         comboBox.getDataController().reset();
+        if (comboBox.getValue() != null) {
+            comboBox.refreshValue();
+        }
```
sissbruecker, line 85: From what I understand the change makes sense, setting a renderer changes the data generator, so you need to run `refreshValue` again which relies on the data generator. 

Since this class is about managing the renderer, I would consider moving this code outside of this class to the call location that sets the renderer (`ComboBoxBase.setRenderer`). A code comment also can't hurt to explain why we need to regenerate the value.

bwajtr: I get your point about moving this code out of that place. But if I put it to the `ComboBoxBase.setRenderer`, then it's code would have to like this:
```
    public void setRenderer(Renderer<TItem> renderer) {
        Objects.requireNonNull(renderer, "The renderer must not be null");

        renderManager.setRenderer(renderer);

        // renderer might add new dataGenerator -> refresh getValue() data
        // so it also contains the new dataGenerator data
        runBeforeClientResponse(ui -> {
            if (getValue() != null) {
                refreshValue();
            }
        });
    }
 ```
 
 that also does not seem right: a lot of details of `setRenderer` is leaked in it - that it might (or might not) set dataGenerator and that it's doe before client response...

sissbruecker: Makes sense, in that case I prefer the original approach, no need to overcomplicate it at this point 🙂. Just adding the comment there would be fine.

bwajtr: Ok, I've just added a comment

sissbruecker (approved):

PR #5850, opened 2023-12-15
docs: Clarify the uploaded file types are checked by web-component

mshabarov:
Upload file types are checked in browser by we-component, not on the server side.

### DIFF
--- a/vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
+++ b/vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
@@ -285,13 +285,21 @@ public class Upload extends Component implements HasSize, HasStyle {
     }
 
     /**
-     * Specify the types of files that the server accepts. Syntax: a MIME type
-     * pattern (wildcards are allowed) or file extensions. Notice that MIME
-     * types are widely supported, while file extensions are only implemented in
-     * certain browsers, so it should be avoided.
+     * Specify the types of files that the Upload web-component accepts. Syntax:
+     * a MIME type pattern (wildcards are allowed) or file extensions. Notice
+     * that MIME types are widely supported, while file extensions are only
+     * implemented in certain browsers, so it should be avoided.
      * <p>
      * Example: <code>"video/*","image/tiff"</code> or
      * <code>".pdf","audio/mp3"</code>
+     * <p>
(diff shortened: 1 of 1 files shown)


vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -285,13 +285,21 @@ public boolean isDropAllowed() {
...
      * Example: <code>"video/*","image/tiff"</code> or
      * <code>".pdf","audio/mp3"</code>
+     * <p>
+     * File format restrictions are checked only on the client side (browser).
+     * They indicate the hints for users as to what file types to upload. Using
+     * this method won’t restrict the uploaded file’s format on the server side.
+     * If required, it’s the responsibility of the application developer to
+     * implement application-specific restrictions on the server side in one or
+     * more of the Upload component’s event listeners (e.g., in
+     * {@link #addSucceededListener}).
```
knoobie, line 302: Suggestions to make it clear (optional, can be discarded if you don't wanna create a lot of fuss): deprecate the method and add another method like setBrowserAcceptedFileTypes or setClientAcceptedFileTypes to make the meaning clear and reach more people with the deprecation notice

Legioth: I agree with deprecating, but I think we should have a slightly more sophisticated replacement API (and we should introduce a replacement before we add any deprecation).

The problem with the current API is that it mixes file extensions and MIME types which in turn makes it impossible to implement solid server-side filtering (since what's intended as a `.pdf` file type requirement could just as well be satisfied by setting the MIME type as `.pdf`).

Instead, we should have one of these API alternatives:
* One method for giving a list of file extensions and another method for a list of MIME types.
* One method for defining accepted [file extension, MIME type] pairs.

In either case, we would not need any separate API for defining the value to send to the browser but could instead derive that value from the primary data used for server-side filtering.

knoobie: Totally agree with everything you said 👍

mshabarov: Good points, these deserve a separate ticket in Flow. Let us make it and implement later.

vursen (approved):

PR #5853, opened 2023-12-18
fix: handle empty node id in component renderer

### LINKED ISSUE #5830: ComboBox with custom renderer shows stale items before query finishes
### Description

When using a custom renderer with a ComboBox, when you open the dropdown you can see stale items before the fetch completes.  In my current project, we've run into multiple situations where this confuses users.

combobox-stale-render

### Expected outcome

If you do not use a renderer, the prev items are cleared when the dropdown is opened which I would expect to happen with a custom renderer as well.

```
var cb = new ComboBox<Person>();
cb.setRenderer(item -> return new new Span(item.getFirstName() + " " + item.getLastName()));
cb.setItems(generateLargePersonBeanList());
```

### Steps to reproduce

1. add above snippet to view
2. open combobox and filter list, scroll down a bit 
3. close combobox
4. open again
5. while the fetch is completing (spinner at top right), you'll be able to see items from your previous filtered list (probably just a split second, longer fetches/slow connection makes longer)

### Environment

Vaadin version(s): 23.3.29
OS: Mac

### Browsers

Chrome, Firefox

ugur-vaadin:
## Description

This PR handles the empty nodeid case for Flow component renderer. This is already [handled](https://github.com/vaadin/flow-components/blob/8cebb28a0f8ecb55dc3e79c78da42b4831bcbb42/vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/flow-component-directive.js#L34-L35) in the newer Flow component directive. With the new change, if a null/undefined node ID is received and there is a rendered child, it is removed.

Note: The test will succeed without the change in v24. The reason is that the Flow component renderer has already been refactored out of the renderer.

Note 2: There is no test added to the renderer package since I could not come up with a good test. This test is added to test at least the use-case provided in the related issue for this change.

Fixes #5830

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/ComponentRendererIT.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/ComponentRendererIT.java
@@ -68,28 +68,24 @@ public class ComponentRendererIT extends AbstractComboBoxIT {
     }
 
     @Test
-    public void throttledDataProvider_scrollDown_close_noItemsWhenReopened()
-            throws InterruptedException {
+    public void multiplePagesOfItems_scrollDown_close_noItemsWhenReopened() {
         ComboBoxElement comboBox = $(ComboBoxElement.class)
-                .id("throttled-data-provider");
+                .id("multiple-pages-of-items");
 
         comboBox.openPopup();
         waitUntilTextInContent("Song");
 
-        int pageSize = 50;
-        int pageCount = 12;
-        for (int i = 0; i < pageCount * pageSize; i += pageSize) {
+        for (int i = 0; i < 600; i += 50) {
             scrollToItem(comboBox, i);
         }
 
         comboBox.closePopup();
 
-        String text = (String) executeScript("arguments[0].open();"
+        String firstItemText = (String) executeScript("arguments[0].open();"
                 + "const item = document.querySelector('vaadin-combo-box-item');"
                 + "const spanInItem = item.querySelector('span');"
                 + "return spanInItem ? spanInItem.innerText : '';", comboBox);
-
-        Assert.assertEquals("", text);
+        Assert.assertEquals("", firstItemText);
     }
 
     private void testItems(TestBenchElement comboBox) {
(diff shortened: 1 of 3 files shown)


vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/flow-component-renderer.js
```diff
@@ -104,7 +104,11 @@ class FlowComponentRenderer extends PolymerElement {
   }
 
   _attachRenderedComponentIfAble() {
-    if (!this.nodeid || !this.appid) {
+    if (this.appid == null) {
+      return;
+    }
+    if (this.nodeid == null) {
+      this.remove(this.firstChild);
```
tomivirkki, line 111: Use `removeChild` instead of [`remove`](https://developer.mozilla.org/en-US/docs/Web/API/Element/remove) (which removes the element itself, in this case, the `<flow-component-renderer>`, from its parent element). Note that `removeChild` throws if the given parameter is `undefined` (no child element).

ugur-vaadin: Done.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/ComponentRendererIT.java
```diff
@@ -66,14 +61,33 @@ public void captionsForItemsExistWhenFirstSettingDataProvider() {
...
+        for (int i = 0; i < 600; i += 50) {
+            scrollToItem(comboBox, i);
+        }
+
+        comboBox.closePopup();
+
+        String firstItemText = (String) executeScript("arguments[0].open();"
+                + "const item = document.querySelector('vaadin-combo-box-item');"
+                + "const spanInItem = item.querySelector('span');"
+                + "return spanInItem ? spanInItem.innerText : '';", comboBox);
```
vursen, line 87: **nit:**

```suggestion
                + "return document.querySelector('vaadin-combo-box-item')?.textContent;", comboBox);
```

vursen (approved):

PR #5881, opened 2023-12-28
fix: prevent context menu on selection column left-click 

### LINKED ISSUE #5833: GridContextMenu does not work with grid in multi-select mode
### Description

When a grid is set in multi-select mode and uses a GridContextMenu with the gridContextMenu.setOpenOnClick(true) you cannot select anything in the grid.

### Expected outcome

Be able to select something in the grid.

```java
Grid grid = new Grid<>();
grid.setSelectionMode(SelectionMode.MULTI);

GridContextMenu gridContextMenu = new new GridContextMenu<>(grid);
gridContextMenu.setOpenOnClick(true);
```

### Steps to reproduce

Try to select any item in the grid.

### Environment

Vaadin version(s):  24.2.6

### Browsers

Issue is not browser related

bwajtr:
## Description

You can set the grid context menu to also open on left-click (instead of right-click which is the default) by calling `gridContextMenu.setOpenOnClick(true)`. However, this interferes with checkboxes in the selection column when in multi-selection grid (only the context menu is opened, but the selection checkbox is not checked -> therefore, no row can be selected)

I described the reasons why this has been working in V24.1 and why it's not working now in [this comment](https://github.com/vaadin/flow-components/issues/5833#issuecomment-1867933676). The comment also presents the context for the solution (and also some worse solutions that I tried).

This PR introduces a way how to prevent calling `e.preventDefault()` inside `contextMenuTargetConnector.js -> openOnHandler` that is responsible for the selection checkbox not being checked. I've introduced a new function that can be defined on the context menu target, that (if implemented) can prevent any context menu processing code. This function is then implemented by the `gridConnector`, where the code specific to the multi-selection grid can be put.

Note that the described issue is specific **only to Flow** usage of the Grid and ContextMenu - [see this comment](https://github.com/vaadin/flow-components/issues/5833#issuecomment-1871127437) related to web component usage

Fixes #5833

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/resources/META-INF/resources/frontend/contextMenuTargetConnector.js
+++ b/vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/resources/META-INF/resources/frontend/contextMenuTargetConnector.js
@@ -12,6 +12,9 @@ import * as Gestures from '@vaadin/component-base/src/gestures.js';
 
     target.$contextMenuTargetConnector = {
       openOnHandler: tryCatchWrapper(function (e) {
+        if (target.preventContextMenu && target.preventContextMenu(e)) {
+          return;
+        }
         e.preventDefault();
         e.stopPropagation();
         this.$contextMenuTargetConnector.openEvent = e;
--- a/vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/resources/META-INF/resources/frontend/contextMenuTargetConnector.js
+++ b/vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/resources/META-INF/resources/frontend/contextMenuTargetConnector.js
@@ -12,6 +12,7 @@ import * as Gestures from '@vaadin/component-base/src/gestures.js';
 
     target.$contextMenuTargetConnector = {
       openOnHandler: tryCatchWrapper(function (e) {
+        // used by Grid to prevent context menu on selection column click
         if (target.preventContextMenu && target.preventContextMenu(e)) {
           return;
         }
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
@@ -1040,6 +1040,15 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
           return { key, columnId };
         });
 
+        grid.preventContextMenu =  tryCatchWrapper(function (event) {
+            const isLeftClick = event.type === 'click';
+            const firstColumnClicked = grid.getEventContext(event).column === grid._getColumns()[0]
+            const isMultiSelectGrid = selectionMode === validSelectionModes[2];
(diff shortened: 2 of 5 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -1053,6 +1053,15 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
           return { key, columnId };
         });
 
+        grid.preventContextMenu =  tryCatchWrapper(function (event) {
```
web-padawan, line 1056: nit: extra whitespace.

```suggestion
        grid.preventContextMenu = tryCatchWrapper(function (event) {
```

vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/resources/META-INF/resources/frontend/contextMenuTargetConnector.js
```diff
@@ -12,6 +12,9 @@ import * as Gestures from '@vaadin/component-base/src/gestures.js';
 
     target.$contextMenuTargetConnector = {
       openOnHandler: tryCatchWrapper(function (e) {
+        if (target.preventContextMenu && target.preventContextMenu(e)) {
```
web-padawan, line 15: Consider adding a comment e.g. "Used by Grid to prevent context menu on selection column click".
Otherwise it could be unclear when looking at this code where `preventContextMenu` comes from.

web-padawan (approved):

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -1053,6 +1053,15 @@ import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
...
+            const isLeftClick = event.type === 'click';
+            const firstColumnClicked = grid.getEventContext(event).column === grid._getColumns()[0]
+            const isMultiSelectGrid = selectionMode === validSelectionModes[2];
+            const selectionColumnClicked = isMultiSelectGrid && firstColumnClicked;
+
+            return isLeftClick && selectionColumnClicked;
```
vursen, line 1062: Would this work?

```suggestion
            const { column } = grid.getEventContext(event);
            return isLeftClick && column instanceOf GridFlowSelectionColumn;
```

vursen: I wonder if the context menu is supposed to open on selection column even on right click.

bwajtr: > wondering if the context menu is intended to open when right-clicking on the selection column

Yes, this also crossed my mind. Let's discuss on daily... On one side there are no issues with it today, on the other side it would be more consistent to not show it at all (on selection column)

bwajtr: > Would this work?

Yes, that works. Much nicer... thanks!

PR #5889, opened 2024-01-01
fix: use a custom Tab id if available

### LINKED ISSUE #5825: TabSheet: setting an id on a Tab does not work and hangs the view
### Description of the bug

When setting an id on a tab, the view will display the loading progress icon forever. Furthermore, the id is overwritten internally in the add() method on TabSheet.

image

### Expected behavior

- the page should at least load.
- the id that is set should not be overwritten in the `add()` method.

```
@PageTitle("Hello World")
@Route(value = "hello")
@RouteAlias(value = "")
public class HelloWorldView extends HorizontalLayout {

  public HelloWorldView() {
    Tab myTab = new Tab("the tab");
    TabSheet myTabSheet = new TabSheet();
    myTabSheet.add(myTab, new Span("content"));
    // remove below line to make it work
    myTab.setId("new-id");
    add(myTabSheet);
  }
}
```

### Versions

Flow: 24.2.4
Vaadin: 24.2.5
Java: JetBrains s.r.o. 17.0.9
OS: amd64 Linux 6.2.0-37-generic
Browser: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:120.0) Gecko/20100101 Firefox/120.0
Live reload: Java active
          (Spring Boot Devtools): Front end active

bwajtr:
## Description

The `TabSheet` component uses `Tab` id attribute to link the associated `content` element with it. When a `Tab` is selected, the element with `tab='tab-id-xxxxxx'` attribute is searched and displayed in the `TabSheet` content pane.

When using `TabSheet` as a web-component, it's the responsibility of the developer to make sure that the `id` of the `vaadin-tab` matches with the `tab` attribute of it's content. See [example here](https://vaadin.com/docs/latest/components/tabs#tab-sheet). However, when used from Flow, the developer does not have to set the `Tab` id at all -> it'll be generated for him.

The problem was when the developer _wanted_ to set a custom `Tab` id. You could call the `Tab.setId()`, but this did not change the `tab` attribute on the content element -> the `Tab` and `content` got disconnected and the whole `TabSheet` functionality stopped working.

This PR introduces two things:
1. The `TabSheet` code does not generate a random `Tab` id if there is already custom Id set on the `Tab`
2. It defers linking the `Tab` with `content` to `beforeClientResponse` so the developer can also set the id after the `Tab` have been added to the `TabSheet`

Note that this fixes only a situation when you call `Tab.setId()` during `TabSheet` "initialization phase" - meaning before the created `TabSheet` is sent to the client for the first time. If you, for example, change the `Tab` id in some Button click handler, then the `TabSheet` will still stop working - because you'd disconnect the `Tab` with the `content`.

After some consideration, I've chosen not to include code fixing this case in this PR. If you change id of the `Tab`, some other things will stop working too: for example the `aria-labelledby` attribute of the `content` element. So more things will have to be changed to fix this properly. On the other hand, how often somebody changes id attribute of an element after it's attached to the DOM?

Fixes #5825

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-tabs-flow-parent/vaadin-tabs-flow/src/main/java/com/vaadin/flow/component/tabs/TabSheet.java
+++ b/vaadin-tabs-flow-parent/vaadin-tabs-flow/src/main/java/com/vaadin/flow/component/tabs/TabSheet.java
@@ -28,6 +28,7 @@ import com.vaadin.flow.component.HasSize;
 import com.vaadin.flow.component.HasStyle;
 import com.vaadin.flow.component.Tag;
 import com.vaadin.flow.component.Text;
+import com.vaadin.flow.component.UI;
 import com.vaadin.flow.component.dependency.JsModule;
 import com.vaadin.flow.component.dependency.NpmPackage;
 import com.vaadin.flow.component.shared.HasPrefix;
@@ -35,6 +36,7 @@ import com.vaadin.flow.component.shared.HasSuffix;
 import com.vaadin.flow.component.shared.HasThemeVariant;
 import com.vaadin.flow.component.shared.SlotUtils;
 import com.vaadin.flow.dom.Element;
+import com.vaadin.flow.function.SerializableConsumer;
 import com.vaadin.flow.shared.Registration;
 
 /**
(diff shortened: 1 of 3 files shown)


tomivirkki (approved):

vaadin-tabs-flow-parent/vaadin-tabs-flow/src/main/java/com/vaadin/flow/component/tabs/TabSheet.java
```diff
@@ -139,10 +141,7 @@ public Tab add(Tab tab, Component content, int position) {
             tabToContent.get(tab).removeFromParent();
         }
 
-        // On the client, content is associated with a tab by id
-        var id = "tabsheet-tab-" + UUID.randomUUID().toString();
-        tab.setId(id);
-        content.getElement().setAttribute("tab", id);
+        linkTabToContent(tab, content);
```
vursen, line 144: What happens if the user changes a tab's id later?

vursen (approved):

vaadin-tabs-flow-parent/vaadin-tabs-flow/src/main/java/com/vaadin/flow/component/tabs/TabSheet.java
tomivirkki: This case is also mentioned in the PR description

PR #5897, opened 2024-01-04
fix: reset fetch range when requested page is out of visible range

### LINKED ISSUE #5865: Vaadin Grid Scrolling Issue
### Description

When I have a Grid with a lot of Items (Paged) and a user is scrolling really fast, because for example he wants to look at the last elements, I get an Exception thrown. 
> 2023-12-19T17:27:14.987+01:00  WARN 19368 --- [nio-8081-exec-8] c.v.flow.data.provider.DataCommunicator  : Attempted to fetch more items from server than allowed in one go: number of items requested '19100', maximum items allowed '500'.

This results in not all element being loaded, only the first 500 Elements get fetched. My Page Size is 50. So 10 Pages get fetched, but the rest is blocked. The user that scrolled too fast is presented with an empty looking grid.

It seems to work better (or rather worse) the more elements that can be shown in the grid. Also I artificially timeout in the demo, to simulate slower read times from different servers.
In the Demo Project that I have linked to, the bug is not as reliable as in my real project, there I have a lot more data. But with enough scrolling and patience it can be reproduced.

I tried to continue my search on the client side, and it looks to me, that a unordered list in the gridConnector.js seems to be a problem.

Well last time I checked this was the problem. The index property on the elements in the visibleRows array were not ordered.
```js
let start = visibleRows.length > 0 ? visibleRows[0].index : 0;
let end = visibleRows.length > 0 ? visibleRows[visibleRows.length - 1].index : 0;
```
But now I checked again and this does not seem to be the origin of the problem.
image
As you can see, the breakpoint is configured to only stop when ve is greater than 10.
When the code breaks on this conditional breakpoint, we can know, that the server will run into an exception, as ve is not allowed to be greater than 10
This state is immediately fallowed by this exception on the server:
image
Because as mentioned above, there is a maximum number of pages that is allowed to being fetched.
```java
private int getMaximumAllowedItems() {
        return Math.max(500, 10 * this.pageSize);
}
```
Hopefully this helps analyse the problem better.
The Code uses cryptic variable names, as somehow I was not able to turn of frontend minimization (I know that I was able to do this at some point in time, but either I forgot or its not possible anymore)
This is the source:
```js
grid.$connector.fetchPage = tryCatchWrapper(function (fetch, page, parentKey) {
          // Determine what to fetch based on scroll position and not only
          // what grid asked for

          // The buffer size could be multiplied by some constant defined by the user,
          // if he needs to reduce the number of items sent to the Grid to improve performance
          // or to increase it to make Grid smoother when scrolling
          const visibleRows = grid._getRenderedRows();
          let start = visibleRows.length > 0 ? visibleRows[0].index : 0;
          let end = visibleRows.length > 0 ? visibleRows[visibleRows.length - 1].index : 0;
          let buffer = end - start;

          let firstNeededIndex = Math.max(0, start - buffer);
          let lastNeededIndex = Math.min(end + buffer, grid._flatSize);

          let firstNeededPage = page;
          let lastNeededPage = page;
          for (let idx = firstNeededIndex; idx <= lastNeededIndex; idx++) {
            const { cache, index } = dataProviderController.getFlatIndexContext(idx);
            // Try to match level by going up in hierarchy. The page range should include
            // pages that contain either of the following:
            //   - visible items of the current cache
            //   - same level parents of visible descendant items
            // If the parent items are not considered, Flow would remove the hidden parent
            // items from the current level cache. This can lead to an infinite loop when using
            // scrollToIndex feature.
            const sameLevelPage = grid.$connector._getSameLevelPage(parentKey, cache, index);
            if (sameLevelPage === null) {
              continue;
            }
            firstNeededPage = Math.min(firstNeededPage, sameLevelPage);
            lastNeededPage = Math.max(lastNeededPage, sameLevelPage);
          }

          let firstPage = Math.max(0, firstNeededPage);
          let lastPage =
            parentKey !== root ? lastNeededPage : Math.min(lastNeededPage, Math.floor(grid.size / grid.pageSize));
          let lastRequestedRange = lastRequestedRanges[parentKey];
          if (!lastRequestedRange) {
            lastRequestedRange = [-1, -1];
          }
          if (lastRequestedRange[0] != firstPage || lastRequestedRange[1] != lastPage) {
            lastRequestedRange = [firstPage, lastPage];
            lastRequestedRanges[parentKey] = lastRequestedRange;
            let count = lastPage - firstPage + 1;
            fetch(firstPage * grid.pageSize, count * grid.pageSize);
          }
});
```

### Expected outcome

No matter how fast the user scrolls, he / she will always see the elements he wants to see. When he / she scrolls all the way to the bottom of the page, he / she wants to see the last elements.
Of course the user could scroll up a little and then down again, but this is still an inconvenience.

Have a look at the fallowing project (reproducible example):

https://github.com/DarioViva42/grid-scrolling-error

There are already 1000 pages of data as an example. Of course in real project the data source is not from files but this is just for demonstrating purposes.
Note: When you want to have more pages, just use the create endpoint with the count query parameter specifying the number of pages (as files) that should be created.

### Steps to reproduce

1. Start the application.
2. Look at the grid.
3. Scroll up and down the grid, very quickly / fast.
4. Look at the log

### Environment

Vaadin version(s): 24.3.1
OS: Windows (probably does not matter)

### Browsers

_No response_

vursen:
## Description

The PR updates the grid connector's `fetchPage` method to reset the fetch range instead of extending it to avoid server exceptions about too large range when the requested page is outside the visible range.

The requested page can end up outside the visible range when the user, for example, scrolls the grid to the end and then back to the start within 150 ms (= root request debouncer timeout). Since the first page is already loaded, there will be only a request for the last page scheduled in the debouncer. The fetch range needs to be reset to ensure the resolution of that "last page" request, as well as any other requests that happen to be hanging, by the `confirm` method. After that, the grid will request the first page or whatever in the viewport again. The obvious drawback is that there will be two round-trips and can be a short blink if the server is slow. However, I propose this as a workaround for the time being. The combo-box seems to work similarly. 

I also considered canceling such requests to avoid blinks 58383b7027ae4239f4905ac4751e984dbc82c615, but that caused some ITs using small page sizes to fail and I couldn't get to the reason in a reasonable amount of time.

Fixes #5865 

## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
@@ -241,22 +241,26 @@ import { GridFlowSelectionColumn } from "./vaadin-grid-flow-selection-column.js"
         });
 
         grid.$connector.fetchPage = tryCatchWrapper(function (fetch, page, parentKey) {
+          // Adjust the requested page to be within the valid range in case
+          // the grid size has changed while fetchPage was debounced.
+          if (parentKey === root) {
+            page = Math.min(page, Math.floor(grid.size / grid.pageSize));
+          }
+
           // Determine what to fetch based on scroll position and not only
           // what grid asked for
+          const visibleRows = grid._getRenderedRows();
+          let start = visibleRows.length > 0 ? visibleRows[0].index : 0;
+          let end = visibleRows.length > 0 ? visibleRows[visibleRows.length - 1].index : 0;
 
           // The buffer size could be multiplied by some constant defined by the user,
           // if he needs to reduce the number of items sent to the Grid to improve performance
           // or to increase it to make Grid smoother when scrolling
-          const visibleRows = grid._getRenderedRows();
-          let start = visibleRows.length > 0 ? visibleRows[0].index : 0;
-          let end = visibleRows.length > 0 ? visibleRows[visibleRows.length - 1].index : 0;
           let buffer = end - start;
-
           let firstNeededIndex = Math.max(0, start - buffer);
           let lastNeededIndex = Math.min(end + buffer, grid._flatSize);
 
-          let firstNeededPage = page;
-          let lastNeededPage = page;
+          let pageRange = [null, null];
           for (let idx = firstNeededIndex; idx <= lastNeededIndex; idx++) {
             const { cache, index } = dataProviderController.getFlatIndexContext(idx);
             // Try to match level by going up in hierarchy. The page range should include
@@ -270,18 +274,26 @@ import { GridFlowSelectionColumn } from "./vaadin-grid-flow-selection-column.js"
             if (sameLevelPage === null) {
               continue;
             }
-            firstNeededPage = Math.min(firstNeededPage, sameLevelPage);
-            lastNeededPage = Math.max(lastNeededPage, sameLevelPage);
+            pageRange[0] = Math.min(pageRange[0] ?? sameLevelPage, sameLevelPage);
+            pageRange[1] = Math.max(pageRange[1] ?? sameLevelPage, sameLevelPage);
           }
 
-          let firstPage = Math.max(0, firstNeededPage);
-          let lastPage =
-            parentKey !== root ? lastNeededPage : Math.min(lastNeededPage, Math.floor(grid.size / grid.pageSize));
+          // When the viewport doesn't contain the requested page or it doesn't contain any items from
+          // the requested level at all, it means that the scroll position has changed while fetchPage
+          // was debounced. For example, it can happen if the user scrolls the grid to the bottom and
+          // then immediately back to the top. In this case, the request for the last page will be left
+          // hanging. To avoid this, as a workaround, we reset the range to only include the requested page
+          // to make sure all hanging requests are resolved. After that, the grid requests the first page
+          // or whatever in the viewport again.
+          if (pageRange.some((p) => p === null) || page < pageRange[0] || page > pageRange[1]) {
+            pageRange = [page, page];
+          }
+
           let lastRequestedRange = lastRequestedRanges[parentKey] || [-1, -1];
-          if (lastRequestedRange[0] != firstPage || lastRequestedRange[1] != lastPage) {
-            lastRequestedRanges[parentKey] = [firstPage, lastPage];
-            let count = lastPage - firstPage + 1;
-            fetch(firstPage * grid.pageSize, count * grid.pageSize);
+          if (lastRequestedRange[0] != pageRange[0] || lastRequestedRange[1] != pageRange[1]) {
+            lastRequestedRanges[parentKey] = pageRange;
+            let pageCount = pageRange[1] - pageRange[0] + 1;
+            fetch(pageRange[0] * grid.pageSize, pageCount * grid.pageSize);
           }
         });
 
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
@@ -240,40 +240,27 @@ import { GridFlowSelectionColumn } from "./vaadin-grid-flow-selection-column.js"
           });
         });
 
-        grid.$connector.cancelRootRequests = tryCatchWrapper(function (predicate) {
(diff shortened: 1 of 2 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.js
```diff
@@ -241,22 +241,26 @@ import { GridFlowSelectionColumn } from "./vaadin-grid-flow-selection-column.js"
         });
 
         grid.$connector.fetchPage = tryCatchWrapper(function (fetch, page, parentKey) {
+          // Adjust the requested page to be within the valid range in case
+          // the grid size has changed while fetchPage was debounced.
+          if (parentKey === root) {
+            page = Math.min(page, Math.floor(grid.size / grid.pageSize));
+          }
```
vursen, line 248: **note:** Replaces this [line](https://github.com/vaadin/flow-components/pull/5897/files#diff-255f48ed31d114d33fd62d5f44b260202e622606b1f6ffb8fa39fbbc96306a84L279).

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
```diff
@@ -241,22 +241,26 @@ import { GridFlowSelectionColumn } from "./vaadin-grid-flow-selection-column.js"
         });
 
         grid.$connector.fetchPage = tryCatchWrapper(function (fetch, page, parentKey) {
+          // Adjust the requested page to be within the valid range in case
+          // the grid size has changed while fetchPage was debounced.
+          if (parentKey === root) {
+            page = Math.min(page, Math.floor((grid.size - 1) / grid.pageSize));
```
vursen, line 247: **note:** I moved this [line](https://github.com/vaadin/flow-components/pull/5897/files#diff-255f48ed31d114d33fd62d5f44b260202e622606b1f6ffb8fa39fbbc96306a84L279) up here for better readability. I also fixed it, as it previously returned the number of last page + 1, which again exceeded the valid range.

vursen, line 247: **note:** This line was previously [here](https://github.com/vaadin/flow-components/pull/5897/files#diff-255f48ed31d114d33fd62d5f44b260202e622606b1f6ffb8fa39fbbc96306a84L279). I moved it up just for better readability. The tests also revealed a bug: the previous version always added +1 to the number of the last page, thus giving an invalid page again.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
```diff
@@ -270,18 +274,26 @@ import { GridFlowSelectionColumn } from "./vaadin-grid-flow-selection-column.js"
...
+
+          // When the viewport doesn't contain the requested page or it doesn't contain any items from
+          // the requested level at all, it means that the scroll position has changed while fetchPage
+          // was debounced. For example, it can happen if the user scrolls the grid to the bottom and
+          // then immediately back to the top. In this case, the request for the last page will be left
+          // hanging. To avoid this, as a workaround, we reset the range to only include the requested page
+          // to make sure all hanging requests are resolved. After that, the grid requests the first page
+          // or whatever in the viewport again.
+          if (pageRange.some((p) => p === null) || page < pageRange[0] || page > pageRange[1]) {
+            pageRange = [page, page];
```
vursen, line 289: **note:** Despite the workaround, there may be still situations when certain requests will end up hanging. For example, it can happen when the user scrolls from index range `[0, 10]` to index range `[10,  20]` and `size` suddenly changes to `10` before the debouncer timeout. In this case, the adjusted requested page will point to the end of the already loaded range `[0, 10]`, so the method will skip loading but the requests for `[10, 20]` still exist.

vursen: To work this around, I considered doing a data communicator reset when the connector decides that the fetch range has not changed. However, when I tried that, I sometimes observed multiple round-trips, which made me afraid that it can result in a performance degradation given how convoluted the connector is at the moment.

sissbruecker (approved):
Change makes sense and seems to work well

tomivirkki (approved):
Didn't find any issues either. lgtm

PR #5916, opened 2024-01-09
fix: prevent grid from requesting too many sub-level items

### LINKED ISSUE #5902: [TreeGrid] Instantly scrolling from expanded item to another item causes grid to load whole sub-level of the expanded item
### Description

Finding from #5865 

When you instantly scroll from expanded item 0 to item 1 (expanded/collapsed doesn't matter) e.g. with `scrollToIndex`, Grid requests the whole sub-level of the expanded item 0 regardless of its size. 

This can be highly inefficient in the case of large grids with thousands of items: processing such a large amount of items can take a significant amount of time, as well as increase memory usage.

### Expected outcome

Grid should properly debounce sub-level requests to send only those that are still within the visible range.

```java
@Route("grid")
    private static class DataProvider
            extends AbstractHierarchicalDataProvider<String, Void> {

        @Override
        public int getChildCount(HierarchicalQuery<String, Void> query) {
            return 1000;
        }

        @Override
        public Stream<String> fetchChildren(
                HierarchicalQuery<String, Void> query) {
            int limit = query.getLimit();
            int offset = query.getOffset();
            if (query.getParent() != null) {
                return IntStream.range(offset, offset + limit)
                    .mapToObj(index -> query.getParent() + "-" + index);
            }

            return IntStream.range(offset, offset + limit)
                    .mapToObj(index -> "Item " + index);

        }

        @Override
        public boolean hasChildren(String item) {
            return true;
        }

        @Override
        public boolean isInMemory() {
            return true;
        }

    }

    public GridView() {
        TreeGrid<String> treeGrid = new TreeGrid<>();
        treeGrid.setDataProvider(new DataProvider());
        treeGrid.addHierarchyColumn(ValueProvider.identity()).setHeader("Name");
        treeGrid.setPageSize(50);
        treeGrid.getStyle().setHeight("800px");


        Button scrollToItem1 = new Button("scroll to Item 1", event -> {
            treeGrid.scrollToIndex(1);
        });

        add(treeGrid, scrollToItem1);
    }
}
```

### Steps to reproduce

1. Expanded `Item 0`
2. Click on the button "Scroll to Item 1"
3. See that Grid requested all 1000 items

image

### Environment

Vaadin version(s): 24.4 and earlier
OS: Mac OS

### Browsers

_No response_

vursen:
## Description

The PR fixes the request debounce logic in the grid connector to prevent it from potentially requesting an excessive number of sub-level items and getting a server exception as a result. The fix is achieved by moving the computation of the visible sub-level range inside the debouncer callback, which essentially replicates the approach already used for the root level.

Fixes #5902 

## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
@@ -215,24 +215,40 @@ import { GridFlowSelectionColumn } from "./vaadin-grid-flow-selection-column.js"
           return false;
         });
 
+        grid.$connector.debounceRootRequest = tryCatchWrapper(function (page) {
+          const delay = grid._hasData ? rootRequestDelay : 0;
+
+          rootRequestDebouncer = Debouncer.debounce(rootRequestDebouncer, timeOut.after(delay), () => {
+            grid.$connector.fetchPage(
+              (firstIndex, size) => grid.$server.setRequestedRange(firstIndex, size),
+              page,
+              root
+            );
+          });
+        });
+
         grid.$connector.flushParentRequests = tryCatchWrapper(function () {
-          let pendingFetches = parentRequestQueue.splice(0, parentRequestBatchMaxSize);
+          const pendingFetches = [];
+
+          parentRequestQueue.splice(0, parentRequestBatchMaxSize).forEach(({ parentKey, page }) => {
+            grid.$connector.fetchPage(
+              (firstIndex, size) => pendingFetches.push({ parentKey, firstIndex, size }),
+              page,
+              parentKey
+            );
+          });
 
           if (pendingFetches.length) {
             grid.$server.setParentRequestedRanges(pendingFetches);
-            return true;
           }
-          return false;
         });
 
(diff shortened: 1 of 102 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
```diff
@@ -215,24 +215,40 @@ import { GridFlowSelectionColumn } from "./vaadin-grid-flow-selection-column.js"
...
-          // add request in queue
-          parentRequestQueue.push({
-            firstIndex: firstIndex,
-            size: size,
-            parentKey: parentKey
-          });
-
+        grid.$connector.debounceParentRequest = tryCatchWrapper(function (parentKey, page) {
+          // Remove any pending requests for the same parentKey.
+          parentRequestQueue = parentRequestQueue.filter((request) => request.parentKey !== parentKey);
```
vursen, line 248: **note:** It now seems that a more appropriate data structure to store scheduled requests would be Map rather than Array. However, I didn't change it to minimize the impact of the changes, considering that the PR will likely need to be backported to early versions.

DiegoCardoso (approved):
LGTM

PR #5927, opened 2024-01-09
fix: wraps client setEditModeRenderer call in a runWhenAttached callback

### LINKED ISSUE #5925: GridPro : Cannot set properties of null (setting 'editModeRenderer') after reattach/detach
### Description

The grid with one column is initially added to the view.
After clicking 'Remove' and then'AddAndRemove' the following error occures in the Console:
image

### Expected outcome

I would expect the Console to not show any errors.
For me it seems to have anything to do with Re-Attaching and Detaching within the same request-response-cycle.

Using the same scenario for a normal grid with DropHandler, we also detected still a problem after (fixed #1434}, but only with Vaadin 23.3.30.
There I could work around the error by calling ComponentUtil.setData("_detachListenerForDropTarget", null);

Don't know , if there are components left, that throw errors in this admittely rare scenario from a vaadin view, but for us this scenario happens often, because we have this when returning to the calling view from a sub call and then branch to another view in the same request/response-cycle.

```
package org.vaadin.example;

import com.vaadin.flow.component.gridpro.GridPro;
//import com.vaadin.flow.component.UI;
//import com.vaadin.flow.component.button.Button;
import com.vaadin.flow.component.textfield.TextField;
import com.vaadin.flow.router.PreserveOnRefresh;

@Route
@PreserveOnRefresh
	
	private GridPro<Object> personGrid;

	public MainView() {

		personGrid = new GridPro<>();
        TextField textField = new TextField();
        personGrid.addEditColumn(item -> {
        	String name = item.toString();
        	return name;
        }).custom(textField, (item, value) -> {
        	System.out.println("Updating value for name");
        });


		add(new Button("Remove", e -> remove(personGrid)));
		add(new Button("AddAndRemove", e -> {
			add(personGrid);
			remove(personGrid);
		}));
		add(personGrid);
	}
	
}
```

### Steps to reproduce

1. Click 'Remove
(2. Clear JavaScript-Console)
3. Click'AddAndRemove

### Environment

Vaadin version(s): 23.3.30, 24.3.2
OS: Windows

### Browsers

Issue is not browser related

DiegoCardoso:
## Description

Wrapps the `setEditModeRenderer` method in a `runWhenAttached` callback. 

Both calls to `setEditModeRenderer` private method in the `EditColumnConfigurator` class due to two factors:
- the attach listener is only called when the whole GridPro is attached to the client, because the listener is not called when the column is created to an already attached GridPro
- because of the previous issue, there's still the need to keep a call to `setEditModeRenderer` inside the `custom` set it when the column is added afterward

To avoid the issue of two calls for the same column/component on the initialization, I implemented the same flags used in the `AbstractColumn` class for the `scheduleHeaderRendering`/`scheduleFooterRendering` methods.
Fixes #5925

## Type of change

- [X] Bugfix
- [ ] Feature

### DIFF
--- a/vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/java/com/vaadin/flow/component/gridpro/EditColumnConfigurator.java
+++ b/vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/java/com/vaadin/flow/component/gridpro/EditColumnConfigurator.java
@@ -37,6 +37,8 @@ public class EditColumnConfigurator<T> implements Serializable {
     private final EditColumn<T> column;
     private Registration attachRegistration;
 
+    private boolean editModeRendererRequested = false;
+
     /**
      * Creates a new configurator for the given column.
      *
@@ -150,21 +152,37 @@ public class EditColumnConfigurator<T> implements Serializable {
             attachRegistration.remove();
             attachRegistration = null;
         }
+        // Need to call on attach to make sure that the edit mode renderer is
+        // set in case the GridPro is detached and attached again
         attachRegistration = column.getElement()
                 .addAttachListener(e -> setEditModeRenderer(component));
 
-        column.getElement().getNode()
-                .runWhenAttached(ui -> ui.beforeClientResponse(column,
-                        context -> setEditModeRenderer(component)));
+        // Calling setEditModeRenderer here in case the GridPro is already
+        // attached and the column is added later
+        // This is needed because in this case the attach listener is not called
+        setEditModeRenderer(component);
+
         return configureColumn(valueProvider, (item, ignore) -> itemUpdater
                 .accept(item, component.getValue()), EditorType.CUSTOM,
                 component);
     }
 
     private <V> void setEditModeRenderer(HasValueAndElement<?, V> component) {
-        UI.getCurrent().getPage().executeJs(
-                "window.Vaadin.Flow.gridProConnector.setEditModeRenderer($0, $1)",
-                column.getElement(), component.getElement());
+        if (editModeRendererRequested) {
+            return;
+        }
+        editModeRendererRequested = true;
+        column.getElement().getNode().runWhenAttached(ui -> {
+            ui.beforeClientResponse(column, context -> {
+                if (!editModeRendererRequested) {
+                    return;
+                }
+                ui.getPage().executeJs(
+                        "window.Vaadin.Flow.gridProConnector.setEditModeRenderer($0, $1)",
+                        column.getElement(), component.getElement());
+                editModeRendererRequested = false;
+            });
+        });
     }
 
     /**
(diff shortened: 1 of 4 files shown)


vursen:
The solution seems to work. I wonder though why there is no same issue for `setHeaderRenderer` / `setFooterRenderer` considering that they are also assigned on attach.

vursen:
> Although, I wonder why there is no same issue for setHeaderRenderer / setFooterRenderer considering that they are also run in addAttachListener and runWhenAttached:

It seems they don't have this issue because they always wrap `renderHeader()` in `runWhenAttached` even when it's called from `addAttachListener`.

ugur-vaadin:
> I have a theory that they don't have this issue because they always wrap `$connector.setHeaderRenderer` in `runWhenAttached` even when it's called from `addAttachListener`. Would it make sense to adopt the same approach for `$connector.setEditModeRenderer`? This could completely prevent its invocation in the issue's scenario.

I can confirm that it works with `runWhenAttached` wrapper without the change to the connector.

vursen:
Another finding not directly related to the issue. The `setEditModeRenderer` method is currently called twice for each column during initialization:

image

DiegoCardoso:
I ended up being able to address this issue on the server level by wrapping the `setEditModeRenderer` in a `runWhenAttached` callback. In the end, I still needed to keep both calls to `setEditModeRenderer` private method in the `EditColumnConfigurator` class due to two factors:
- the attach listener is only called when the whole GridPro is attached to the client, because the listener is not called when the column is created to an already attached GridPro
- because of the previous issue, there's still the need to keep a call to `setEditModeRenderer` inside the `custom` set it when the column is added afterward

To avoid the issue of two calls for the same column/component on the initialization, I implemented the same flags used in the `AbstractColumn` class for the `scheduleHeaderRendering`/`scheduleFooterRendering` methods.

vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/java/com/vaadin/flow/component/gridpro/EditColumnConfigurator.java
```diff
@@ -150,21 +152,37 @@ public <V> Column<T> custom(HasValueAndElement<?, V> component,
...
-                column.getElement(), component.getElement());
+        if (editModeRendererRequested) {
+            return;
+        }
+        editModeRendererRequested = true;
+        column.getElement().getNode().runWhenAttached(ui -> {
+            ui.beforeClientResponse(column, context -> {
+                if (!editModeRendererRequested) {
+                    return;
+                }
```
vursen, line 179: **nit:** Is this check necessary? Because `runWhenAttached` is the only place where `editModeRendererRequested` can be set to `false`.

vursen (approved):

vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/java/com/vaadin/flow/component/gridpro/EditColumnConfigurator.java
DiegoCardoso: This is the same case in the `scheduleHeaderRendering`. My understanding is to safeguard any unlike case of two callbacks scheduled even with the early exit at the beginning of the method.

vursen (approved):

PR #5930, opened 2024-01-11
fix: do not render editor for invisible rows

### LINKED ISSUE #5898: Grid's Editor row may render empty when opening the Editor after column order is changed and filter is applied
### Description

This is a corner case that happens only with rather complex setups. I have attached source code that demonstrates the problem.

image

Preliminary analysis of the problem
- When Grid's editor is closed, the __editing property of the cell content stays in true value, this closing editor should clear it
- This can lead to situation that there are multiple rows where __editing property is true
- When filter is applied, some rows are hidden, some of the hidden rows may have __editing property true
- When Editor is opened it may be rendered on the row, that is actually hidden, while it should not, hence the visible row has empty rendering

### Expected outcome

Editor is rendered correctly when opened.

[grid-with-filters.zip](https://github.com/vaadin/flow-components/files/13827985/grid-with-filters.zip)

### Steps to reproduce

Use the provided code, do the following

- Click "Close editor" checkbox to enable editor closing while column order is changed
- Scroll down the Grid and click a row to open an Editor, e.g. "Lydia House"
- Type "Ho" in filter, this will match more than one row
- Then click "Change Column Order"
- Editor row is now empty like in the screenshot (some times you need to click the button more than once)
- This can be replicated with other items too, the important factors are: 1. you need to scroll to next page, 2. you need to use filter that matches more than one item, but less items visible than max row count (like in the example on the screen). Specifically if only one row matches filter, the problem does not occur.

### Environment

Vaadin version(s): Vaadin 24.3.0 

### Browsers

_No response_

ugur-vaadin:
## Description

WIP

Fixes #5898

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridEditorFilteringPage.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridEditorFilteringPage.java
@@ -1,0 +1,190 @@
+/*
+ * Copyright 2000-2024 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


ugur-vaadin:
Closing in favour of [7070](https://github.com/vaadin/web-components/pull/7070).

PR #5962, opened 2024-01-16
feat: add target API to SideNavItem

### LINKED ISSUE #5091: SideNavItem should have API for setting target
### Describe your motivation

Currently, if I add an external link to SideNav as a SideNavItem, it opens in the same tab. There is  no API to tell it to open in a new tab instead (`<a ... target="_blank">..</a>`).

A way to provide target type or indicate that it should be opened in a new tab through SideNavItem API. 

Current workaround seems to be:
```
navItem.getElement().executeJs("requestAnimationFrame(() => this.shadowRoot.querySelector('a').target = '_blank')")
```

### Additional context

_No response_

DiegoCardoso:
## Description

Add two new API for setting/getting the `target` property of a `SideNavItem`, whose value is used to set the `target` attribute in the internal anchor element inside the web-component:
- `setTarget(String)`/`getTarget()` - which receives any string as argument
- `setOpenInNewBrowserTab(boolean)`/`isOpenInNewBrowserTab()`- a convenience method that internally calls `setTarget` with `_blank` as value when called with `true`.

Depends on vaadin/web-components#7088
Fixes #5091
AC https://github.com/vaadin/platform/issues/4850

## Type of change

- [ ] Bugfix
- [X] Feature

### DIFF
--- a/vaadin-side-nav-flow-parent/vaadin-side-nav-flow/src/main/java/com/vaadin/flow/component/sidenav/SideNavItem.java
+++ b/vaadin-side-nav-flow-parent/vaadin-side-nav-flow/src/main/java/com/vaadin/flow/component/sidenav/SideNavItem.java
@@ -357,6 +357,66 @@ public class SideNavItem extends SideNavItemContainer
         }
     }
 
+    /**
+     * Gets the target of this item.
+     *
+     * @return the target of this item
+     */
+    public String getTarget() {
+        return getElement().getProperty("target");
+    }
+
+    /**
+     * Where to display the linked URL, as the name for a browsing context.
+     * <p>
+     * The following keywords have special meanings for where to load the URL:
+     * <ul>
+     * <li><code>_self</code>: the current browsing context. (Default)</li>
+     * <li><code>_blank</code>: usually a new tab, but users can configure
+     * browsers to open a new window instead.</li>
+     * <li><code>_parent</code>: the parent browsing context of the current one.
+     * If no parent, behaves as <code>_self</code>.</li>
+     * <li><code>_top</code>: the topmost browsing context (the "highest"
+     * context that’s an ancestor of the current one). If no ancestors, behaves
+     * as <code>_self</code>.</li>
+     * </ul>
+     * </p>
+     *
+     * @param target
+     *            the target of this item
+     */
+    public void setTarget(String target) {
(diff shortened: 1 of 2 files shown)


vaadin-side-nav-flow-parent/vaadin-side-nav-flow/src/main/java/com/vaadin/flow/component/sidenav/SideNavItem.java
```diff
@@ -357,6 +357,67 @@ public void setPathAliases(Set<String> pathAliases) {
...
+     * <p>
+     * This is a convenience method for setting the target to
+     * <code>_blank</code>. See {@link #setTarget(String)} for more information.
+     * </p>
+     *
+     * @param openInNewBrowserTab
+     *            true if the target URL should be opened in a new browser tab,
+     *            false otherwise
+     */
+    public void setOpenInNewBrowserTab(boolean openInNewBrowserTab) {
```
sissbruecker, line 407: `Browser` is maybe a bit redundant? `setOpenInNewTab` would be enough IMO.

sissbruecker (approved):

vaadin-side-nav-flow-parent/vaadin-side-nav-flow/src/main/java/com/vaadin/flow/component/sidenav/SideNavItem.java
DiegoCardoso: The name has been taken from the AC ticket: https://github.com/vaadin/platform/issues/4850
(ping @rolfsmeds)

rolfsmeds: `setOpenInNewBrowsingContext` would be the most technically correct terminology, I guess, but _very_ unintuitive to most developers.

DiegoCardoso: `Window` seems more misleading to me since opening in a new tab is the default behavior in most browsers.

I agree that `Context` is unintuitive.

PR #5979, opened 2024-01-23
fix: Remove connector only if it still exists

### LINKED ISSUE #5283: Error logged when adding back a previously added component
### Description of the bug

The error below is logged in an application that starts with a single component with a `ContextMenu` and the following is performed:

1. The application to another component by:
  a. Adding the next component
  b. Calls `ContextMenu.setTarget` to change the target to the next component
  c. Removes the first component
2. The application then back to the first component by repeating the above

> The error has occurred in the JS code: '$0, return 
> $0.$contextMenuTargetConnector.removeConnector()' TypeError: $0.$contextMenuTargetConnector is undefined
> Uncaught TypeError: Cannot read properties of undefined (reading 'removeConnector')

This is the stack trace from the browser:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'removeConnector')
    at Object.eval (eval at Gu (FlowClient.js:1090:101), <anonymous>:3:39)
    at Gu (FlowClient.js:1090:166)
    at Fu (FlowClient.js:1039:277)
    at Du (FlowClient.js:643:57)
    at hs (FlowClient.js:550:18)
    at Es.Fs [as hb] (FlowClient.js:1100:21583)
    at yC (FlowClient.js:1008:221)
    at ks (FlowClient.js:1096:303)
    at As.Bs [as I] (FlowClient.js:1100:21349)
    at rl (FlowClient.js:459:29)
    at cs (FlowClient.js:1098:1159)
    at ds (FlowClient.js:1079:403)
    at ru.tu [as rb] (FlowClient.js:1100:24815)
    at dD.eD [as O] (FlowClient.js:1100:38308)
    at XMLHttpRequest.<anonymous> (FlowClient.js:649:63)
    at Eb (FlowClient.js:482:29)
    at Hb (FlowClient.js:959:56)
    at XMLHttpRequest.<anonymous> (FlowClient.js:676:46)
```

### Expected behavior

No error to be logged.

Here is a sample code to reproduce this issue:

```java
import com.vaadin.flow.component.contextmenu.ContextMenu;
import com.vaadin.flow.component.html.Label;
import java.util.Objects;

@PageTitle("Main")
@Route(value = "")

  public MainView() {
    Label label1 = new Label("Label 1");
    Label label2 = new Label("Label 2");

    ContextMenu menu = new ContextMenu();
    menu.addItem("Context menu");
    menu.setTarget(label1);

    Button button =
        new Button(
            "Swap",
            event -> {
              if (getChildren().anyMatch(c -> Objects.equals(label1, c))) {
                add(label2);
                menu.setTarget(label2);
                remove(label1);
              } else {
                add(label1);
                menu.setTarget(label1);
                remove(label2);
              }
            });

    add(button, label1);
  }
}
```

The app contains a button that swaps a label. The labels, themselves, use the same `ContextMenu`. 

To see the issue, click on the button twice.

### Versions

Vaadin: 23.3.16
Flow: 23.3.15
Java: Eclipse Adoptium 17.0.4.1
OS: amd64 Windows 10 10.0
Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36 Edg/114.0.1823.79

TatuLund:
fixes #5283

### DIFF
--- a/vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/java/com/vaadin/flow/component/contextmenu/ContextMenuBase.java
+++ b/vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/java/com/vaadin/flow/component/contextmenu/ContextMenuBase.java
@@ -120,8 +120,8 @@ public abstract class ContextMenuBase<C extends ContextMenuBase<C, I, S>, I exte
         if (getTarget() != null) {
             targetBeforeOpenRegistration.remove();
             targetAttachRegistration.remove();
-            getTarget().getElement().callJsFunction(
-                    "$contextMenuTargetConnector.removeConnector");
+            getTarget().getElement().executeJs(
+                    "if (this.$contextMenuTargetConnector) { this.$contextMenuTargetConnector.removeConnector() }");
             if (isTargetJsPending()) {
                 targetJsRegistration.cancelExecution();
                 targetJsRegistration = null;
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/contextmenu/GridContextMenuTest.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/contextmenu/GridContextMenuTest.java
@@ -119,7 +119,7 @@ public class GridContextMenuTest {
         gridContextMenu.setTarget(null);
 
         Mockito.verify(registration).remove();
-        Mockito.verify(element)
-                .callJsFunction("$contextMenuTargetConnector.removeConnector");
+        Mockito.verify(element).executeJs(
+                "if (this.$contextMenuTargetConnector) { this.$contextMenuTargetConnector.removeConnector() }");            
     }
 }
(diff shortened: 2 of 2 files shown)


vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/java/com/vaadin/flow/component/contextmenu/ContextMenuBase.java
```diff
@@ -120,8 +120,8 @@ public void setTarget(Component target) {
         if (getTarget() != null) {
             targetBeforeOpenRegistration.remove();
             targetAttachRegistration.remove();
-            getTarget().getElement().callJsFunction(
-                    "$contextMenuTargetConnector.removeConnector");
+            getTarget().getElement().executeJs(
+                    "if (this.$contextMenuTargetConnector) { this.$contextMenuTargetConnector.removeConnector() }");
```
sissbruecker, line 124: It seems the problem is that the JS call is only actually run after a label has been added to the page again, at which point the connector instance doesn't exist anymore because the DOM element is a new instance.

For example:
- Change target to label 2
- Schedules JS to remove connector for label 1
- Label 1 is removed
- JS doesn't run, because element isn't attached, but JS call also isn't discarded
- ...
- Add label 1 again
- JS call runs

While the fix here would be OK, I wonder if there isn't a better way to handle this? Should Flow discard JS calls when an element is removed, since those are not part of the element's state? Is there already a way to do so without making framework users write complex code with checking for detach etc.?

sissbruecker: Let's go ahead with the proposed fix for now.

sissbruecker (approved):

PR #5989, opened 2024-01-25
fix: re-render components on attach

### LINKED ISSUE #5732: Grid items disappear after grid is detached and attached in the same round-trip
### Description

The elements in `Grid` can disappear in case the component is detached and re-attached in the same round-trip. One real case example where this happens is if there's a `Grid` element already inside a `SplitLayout` and another element is added to the other slot, since internally, `SplitLayout` removes all elements and adds them back to the right slots.

Below you can see the buggy behavior:

https://github.com/vaadin/flow-components/assets/262432/2741c9e4-0016-4ca6-9992-6f6e016902f4

I was able to reproduce the example from V24.0. In V23.3 the error is not present.

### Expected outcome

Grid items do not disappear in such cases.

```java
var grid = new Grid<String>();
grid.addComponentColumn(Span::new).setHeader("items");
grid.setItems("foo", "bar", "baz");

var toggleGridButton = new Button("detach/attach grid", e -> { 
	remove(grid);
	add(grid);
});

var detachGridButton = new Button("detach grid", e -> {
	remove(grid);
});

var attachGridButton = new Button("attach grid", e -> {
	add(grid);
});

add(newtoggleGridButton, detachGridButton, attachGridButton, grid);
```

### Steps to reproduce

Example:
1. Add the snippet example to a view
2. Click on the "detach/attach grid" button
3. See items disappear
4. However, if detach and re-attach are done in separate round-trips, the items as shown as expected.

### Environment

Vaadin version(s): V24.0+
OS: N/A

### Browsers

Chrome, Firefox, Safari, Edge

sissbruecker:
When a grid component is detached, its data communicator will clear all data generators, which also removes all components rendered by component renderers. If the grid is then reattached in the same roundtrip, the components are not rendered again, causing them to disappear on the client-side.

This fixes the issue by resetting the data communicator on attach, which forces items to be regenerated, thus triggering a new render of the component renderers. This causes additional roundtrips in the detach / reattach scenario, but doesn't change how the component behaves when detaching / attaching in separate roundtrips, or when attaching the component for the first time.

Fixes #5732

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
@@ -1677,10 +1677,6 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
         updateMultiSortPriority(defaultMultiSortPriority);
 
         initSelectionPreservationHandler();
-
-        addAttachListener(e -> {
-            dataCommunicator.reset();
-        });
     }
 
     private void generateUniqueKeyData(T item, JsonObject jsonObject) {
@@ -3567,12 +3563,16 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
 
     @Override
     protected void onAttach(AttachEvent attachEvent) {
-        super.onAttach(attachEvent);
         updateClientSideSorterIndicators(sortOrder);
         updateSelectionModeOnClient();
         if (getDataProvider() != null) {
             handleDataProviderChange(getDataProvider());
         }
+        // When the component is detached and reattached in the same roundtrip,
+        // data communicator will clear all data generators, which will also
+        // remove all components rendered by component renderers. Thus reset the
+        // data communicator to re-render components.
+        dataCommunicator.reset();
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -3563,12 +3563,16 @@ public void sort(List<GridSortOrder<T>> order) {
...
         updateClientSideSorterIndicators(sortOrder);
         updateSelectionModeOnClient();
         if (getDataProvider() != null) {
             handleDataProviderChange(getDataProvider());
         }
+        // When the component is detached and reattached in the same roundtrip,
+        // data communicator will clear all data generators, which will also
+        // remove all components rendered by component renderers. Thus reset the
+        // data communicator to re-render components.
+        dataCommunicator.reset();
```
tomivirkki, line 3575: This seems to fix the issue but maybe causes something else to break (the CI build failed).

Discovered another regression with detach-reattach while testing the change: https://github.com/vaadin/flow-components/issues/5994

tomivirkki (approved):

ugur-vaadin (approved):

PR #6060, opened 2024-02-19
fix: sync default header row for selection row

### LINKED ISSUE #1054: appendHeaderRow is pushing sort and resize to the new header row.
https://vaadin.com/components/vaadin-grid/java-examples/filtering

base on above link, when create filter no  grid, the appendHeaderRow is using.
in such case, the expected behave is that resize ,sort and multi-select checkbox will stay on the read header and not push down to the filter row.

This behave not make sense to me. when we add headers that are not filter, it will usually aggregate data and then will use prepandHeaderRow and not appendHeaderRow

any hot fix solution will be appriciated
Guy

ugur-vaadin:
## Description

The selection column in a Grid is not a part of the column layer structure, therefore cannot track the default header. When the index of the default row changes, the sorting indicators are updated, but there is nothing that handles this change for the select all checkbox. This PR introduces a synchronizing mechanism for the Flow selection column. When a header row is appended, prepended, or removed, the default header row index is sent to the client. There is a proposed custom selection column header cell provider, which does not exist by default. When it exists, the decision to set a cell as a header cell while updating the rows is delegated to the custom provider for that column. This way, the checkbox is only rendered in the header cell that corresponds to the default header row.

Note: The proposed custom selection column header cell provider should be in the web component. Currently, for discussion purposes, the relevant code is copied from the web component and modified in the connector.

Fixes #1054

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
@@ -524,6 +524,7 @@ import { GridFlowSelectionColumn } from "./vaadin-grid-flow-selection-column.js"
                 // Header & footer
                 const tagName = section === 'header' ? 'th' : 'td';
 
+                // The change here is for discussion and should be applied to the web component
                 let isHeaderOrFooterCell;
                 if (section === 'header' && column._customSelectionColumnHeaderHandler) {
                   const headerRowIndex = Array.from(grid.$.header.children).indexOf(row);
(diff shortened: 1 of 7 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
```diff
@@ -455,6 +456,120 @@ import { GridFlowSelectionColumn } from "./vaadin-grid-flow-selection-column.js"
           Grid.prototype.collapseItem.call(grid, item);
         });
 
+        grid._updateRow = tryCatchWrapper(function (row, columns, section = 'body', isColumnRow = false, noNotify = false) {
```
ugur-vaadin, line 459: This part is a copy from the web component for draft PR discussion purposes.

ugur-vaadin:
Closing as the it was decided that the solution for the issue needs more research.

PR #6068, opened 2024-02-20
feat: allow using live directive in LitRenderer

DiegoCardoso:
## Description

Inject the `live` directive into the closure where the render is executed.

It's beneficial for cases where components can have their state changed by users to reset their value when the render is called.

### DIFF
--- a/vaadin-renderer-flow-parent/vaadin-renderer-flow-integration-tests/src/main/java/com/vaadin/flow/data/renderer/tests/LitRendererPage.java
+++ b/vaadin-renderer-flow-parent/vaadin-renderer-flow-integration-tests/src/main/java/com/vaadin/flow/data/renderer/tests/LitRendererPage.java
@@ -50,6 +50,14 @@ public class LitRendererPage extends Div {
         setSimpleLitRendererButton.setId("setSimpleLitRendererButton");
         add(setSimpleLitRendererButton);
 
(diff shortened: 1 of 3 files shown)


vaadin-renderer-flow-parent/vaadin-renderer-flow-integration-tests/src/main/java/com/vaadin/flow/data/renderer/tests/LitRendererPage.java
```diff
@@ -87,6 +87,13 @@ public LitRendererPage() {
         toggleAttachedButton.setId("toggleAttachedButton");
         add(toggleAttachedButton);
 
+        NativeButton setCheckboxRenderer = new NativeButton("Set renderer with checkbox", e -> {
```
tomivirkki, line 90: Nit: Would it make sense to place this button after the "Set simple LitRenderer" button instead?

tomivirkki (approved):

vaadin-renderer-flow-parent/vaadin-renderer-flow-integration-tests/src/main/java/com/vaadin/flow/data/renderer/tests/LitRendererPage.java
DiegoCardoso: Done.

PR #6100, opened 2024-03-06
fix: enforce current year if date format does not contain year

### LINKED ISSUE #6099: DatePicker still parses short dates incorrectly in 24.3.6 if reference date is set
### Describe the bug

With reference date set to today-40 years, "1" is expanded to "01.03.1984"

### Expected-behavior

When year is not included at all, it should always be this year.
Only when 2-digit year is included should it use the reference date. 
Whatever the reference date is, it should only affect the century. Not the decade.
Ie "1" should expand to either "01.03.1924" or "01.03.2024"

### Reproduction

```var datePicker = new DatePicker();
datePicker.setI18n
    ( new DatePickerI18n()
        ...
        .setDateFormats("dd.MM.yyyy", "dd", "ddMM", "ddMMyy", "ddMMyyyy", "yyyy-MM-dd", "MM/dd/yyyy")
        .setReferenceDate(LocalDate.now().minusYears(40)) 
    );

add(datePicker);
```

### System Info

Vaadin 24.3.6

web-padawan:
## Description

Fixes #6099

Follow-up to #6088 which was an incomplete fix.

This change enforces the behavior as described in the issue:

> When year is not included at all, it should always be this year.
> Only when 2-digit year is included should it use the reference date.

As `date-fns` parse function requires a reference date, the current date is used in case format like `ddMM` is passed.

## Type of change

- Bugfix

### DIFF
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
@@ -65,6 +65,10 @@ import { extractDateParts, parseDate as _parseDate } from '@vaadin/date-picker/s
             return undefined;
           }
 
+          function isFormatWithoutYears(format) {
+            return !format.includes('y') && !format.includes('Y');
+          }
+
           function isShortYearFormat(format) {
             // Format is long if it includes a four-digit year.
             return !format.includes('yyyy') && !format.includes('YYYY');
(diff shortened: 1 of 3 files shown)


vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
```diff
@@ -82,20 +86,24 @@ import { extractDateParts, parseDate as _parseDate } from '@vaadin/date-picker/s
 
             // Update century if the last parsed date is the same except the century.
             if (datepicker.$connector._lastParseStatus === 'successful') {
-              if (datepicker.$connector._lastParsedDate.day === date.getDate() &&
+              if (
+                datepicker.$connector._lastParsedDate.day === date.getDate() &&
```
web-padawan, line 90: Note: some changes are caused by Prettier. I decided to commit them as they are quite small.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js
```diff
@@ -108,7 +116,9 @@ import { extractDateParts, parseDate as _parseDate } from '@vaadin/date-picker/s
           }
 
           function doParseDate(dateString, format, referenceDate) {
-            const date = dateFnsParse(dateString, format, referenceDate);
+            // When format does not contain a year, then current year should be used.
+            const refDate = isFormatWithoutYears(format) ? new Date() : referenceDate;
+            const date = dateFnsParse(dateString, format, refDate);
```
web-padawan, line 121: This is the actual fix. We could also try `date.setFullYear(new Date().getFullYear())` as an alternative.

ugur-vaadin (approved):
LGTM.

vursen (approved):

Issue #6098, opened 2024-03-05
Login custom area text field XHR event interferes with login process 


apryamostanov:
### Description of the bug

In Vaadin 24.2, Custom Form area was added into Login component (https://vaadin.com/docs/latest/components/login#custom-form-area).
If I put text area inside it (e.g. OTP field), when I press Login button - focus is changed and sync event is fired resulting in XHR. At same time Login form submit is processed. XHR interferes with login process and cancels resource fetch from service worker.
This is not visible locally/on fast connections as XHR is processed fast. But on slower connections it results in failure to login.
You can simulate this by enabling throttling in Chrome.

### Expected behavior

Login form with TextField in Custom Area should work consistently and there should not be cancelled requests.

### Minimal reproducible example

TBD

### Versions

- Vaadin / Flow version: 24.2
- Java version: 17
- OS version: Mac OS 14.1.2
- Browser version (if applicable): Chrome 122.0.6261.94 (Official Build) (arm64)
- Application Server (if applicable): Spring Boot
- IDE (if applicable): N/A

mcollovati:
I noticed the same issue yesterday by testing an application created with Vaadin CLI (`npx @vaadin/cli init login-form --auth`).
To reproduce, start the application and try to log in with the wrong credentials: you'll see two requests (POST + UIDL) before being redirected to login view with error parameter.

![image](https://github.com/vaadin/flow-components/assets/4648894/fcfe8237-c426-416f-a3bb-e0f73aaf84c0)



This is the UIDL request sent in parallel with the login POST request

```
{
    "csrfToken": "fdfa01a3-7cf3-4bcf-b8d9-a1f906bcb6c0",
    "rpc": [
        {
            "type": "mSync",
            "node": 4,
            "feature": 1,
            "property": "disabled",
            "value": true
        },
        {
            "type": "event",
            "node": 4,
            "event": "login",
            "data": {
                "event.detail.username": "xxxxx",
                "event.detail.password": "xxxxx"
            }
        }
    ],
    "syncId": 2,
    "clientId": 2
}
```

web-padawan:
> This is the UIDL request sent in parallel with the login POST request

[...]

PR #6116, opened 2024-03-11
docs: add note about fields in LoginOverlay custom form area

web-padawan:
## Description

Related to #6098

Updated JavaDoc of `getCustomFormArea()` as suggested by https://github.com/vaadin/flow-components/issues/6098#issuecomment-1988556787.

I will also submit a PR to the docs project, as the original issue isn't obvious and deserves a note.

## Type of change

- Documentation

### DIFF
--- a/vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/LoginOverlay.java
+++ b/vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/LoginOverlay.java
@@ -242,6 +242,11 @@ public class LoginOverlay extends AbstractLogin implements HasStyle {
      * overlay custom form area. This area is displayed only if there's at least
      * one component added with {@link LoginOverlayContent#add(Component...)}.
      *
+     * Fields that are part of custom form area are not automatically submitted
+     * as part of the {@link LoginForm.LoginEvent}, and are not supported when
+     * setting {@code action} as their values will not be part of the login
+     * request.
+     *
      * @since 24.2
      * @return the custom form area object
      */


vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/LoginOverlay.java
```diff
@@ -242,6 +242,10 @@ public String getDescription() {
      * overlay custom form area. This area is displayed only if there's at least
      * one component added with {@link LoginOverlayContent#add(Component...)}.
      *
+     * Fields that are part of custom form area are not automatically submitted
+     * as part of the {@link LoginForm.LoginEvent} and their values need to be
+     * synced using separate events, which interferes with the POST request.
```
sissbruecker, line 247: This mixes several issues which makes understanding this confusing:
- Custom fields are not part of the `LoginEvent`. We can definitely mention that here, and that you need to access their values separately from the event.
- The POST request itself is not related to the `LoginEvent`. When you set an `action` the form just does a POST to whatever URL you set there (so some Spring security handler probably), but you don't handle the login event itself in that case.
- From testing this I can at least see that I get JS errors if Flow does a sync request for the custom field while the browser does the login request in parallel. The other thing is that the custom field is not part of the form, so whatever handles the login request can not use it. So from a practical perspective, the second thing to mention is that custom fields are straight up not supported when setting an `action` on the LoginOverlay, as their values will not be part of the login request.

web-padawan: Thank you for the note. Updated the JavaDoc to also include the following:

>  not supported when setting an action on the LoginOverlay, as their values will not be part of the login request.

web-padawan: > From testing this I can at least see that I get JS errors if Flow does a sync request for the custom field while the browser does the login request in parallel. 

Is that the same as reported in https://github.com/vaadin/flow-components/issues/6098#issuecomment-1980190986? I think this is the actual issue that needs to be fixed.

sissbruecker: > Is that the same as reported in https://github.com/vaadin/flow-components/issues/6098#issuecomment-1980190986? I think this is the actual issue that needs to be fixed.

I'm not sure, the issue description is quite vague in what "interferes with the POST request" means, and whether the login effectively works or not. For me it was working, there was just an error to go with it.

sissbruecker (approved):

PR #6125, opened 2024-03-14
fix: prevent sorters change when updating multiple headers

### LINKED ISSUE #6122: setHeader on grid columns may trigger sort events
### Description

Calling setHeader on grid columns can trigger sort events. This seems to happen when both of these are true:

- setHeader is called on more than one column
- the grid has a non-empty sort order

This behavior is new in 24.3.0+ and doesn't occur in previous versions I've tested (24.2.11 and below). The number of sort events seems to depend on the number of columns/times setHeader is called.

### Expected outcome

I wouldn't expect setHeader to trigger any sort events since no sorting settings have been changed.

Demo: [grid-sort-event-v2.zip](https://github.com/vaadin/flow-components/files/14592901/grid-sort-event-v2.zip)
[Screen recording](https://github.com/vaadin/flow-components/assets/33359250/48f6d410-d1da-4602-9377-8d482c457d35)

### Steps to reproduce

1. Click the "Change Header Text" button
2. Text below each grid shows how many sort events have been triggered for that grid. Notice how the first grid (with 1 column) does not have any sort events after the first one (which sets the grid's initial sort order). However, the number of sort events for grids 2-4 increments by 2, 4, and 6 respectively, each time the button is pressed.

### Environment

Vaadin version(s): 24.3.0-24.3.7
OS: macOS 13.6.4

Tested in Chrome 122.0.6261.112, Safari 16.6, and Firefox 123.0

### Browsers

Chrome, Firefox, Safari

web-padawan:
## Description

Fixes #6122

When calling `setHeader()` on the column, Grid updates sorter indicators (as implemented in #1841). Currently it is done by re-setting `direction` to `null` in the connector and then setting directions in the right order.

The logic was changed to reset `direction` for all sorters in #5822, which introduced the regression: when calling `setHeader()` for more than one column in one request, it caused grid to incorrectly apply sorters change.

Updated the connector to add a check for this case, and added new ITs to verify that sort listeners are not invoked.

## Type of change

- Bugfix

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
@@ -375,8 +375,11 @@ import { GridFlowSelectionColumn } from "./vaadin-grid-flow-selection-column.js"
                   }
                 });
 
+                // Only reset direction for sorters that no longer apply.
                 sorters.forEach((sorter) => {
-                  sorter.direction = null;
+                  if (!directions.some(({ column }) => column === sorter.getAttribute('path'))) {
+                    sorter.direction = null;
+                  }
                 });
 
                 // Apply directions in correct order, depending on configured multi-sort priority.
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
@@ -375,12 +375,15 @@ import { GridFlowSelectionColumn } from "./vaadin-grid-flow-selection-column.js"
                   }
                 });
 
-                // Only reset direction for sorters that no longer apply.
-                sorters.forEach((sorter) => {
-                  if (!directions.some(({ column }) => column === sorter.getAttribute('path'))) {
+                // When changing header components, this method can be invoked multiple times with
+                // the same directions. In this case, we should avoid re-setting sorters state.
+                if (JSON.stringify(this._previousDirections) !== JSON.stringify(directions)) {
+                  sorters.forEach((sorter) => {
                     sorter.direction = null;
-                  }
-                });
+                  });
+                }
+
+                this._previousDirections = [...directions];
 
                 // Apply directions in correct order, depending on configured multi-sort priority.
                 // For the default "prepend" mode, directions need to be applied in reverse, in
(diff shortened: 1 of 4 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
sissbruecker, line 1: As an alternative to trying to deal with multiple calls on the client side, how about updating the server-side logic so that:
- it tracks whether `updateClientSideSorterIndicators` has already scheduled a client-side call
- if so, cancel that
- schedule a new client-side call

So the server would only ever schedule one client-side sorter update for each roundtrip. IMO the logic would be more appropriate there. Might also be testable with a unit test then, if we can check how many JS calls are scheduled by the component.

sissbruecker: There's a similar approach in context menu:
https://github.com/vaadin/flow-components/blob/dedb257731a78f7874ede961a05c9df639040217/vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/java/com/vaadin/flow/component/contextmenu/ContextMenuBase.java#L423-L433

web-padawan: Good point, thanks. I will try to refactor the code accordingly.

web-padawan: BTW, we no longer defer updating client-side sorter indicators since https://github.com/vaadin/flow-components/pull/6030. So I'm not sure how to schedule this update as apparently `beforeClientResponse` is generally used for that, but in this case it's not an option.

sissbruecker: Every JS call is technically "scheduled", it will only be sent to the client at the end of the roundtrip.`beforeClientResponse` is only necessary if you want to make sure your code is run after all the other scheduled JS calls have been executed. 

So in this case we could add a field to the grid like so:
```java
private PendingJavaScriptResult pendingSorterUpdate;
``` 

In `updateClientSideSorterIndicators`, we would first check if there is a pending call, and if so, cancel it:
```java
if (pendingSorterUpdate != null && !pendingSorterUpdate.isSentToBrowser()) {
    pendingSorterUpdate.cancelExecution();
}
```

And when scheduling, assign the result to the field:
```java
            this.pendingSorterUpdate = getElement().executeJs(
                    "if (this.$connector) { this.$connector.setSorterDirections($0) }",
                    directions);
```

sissbruecker: From just looking at the linked issue I'm not sure if re-scheduling the calls would be an issue, but I guess we can try it out and see if the tests pass 🙂.

web-padawan: Updated the code as you suggested and confirmed that the problem is fixed. Also reverted the connector change.

sissbruecker (approved):

Issue #6066, opened 2024-02-19
Grid-Component : Errors with Drag-And-Drop and Item-Selection


SonReimer:
### Description of the bug

A View has a grid for which a Drag-And-Drop with DragMode.BETWEEN is configured and that will navigate to another view on item click. Clicking on an item leads to several javascript erros in the browser console.
In our application this had the effect, that some checkboxes could not be clicked afterwards (something got broken), but I could not reproduce this in the showcase.
if DragAndDrop is deactivated (see if(true)-Statement in class PersonGridDragDrop no errors occur.
[example4.zip](https://github.com/vaadin/flow/files/14329862/example4.zip)

### Expected behavior

I expect that no javascript errors occur on the console

### Minimal reproducible example

See attached example

### Versions

- Vaadin / Flow version: 23.3.33
- Java version: 17
- OS version: Windows
- Browser version (if applicable): Firefox
- Application Server (if applicable): Tomcat
- IDE (if applicable): Eclipse

vursen:
The issue isn't reproducible starting from `24.1`. I'm currently trying to identify the specific commit that fixed it.

vursen:
This issue has been fixed in `24.1` by [the removal of the dnd polyfill](https://github.com/vaadin/flow-components/pull/5686). The polyfill, however, wasn't removed from `23.3`, as this version still supports Safari 14.1. For more insights into what happens in the issue scenario, see @DiegoCardoso's detailed explanation in https://github.com/vaadin/flow-components/issues/1434#issuecomment-1807908656.

sissbruecker:
Simplified reproduction:
```java
Grid<String> grid = new Grid<>();
grid.addColumn(s -> s).setHeader("Item");
grid.setItems("foo", "bar", "baz");
grid.setDropMode(GridDropMode.BETWEEN);

grid.addItemClickListener(event -> {
    UI.getCurrent().navigate(OtherView.class);
});
add(grid);
```

sissbruecker:
Moving this to the Flow repo, as it turns out to be an issue with how Flow handles detaching of child nodes.

PR #6126, opened 2024-03-15
fix: add mobile polyfill without creating drop target

sissbruecker:
## Description

Currently clicking on an item in a grid results in Javascript errors if drag and drop is enabled and the click results in navigation to a different view. It's caused by the Grid creating a drop target when enabling drag and drop, which is only necessary to install a mobile DND polyfill for iOS. The drop target is deactivated right after creating it.

The issue can be fixed by not creating a drop target and instead installing the polyfill directly using the respective `DndUtil` class. The code comment mentions that using this class should be avoided since it's marked as internal, but I think it's less of an issue now. It's unlikely that this class or the method will be removed in v23, and even if its modified we'll catch the issue in platform validation and can adapt. In v24 the mobile polyfill has been removed entirely already.

Fixes https://github.com/vaadin/flow-components/issues/6066

## Type of change

- Bugfix


sissbruecker:
Seems the fix is not viable as the internal class is not availabe in OSGI environments: https://github.com/vaadin/vaadin-grid-flow/pull/863, which is still the case for v23.

PR #6138, opened 2024-03-18
chore: add note about TextField value change modes in TextFieldElement

TatuLund:


vaadin-text-field-flow-parent/vaadin-text-field-testbench/src/main/java/com/vaadin/flow/component/textfield/testbench/TextFieldElement.java
```diff
@@ -36,6 +36,11 @@ public class TextFieldElement extends TestBenchElement
      * Emulates the user changing the value, which in practice means setting
      * {@code value} of the {@code input} element to the given value and then
      * triggering {@code input}, {@code change} and {@code focusout} DOM events.
+     * <p>
+     * <em>Note:</em>Note: If you are using <code>ValueChangeMode.LAZY</codE> or
+     * <code>ValueChangeMode.TIMEOUT</code>, value change event is triggered
+     * also without field losing focus, and hence use of <code>sendKeys</code>
+     * may be better choice in testing.
```
sissbruecker, line 43: How is it better? Wouldn't that mean you also have to add a `waitUntil` after using `sendKeys`, otherwise you end up with a flaky test?

TatuLund: My first idea was, that sometimes you should avoid triggering the blur, and just use sendKeys and let the field trigger the value change after timeout. And yes it is then an additional point that whether some wait is needed in some cases as well. But the whole point here is that setValue is mostly applicable only when using ValueChangeMode.BLUR and ValueChangeMode.EAGER, otherwise risk of flake in a way or other is possible. I just encountered this problem recently in my demo project.

sissbruecker: Sorry, I still don't quite understand what the original issue is that you are trying to address. In which way is `setValue` problematic when using TIMEOUT / LAZY?

vursen: The `setValue` method triggers artificial `input`, `change`, and `focusout` events on the input element. The `focusout` event was added for IntegerField, whose web component commits the value on `focusout` rather than on input element change. The effect of `setValue` is similar to what happens when you click on increment or decrement button of a number field. If there are situations where it doesn't work with lazy modes, it should be considered a bug. 

As for testing, there should be no need to test the behavior of lazy modes in user applications, since it's already tested in our repository. For user applications, it should be enough to make sure that the field is set to the desired mode using unit tests.

TatuLund: I had some trouble in my project (stale element exceptions if I recall right), hence I see this note is appropriate. I have added similar note also in Vaadin 8 version.

vursen: One thing I've noticed is that the `setValue` method – and basically any action that leads to a change event – currently doesn't cause a value debouncer flush when using a lazy value change mode. This feels not right, so I opened https://github.com/vaadin/flow-components/issues/6265. 

If we fix this issue, `setValue` will start sending the value to the server immediately. Then we could rephrase this comment to say something like "Note that setValue causes an immediate value flush in lazy modes. If you want to avoid this, consider using sendKeys instead".

vursen: I've also updated the setValue's JavaDoc to mention sendKeys that can be used for more complex scenarios: https://github.com/vaadin/flow-components/pull/6263

vaadin-text-field-flow-parent/vaadin-text-field-testbench/src/main/java/com/vaadin/flow/component/textfield/testbench/TextFieldElement.java
```diff
@@ -40,6 +40,11 @@ public class TextFieldElement extends TestBenchElement
      * <p>
      * For more complex scenarios that require a full browser simulation of
      * typing, use {@link #sendKeys(CharSequence...)} instead.
+     * <p>
+     * Note, if you are using {@code ValueChangeMode.LAZY} or
+     * {@code ValueChangeMode.TIMEOUT}, value change event is triggered also
+     * without field losing focus, and hence use of
+     * {@link #sendKeys(CharSequence...)} may be better choice in testing.
```
vursen, line 47: I've rebased the PR and now the latest revision of setValue already has one mention of sendKeys here:

https://github.com/vaadin/flow-components/blob/e4008920d352eabf33d82cc9fc9845f97420d2c1/vaadin-text-field-flow-parent/vaadin-text-field-testbench/src/main/java/com/vaadin/flow/component/textfield/testbench/TextFieldElement.java#L40-L42

If you think mentioning value change modes is still helpful, I'd like the comment to explain why using sendKeys is a better option in those modes.

yuriy-fix:
Closing as per @vursen comment.

PR #6142, opened 2024-03-19
fix: enable invisible button after disable on click

### LINKED ISSUE #5924: Problem when update component after async execution
### Description

I added 1 button to my screen, set the disableOnClick to true and add a listener to execute a function in async mode. when my async methos end I set the button to enabled, but if I click the button a second time (after the process finish) the button still disabled, the button.setEnabled(true) dont works.

### Expected outcome

Every time that I click in button and the process has finished the button should be enabled

```

@PageTitle("Hello World")
@Route(value = "hello", layout = MainLayout.class)
@RouteAlias(value = "", layout = MainLayout.class)
public class HelloWorldView extends HorizontalLayout {

    private Button btnFinish = new Button("Click");
    private ProgressBar progress = new ProgressBar();
    private final ConferenciaMatriculaService conferenciaMatriculaService;
    public HelloWorldView(ConferenciaMatriculaService conferenciaMatriculaService) {
        this.conferenciaMatriculaService = conferenciaMatriculaService;
        btnFinish.addClickListener(this::btnFinishClickListener);
        btnFinish.setDisableOnClick(true);
        progress.setIndeterminate(true);
        progress.setVisible(false);

        setMargin(true);

        add(btnFinish, progress);
    }

    private void btnFinishClickListener(ClickEvent<Button> buttonClickEvent) {

        progress.setVisible(true);
        btnFinish.setVisible(false);
        UI ui = UI.getCurrent();
        CompletableFuture.runAsync(()-> {
            try {
                Thread.sleep(6000);
            } catch (InterruptedException e) {
                System.out.println("deu erro");
            }
        }).whenComplete((a,b) -> {
            ui.access(() -> {
                btnFinish.setEnabled(true);
                btnFinish.setVisible(true);
                progress.setVisible(false);
                Notification.show("Error: " + b.getMessage());

            });
        });

    }

}

```

### Steps to reproduce

1. Create the view with code above
2. Click in button and wait the process;
3. Click again in button (The will stay disabled)

### Environment

Vaadin version(s): 
OS: Windows 11 
Computer: Acert Nitro 5

### Browsers

Chrome

sissbruecker:
When using disable on click, `Button` has a workaround for the case where the button is enabled again in the same roundtrip that it was disabled. Basically the client-side component disables itself right after the click, the server-side component disables itself, and then the application code enables the button again in the same roundtrip. In that case Flow doesn't detect any changes to the disabled state and wouldn't enable the client-side component again. This is the original issue that caused the workaround to be introduced: https://github.com/vaadin/vaadin-button-flow/issues/115.

The existing workaround does some tampering with state nodes that doesn't seem to work if the component is invisible, probably because Flow ignores updates to invisible components. This change replaces the workaround with a different solution that forces the client-side `disabled` property to be updated using JS whenever `setEnabled` is called. That seems to work in both cases.

Fixes #5924

### DIFF
--- a/vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/main/java/com/vaadin/flow/component/button/tests/ButtonView.java
+++ b/vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/main/java/com/vaadin/flow/component/button/tests/ButtonView.java
@@ -50,6 +50,7 @@ public class ButtonView extends Div {
         createDisabledButton();
         createButtonWithDisableOnClick();
         createButtonWithDisableOnClickThatEnablesInSameRoundtrip();
+        createButtonWithDisableOnClickThatIsHidden();
         addVariantsFeature();
         createButtonsWithShortcuts();
 
@@ -260,6 +261,22 @@ public class ButtonView extends Div {
         button.setId("disable-on-click-re-enable-button");
         addCard("Button disabled on click and re-enabled in same roundtrip",
                 button);
+    }
+
+    private void createButtonWithDisableOnClickThatIsHidden() {
+        Button button = new Button("Disabled on click and hidden", event -> {
+            event.getSource().setVisible(false);
+        });
+        button.setDisableOnClick(true);
+        button.setId("disable-on-click-hidden-button");
+
+        Button enableButton = new Button("Enable hidden button", event -> {
+            button.setVisible(true);
+            button.setEnabled(true);
+        });
+        enableButton.setId("enable-hidden-button");
+
+        addCard("Button disabled on click and hidden", button, enableButton);
     }
 
     private void addCard(String title, Component... components) {
--- a/vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/main/java/com/vaadin/flow/component/button/tests/ButtonView.java
+++ b/vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/main/java/com/vaadin/flow/component/button/tests/ButtonView.java
@@ -270,10 +270,11 @@ public class ButtonView extends Div {
         button.setDisableOnClick(true);
         button.setId("disable-on-click-hidden-button");
 
-        Button enableButton = new Button("Enable hidden button and show", event -> {
-            button.setEnabled(true);
-            button.setVisible(true);
-        });
+        Button enableButton = new Button("Enable hidden button and show",
+                event -> {
+                    button.setEnabled(true);
+                    button.setVisible(true);
+                });
         enableButton.setId("enable-hidden-button");
 
         addCard("Button disabled on click and hidden", button, enableButton);
--- a/vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/main/java/com/vaadin/flow/component/button/tests/ButtonView.java
+++ b/vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/main/java/com/vaadin/flow/component/button/tests/ButtonView.java
@@ -49,6 +49,7 @@ public class ButtonView extends Div {
         createButtonsWithTabIndex();
         createDisabledButton();
         createButtonWithDisableOnClick();
+        createButtonWithDisableOnClickThatEnablesInSameRoundtrip();
         addVariantsFeature();
         createButtonsWithShortcuts();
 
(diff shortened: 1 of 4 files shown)


vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/test/java/com/vaadin/flow/component/button/tests/ButtonIT.java
```diff
@@ -276,6 +276,37 @@ public void removeDisabled_buttonWorksNormally() {
 
     }
 
+    @Test
+    public void disableOnClick_enableInSameRoundtrip_clientSideButtonIsEnabled() {
```
sissbruecker, line 280: This tests that the workaround for disabling and enabling in the same roundtrip still works. After just removing the workaround none of the ITs failed otherwise 🤷‍♂️

vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/test/java/com/vaadin/flow/component/button/tests/ButtonIT.java
```diff
@@ -276,6 +276,37 @@ public void removeDisabled_buttonWorksNormally() {
...
+                    "arguments[0].click(); return arguments[0].disabled",
+                    button);
+            Assert.assertTrue(disabled);
+
+            waitUntil(ExpectedConditions.elementToBeClickable(button));
+        }
+    }
+
+    @Test
+    public void disableOnClick_hideWhenDisabled_showWhenEnabled_clientSideButtonIsEnabled() {
```
sissbruecker, line 294: This is a regression test for the new issue.

vaadin-button-flow-parent/vaadin-button-flow/src/test/java/com/vaadin/flow/component/button/tests/ButtonTest.java
```diff
@@ -314,12 +314,6 @@ public void disableOnClick_click_componentIsDisabled() {
         Assert.assertFalse(
                 "Button should have been disabled when event has been fired",
                 buttonIsEnabled.get());
-
-        StateNode node = button.getElement().getNode();
-        HashMap<String, Serializable> changeTracker = node.getChangeTracker(
-                node.getFeature(ElementAttributeMap.class), () -> null);
-        Assert.assertEquals("Disabled attribute should be set to true", "true",
-                changeTracker.get("disabled"));
```
sissbruecker, line 322: This was testing implementation details of the old workaround that don't apply anymore. I think it's safe to remove with having integration tests in place.

vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/main/java/com/vaadin/flow/component/button/tests/ButtonView.java
```diff
@@ -250,6 +252,33 @@ private void createButtonWithDisableOnClick() {
...
+                    event.getSource().setEnabled(true);
+                });
+        button.setDisableOnClick(true);
+        button.setId("disable-on-click-re-enable-button");
+        addCard("Button disabled on click and re-enabled in same roundtrip",
+                button);
+    }
+
+    private void createButtonWithDisableOnClickThatIsHidden() {
+        Button button = new Button("Disabled on click and hidden", event -> {
```
vursen, line 267:
```suggestion
        Button button = new Button("Disabled on click and hide", event -> {
```

vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/main/java/com/vaadin/flow/component/button/tests/ButtonView.java
```diff
@@ -250,6 +252,33 @@ private void createButtonWithDisableOnClick() {
...
+            event.getSource().setVisible(false);
+        });
+        button.setDisableOnClick(true);
+        button.setId("disable-on-click-hidden-button");
+
+        Button enableButton = new Button("Enable hidden button", event -> {
```
vursen, line 273:
```suggestion
        Button enableButton = new Button("Enable hidden button and show", event -> {
```

vursen (approved):
Seems to work well. The only thing to note is that Flow accumulates any executeJs calls while the element is hidden, and then sends them all at once when the element becomes visible. But I guess it's not a cause for a concern as long as the client cannot "undisable" the button once it's been disabled on click.

PR #6165, opened 2024-04-02
feat: implement proper readonly state in CheckboxGroup

web-padawan:
## Description

Part of https://github.com/vaadin/platform/issues/5178

Depends on https://github.com/vaadin/web-components/pull/7199

Updated `CheckboxGroup.setReadonly(true)` to use `readonly` property on the web component instead of disabling.

## Type of change

- Feature

### DIFF
--- a/vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/CheckboxGroup.java
+++ b/vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/CheckboxGroup.java
@@ -105,8 +105,6 @@ public class CheckboxGroup<T>
     private final AtomicReference<DataProvider<T, ?>> dataProvider = new AtomicReference<>(
             DataProvider.ofItems());
 
-    private boolean isReadOnly;
-
     private SerializablePredicate<T> itemEnabledProvider = item -> isEnabled();
 
     private ItemLabelGenerator<T> itemLabelGenerator = String::valueOf;
@@ -461,26 +459,21 @@ public class CheckboxGroup<T>
 
     @Override
     public void onEnabledStateChanged(boolean enabled) {
-        if (isReadOnly()) {
-            setDisabled(true);
-        } else {
-            setDisabled(!enabled);
-        }
+        setDisabled(!enabled);
         getCheckboxItems().forEach(this::updateEnabled);
     }
 
     @Override
     public void setReadOnly(boolean readOnly) {
-        isReadOnly = readOnly;
+        getElement().setProperty("readonly", readOnly);
         if (isEnabled()) {
-            setDisabled(readOnly);
             refreshCheckboxes();
         }
     }
 
     @Override
     public boolean isReadOnly() {
-        return isReadOnly;
+        return getElement().getProperty("readonly", false);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/CheckboxGroup.java
```diff
@@ -461,26 +459,21 @@ public Registration addSelectionListener(
...
         getCheckboxItems().forEach(this::updateEnabled);
     }
 
     @Override
     public void setReadOnly(boolean readOnly) {
-        isReadOnly = readOnly;
+        getElement().setProperty("readonly", readOnly);
         if (isEnabled()) {
-            setDisabled(readOnly);
             refreshCheckboxes();
```
sissbruecker, line 470: Is this still required? Before the change, running this would disable individual checkboxes as well. After the change, individual checkboxes are not set as readonly. It looks like you can not tamper with the `CheckBoxGroup` value by making a checkbox writable on the client and changing it's value, as the `CheckBoxGroup` being readonly prevents its value from changing. So maybe marking individual checkboxes as readonly is not required.

So either we:
- Remove this part, at which point the override for `setReadOnly`, `getReadOnly` becomes unnecessary
- Or we update `refreshCheckboxes` to make individual checkboxes readonly as well

vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/test/java/com/vaadin/flow/component/checkbox/tests/CheckboxGroupTest.java
```diff
@@ -78,19 +78,14 @@ public void setValueNull_throws() {
...
         group.setReadOnly(true);
         Assert.assertTrue(group.isReadOnly());
 
         Assert.assertEquals(Boolean.TRUE.toString(),
-                group.getElement().getProperty("disabled"));
-
-        long disabledChildCount = group.getChildren().filter(
-                child -> child.getElement().getProperty("disabled", false))
-                .count();
-        Assert.assertEquals(group.getChildren().count(), disabledChildCount);
```
sissbruecker, line 93: Depending on which solution you decide on above, we might want to restore checking individual checkboxes to be readonly as well.

vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/test/java/com/vaadin/flow/component/checkbox/tests/CheckboxGroupTest.java
```diff
@@ -118,18 +113,12 @@ public void unsetReadOnlyDisabledCheckboxGroup_groupIsDisabledAndNotReadonly() {
     }
 
     @Test
-    public void setReadOnlyEnabledCheckboxGroup_groupIsDisabledAndNotReadonly() {
+    public void setReadOnlyEnabledCheckboxGroup_groupIsNotDisabled() {
```
sissbruecker, line 116: I feel this test is not needed anymore as it was just testing the previous workaround.

web-padawan: Right, thanks. Removed the test.

sissbruecker (approved):

PR #6182, opened 2024-04-09
fix: reset communicator regardless of filter

### LINKED ISSUE #5468: ComboBox with BackendDataProvider Fails to Load Data
### Description

When using Vaadin's ComboBox with a BackendDataProvider if a user enters a filter and waits for the results to load then changes and reverts the filter within the debounder timeout (0.5s) the ComboBox never loads the results 

### Expected outcome

The ComboBox should load the results when the filter is changed from and to the original filter within the debouncer timeout

```java

class AutoCompleteProvier extends AbstractBackEndDataProvider<Integer, String> {

  @Serial private static final long serialVersionUID = 2;

  private Map<Integer, String> data;
  /** The maximum length of the summary field */
  private static final int MAX_SUMMARY_LENGTH = 100;

  /** The maximum number of autocomplete suggestions to load */
  private static final int MAX_RESULTS = 10;

  /** Constructor of the class. */
  public AutoCompleteProvier() {
    this.data = new HashMap();
    for (int i = 0; i < 100; i++) {
      this.data.put(Integer.valueOf(i), i + " value");
    }
  }

  @Override
  protected int sizeInBackEnd(Query<Integer, String> query) {
    String filterText = query.getFilter().orElse(null);

    if (filterText == null) {
      return 0;
    }

    String finalFilterText = filterText;
    Map<Integer, String> filtered =
        this.data.entrySet().stream()
            .filter(a -> a.getValue().contains(finalFilterText))
            .collect(Collectors.toMap(Entry::getKey, Entry::getValue));

    int count = filtered.size();

    return min(MAX_RESULTS, count);
  }

  @Override
  protected Stream<Integer> fetchFromBackEnd(Query<Integer, String> query) {

    String filterText = query.getFilter().orElse(null);

    if (filterText == null) {
      return Stream.empty();
    }

    String finalFilterText = filterText;
    Map<Integer, String> filtered =
        this.data.entrySet().stream()
            .filter(a -> a.getValue().contains(finalFilterText))
            .collect(Collectors.toMap(Entry::getKey, Entry::getValue));

    int limit = query.getLimit();

    query.getOffset();
    query.getPageSize();
    return filtered.keySet().stream().limit(limit);
  }
}

@Route("numbercombo")
@PermitAll
public class NumberComboTest extends VerticalLayout {

  private NumberComboTest(@Nonnull EntityService entityService) {

    ComboBox<Integer> combo = newComboBox<>();

    AutoCompleteProvier dataProvider = new AutoCompleteProvier();
    combo.setPlaceholder("Search Numbers");

    combo.setItemLabelGenerator(value -> value == null ? "" : value.toString());
    combo.setAllowCustomValue(true);
    combo.setItems(dataProvider);
    combo.setWidthFull();
    combo.setAutofocus(false);

    add(new Span("Numbers:"), catalogCombo);
  }
}

```

### Steps to reproduce

1. Launch to the numbercombo url. 
2. Enter a number ie '1' and wait for results to load
3. Once loaded very quickly (<0.5s) delete the '1' and enter it again

If done quickly enough, the combo box should never loads the results and will spin forever

### Environment

- Vaadin / Flow version: 23.3.20
- Java version: openjdk 17.0.8 2023-07-18 LTS
- OS version: Windows 10
- Browser version (if applicable): Google Chrome Version 111.0.5563.65 (Official Build) (64-bit)
- Application Server (if applicable):
- IDE (if applicable): Intellij IDEA

### Browsers

Chrome, Firefox

### Workaround 
This issue can be worked around by creating a subclass that extends ComboBox and uses the function decorator `@ClientCallable` to "overwrite" `ComboBoxBase.resetDataCommunicator()`. Then using `getDataCommunicator()` to explicity call `dataCommunicator.reset()`.
```java
public class ComboBoxFix<T> extends ComboBox<T> {
  @ClientCallable
  private void resetDataCommunicator() {
    getDataCommunicator().reset();
  }
}
````

### PR DESCRIPTION
**Pull request by ugur-vaadin**
## Description

Currently, when a filter is quickly changed, the combobox is left in a loading state. The reason is the data communicator is never reset because the new filter applied is the same one as before. Even if the connector explicitly [handles](https://github.com/vaadin/flow-components/blob/main/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js#L99-L105) this case, the server side reset does not happen. The client-callable `resetDataCommunicator` is only used for this purpose, and calling it from the client side should be enough to trigger the reset in any case.

This PR removes the legacy workaround in the reset logic in the controller and directly resets the communicator.

Fixes #5468 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxDataController.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxDataController.java
@@ -152,6 +152,8 @@ class ComboBoxDataController<TItem>
     private Registration clearFilterOnCloseRegistration;
     private Registration dataProviderListener = null;
 
+    private boolean forceResetDueToClientSideFilterChange = false;
+
     /**
      * Creates a new data controller for that combo box
      *
@@ -284,6 +286,8 @@ class ComboBoxDataController<TItem>
         }
         dataCommunicator.setRequestedRange(start, length);
         filterSlot.accept(filter);
+        forceResetDueToClientSideFilterChange = Objects.equals(filter,
+                lastFilter);
     }
 
     /**
@@ -297,9 +301,7 @@ class ComboBoxDataController<TItem>
          * handling methods. Thus, if the current client filter is not empty,
          * then we need to re-set it in the data communicator.
          */
-        if (lastFilter == null || lastFilter.isEmpty()) {
-            dataCommunicator.reset();
-        } else {
+        if (lastFilter != null && !lastFilter.isEmpty()) {
             String filter = lastFilter;
             lastFilter = null;
             /*
@@ -308,6 +310,15 @@ class ComboBoxDataController<TItem>
              * consumer, so we don't need to explicitly call it.
              */
             filterSlot.accept(filter);
+        }
+        /*
+         * The filter slot will not call 'DataCommunicator::reset' when the
+         * applied filter is the same as the previously applied filter. This has
+         * to be done manually here.
+         */
+        if (forceResetDueToClientSideFilterChange) {
+            dataCommunicator.reset();
+            forceResetDueToClientSideFilterChange = false;
         }
     }
 
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxDataController.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxDataController.java
@@ -152,7 +152,7 @@ class ComboBoxDataController<TItem>
     private Registration clearFilterOnCloseRegistration;
     private Registration dataProviderListener = null;
 
-    private boolean forceResetDueToClientSideFilterChange = false;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBoxDataController.java [hunk 1]
```diff
@@ -284,6 +286,8 @@ void setRequestedRange(int start, int length, String filter) {
         }
         dataCommunicator.setRequestedRange(start, length);
         filterSlot.accept(filter);
+        forceResetDueToClientSideFilterChange = Objects.equals(filter,
```
vursen, line 289: The approach is currently based on an assumption about the timing and order in which the connector requests data. I wonder if there is a way to avoid an inter-request flag and rather limit the fix to the `resetDataCommunicator` method only.

ugur-vaadin: Good point. Updated the logic to use a flag that is updated on data communicator reset. This way, whenever it is called should not matter anymore.

vursen (approved):
LGTM, couldn't find any issues with the refactored version of `resetDataCommunicator()`.

PR #6201, opened 2024-04-15
fix: prevent setting disabled to false with pointer events: auto

### LINKED ISSUE #6183: Button's setDisabledOnClick(true) has an enabling side effect when receiving click events due to pointer-events: auto (instead of none)
### Description of the bug

Any button that has `setDisabledOnCLick(true)` could possibly wrongly set a button to be enabled due to this code snippet:

```
function disableOnClickListener({
  currentTarget: E
}) {
  E.disabled = E.hasAttribute("disableOnClick")
}
```

As one can see, the code assumes it is only executed when the button is enabled. This does not have to be the case, e.g. when setting pointer-events: auto then the click listener is executed too -- and it has the (wrong) side effect of enabling a button that was not enabled before (and should stay disabled).

better:
```
if ( E.hasAttribute("disableOnClick" ) {
  E.disabled = true;
}
```
After all, the better code knows there was a click and it knows it should disable upon a click. There is no side effect when the click happens and the button was not enabled. I consider this to be more robust.

### Expected behavior

Button should not surprisingly enable itself.

Set the `pointer-events: auto` for the disabled button.

```
[disabled] {
  cursor: default;
  pointer-events: auto;
}
```

### Versions

- Vaadin / Flow version: 24.3
- Java version: 17
- OS version: n/a
- Browser version (if applicable): Chrome latest / Firefox latest
- Application Server (if applicable): n/a
- IDE (if applicable): n/a

web-padawan:
## Description

Fixes #6183

This changes the client side JS listener as suggested in the issue to only set `disabled` to `true` and not `false`.

## Type of change

- Bugfix

## Note

The IT setup represents an edge case as it uses `setDisableOnClick(true)` followed by `setDisableOnClick(false)` and is based on the code example https://github.com/vaadin/flow-components/issues/6183#issuecomment-2046669463. Also, clicking the button in the IT requires firing an event manually.

### DIFF
--- a/vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/test/java/com/vaadin/flow/component/button/tests/ButtonIT.java
+++ b/vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/test/java/com/vaadin/flow/component/button/tests/ButtonIT.java
@@ -308,6 +308,21 @@ public class ButtonIT extends AbstractComponentIT {
     }
 
     @Test
+    public void disabledAndPointerEventsAuto_disableOnClick_clientSideButtonIsDisabled() {
+        WebElement button = layout
+                .findElement(By.id("disable-on-click-pointer-events-auto"));
+
+        scrollToElement(button);
+        executeScript("arguments[0].dispatchEvent(new MouseEvent(\"click\"));",
+                button);
+
+        String disabled = button.getAttribute("disabled");
+        Assert.assertTrue(
+                "The button should contain the 'disabled' attribute after click",
+                disabled != null);
+    }
+
+    @Test
     public void buttonShortcuts_shortcutsWork() {
         WebElement button = findElement(By.id("shortcuts-enter-button"));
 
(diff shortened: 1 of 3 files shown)


vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/test/java/com/vaadin/flow/component/button/tests/ButtonIT.java
```diff
@@ -307,6 +307,21 @@ public void disableOnClick_hideWhenDisabled_showWhenEnabled_clientSideButtonIsEn
...
+                .findElement(By.id("disable-on-click-pointer-events-auto"));
+
+        scrollToElement(button);
+        executeScript("arguments[0].dispatchEvent(new MouseEvent(\"click\"));",
+                button);
+
+        String disabled = button.getAttribute("disabled");
+        Assert.assertTrue(
+                "The button should contain the 'disabled' attribute after click",
+                disabled != null);
```
vursen, line 322: **nit:**

```suggestion
        Assert.assertNotNull(
                "The button should contain the 'disabled' attribute after click",
                button.getAttribute("disabled"));
```

vursen (approved):

PR #6208, opened 2024-04-16
fix: update formula bar after the other executions are done

### LINKED ISSUE #5731: Spreadsheet:  Incorrect number formatting in formula - locale and precision is lost
### Description

When you enter a number with lots of digits to the cell, then the actual value of the number is truncated by `CellValueManager.EXCEL_FORMULA_BAR_DECIMAL_FORMAT`. That is very similar behavior as in Excel 2019, where the number is trucated so it has 16 digits (entering `=0,123456789012345678` to a cell in Excel 2019 turns into `=0,123456789012345` in the formula bar).

This behavior works correctly in `Spreadsheet` only when you enter a plain number to a cell (as shown in the first part of the video). Once you enter the number as a formula (with a leading `=`), then strange things start to happen. The value in the cell is displayed correctly, but when you go to the formula bar then:
1. The number is truncated to only three decimal places in the formula bar. So the precision is lost - you cannot edit the original number!
2. Select the cell and go to another sheet, then return back -> suddenly, the number in the formula bar has the original precision, but it's not formatted using the current UI locale!

See it on this video:

https://github.com/vaadin/flow-components/assets/1890879/854963f9-d0ab-4156-bc9a-fb0414a707c8

### Expected outcome

I'd expect numbers, that are part of formulas, to be always formatted using the current UI locale and I'd expect, that the numbers in the formula bar show maximum possible precision - regardless of whether they are part of a formula or not

See the behavior on the attached video

### Steps to reproduce

See the behavior on the attached video

### Environment

Vaadin version(s): 24.3
OS: MacOs
Not browser related

### Browsers

_No response_

ugur-vaadin:
## Description

This PR defers the update of the formula bar so that all the calculations are done beforehand. This fixes the case where the formula bar displays the number using the wrong locale when the cell is selected using the address field. This also applies for the case changing sheet after setting the value and returning to the sheet.

Partially fixes #5731

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-integration-tests/src/test/java/com/vaadin/flow/component/spreadsheet/test/FormulaFieldFormatIT.java
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-integration-tests/src/test/java/com/vaadin/flow/component/spreadsheet/test/FormulaFieldFormatIT.java
@@ -37,6 +37,16 @@ public class FormulaFieldFormatIT extends AbstractSpreadsheetIT {
     }
 
     @Test
+    public void formulaLocaleFormatting_selectCellUsingAddressField_formulaFieldContentsFormattedForLocale() {
+        setLocale(Locale.ITALY);
+        createNewSpreadsheet();
+        setCellValue("A1", "=0,123456");
+        clickCell("B1");
+        setAddressFieldValue("A1");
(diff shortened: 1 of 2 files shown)


yuriy-fix:
@ugur-vaadin , would you like to proceed with this PR in the future and open it for the review?

ugur-vaadin:
I can revisit this one during the upcoming CF and see what remains to be done

DiegoCardoso (approved):

PR #6216, opened 2024-04-18
fix: mark `DataSeriesItemSankey` as customized to ensure its proper serialization

DiegoCardoso:
## Description

Without this call, the object would be serialized as [an array containing two values](https://github.com/vaadin/flow-components/blob/main/vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/model/serializers/DataSeriesItemBeanSerializer.java#L75-L80) from `bean.getLow()` and `bean.getHigh()`, which are not usually set on the Sankey type, so the result would be `[null, null]`, instead of the actual JSON string expected (containing the `from`, `to`, and `weight` keys).

The Sankey example in the IT works because it uses `setDataLabels`, which in turn [calls `makeCustomized`](https://github.com/vaadin/flow-components/blob/main/vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/model/DataSeriesItem.java#L366-L369) and then make the instance to be properly serialized.

## Type of change

- [X] Bugfix
- [ ] Feature


alvarezguille (approved):
:shipit:

web-padawan (approved):

Issue #6221, opened 2024-04-17
[PT-PiT 24.4] V24.4 alpha 22 testing: "Upload" component missing some methods 


bennewi:
### Describe the bug

doesn't have some V24.3 like 
method fireUpdateProgress(long, long) 
constructor FileRejectedEvent(Upload, String)

### Expected-behavior

should have the V24.3 methods unless marked as deprecated, right?

### Reproduction

x

### System Info

x

PR #6235, opened 2024-04-24
chore: overload constructor in order not to break API

manolo:
fixes: https://github.com/vaadin/flow-components/issues/6221

### DIFF
--- a/vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/FileRejectedEvent.java
+++ b/vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/FileRejectedEvent.java
@@ -33,6 +33,23 @@ public class FileRejectedEvent extends ComponentEvent<Upload> {
      * Creates a new event using the given source and indicator whether the
      * event originated from the client side or the server side.
      *
+     * @deprecated since 24.4. Use
+     *             {@link #FileRejectedEvent(Upload, String, String)}
+     *
+     * @param source
+     *            the source component
+     * @param errorMessage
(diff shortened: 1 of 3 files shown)


DiegoCardoso:
Wouldn't this partially fix #6221?
The protected method `Upload#fireUpdateProgress(long, long)` is still missing.

manolo:
@DiegoCardoso updated

DiegoCardoso:
I just noticed that the constructor of `ProgressUpdatEvent` has been also changed too on the same PR:

```diff
public ProgressUpdateEvent(Upload source, long readBytes,
-            long contentLength) {
+           long contentLength, String fileName) {
```

Something we have discussed is that while this PR fixes the binary compatibility of the classes, it still breaks the behavior of anyone who currently extends `Upload` and overrides the `fireUpdateProgress` protected method. We need to ensure that this behavioral change is mentioned in the release notes.

manolo:
Can you explain the problem with an example?

DiegoCardoso:
If someone creates a class like

```java
class MyUpload extends Upload {
	protected void fireUpdateProgress(long totalBytes, long contentLength) {
		super(totalBytes, contentLength);
		// some custom logic
	}
}
```

Before, that method would be called from `DefaultStreamVariable#onProgress` method, but now as the new `fireUpdateProgress` containing the `fileName` parameter is called instead, the old method with two arguments will not be called anymore, so anyone relying on that would see a difference in behavior.

manolo:
Then if this is need to be documented, probably it's better just mention in the release notes the change and not overload any method. Still is an API change in a minor. DS team should decide what's the best option then

DiegoCardoso (approved):

PR #6242, opened 2024-04-24
chore: bump Testbench to 9.3.0.alpha1

mcollovati:
Fixed tests to adapt to the new semantic of 'id()' query selector throwing an exception when there are more elements with the same id.
   
For CompositeIT, two modal dialogs are opened, adding two elements with 'overlay' id. The test now takes the last one, that corresponds to the Language edit dialog.

### DIFF
--- a/vaadin-crud-flow-parent/vaadin-crud-flow-integration-tests/src/test/java/com/vaadin/flow/component/crud/tests/CompositeIT.java
+++ b/vaadin-crud-flow-parent/vaadin-crud-flow-integration-tests/src/test/java/com/vaadin/flow/component/crud/tests/CompositeIT.java
@@ -28,7 +28,8 @@ public class CompositeIT extends AbstractParallelTest {
         $(TextFieldElement.class).attribute("editor-role", "language-field")
                 .first().setValue("English");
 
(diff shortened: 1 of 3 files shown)


mshabarov (approved):

vaadin-crud-flow-parent/vaadin-crud-flow-integration-tests/src/test/java/com/vaadin/flow/component/crud/tests/CompositeIT.java
```diff
@@ -28,8 +28,8 @@ public void compositeTouchesDirtyState() {
         $(TextFieldElement.class).attribute("editor-role", "language-field")
                 .first().setValue("English");
 
-        $(TestBenchElement.class).id("overlay").$(ButtonElement.class).first()
-                .click();
+        $(TestBenchElement.class).withId("overlay").last()
```
ZheSun88, line 31: i read this from the TB release note
```
Differently from the previous versions, ElementQuery.id(...) will now throw an exception if multiple elements with the same id are present, instead of returning the first matching item. To switch back to the old behavior, you have to change the id(...) call to withId(...).first().
```

but here we use `withId(...).last()` ?

Issue #6227, opened 2024-04-22
Upload: Switching to Single file receiver and then back to MultiFileReceiver prevents adding multiple files


HerbertsVaadin:
### Description

If I set receiver to instance of `MemoryBuffer` and then switch it to instance of `MultiFileReceiver` (for example `MultiFileMemoryBuffer`), the upload component still only allows to upload a single file. 
![bug-switching-upload-receiver](https://github.com/vaadin/flow-components/assets/80950643/e02cdf4f-2763-4f0b-b73c-acae6f408576)

**Workaround**: 
Manually set the max allowed files
`upload.setMaxFiles(12);`

### Expected outcome

When switching to a `MultiFileReceiver` it should allow to upload multiple files.

### Minimal reproducible example

```java
package com.example.application.views.random;

import com.example.application.views.MainLayout;
import com.vaadin.flow.component.orderedlayout.VerticalLayout;
import com.vaadin.flow.component.radiobutton.RadioButtonGroup;
import com.vaadin.flow.component.upload.Upload;
import com.vaadin.flow.component.upload.receivers.MemoryBuffer;
import com.vaadin.flow.component.upload.receivers.MultiFileMemoryBuffer;
import com.vaadin.flow.router.PageTitle;
import com.vaadin.flow.router.Route;
import jakarta.annotation.security.PermitAll;

import static com.example.application.views.random.UploadComponentView.Type.MULTI;
import static com.example.application.views.random.UploadComponentView.Type.SINGLE;

@PageTitle("Upload")
@Route(value = "upload", layout = MainLayout.class)
@PermitAll
public class UploadComponentView extends VerticalLayout {

    public UploadComponentView() {
        var upload = new Upload();

        var radioButtons = new RadioButtonGroup<Type>();
        radioButtons.setItems(Type.values());
        radioButtons.setItemLabelGenerator(Type::name);
        radioButtons.addValueChangeListener(valueChange -> {
            if (valueChange.getValue() == SINGLE) {
                upload.setReceiver(new MemoryBuffer());
            } else {
                upload.setReceiver(new MultiFileMemoryBuffer());
            }
        });
        radioButtons.setValue(MULTI);

        add(radioButtons, upload);
    }

    enum Type {
        SINGLE, MULTI
    }
}
```

### Steps to reproduce

1. Use the example above.
2. Switch to SINGLE radio button
3. Switch back to MULTI radio button
4. Notice that Upload component only allows to upload a single file

### Environment

[...]

PR #6253, opened 2024-04-26
fix: set maxFiles to Infinity when setting a MultiFileReceiver

vursen:
## Description

Fixes an issue where the user was unable to select more than one file after switching from a single-file receiver to a multi-file one. It was because the Flow component didn't set the `maxFiles` client-side property to `Infinity` – the web component's default value for that property – when a multi-file receiver was set.

Fixes https://github.com/vaadin/flow-components/issues/6227

## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
+++ b/vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
@@ -15,7 +15,6 @@
  */
 package com.vaadin.flow.component.upload;
 
-import java.awt.geom.Arc2D;
 import java.io.OutputStream;
 import java.util.ArrayDeque;
 import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -616,10 +616,11 @@ public Receiver getReceiver() {
      */
     public void setReceiver(Receiver receiver) {
         this.receiver = receiver;
-        if (!(receiver instanceof MultiFileReceiver)) {
-            setMaxFiles(1);
+        if (receiver instanceof MultiFileReceiver) {
+            getElement().removeProperty("maxFiles");
+            getElement().executeJs("this.maxFiles = Infinity");
```
vursen, line 621: Should we rather change the default value for this property?

vursen: We discussed this internally and decided to think about changing the web component's default value later.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -15,6 +15,7 @@
  */
 package com.vaadin.flow.component.upload;
 
+import java.awt.geom.Arc2D;
```
web-padawan: Unused import.

web-padawan (approved):

PR #6273, opened 2024-05-06
fix!: throw when setting value that is not present in key mapper

vursen:
## Description

Setting values that aren't present in the key mapper shouldn't be allowed, as they cannot be represented on the client-side.

Related to https://github.com/vaadin/flow/pull/19310

Fixes https://github.com/vaadin/flow-components/issues/6605

## Type of change

- [x] Bugfix


vursen:
ComboBox is another candidate for this fix.

yuriy-fix:
@vursen, please create a ticket for this PR to address. Discussion could be started within the ticket.

vursen:
> please create a ticket for this PR to address. Discussion could be started within the ticket.

PR #6284, opened 2024-05-10
feat: add API to set color options to RichTextEditor

web-padawan:
## Description

Related to https://github.com/vaadin/web-components/pull/7392

## Type of change

- Enhancement

## Note

Currently `getColorOptions()` doesn't return the default value of the corresponding web component property.
This could be probably workarounded by adding `notify: true` and using `@Synchronize` annotation.

### DIFF
--- a/vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java
+++ b/vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java
@@ -9,6 +9,7 @@
 package com.vaadin.flow.component.richtexteditor;
 
 import java.io.Serializable;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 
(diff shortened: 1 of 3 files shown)


vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java
```diff
@@ -269,6 +272,32 @@ private String getDeltaValue() {
...
+     * @since 24.5
+     * @return the list of colors options
+     */
+    public List<String> getColorOptions() {
+        return JsonSerializer.toObjects(String.class,
+                (JsonArray) getElement().getPropertyRaw("colorOptions"));
+    }
+
+    /**
+     * Sets the list of colors used by the text color picker and background
```
knoobie, line 288: What format is required here to be considered a color? hex, hsl, name.. something else? 😀

web-padawan: Added a mention that colors should be in HEX format (the web component popup uses that).
Note: the actual HTML value produced by Quill uses `rgb()` so the output looks like this:

```html
<p><span style="color: rgb(230, 0, 0);">Foo</span></p>
```

vaadin-rich-text-editor-flow-parent/vaadin-rich-text-editor-flow/src/main/java/com/vaadin/flow/component/richtexteditor/RichTextEditor.java
```diff
@@ -268,6 +270,32 @@ private String getDeltaValue() {
         return getElement().getProperty("value");
     }
 
+    /**
+     * Gets the list of colors in HEX format used by the text color picker and
+     * background color picker controls of the text editor.
```
sissbruecker, line 275: Let's mention that this doesn't have a default value and returns null by default. Maybe also elaborate that in this case the web component shows a default palette.

Apart from that modifying the list returned by the method doesn't do anything. Let's return an unmodifyable list which would trigger a respective exception when trying to modify it, and mention in the JavaDoc that this returns an umnodifyable list. There are some similar use cases in the repo already (e.g. setting avatar items).

web-padawan: Thanks, updated.

Issue #4618, opened 2023-01-30
Error messages for constraints


vursen:
### Describe your motivation

A follow-up to https://github.com/vaadin/platform/issues/3066

Every field component provides its own set of constraints such as `setMinLength(...)`, `setPattern(...)`, etc. Not long ago, we integrated component constraints into Binder validation but there's been no way to set up error messages for them since then. A red highlight is the only indication of constraint violations. That said, it's been possible to set up error messages for custom validators. As a result, the error messages may be inconsistent.

Error messages are integral in forms and constraints aren't an exception:

**Example:** You accidentally added extra spaces to a field that doesn't allow them due to a pattern constraint. It is really hard to notice the mistake without an explicit error message. 

**Example:** DatePicker cannot parse “11th of December, 2023” because it is an invalid format. It looks however like a valid date, so just a red highlight isn't enough to figure out the mistake.

### Describe the solution you'd like

We could add the second argument to constraint setters that would be used by `getDefaultValidator()` as an error message when returning a `ValidationResult`:

```java
textField.setPattern("^\\d+$", "Must consist of digits");
```

```java
passwordField.setMinLength(6, "Must have at least 6 characters");
```

The error messages added this way could be also shown during the built-in component constraint validation.

### Additional context

We need to consider that some might want to provide constraint error messages at the global level rather than for each individual field, so the implementation should apparently take this into account.

Related AC: https://github.com/vaadin/platform/issues/6707

knoobie:
Probably related with additional API suggestions: https://github.com/vaadin/flow-components/pull/3406#discussion_r925934529

knoobie:
Looking back at my comment from some month ago - instead of just another BiFunction, I would suggest to reuse Flow's `ErrorMessageProvider`as second param (for easier usage a simple String as second param should also work as overloaded method. To give Developers the full flexibility they are used by the `Binder::withValidator` API where this interace is also used.

gert-dubois:
What is the status of this issue? For our current project being able to customize these error messages (especially the non-parsable error message) is considered quite import.

vursen:
> What is the status of this issue? For our current project being able to customize these error messages (especially the non-parsable error message) is considered quite important.

Unfortunately, we are failing to make it to 24.2, but it's on our backlog and we are aiming at getting it into 24.3.

vursen:
I've opened a Github discussion with a proposal for error messages: https://github.com/orgs/vaadin/discussions/4664. 

[...]

PR #6330, opened 2024-05-29
feat: expose i18n object by adding setter and getter

vursen:
## Description

The PR provides public access to the i18n object of DateTimePicker by adding setter and getter methods. 

> [!NOTE]
> The i18n object has only two properties at the moment: dateLabel and timeLabel. These properties specify a suffix to be added to the aria-label of DatePicker and TimePicker, where the first part of that aria-label is the aria-label or label of DateTimePicker. Due to an API design issue, we also have methods setDateAriaLabel and setTimeAriaLabel set these suffixes despite their name suggesting they set the whole aria-label (overriding the default aria-label format). Ideally, they would be changed to set the whole aria-label while suffixes would only be set through i18n. However, this would be a significant breaking change, which is too big to be included in a minor version. So, for now, it was decided to keep these properties in i18n and also retain these methods, with the difference that the methods will store their labels separately from i18n and their labels will take precedence over the labels set through i18n to avoid undesired side effects.

Part of #4618 

## Type of change

- [x] Feature


vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/datetimepicker/DateTimePickerI18nIT.java
```diff
@@ -0,0 +1,93 @@
...
+    private DateTimePickerElement dateTimePicker;
+
+    @Before
+    public void init() {
+        open();
+        dateTimePicker = $(DateTimePickerElement.class).first();
+    }
+
+    @Test
+    public void setAndRemoveDateAriaLabel_assertI18nProperty() {
```
vursen, line 38: It seems that these tests could be rewritten into unit tests, as the only thing that check is the i18n property. I'll give it a try.

vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/datetimepicker/DateTimePickerI18nIT.java
```diff
@@ -0,0 +1,93 @@
...
+        Assert.assertEquals("Custom date",
+                dateTimePicker.getPropertyString("i18n", "dateLabel"));
+
+        $("button").id("remove-date-aria-label").click();
+        Assert.assertEquals("date",
+                dateTimePicker.getPropertyString("i18n", "dateLabel"));
+    }
+
+    @Test
+    public void setI18n_setAndRemoveTimeAriaLabel_assertI18nProperty() {
```
yuriy-fix, line 82: Nit: We should probably add one more test scenario. If the property is set before the i18n, we should ensure that the set property takes priority over the value provided within i18n.

vursen: Thanks, added.

vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/datetimepicker/DateTimePickerI18nIT.java
vursen: Done.

vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow/src/main/java/com/vaadin/flow/component/datetimepicker/DateTimePicker.java
```diff
@@ -890,6 +864,49 @@ public void setDatePickerI18n(DatePickerI18n i18n) {
         datePicker.setI18n(i18n);
     }
 
+    /**
+     * Gets the internationalization object previously set for this component.
+     * <p>
+     * Note: updating the object content that is gotten from this method will
+     * not update the lang on the component if not set back using
+     * {@link DateTimePicker#setI18n(DateTimePickerI18n)}
```
sissbruecker, line 872:
```suggestion
     * Note: updating the instance that is returned from this method will
     * not update the the component if not set back using
     * {@link DateTimePicker#setI18n(DateTimePickerI18n)}
```

vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow/src/main/java/com/vaadin/flow/component/datetimepicker/DateTimePicker.java
```diff
@@ -416,97 +420,67 @@ public Optional<String> getAriaLabel() {
...
      *
      * @return an optional label or an empty optional if no label has been set
+     *         with this method before.
      */
     public Optional<String> getDateAriaLabel() {
-        if (i18n == null) {
-            return Optional.empty();
-        }
-        return Optional.ofNullable(i18n.getDateLabel());
+        return Optional.ofNullable(dateAriaLabel);
```
sissbruecker, line 451: Do we consider this a minor breaking change here?

vursen: Not sure, could you explain what the breaking change is here?

sissbruecker: Ah nevermind, since the i18n instance was not accessible before this can not result in a change in behavior.

sissbruecker (approved):

vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow/src/main/java/com/vaadin/flow/component/datetimepicker/DateTimePicker.java
```diff
@@ -890,6 +864,49 @@ public void setDatePickerI18n(DatePickerI18n i18n) {
         datePicker.setI18n(i18n);
     }
 
+    /**
+     * Gets the internationalization object previously set for this component.
+     * <p>
+     * Note: updating the instance that is returned from this method will not
+     * update the component if not set back using
```
vursen, line 871:
```suggestion
     * NOTE: Updating the instance that is returned from this method will not
     * update the component if not set again using
```

DiegoCardoso (approved):

PR #6337, opened 2024-05-31
feat: add Popover opened state and related API

web-padawan:
## Description

Depends on #6332

Part of https://github.com/vaadin/platform/issues/5271

- Added logic to handle `opened` state and related event
- Added `setCloseOnEsc()` and `isCloseOnEsc()`
- Added `setCloseOnOutsideClick()` and `isCloseOnOutsideClick()`

The code is mostly copied from dialog with some exceptions:

- The "auto adding to the UI" logic is not implemented
- Modality and related logic is missing, to be added later

## Type of change

- Feature


DiegoCardoso (approved):

vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
```diff
@@ -117,6 +202,56 @@ public String getOverlayRole() {
         return getElement().getProperty("overlayRole");
     }
 
+    /**
+     * Gets whether this popover can be closed by pressing the Esc key or not.
+     * <p>
+     * By default, the popover is closable with esc.
```
web-padawan, line 208:
```suggestion
     * By default, the popover is closable with Esc.
```

vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
```diff
@@ -70,10 +73,92 @@ public class Popover extends Component implements HasAriaLabel, HasComponents {
...
+        getElement().setProperty("opened", opened);
+        fireEvent(new OpenedChangeEvent(this, fromClient));
+    }
+
+    /**
+     * Gets the open state from the popover.
+     *
+     * @return the {@code opened} property from the popover
+     */
+    @Synchronize(property = "opened", value = "opened-changed")
```
sissbruecker, line 145: So this has a similar issue to what was fixed for Dialog in https://github.com/vaadin/flow-components/pull/5258.

This is reproducable as such:
- Configure popover to open / close on hover
- Add a button to the popover to open a modal dialog
- Clicking the button opens the dialog and closes the popover on the client
- However on the server the opened changed event for the popover is ignored due to server-side modality
- This results in the popover staying open on the server, while it is actually closed on the client

Not really sure what to do about this. For Dialog we fixed it so that a dialog doesn't close itself on the client-side on outside click if another modal dialog is opened as part of that click. That doesn't really make sense for the popover though, which should always close when it is not hovered. 

We could ignore the issue for now and document it in a ticket. I think in that case the property listener in the constructor and the `doSetOpened` method are not necessary, as those were introduced as part of the fix mentioned above.

web-padawan: Thanks, I missed this. Reported #6433 and removed the property listener / `doSetOpened` method for now.

vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
```diff
@@ -70,10 +73,80 @@ public class Popover extends Component implements HasAriaLabel, HasComponents {
...
         updateTrigger();
         setOverlayRole("dialog");
     }
 
+
+    /**
+     * {@code opened-changed} event is sent when the overlay opened state
+     * changes.
+     */
+    public static class OpenedChangeEvent extends ComponentEvent<Popover> {
```
sissbruecker, line 88: This needs to be annotated with `@DomEvent("opened-changed")` now, otherwise the event listener won't work.

sissbruecker (changes requested):

sissbruecker (approved):

PR #6322, opened 2024-05-24
refactor(ValidationUtil): add methods accepting error message, deprecate old ones

vursen:
## Description

- Added new methods `validateRequiredConstraint`, `validateMinConstraint` and `validateMaxConstraint` to `ValidationUtil`. They accept a string error message to return in the `ValidationResult` when the check fails.
- Deprecated the previous methods without an error message in favor of the new ones.

Part of https://github.com/vaadin/flow-components/issues/4618

## Type of change

- [x] Refactor

PR #6343, opened 2024-06-03
feat: add i18n error message support for TextField

vursen:
## Description

The PR introduces TextFieldI18n with methods for setting error messages and updates the TextField's validation logic to show these error messages when validation fails.

> [!NOTE]

Depends on
- https://github.com/vaadin/flow-components/pull/6322

Part of #4618 

## Type of change

- [x] Feature

### DIFF
--- a/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/ValidationUtil.java
+++ b/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/ValidationUtil.java
@@ -170,4 +170,72 @@ public class ValidationUtil {
         return isError ? ValidationResult.error(errorMessage)
                 : ValidationResult.ok();
     }
+
+    /**
+     * Checks if the value satisfies the minimum length constraint and returns a
+     * {@code ValidationResult.ok()} or {@code ValidationResult.error()} with
+     * the given error message depending on the result.
+     *
+     * @param errorMessage
+     *            the error message to return if the check fails
+     * @param value
+     *            the value to validate
+     * @param minLength
+     *            the minimum allowed length
+     * @return {@code ValidationResult.ok()} if the value is shorter than or
+     *         equal to the minimum length, {@code ValidationResult.error()}
+     *         otherwise
+     */
+    public static ValidationResult validateMinLengthConstraint(
+            String errorMessage, String value, Integer minLength) {
+        boolean isError = value != null && !value.isEmpty() && minLength != null
+                && value.length() < minLength;
+        return isError ? ValidationResult.error(errorMessage)
+                : ValidationResult.ok();
+    }
+
+    /**
+     * Checks if the value satisfies the maximum length constraint and returns a
+     * {@code ValidationResult.ok()} or {@code ValidationResult.error()} with
+     * the given error message depending on the result.
+     *
+     * @param errorMessage
+     *            the error message to return if the check fails
+     * @param value
+     *            the value to validate
+     * @param maxLength
+     *            the maximum allowed length
+     * @return {@code ValidationResult.ok()} if the value is longer than or
+     *         equal to the minimum length, {@code ValidationResult.error()}
+     *         otherwise
+     */
+    public static ValidationResult validateMaxLengthConstraint(
+            String errorMessage, String value, Integer maxLength) {
+        boolean isError = value != null && maxLength != null
+                && value.length() > maxLength;
+        return isError ? ValidationResult.error(errorMessage)
+                : ValidationResult.ok();
+    }
+
+    /**
+     * Checks if the value satisfies the pattern constraint and returns a
+     * {@code ValidationResult.ok()} or {@code ValidationResult.error()} with
+     * the given error message depending on the result.
+     *
+     * @param errorMessage
+     *            the error message to return if the check fails
+     * @param value
+     *            the value to validate
+     * @param pattern
+     *            the pattern to match
+     * @return {@code ValidationResult.ok()} if the value matches the pattern,
+     *         {@code ValidationResult.error()} otherwise
+     */
+    public static ValidationResult validatePatternConstraint(
+            String errorMessage, String value, String pattern) {
+        boolean isError = value != null && !value.isEmpty() && pattern != null
+                && !pattern.isEmpty() && !value.matches(pattern);
+        return isError ? ValidationResult.error(errorMessage)
+                : ValidationResult.ok();
+    }
 }
--- a/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/ValidationUtil.java
+++ b/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/ValidationUtil.java
@@ -182,7 +182,7 @@ public class ValidationUtil {
      *            the value to validate
      * @param minLength
      *            the minimum allowed length
-     * @return {@code ValidationResult.ok()} if the value is shorter than or
+     * @return {@code ValidationResult.ok()} if the value is longer than or
      *         equal to the minimum length, {@code ValidationResult.error()}
      *         otherwise
      */
@@ -205,8 +205,8 @@ public class ValidationUtil {
      *            the value to validate
      * @param maxLength
      *            the maximum allowed length
-     * @return {@code ValidationResult.ok()} if the value is longer than or
-     *         equal to the minimum length, {@code ValidationResult.error()}
+     * @return {@code ValidationResult.ok()} if the value is shorter than or
+     *         equal to the maximum length, {@code ValidationResult.error()}
      *         otherwise
      */
     public static ValidationResult validateMaxLengthConstraint(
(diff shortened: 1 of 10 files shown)


vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/TextField.java
```diff
@@ -337,4 +360,169 @@ protected void onAttach(AttachEvent attachEvent) {
...
+        private String maxLengthErrorMessage;
+        private String patternErrorMessage;
+
+        /**
+         * Gets the error message displayed when the field is required but
+         * empty.
+         *
+         * @return the error message or {@code null} if not set
+         * @see TextField#isRequiredIndicatorVisible()
+         * @see TextField#setRequiredIndicatorVisible(boolean)
```
vursen, line 426: **question:** Should we reference `setRequired(boolean)` and `isRequired()` here as well?

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/TextField.java
```diff
@@ -304,31 +288,71 @@ public String getValue() {
...
+                            isRequiredIndicatorVisible(), value,
+                            getEmptyValue());
+            if (requiredResult.isError()) {
+                return requiredResult;
+            }
+        }
+
+        ValidationResult maxLengthResult = ValidationUtil
+                .validateMaxLengthConstraint(getMaxLengthErrorMessage(), value,
+                        hasMaxLength() ? getMaxLength() : null);
```
vursen, line 315: **note:** `getMaxLength()` returns `0` if the maxLength property is `null`. I added this check to ensure the max length is only considered when the user has explicitly set it with `setMinLength(int minLength)`.

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/ValidationUtil.java
```diff
@@ -170,4 +170,72 @@ public static <V extends Comparable<V>> ValidationResult validateMinConstraint(
...
+     * the given error message depending on the result.
+     *
+     * @param errorMessage
+     *            the error message to return if the check fails
+     * @param value
+     *            the value to validate
+     * @param maxLength
+     *            the maximum allowed length
+     * @return {@code ValidationResult.ok()} if the value is longer than or
+     *         equal to the minimum length, {@code ValidationResult.error()}
```
tomivirkki, line 209: `minimum` -> `maximum`

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/TextField.java
```diff
@@ -304,31 +288,71 @@ public String getValue() {
...
+            ValidationResult requiredResult = ValidationUtil
+                    .validateRequiredConstraint(getRequiredErrorMessage(),
+                            isRequiredIndicatorVisible(), value,
+                            getEmptyValue());
+            if (requiredResult.isError()) {
+                return requiredResult;
+            }
+        }
+
+        ValidationResult maxLengthResult = ValidationUtil
```
tomivirkki, line 313: I wonder if it would make sense to reuse `validateMaxConstraint`?
```java
ValidationResult maxLengthResult = ValidationUtil.validateMaxConstraint(
                getMaxLengthErrorMessage(), value.length(),
                hasMaxLength() ? getMaxLength() : null);
```

tomivirkki: I don't have a strong preference for either approach so this isn't a blocker for the PR

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/ValidationUtil.java
vursen: Thanks, fixed this and also another typo.

tomivirkki (approved):

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/TextField.java
```diff
@@ -337,4 +361,167 @@ protected void onAttach(AttachEvent attachEvent) {
...
+
+    private String getMaxLengthErrorMessage() {
+        return Optional.ofNullable(i18n)
+                .map(TextFieldI18n::getMaxLengthErrorMessage).orElse("");
+    }
+
+    private String getPatternErrorMessage() {
+        return Optional.ofNullable(i18n)
+                .map(TextFieldI18n::getPatternErrorMessage).orElse("");
+    }
```
DiegoCardoso, line 407: Could use the same approach as suggested in https://github.com/vaadin/flow-components/pull/6365#discussion_r1675358980

vursen: Updated.

vursen:
During our internal discussions, we found a way to keep error messages that are set with `setErrorMessage`. They now take priority over i18n error messages, which allows developers to use the old straightforward approach to set a single error message for all constraints. I've updated the PR and also removed the breaking mark.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/TextField.java
```diff
@@ -40,12 +47,13 @@
...
 
-    private boolean isConnectorAttached;
-
-    private TextFieldValidationSupport validationSupport;
+    private TextFieldI18n i18n;
 
     private boolean manualValidationEnabled = false;
 
+    private String customErrorMessage;
+    private String constraintErrorMessage;
```
sissbruecker, line 55: What do you think about extracting this somehow? There could be a class that:
- Manages custom + constraint error message states
- Updates the effective error message property on the component when either is set

It wouldn't do a whole lot, but it looks like you have to replicate the logic quite a few times for different components. There's also `checkValidity` which is always the same though that one needs to access more stuff from the component. Anyway, maybe something to think about, we can also do this after updating all components and then see if there is something that works for all of them.

vursen: Are you suggesting extracting this only for text fields or for all field components?

vursen: I guess we could have something like `ErrorMessageController`:

```java
class ErrorMessageController {
  private Component component;

  public ErrorMessageController(HasValidation component) {
    this.component = component;
  }

  public void setCustomErrorMessage() {
    ...
    this.updateErrorMessageProperty();    
  }

  public void setConstraintErrorMessage() {
    ...
    this.updateErrorMessageProperty();
  }

  public void updateErrorMessageProperty() {
    this.component.getElement().setProperty("errorMessage");
  }
}
```

vursen: > There's also checkValidity which is always the same though that one needs to access more stuff from the component. 

I discussed this with @yuriy-fix some time ago, and we concluded that it would currently complicate things especially since the whole i18n needs to be passed to that helper class or the class would need to provide an API to abstract itself from i18n. We agreed to revisit this in the future when we get to implementing error message providers.

sissbruecker: Yep, something like what you suggested with the controller. It seems like that would be applicable to all components, but not sure. But since there are other potential opportunities for code reuse, let's maybe finish changing the components first, and then see what we can extract as common logic?

vursen: Agreed 👍

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/TextField.java
```diff
@@ -304,31 +328,79 @@ public String getValue() {
...
+    }
+
     /**
-     * Performs server-side validation of the current value and the validation
-     * constraints of the field, such as {@link #setPattern(String)}. This is
-     * needed because it is possible to circumvent the client-side validation
-     * constraints using browser development tools.
+     * Validates the current value against the constraints and sets the
+     * {@code invalid} property and the {@code errorMessage} property using the
+     * error messages defined in the i18n object.
```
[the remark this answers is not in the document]
vursen: Done.

sissbruecker (approved):

PR #5990, opened 2024-01-26
feat: add remove functionality for grid header and footers

ugur-vaadin:
## Description

This PR adds 4 new APIs to Grid:
- `removeHeaderRow(HeaderRow)`
- `removeAllHeaderRows()`
- `removeFooterRow(FooterRow)`
- `removeAllFooterRows()`

Notes:
- The first column layer is never removed. Instead, the content on it is manipulated.
- The default header row can only be removed only when there is no other header row.

Fixes #1538 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [x] I have read the contribution guide: https://vaadin.com/docs/latest/contributing/overview
- [x] I have added a description following the guideline.
- [x] The issue is created in the corresponding repository and I have referenced it.
- [x] I have added tests to ensure my change is effective and works as intended.
- [x] New and existing tests are passing locally with my change.
- [x] I have performed self-review and corrected misspellings.
- [ ] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.

PR #6344, opened 2024-06-03
feat: add remove functionality for grid header and footers (CP: #5990)

yuriy-fix:
This a cherry-pick of the https://github.com/vaadin/flow-components/pull/5990


ugur-vaadin:
In v24, the `hidden` attribute is set to the `tr` element for header and footer whenever the content is removed using the new API. For some reason, it is not the case with v23 back-port. Needs further investigation.

PR #6346, opened 2024-06-04
feat: add i18n error message support for NumberField

vursen:
## Description

The PR introduces NumberFieldI18n with methods for setting error messages and updates the NumberField's validation logic to show these error messages when validation fails. 

> [!NOTE]

Part of #4618 

## Type of change

- [x] Feature


vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/AbstractNumberField.java
```diff
@@ -407,4 +427,164 @@ protected void onAttach(AttachEvent attachEvent) {
...
+    private String getStepErrorMessage() {
+        return Optional.ofNullable(i18n)
+                .map(AbstractNumberFieldI18n::getStepErrorMessage).orElse("");
+    }
+
+    /**
+     * The abstract internationalization properties for {@link AbstractNumberField}.
+     */
+    public static abstract class AbstractNumberFieldI18n
+            implements Serializable {
```
knoobie, line 484: What are the benefits of the abstract class if the implementing class has to do all the "heavily" lifting? (Like overriding all methods)

vursen: Mmm, right. I think I could avoid needing an abstract class by making the visibility of getRequiredErrorMessage-like methods package-private and overriding them in the implementing classes to return error messages from their i18n objects. Not sure if it's the best alternative, but this should at least save us from having this abstract class which is public.

knoobie: Just asking without much thought: Wouldn't it be possible to create a shared instance `NumberFieldI18n` which is used in all different number fields? It feels kinda overwhelming to e.g. maintain different classes for a field "type" that's so similar.

vursen: I brought this up in our internal discussion and we leaned toward proceeding with separate i18n classes for each number field class to keep it consistent with other fields. Using a single class could lead to breaking changes in the setI18n / getI18n signatures if we later need to add properties specific to a single number field, even though this is unlikely. Given that we already use AbstractNumberField, using AbstractNumberFieldI18n feels more consistent. Package-private methods are another option, but the abstract class seems more appropriate.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/AbstractNumberField.java
```diff
@@ -388,15 +414,185 @@ private boolean isValidByStep(T value) {
...
+        public abstract AbstractNumberFieldI18n setBadInputErrorMessage(
+                String errorMessage);
+
+        /**
+         * Gets the error message displayed when the field is required but
+         * empty.
+         *
+         * @return the error message or {@code null} if not set
+         * @see AbstractNumberField#isRequired()
+         * @see AbstractNumberField#setRequired(boolean)
```
vursen, line 514: **question:** Should we mention `setRequiredIndicatorVisible` and `isRequiredIndicatorVisible` here as well?

vaadin-text-field-flow-parent/vaadin-text-field-flow-integration-tests/src/test/java/com/vaadin/flow/component/textfield/tests/validation/NumberFieldBasicValidationIT.java
```diff
@@ -63,11 +71,25 @@ public void required_changeValue_assertValidity() {
...
         assertValidationCount(1);
         assertServerInvalid();
         assertClientInvalid();
+        assertErrorMessage(REQUIRED_ERROR_MESSAGE);
+
+        testField.sendKeys("--2", Keys.ENTER);
+        assertValidationCount(1);
+        assertServerInvalid();
+        assertClientInvalid();
+        assertErrorMessage(BAD_INPUT_ERROR_MESSAGE);
```
vursen, line 86: **note:** This ensures that the required validator executes after the bad input validator (previously, it was the other way around).

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/AbstractNumberField.java
DiegoCardoso: As a debate: I wondered why this class couldn't have a generic type like `<T extends AbstractNumberFieldI18n<T>>` and then use it as the return type of the setters. From what I can see, the current contract could allow a child class to return an instance of a second child class from a setter.

vursen: We talked about this approach internally and found that the developer can still omit the generic <T> parameter. Without this parameter, there is no limit on the return type. So, the approach doesn't offer much benefit after all.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/NumberField.java
sissbruecker, line 1: Regarding:
> The PR is marked as breaking because it causes the component to start overriding any custom error messages that were set manually with the `setErrorMessage` method before validation. If you need to set error messages manually, use `setManualValidation(true)`, which prevents the component from updating the `invalid` and `errorMessage` properties, allowing you to control them manually.

Do I understand this correctly that a simple use case like:
```java
var field = new NumberField();
// set up min/max/required
field.setErrorMessage("Please enter a number between 1 and 10");
```

Now becomes:
```java
var field = new NumberField();
// set up min/max/required
var i18n = new NumberField.NumberFieldI18n()
i18n.setRequiredErrorMessage("Please enter a number between 1 and 10");
i18n.setMinErrorMessage("Please enter a number between 1 and 10");
i18n.setMaxErrorMessage("Please enter a number between 1 and 10");
field.setI18n(i18n);
```

Or, alternatively implementing required, min and max validation myself?

vursen: That's correct. We didn't think it could be a major problem since we thought it was a rare case. However, looking at your example, it seems like it could be common when you only have one error message to display.

vursen: There will be a warning in the JavaDoc of `setErrorMessage` about this behavior:

https://github.com/vaadin/flow-components/pull/6343/files#diff-0db4bfd08557d0764d0b801e103dc9e0cbbf8d9b2ed5e9602f4cb0591cb87660R31-R35

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/AbstractNumberField.java
```diff
@@ -400,4 +418,157 @@ protected void onAttach(AttachEvent attachEvent) {
...
+
+    private String getMaxErrorMessage() {
+        return Optional.ofNullable(i18n)
+                .map(AbstractNumberFieldI18n::getMaxErrorMessage).orElse("");
+    }
+
+    private String getStepErrorMessage() {
+        return Optional.ofNullable(i18n)
+                .map(AbstractNumberFieldI18n::getStepErrorMessage).orElse("");
+    }
```
DiegoCardoso, line 471: Could use the same approach as suggested in https://github.com/vaadin/flow-components/pull/6365#discussion_r1675358980

vursen: Thanks, updated.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/AbstractNumberField.java
```diff
@@ -16,7 +16,11 @@
 package com.vaadin.flow.component.textfield;
 
 import java.math.BigDecimal;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.function.Function;
+import java.io.Serializable;
```
sissbruecker, line 23: Unused import

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/AbstractNumberField.java
```diff
@@ -99,6 +107,30 @@ public AbstractNumberField(SerializableFunction<String, T> parser,
         });
     }
 
+    @Override
+    public void setErrorMessage(String errorMessage) {
```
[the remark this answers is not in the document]
vursen: Good point. I'll add this to the JavaDoc.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/AbstractNumberFieldI18n.java
```diff
@@ -0,0 +1,63 @@
...
+ * the License.
+ */
+package com.vaadin.flow.component.textfield;
+
+import java.io.Serializable;
+
+/**
+ * The internationalization properties for {@link AbstractNumberField}.
+ */
+public interface AbstractNumberFieldI18n extends Serializable {
```
DiegoCardoso, line 24: I don't think the `Abstract` prefix fits right now because it's an interface instead of an abstract class. On the other hand, I have been trying to come up with another name but wasn't able to have many ideas. Maybe `BaseNumberFieldI18n` or just `NumberFieldI18n` (which would require changing the name of its default implementation)? Or even `HasNumberFieldErrorMessages`?

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/NumberField.java
```diff
@@ -197,6 +197,182 @@ public double getStep() {
...
+     *
+     * @param i18n
+     *            the i18n object, not {@code null}
+     */
+    public void setI18n(NumberFieldI18n i18n) {
+        super.setI18n(i18n);
+    }
+
+    /**
+     * The abstract internationalization properties for {@link NumberField}.
```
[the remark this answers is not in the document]
vursen: Thanks, fixed.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/AbstractNumberField.java
vursen: Removed.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/AbstractNumberField.java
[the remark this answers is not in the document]
vursen: Done.

sissbruecker (approved):

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/AbstractNumberFieldI18n.java
vursen: When we were talking with @sissbruecker about converting the class into an interface, we also had trouble finding a good name. We concluded that `AbstractNumberFieldI18n` is better because it's more aligned with `AbstractNumberField`, which it is mainly associated with.

DiegoCardoso (approved):

PR #6360, opened 2024-06-10
feat!: support setting error messages with DatePickerI18n

vursen:
## Description

The PR introduces DatePickerI18n with methods for setting error messages and updates the DatePicker's validation logic to show these error messages when validation fails.

> [!WARNING]
> The PR is marked as breaking because it causes the component to start overriding any custom error messages that were set manually with the `setErrorMessage` method before validation. If you need to set error messages manually, use `setManualValidation(true)`, which prevents the component from updating the `invalid` and `errorMessage` properties, allowing you to control them manually.

Part of #4618 

## Type of change

- [x] Feature

### DIFF
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
@@ -496,6 +496,14 @@ public class DatePicker
             i18nObject.put("referenceDate",
                     i18n.getReferenceDate().format(DateTimeFormatter.ISO_DATE));
         }
+
+        // Remove the error message properties because they aren't used on
+        // the client-side.
+        i18nObject.remove("badInputErrorMessage");
+        i18nObject.remove("requiredErrorMessage");
+        i18nObject.remove("minErrorMessage");
+        i18nObject.remove("maxErrorMessage");
+
         // Remove properties with null values to prevent errors in web component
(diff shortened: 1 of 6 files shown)


vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
```diff
@@ -846,6 +864,26 @@ public Registration addInvalidChangeListener(
...
+
+    private String getMinErrorMessage() {
+        return Optional.ofNullable(i18n).map(DatePickerI18n::getMinErrorMessage)
+                .orElse("");
+    }
+
+    private String getMaxErrorMessage() {
+        return Optional.ofNullable(i18n).map(DatePickerI18n::getMaxErrorMessage)
+                .orElse("");
+    }
```
DiegoCardoso, line 885: Could use the same approach suggested in https://github.com/vaadin/flow-components/pull/6365#discussion_r1675358980

vursen: Thanks, updated.

vursen:
During our internal discussion, we discovered a way to keep error messages that are set with `setErrorMessage`. They now take priority over i18n error messages, which allows developers to continue using the old familiar method, setting a single error message for all constraints. I've updated the PR and also removed the breaking change mark.

DiegoCardoso (approved):

sissbruecker (approved):

PR #6365, opened 2024-06-11
feat: add i18n error message support for TimePicker

vursen:
## Description

The PR introduces TimePickerI18n with methods for setting error messages and updates the TimePicker's validation logic to show these error messages when validation fails.

> [!NOTE]

Part of #4618 

## Type of change

- [x] Feature

### DIFF
--- a/vaadin-time-picker-flow-parent/vaadin-time-picker-flow/src/main/java/com/vaadin/flow/component/timepicker/TimePicker.java
+++ b/vaadin-time-picker-flow-parent/vaadin-time-picker-flow/src/main/java/com/vaadin/flow/component/timepicker/TimePicker.java
@@ -699,20 +699,19 @@ public class TimePicker
      * <p>
      * NOTE: Updating the instance that is returned from this method will not
      * update the component if not set again using
-     * {@link TimePicker#setI18n(TimePickerI18n)}
+     * {@link #setI18n(TimePickerI18n)}
      *
-     * @return the i18n object. It will be {@code null}, If the i18n properties
-     *         weren't set.
+     * @return the i18n object or {@code null} if no i18n object has been set
      */
     public TimePickerI18n getI18n() {
         return i18n;
     }
 
     /**
-     * Sets the internationalization properties for this component.
+     * Sets the internationalization object for this component.
      *
      * @param i18n
-     *            the internationalized properties, not {@code null}
+     *            the i18n object, not {@code null}
      */
     public void setI18n(TimePickerI18n i18n) {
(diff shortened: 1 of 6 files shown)


vaadin-time-picker-flow-parent/vaadin-time-picker-flow/src/main/java/com/vaadin/flow/component/timepicker/TimePicker.java
```diff
@@ -680,4 +693,163 @@ public static Stream<Locale> getSupportedAvailableLocales() {
...
+
+    private String getMinErrorMessage() {
+        return Optional.ofNullable(i18n).map(TimePickerI18n::getMinErrorMessage)
+                .orElse("");
+    }
+
+    private String getMaxErrorMessage() {
+        return Optional.ofNullable(i18n).map(TimePickerI18n::getMaxErrorMessage)
+                .orElse("");
+    }
```
DiegoCardoso, line 739: Nit: as this pattern seems to be something used multiple times not only in this file, but in other components too, would it make sense to extract it as helper function like:

```java
private <T> String getMessageOrEmptyString(T object, Function<T, String> getter) {
    return Optional.ofNullable(object).map(getter).orElse("");
}
```

It would be called as `getMessageOrEmptyString(i18n, TimePickerI18n::getMinErrorMessage)`, for instance.

vursen: It's a good point. How about this?

```java
private String getI18nErrorMessage(
        Function<DateTimePickerI18n, String> getter) {
    return Optional.ofNullable(i18n).map(getter).orElse("");
}
```

DiegoCardoso: For using it just inside of this class, then the generic is not needed 👍🏻

vursen: Updated.

DiegoCardoso (approved):

vursen:
During our internal discussion, we discovered a way to keep error messages that are set with `setErrorMessage`. They now take priority over i18n error messages, which allows developers to continue using the old familiar method, setting a single error message for all constraints. I've updated the PR and also removed the breaking change mark.

DiegoCardoso (approved):

sissbruecker (approved):

PR #6398, opened 2024-07-03
feat: add convenience method for setting label width

### LINKED ISSUE #6223: Form: Add convenient way to set width of label from Java
### Describe your motivation

To change the default width of label in an aside-based form the usage of a CSS property in CSS or Java is required. This feels "overly" complicated for a simple Java Dev.

```java

public void setLabelAsideWidth(String width);
// or
public void setLabelWidth(String width);

```

Both methods would delegate to `this.getStyle().set("--vaadin-form-item-label-width", width);`

FrediWa:
## Description
Add a convenience method for setting label width. 

Fixes #6223

## Type of change

- [ ] Bugfix
- [X] Feature

### DIFF
--- a/vaadin-form-layout-flow-parent/vaadin-form-layout-flow/src/main/java/com/vaadin/flow/component/formlayout/FormLayout.java
+++ b/vaadin-form-layout-flow-parent/vaadin-form-layout-flow/src/main/java/com/vaadin/flow/component/formlayout/FormLayout.java
@@ -451,7 +451,7 @@ public class FormLayout extends Component
     }
 
     /**
-     * Set the width of side-positioned labels.
+     * Set the width of side-positioned label.
      * 
      * @param width
      *          the value and CSS unit as a string
@@ -463,7 +463,7 @@ public class FormLayout extends Component
     }
 
     /**
-     * Get the width of side-positioned labels.
+     * Get the width of side-positioned label.
      * 
      * @return 
      *          the value and CSS unit as a string
--- a/vaadin-form-layout-flow-parent/vaadin-form-layout-flow/src/main/java/com/vaadin/flow/component/formlayout/FormLayout.java
+++ b/vaadin-form-layout-flow-parent/vaadin-form-layout-flow/src/main/java/com/vaadin/flow/component/formlayout/FormLayout.java
@@ -449,4 +449,16 @@ public class FormLayout extends Component
         add(formItem);
         return formItem;
     }
+
+    /**
+     * Set the width of the label when aside.
+     * 
+     * @param width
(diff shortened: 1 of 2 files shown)


vaadin-form-layout-flow-parent/vaadin-form-layout-flow/src/main/java/com/vaadin/flow/component/formlayout/FormLayout.java
```diff
@@ -449,4 +449,16 @@ public FormItem addFormItem(Component field, Component label) {
         add(formItem);
         return formItem;
     }
+
+    /**
+     * Set the width of the label when aside.
```
sissbruecker, line 454:
```suggestion
     * Set the width of side-positioned labels.
```
Might be a bit clearer?

vaadin-form-layout-flow-parent/vaadin-form-layout-flow/src/main/java/com/vaadin/flow/component/formlayout/FormLayout.java
```diff
@@ -449,4 +449,16 @@ public FormItem addFormItem(Component field, Component label) {
...
+     * 
+     * @param width
+     *          the value and CSS unit
+     * @see <a href=
+     *      "https://vaadin.com/docs/latest/components/form-layout#label-position">Label position</a>
+     */
+    public void setLabelWidth(String width) {
```
sissbruecker, line 461: Should we go with the naming suggestion by Rolf? Apart from the JavaDoc it's not obvious otherwise that this only applies to specific labels / with a specific configuration. Personally I'm also fine with the current naming. @DiegoCardoso Do you have an opinion on this?

FrediWa: I intentionally went with `setLabelWidth` as I don't think `setLabelAsideWidth` is any better. I don't think `setLabelColumnWidth` is too good either as FormLayout is based around columns, and this, the way I see, has nothing to do with said columns. But you're right that the method name does not help the user realize it only applies to side-positioned labels.

DiegoCardoso: I think the name is fine and the JavaDoc is there to help. 

Also, I think a getter should be added too.

sissbruecker (approved):

DiegoCardoso (approved):

PR #6412, opened 2024-07-05
fix: do not fire ItemClickListener on focusable element click

### LINKED ISSUE #4065: Clicking on the Icon of a Button or an Icon inside a Grid's ComponentRenderer the ItemClickListener of the Grid is also called
### Description

I have a Button with only an Icon inside rendered inside a grid column. When clicking on the Icon, the Button's ClickListener and the Grid's ItemClickListener is called. If clicking on the text or the margin of the button, it works.
It was working in vaadin 14.x.

### Same issue for Icon

I could reproduce the same issue if the Button is replaced with an Icon with an added ClickListener.

### Expected outcome

Only the button's ClickListener is called.

```kotlin
            val grid = Grid<String>()
            grid.addComponentColumn {
                Button("Test", VaadinIcon.LOCK.create()) {
                    println("Button Click")
                }
//                VaadinIcon.LOCK.create().apply {
//                    addClickListener {
//                        println("Icon Click")
//                    }
//                }
            }
            grid.addItemClickListener {
                println("Item Click")
            }
            grid.setItems("ABC", "DEF")
```

### Steps to reproduce

1. Click directly on the Icon of the Button
2. `Button Click` and `Item Click` should be in the console.
3. Clicking on the button's text
4. `Button Click` is in the console.

### Environment

Vaadin version(s): 23.1.3
OS: MacOS Ventura

### Browsers

_No response_

web-padawan:
## Description

Fixes #4065

This fixes the case with "a Button with only an Icon inside" to make sure clicking the `vaadin-icon` element doesn't cause `ItemClickListener` to be fired, making it consistent with a regular `vaadin-button` element click.

## Type of change

- Bugfix

## Note

I don't think we should add special logic to stop `ItemClickListener` for `Icon` component when it's not inside a focusable element like `Button`. In fact, a standalone `Icon` should be used inside a `Button` to convey its interactive meaning.

As mentioned in https://github.com/vaadin/flow-components/issues/4065#issuecomment-2210829465, as a workaround it's possible to use `.addEventData("event.preventDefault()")` on the `Icon` click listener in this case, in order to prevent `ItemClickListener` from being fired:

```java
grid.addColumn(new ComponentRenderer<>(person -> {
    Icon icon = new Icon(VaadinIcon.PENCIL);
    icon.getElement().addEventListener("click", e -> {
    }).addEventData("event.preventDefault()");
    return icon;
})).setHeader("Icon");
```

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/pom.xml
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/pom.xml
@@ -98,6 +98,11 @@
         </dependency>
         <dependency>
             <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-icons-flow</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
             <artifactId>vaadin-testbench-core</artifactId>
             <scope>test</scope>
             <exclusions>
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridViewClickListenersPage.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridViewClickListenersPage.java
@@ -15,10 +15,12 @@
  */
 package com.vaadin.flow.component.grid.it;
 
+import com.vaadin.flow.component.button.Button;
 import com.vaadin.flow.component.grid.Grid;
 import com.vaadin.flow.component.grid.Grid.SelectionMode;
 import com.vaadin.flow.component.html.Div;
 import com.vaadin.flow.component.html.Span;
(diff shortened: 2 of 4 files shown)


web-padawan:
WTR tests failed with the following error:

```
grid connector - item click > should dispatch an item-click event
    AssertionError: expected false to be true
    + expected - actual

    -false
    +true

    at n.<anonymous> (test/grid-connector-item-click.test.ts:35:33)
grid connector - item click > should dispatch an item-click event
    AssertionError: expected false to be true
    + expected - actual

    -false
    +true

    at n.<anonymous> (test/grid-connector-item-click.test.ts:35:33)
```

And same for `item-double-click`. However they passed for me locally 🤔

sissbruecker:
> I don't think we should add special logic to stop ItemClickListener for Icon component when it's not inside a focusable element like Button. In fact, a standalone Icon should be used inside a Button to convey its interactive meaning.

Makes sense IMO 👍

sissbruecker (approved):

PR #6420, opened 2024-07-08
feat: add methods to configure Popover trigger mode

web-padawan:
## Description

Added the API to set `trigger` property on the web component and corresponding `PopoverTrigger` enum.

It is possible to use `null` or `Collections.emptySet()` to only allow opening / closing programmatically.

## Type of change

- Feature


vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
```diff
@@ -60,6 +68,48 @@ public class Popover extends Component implements HasComponents {
...
+
+        setTrigger(Set.of(PopoverTrigger.CLICK));
+    }
+
+    /**
+     * Sets the popover trigger mode describing how the overlay is opened or
+     * closed on user interaction with the target.
+     *
+     * When passing null or an empty set, the popover can only be opened and
+     * closed programmatically.
```
sissbruecker, line 80: This is to be added in a follow-up PR?

vaadin-popover-flow-parent/vaadin-popover-flow/src/test/java/com/vaadin/flow/component/popover/PopoverTest.java
```diff
@@ -65,4 +69,62 @@ public void setPosition_getPosition() {
...
         Assert.assertEquals(null, popover.getPosition());
     }
+
+    @Test
+    public void getTrigger_defaultValue_click() {
+        Assert.assertEquals(Set.of(PopoverTrigger.CLICK), popover.getTrigger());
+    }
+
+    @Test
+    public void getTriggerProperty_defaultValue_click() {
```
sissbruecker, line 79: Nit: IMO we could move raw property assertions into the respective tests that checks the property getter

vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
```diff
@@ -60,6 +68,48 @@ public class Popover extends Component implements HasComponents {
...
+            triggerAsJson = JsonSerializer.toJson(trigger.stream()
+                    .map(PopoverTrigger::toString).collect(Collectors.toSet()));
+        }
+
+        getElement().setPropertyJson("trigger", triggerAsJson);
+    }
+
+    /**
+     * Gets the popover trigger mode describing how the overlay is opened or
+     * closed on user interaction with the target.
```
sissbruecker, line 98: Let's mention what the default configuration for the trigger is.

vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
```diff
@@ -60,6 +68,48 @@ public class Popover extends Component implements HasComponents {
...
+     * Sets the popover trigger mode describing how the overlay is opened or
+     * closed on user interaction with the target.
+     *
+     * When passing null or an empty set, the popover can only be opened and
+     * closed programmatically.
+     *
+     * @param trigger
+     *            the set of trigger values to set
+     */
+    public void setTrigger(Set<PopoverTrigger> trigger) {
```
sissbruecker, line 85: Just curious if we have any other API that allows to set multiple enum values? This way of doing it is fine IMO, just checking if there is an existing approach maybe. There's theme variants, but the methods for those are for adding / removing, rather than replacing.

An alternative could be to use variable arguments, which would not require to construct a set. No need to change this immediately, just opening a discussion.

vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
web-padawan: There is already #6337 that adds API to set `opened` state.

vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
web-padawan: I'm also not sure about the proper API for this. I can think of the following instead of using the enum:

- `setOpenOnHover(true)` / `isOpenOnHover()`
- `setOpenOnFocus(true)` / `isOpenOnFocus()`
- `setOpenOnClick(true)` / `isOpenOnClick()`

This would be aligned with the following APIs:

- [`ContextMenu#setOpenOnClick()`](https://github.com/vaadin/flow-components/blob/60244a7113d203871efe5da6735927ce370d9512/vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/java/com/vaadin/flow/component/contextmenu/ContextMenuBase.java#L175)
- [MenuBar#setOpenOnHover()](https://github.com/vaadin/flow-components/blob/60244a7113d203871efe5da6735927ce370d9512/vaadin-menu-bar-flow-parent/vaadin-menu-bar-flow/src/main/java/com/vaadin/flow/component/menubar/MenuBar.java#L352)

vaadin-popover-flow-parent/vaadin-popover-flow/src/test/java/com/vaadin/flow/component/popover/PopoverTest.java
web-padawan: Updated tests.

vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
web-padawan: Removed this block so it's no longer relevant.

vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
web-padawan: Updated the PR to use this approach. We will try to get some feedback about this during DX tests.

sissbruecker (approved):

PR #6446, opened 2024-07-15
feat: add server-side modality support to Popover

web-padawan:
## Description

Mostly based on `Dialog` / `ConfirmDialog` implementation (and `ModalityIT` from `ConfirmDialog`).
The main difference is that `vaadin-popover` has `modal` property which defaults to `false`.

## Type of change

- Feature


web-padawan:
As discussed internally, server-side modality is unncessary for modal popovers due to the (typically) different purpose of modality in a popover (in that it's typically not used to prevent anything else from happening while it's open, but rather as a UX enhancement for close-on-outside-click).

It also makes sense to NOT use it because we have other components e.g. `ComboBox`, `DatePicker`, `ContextMenu` and `MenuBar` with overlays that are essentially modal but do not use server-side modality.

I will create a separate PR to just set `modal` and `withBackdrop` properties on the web component for now.

PR #6473, opened 2024-07-24
feat: add method to customize Dialog overlay ARIA role

web-padawan:
## Description

Part of https://github.com/vaadin/web-components/issues/4977

Depends on https://github.com/vaadin/web-components/pull/7582

Added `setOverlayRole()` and `getOverlayRole()` methods to `Dialog` (the logic is copied from the `Popover`).

## Type of change

- Feature

### DIFF
--- a/vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
+++ b/vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
@@ -126,6 +126,8 @@ public class Dialog extends Component implements HasComponents, HasSize,
             width = event.getWidth();
             height = event.getHeight();
         });
+
+        setOverlayRole("dialog");
     }
 
     /**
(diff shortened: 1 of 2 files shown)


vursen (approved):

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
```diff
@@ -968,6 +970,28 @@ protected void onAttach(AttachEvent attachEvent) {
         registerClientCloseHandler();
     }
 
+    /**
+     * Sets the ARIA role for the overlay element, used by screen readers.
+     *
+     * @param role
+     *            the role to set
+     */
+    public void setOverlayRole(String role) {
```
knoobie, line 979: I'm wondering if this should be an enum on the flow side with both possible values (dialog / alertdialog) to ensure the typical type safety on java side.. those values don't change often and people could theoretically always use the element API if a new value comes out of the blue and you could not update in time.

web-padawan: Good point. Maybe we could add an override for this method to Dialog and also Popover.

knoobie: That you have to decide :) While in alpha.. I would just ditch the old methods completely before you have to maintain them for decades

web-padawan: We are about to release beta so I'm afraid for now we'd have to leave these methods as they are. 
Created #6645 so that we don't forget about this idea, and maybe consider it for the next minor.

PR #6476, opened 2024-07-24
fix: allow editing items outside grid's active range

tomivirkki:
## Description

Fixes https://github.com/vaadin/flow-components/issues/6384

The current Flow Grid editor implementation has a [validation rule](https://github.com/vaadin/flow-components/blob/403b1809b5058de2537b5dc2ca979e61b228d4f3/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/editor/EditorImpl.java#L220-L223) that throws an exception when [`editItem(item)`](https://github.com/vaadin/flow-components/blob/403b1809b5058de2537b5dc2ca979e61b228d4f3/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/editor/EditorImpl.java#L130) is called with an item that is not in the active range (i.e., there is no key for it in the key mapper). However, the exception is [documented](https://github.com/vaadin/flow-components/blob/403b1809b5058de2537b5dc2ca979e61b228d4f3/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/editor/Editor.java#L133-L134) to be thrown if `"the item is not in the backing data provider"`, which is different from the item not having yet been loaded.

Due to this rule, it is not possible, for example, to instantiate a grid with 100 items and have the last one open for editing:

```java
grid.setItems(listOf100Items);
grid.getEditor().editItem(listOf100Items.get(listOf100Items.size() - 1));
```

The [referenced issue](https://github.com/vaadin/flow-components/issues/6384) documents a similar scenario where this bug occurs.

This PR fixes the issue by removing the validation rule as it does not serve a significant purpose. If the application logic attempts to programmatically edit an item that is not included in the item set of the active data provider, it is sufficient for nothing to happen, rather than having an exception thrown when editing items that just haven't been loaded yet.

## Type of change

Bugfix

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/editor/EditorImpl.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/editor/EditorImpl.java
@@ -216,11 +216,6 @@ public class EditorImpl<T> extends AbstractGridExtension<T>
             throw new IllegalStateException("Editing item " + item
                     + " failed. Item editor is already editing item " + edited);
         }
-
-        if (!getGrid().getDataCommunicator().getKeyMapper().has(item)) {
-            throw new IllegalStateException("The item " + item
-                    + " is not in the backing data provider");
-        }
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


web-padawan (approved):

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/editor/EditorImpl.java
```diff
@@ -216,11 +216,6 @@ private void validate(T item) {
             throw new IllegalStateException("Editing item " + item
                     + " failed. Item editor is already editing item " + edited);
         }
-
-        if (!getGrid().getDataCommunicator().getKeyMapper().has(item)) {
-            throw new IllegalStateException("The item " + item
-                    + " is not in the backing data provider");
-        }
```
vursen, line 223: I guess, it would ideally make a request to the data provider to check if the item exists, but not sure how easy it is to do.

vursen (approved):

PR #6480, opened 2024-07-25
refactor: extract common validation logic into ValidationController

vursen:
## Description

The PR moves common validation state and logic into a shared `ValidationController` class and updates text fields to use this new class. 

Additionally, the behavior of custom error messages has been modified to reduce the need for method overrides in components. Now, when a custom error message is cleared by setting it to `null` or an empty string, the error message no longer reverts to the last validation result automatically. Instead, it stays empty until the next validation.

Part of #4618 

## Type of change

- [x] Refactor


vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/ValidationController.java
```diff
@@ -0,0 +1,122 @@
...
+     * Validates the given value using the component's validator and sets the
+     * {@code invalid} and {@code errorMessage} properties based on the result.
+     * If a custom error message is provided with
+     * {@link #setErrorMessage(String)}, it is used. Otherwise, the error
+     * message from the validator is used.
+     * <p>
+     * The method skips validation if the manual validation mode is enabled, see
+     * {@link #setManualValidation(boolean)}.
+     */
+    public void validate(V value) {
```
sissbruecker, line 107: Nit: Document the param

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/ValidationController.java
```diff
@@ -0,0 +1,122 @@
...
+        this.component = component;
+    }
+
+    /**
+     * Sets whether manual validation mode is enabled.
+     * <p>
+     * When enabled, the {@link #validate(V)} method skips validation, allowing
+     * the invalid state and error messages to be controlled manually.
+     */
+    public void setManualValidation(boolean enabled) {
```
sissbruecker, line 36: Nit: Document the param

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/AbstractNumberField.java
```diff
@@ -57,14 +58,57 @@ public abstract class AbstractNumberField<C extends AbstractNumberField<C, T>, T
...
-    private String constraintErrorMessage;
+    private Validator<T> defaultValidator = (value, context) -> {
+        boolean hasBadInput = valueEquals(value, getEmptyValue())
+                && isInputValuePresent();
+        if (hasBadInput) {
+            return ValidationResult.error(getI18nErrorMessage(
+                    AbstractNumberFieldI18n::getBadInputErrorMessage));
+        }
+
+        if (context == null) {
```
sissbruecker, line 73: Unrelated, but this is puzzling without context (🥁). Might be worth a comment when context can be null, and why to exclude the required validator then.

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/ValidationController.java
vursen: Done.

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/ValidationController.java
vursen: Done.

vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/AbstractNumberField.java
vursen: Good point. I've added a comment.

sissbruecker (approved):

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/ValidationController.java
```diff
@@ -125,4 +68,18 @@ public void validate(V value) {
...
+
+    private void setInvalid(boolean invalid) {
+        component.setInvalid(invalid);
+    }
+
+    private void setValidationResultErrorMessage(String errorMessage) {
+        if (component.getErrorMessage() == null
+                || component.getErrorMessage().isEmpty()
+                || component.getErrorMessage()
+                        .equals(validationResultErrorMessage)) {
```
sissbruecker, line 80: It's not immediately obvious that this is used to detect whether the current error message is a custom one or from a constraint. Maybe:
- Rename `validationResultErrorMessage` to `lastValidationResultErrorMessage`
- Extract the condition into a variable with a helpful name, or add a comment explaining what this check does

vursen: Thanks, done.

sissbruecker (approved):

knoobie:
Note: This raises the problem that it is not extendable - even tho I have overwritten `validate()` I don't have access to the newly created controller because it's a private field. 

Additionally: If you change the field to protected.. the user can access it.. but he still can't overwrite it.. so IMHO. it should be an interface so that a user can just pass their own implementation of a controller to the components doing the heavy lifting.

vursen:
> Note: This raises the problem that it is not extendable - even tho I have overwritten validate() I don't have access to the newly created controller because it's a private field.

In what situations would you need access to the controller? If you override `validate()`, I'm not sure what the benefit is of reusing the validation controller...

knoobie:
The idea came from: instead of overwriting / creating a custom component - the developer could just pass a controller to all fields to reduce their implementation overhead when they have to customize it. 

`field.setValidationController(myController)` instead of overwriting multiple methods. 

Additionally the `ValidationController` is within the `component.shared` folder, making me thinking: Oh that's public API; sooo.. how do I access / customize it. 

After your comment, it feels like a totally internal "shared" component. That might be within `component.shared.internal`so that's not part of the public API users wanna customize.

vursen:
> The idea came from: instead of overwriting / creating a custom component - the developer could just pass a controller to all fields to reduce their implementation overhead when they have to customize it.

We might consider that in the future, but for now we would like to keep it internal.

> After your comment, it feels like a totally internal "shared" component. That might be within component.shared.internalso that's not part of the public API users wanna customize.

Good point! Let me move it under the internal namespace.

PR #6485, opened 2024-07-27
refactor: always set flex styles in server state, without delay

sissbruecker:
## Description

The current mechanism for setting the splitter position is problematic:
- It relies on the `setInnerComponentStyle` method
- That method either sets the `flex` style for the primary/secondary child in the server state if they exist, or executes some JS to update the style on the client, I guess in case children get added after setting the splitter position
- There's several issues with this:
  - Updating the `flex` style, regardless whether on server or client, is always delayed, as everything is wrapped into `beforeClientResponse`
  - Server-side state might not get set all 
  - Setting splitter position and adding a child in a later roundtrip does not apply the correct `flex` style to it
  - Removing children, and adding new children does not restore the `flex` style on them, so whatever splitter position is stored on the server gets out of sync with the children styles

This replaces the mechanism with a simpler one:
- The `flex` styles for primary/secondary children are stored in a state whenever the splitter position changes
- Whenever a child is added, the style is applied to it again

Based on changes from https://github.com/vaadin/flow-components/pull/6484

### DIFF
--- a/vaadin-split-layout-flow-parent/vaadin-split-layout-flow/src/main/java/com/vaadin/flow/component/splitlayout/SplitLayout.java
+++ b/vaadin-split-layout-flow-parent/vaadin-split-layout-flow/src/main/java/com/vaadin/flow/component/splitlayout/SplitLayout.java
@@ -27,13 +27,11 @@ import com.vaadin.flow.component.EventData;
 import com.vaadin.flow.component.HasSize;
 import com.vaadin.flow.component.HasStyle;
 import com.vaadin.flow.component.Tag;
-import com.vaadin.flow.component.UI;
 import com.vaadin.flow.component.dependency.JsModule;
 import com.vaadin.flow.component.dependency.NpmPackage;
 import com.vaadin.flow.component.html.Div;
 import com.vaadin.flow.component.shared.HasThemeVariant;
 import com.vaadin.flow.component.shared.SlotUtils;
-import com.vaadin.flow.internal.StateTree;
 import com.vaadin.flow.shared.Registration;
 
 /**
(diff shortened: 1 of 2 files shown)


vursen:
> That method either sets the flex style for the primary/secondary child in the server state if they exist, or executes some JS to update the style on the client, I guess in case children get added after setting the splitter position

It looks like this part wasn't added to allow children to be added later, but to support setting the splitter position when children are defined in a template. Here is a [related issue](https://github.com/vaadin/vaadin-split-layout-flow/issues/72). I tested it and the following example no longer sets the splitter position with the PR change:

```java
@Route("vaadin-split-layout/split-layout-in-lit-template")
public class SplitLayoutInLitTemplatePage extends Div {
    @JsModule("./src/split-layout-lit-wrapper.js")
    @Tag("split-layout-lit-wrapper")
    public class SplitLayoutInLitTemplate extends LitTemplate {
        @Id("split-layout")
        private SplitLayout splitLayout;

        public SplitLayoutInLitTemplate() {
            splitLayout.setSplitterPosition(30);
        }
    }

    public SplitLayoutInLitTemplatePage() {
        add(new SplitLayoutInLitTemplate());
    }
}
```

web-padawan (approved):

sissbruecker:
@vursen Thanks for catching that! I guess there is no way to avoid a JS call then. Reverting this to a draft for now, will need to take another look at this.

vursen:
Would it be possible to set the splitter position using a CSS variable? This way we wouldn't need to rely on references to the primary and secondary elements.

yuriy-fix:
@sissbruecker , would you like to proceed with PR further. For example, with trying out the idea suggested by @vursen?

sissbruecker:
I briefly looked into using a CSS variable but ran into one case where that didn't work. Closing for now.

PR #6494, opened 2024-08-02
feat: make dialogheaderfooter implement hascomponents

### LINKED ISSUE #5735: DialogHeaderFooter should implement HasComponents
### Describe your motivation

I think the Header and Footer of the Dialog should implement the HasComponents interface. It would simplify some things, also because I use Kotlin with extensions, which would then also be available for the footer and header.
Also methods like addComponentAtIndex/asFirst are missing at the moment.

Dialog Header and Footer implements HasComponents

_No response_

### Additional context

_No response_

ugur-vaadin:
## Description

This PR makes `DialogHeaderFooter` implement `HasComponents`. This introduces the following APIs to both `DialogHeader` and `DialogFooter`:

- `add(Collection<Component> components)`
- `add(String text)`
- `remove(Collection<Component> components)`
- `removeAll()`
- `addComponentAtIndex(int index, Component component`
- `addComponentAsFirst(Component component)`

NOTE:
This might be breaking if `dialog.getHeader().add(null);` is used anywhere since it is going to be ambiguous after the refactoring. However, this is unlikely in practice since it is not supported and would throw a `NullPonterException` anyway.

Fixes #5735 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
+++ b/vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
@@ -635,59 +635,40 @@ public class Dialog extends Component implements HasComponents, HasSize,
 
         @Override
         public void add(Component... components) {
-            Objects.requireNonNull(components, "Components should not be null");
-            for (Component component : components) {
-                Objects.requireNonNull(component,
-                        "Component to add cannot be null");
-                root.appendChild(component.getElement());
-            }
-            if (!isRendererCreated()) {
-                initRenderer();
-            }
+            HasComponents.super.add(components);
+            updateRendererState();
         }
 
         @Override
         public void remove(Component... components) {
-            Objects.requireNonNull(components, "Components should not be null");
-            for (Component component : components) {
-                Objects.requireNonNull(component,
-                        "Component to remove cannot be null");
-                if (root.equals(component.getElement().getParent())) {
-                    root.removeChild(component.getElement());
-                }
-            }
-            if (root.getChildCount() == 0) {
-                dialog.getElement()
-                        .executeJs("this." + rendererFunction + " = null;");
-                setRendererCreated(false);
-            }
+            HasComponents.super.remove(components);
+            updateRendererState();
         }
 
         @Override
         public void removeAll() {
-            root.removeAllChildren();
-            dialog.getElement()
-                    .executeJs("this." + rendererFunction + " = null;");
-            setRendererCreated(false);
+            HasComponents.super.removeAll();
+            updateRendererState();
         }
 
         @Override
         public void addComponentAtIndex(int index, Component component) {
-            Objects.requireNonNull(component, "Component should not be null");
-            if (index < 0) {
-                throw new IllegalArgumentException(
-                        "Cannot add a component with a negative index");
-            } else {
-                root.insertChild(index, component.getElement());
-            }
-            if (!isRendererCreated()) {
-                initRenderer();
-            }
+            HasComponents.super.addComponentAtIndex(index, component);
+            updateRendererState();
         }
 
         @Override
         public void addComponentAsFirst(Component component) {
-            this.addComponentAtIndex(0, component);
+            HasComponents.super.addComponentAsFirst(component);
+            updateRendererState();
+        }
+
+        private void updateRendererState() {
+            if (root.getChildCount() == 0) {
+                removeRenderer();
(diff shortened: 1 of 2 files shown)


vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
```diff
@@ -620,7 +620,7 @@ private DialogFooter(Dialog dialog) {
      * components will be attached to as well as the renderer function used by
      * the dialog.
      */
-    abstract static class DialogHeaderFooter implements Serializable {
+    abstract static class DialogHeaderFooter implements HasComponents {
```
sissbruecker, line 623: Looks like several more APIs from `HasComponents` (add with text, add with collection, same with remove, ...) need to be overriden to then either initialize or remove the client-side renderer.

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
```diff
@@ -681,13 +683,49 @@ public void remove(Component... components) {
...
+         */
+        @Override
+        public void addComponentAtIndex(int index, Component component) {
+            Objects.requireNonNull(component, "Component should not be null");
+            if (index < 0) {
+                throw new IllegalArgumentException(
+                        "Cannot add a component with a negative index");
+            } else {
+                root.insertChild(index, component.getElement());
+            }
```
sissbruecker, line 712: Can we not call `HasComponents.addComponentAtIndex` here to reuse it? Same with all the other methods.

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
```diff
@@ -681,13 +683,49 @@ public void remove(Component... components) {
...
+            if (!isRendererCreated()) {
+                initRenderer();
+            }
```
sissbruecker, line 715: There will be quite a bit of duplication of this. It looks like we could introduce a single method for all APIs, that both sets or removes the renderer:
- It checks if the element has children or not
- If it does, it sets the renderer unless it already exists
- If not, it removes the renderer

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
```diff
@@ -681,13 +683,49 @@ public void remove(Component... components) {
...
+        /**
+         * Adds the given component to the container at the specific index.
+         *
+         * @param index
+         *            the index, where the component will be added. The index
+         *            must be non-negative and may not exceed the children count
+         *            component – the component to add, value should not be null
+         * @param component
+         *            the component to be added.
+         */
```
sissbruecker, line 703: No need to duplicate JavaDocs at this point? Looks like we can remove all of them for overriden methods.

ugur-vaadin: Removed the JavaDocs.

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
ugur-vaadin: Extracted the logic to `updateRendererState` and it is called in all overridden methods.

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
ugur-vaadin: Reused the default methods in all overridden methods.

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
ugur-vaadin: Added the missing `HasComponents` APIs.

sissbruecker (approved):

PR #6504, opened 2024-08-06
feat: add units size base apis to uploadi18n

### LINKED ISSUE #6404: No Java API for UploadI18N sizeBase
### Describe your motivation

There is no Java API for `i18n.units.sizeBase` in `UploadI18N.Units`.
https://github.com/vaadin/web-components/blob/6ff1635f7d1f5c387047bd4d9e20a5b30817760e/packages/upload/src/vaadin-upload-mixin.js#L455-L464### 

I would like to set `i18n.units.sizeBase` as an attribute of `UploadI18N.Units`.

As a workaround, use `upload.getElement().executeJs("this.i18n.units.sizeBase=1024");`

### Additional context

_No response_

ugur-vaadin:
## Description

This PR introduces a way to set units size base in `UploadI18N`. The changes are:

- New field `sizeBase` in `Units` with getter and setter
- New constructor in `Units` including `sizeBase` as a parameter.
- New `setUnits` method in `UploadI18N` including `sizeBase` as a parameter.

Fixes #6404 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/test/java/com/vaadin/flow/component/upload/tests/UploadI18nIT.java
+++ b/vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/test/java/com/vaadin/flow/component/upload/tests/UploadI18nIT.java
@@ -114,6 +114,7 @@ public class UploadI18nIT extends AbstractUploadIT {
                 .toJson(fullTranslation);
         deeplyRemoveNullValuesFromJsonObject(fullTranslationJson);
         Map<String, String> fullTranslationMap = jsonToMap(fullTranslationJson);
+        UploadTestsI18N.OPTIONAL_KEYS.forEach(fullTranslationMap::remove);
 
         assertTranslationMapsHaveSameKeys(fullTranslationMap, translationMap);
         assertTranslationMapHasNoMissingTranslations(translationMap);
--- a/vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/test/java/com/vaadin/flow/component/upload/tests/UploadI18nIT.java
+++ b/vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/test/java/com/vaadin/flow/component/upload/tests/UploadI18nIT.java
@@ -69,6 +69,7 @@ public class UploadI18nIT extends AbstractUploadIT {
 
         UploadI18N expected = UploadTestsI18N.RUSSIAN_FULL;
         JsonObject expectedJson = (JsonObject) JsonSerializer.toJson(expected);
+        deeplyRemoveNullValuesFromJsonObject(expectedJson);
         Map<String, String> expectedMap = jsonToMap(expectedJson);
 
         assertTranslationMapsAreEqual(expectedMap, translationMap);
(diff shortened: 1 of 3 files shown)


vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/test/java/com/vaadin/flow/component/upload/tests/UploadI18nIT.java
```diff
@@ -69,6 +69,7 @@ public void testFullI18nShouldOverrideCompleteConfigurationInWebComponentPropert
 
         UploadI18N expected = UploadTestsI18N.RUSSIAN_FULL;
         JsonObject expectedJson = (JsonObject) JsonSerializer.toJson(expected);
+        deeplyRemoveNullValuesFromJsonObject(expectedJson);
```
ugur-vaadin, line 72: These two tests expect keys for null values to be present in the actual map. However, they are removed from the map when setting the i18n.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/UploadI18N.java
```diff
@@ -421,6 +421,8 @@ public static class Units implements Serializable {
         private List<String> size = Arrays.asList("B", "kB", "MB", "GB", "TB",
                 "PB", "EB", "ZB", "YB");
 
+        private Integer sizeBase;
```
vursen, line 424: I wonder why `size` needs a default value. I notice that we tend to avoid default values in i18n to keep those from the web component. But since this one has, I wonder if it would make sense for sizeBase to have it too.

ugur-vaadin: My reasoning to keep `sizeBase` without an explicit default value in the Flow counterpart was exactly that. The default value is already handled in the web component and an i18n configuration without an explicit `sizeBase` value is supported. As per why there is one for `size`, I cannot think of any good reason. The value is exactly the same as in the web component. The difference might be that it is a required parameter and the fall back is not handled as it is done with the `sizeBase`.

vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/test/java/com/vaadin/flow/component/upload/tests/UploadI18nIT.java
vursen, line 72: Please add the new property to `UploadTestsI18N.RUSSIAN_FULL` because otherwise it's not covered by these tests.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/UploadI18N.java
vursen: As discussed internally, the size property has a default value apparently due to an oversight during review when it was contributed. The correct approach is to let the web component handle the default values for i18n propertiesl

vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/test/java/com/vaadin/flow/component/upload/tests/UploadI18nIT.java
ugur-vaadin: Added the property to full translation. Updated the partial i18n test to not expect `sizeBase` to be present in the resulting translation map.

vursen (approved):

PR #6518, opened 2024-08-09
refactor: simplify Lit renderFunction

mcollovati:
Simplifies renderFunction for lit renderers, reducing the amount of code of the runtime generated function.
This change make it easier handling lit renderers for applications using strict CSP, when applying the approach described in vaadin/flow#10810.


vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/lit-renderer.ts
```diff
@@ -37,28 +37,24 @@ _window.Vaadin.setLitRenderer = (
...
-              returnChannel('${clientCallable}', itemKey, args[0] instanceof Event ? [] : [...args]);
-            }
-          }`;
-        })
-        .join('')}
-
-      render(html\`${templateExpression}\`, root)
-    }
-  `)(render, html, live, returnChannel);
+  const callablesCreator = (itemKey) => {
```
tomivirkki, line 41: Avoid `any` type: `(itemKey: string)`

vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/lit-renderer.ts
```diff
@@ -37,28 +37,24 @@ _window.Vaadin.setLitRenderer = (
 ) => {
   // Dynamically created function that renders the templateExpression
   // inside the given root element using Lit
```
tomivirkki, line 39: This comment is no longer attached to `const renderFunction`, but `const callablesCreator`. Consider relocating or removing it.

vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/lit-renderer.ts
tomivirkki, line 1: JS/TS files should be formatted with Prettier as there's a `.prettierrc` file in the project root

vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/lit-renderer.ts
```diff
@@ -37,28 +37,24 @@ _window.Vaadin.setLitRenderer = (
...
+      if (itemKey !== undefined) {
+        returnChannel(clientCallable, itemKey, args[0] instanceof Event ? [] : [...args]);
+      }
+    });
+  }
+  const fnArgs = [
+    'html', 'live', 'item', 'index',
+    ...clientCallables, `return html\`${templateExpression}\``];
+  const htmlGenerator = Function.apply(null, fnArgs);
+  const renderFunction = (root, model, itemKey) => {
```
tomivirkki, line 52: Avoid `any` type: `(root: RenderRoot, model: ItemModel, itemKey: string)`

vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/lit-renderer.ts
```diff
@@ -37,28 +37,24 @@ _window.Vaadin.setLitRenderer = (
...
+    return clientCallables.map(clientCallable => (...args) => {
+      if (itemKey !== undefined) {
+        returnChannel(clientCallable, itemKey, args[0] instanceof Event ? [] : [...args]);
+      }
+    });
+  }
+  const fnArgs = [
+    'html', 'live', 'item', 'index',
+    ...clientCallables, `return html\`${templateExpression}\``];
+  const htmlGenerator = Function.apply(null, fnArgs);
```
tomivirkki, line 51: Should we prefer `new Function(...fnArgs)` here? It also results with `htmlGenerator` having type `Function` rather than `any`

vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/lit-renderer.ts
```diff
@@ -37,28 +37,24 @@ _window.Vaadin.setLitRenderer = (
 ) => {
   // Dynamically created function that renders the templateExpression
   // inside the given root element using Lit
-  const renderFunction = Function(`
-    "use strict";
 
-    const [render, html, live, returnChannel] = arguments;
-
-    return (root, model, itemKey) => {
```
tomivirkki, line 45: These arguments are no longer available from the template expression. Especially not having `model` available is problematic, the tests in `TreeComponentColumnsIT` are failing due to this.

tomivirkki (changes requested):

vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/lit-renderer.ts
Legioth: Are those arguments part of the API contract? (Though I guess we might have to preserve them regardless for backward compatibility reasons...)

tomivirkki: [`model` is at least](https://vaadin.com/docs/latest/components/grid/renderers#accessing-model-properties). I'm not sure if anyone's using `root` or `itemKey` for anything but maybe it's safer to include them also

vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/lit-renderer.ts
tepi: Removed comment

vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/lit-renderer.ts
tepi: Thanks, added types

vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/lit-renderer.ts
tepi: Thanks, added type

vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/lit-renderer.ts
tepi: Changed as suggested

vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/lit-renderer.ts
tepi: Added arguments, hopefully it's ok now. `TreeComponentColumnsIT` seems to pass now at least

tomivirkki (approved):

vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/lit-renderer.ts
tomivirkki: Looks good. `html, live, itemKey, model, item, index` would be a nicer order for the args but not a blocker for merging.

PR #6620, opened 2024-09-06
refactor!: extract auto add behavior into controller

sissbruecker:
Extracts duplicated code for the auto-add behavior of overlays into a controller. This covers:
- ConfirmDialog
- Dialog
- LoginOverlay
- Notification

This does not cover:
- ContextMenu - Uses a different event with additional conditionals for auto-adding
- Popover - Gets auto-added / removed when setting / removing the target

The latter two could maybe be covered by extracting a lower-level controller that allows to explicitly auto-add / remove from the component. For now I'd leave them as is and maybe cover them with a future PR.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -553,7 +553,7 @@
             <dependency>
                 <groupId>org.apache.commons</groupId>
                 <artifactId>commons-lang3</artifactId>
-                <version>3.16.0</version>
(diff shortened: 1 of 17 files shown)


vursen (approved):
I would probably still consider adding `!` given that it changes the order of events in some components, but otherwise LGTM.

PR #6632, opened 2024-09-10
fix: pass id provider to keymapper also

tepi:
Fixes https://github.com/vaadin/flow/issues/19361


mshabarov (commented):
Looks good to me, however, I'd recommend to add an end-to-end test that asserts the MSCB marks all duplicates checked, when identity provider is defined.

That's the point of this fix - selection for duplicates. Before it, MSCB didn't tick duplicates.

tepi:
> Looks good to me, however, I'd recommend to add an end-to-end test that asserts the MSCB marks all duplicates checked, when identity provider is defined.
> 
> That's the point of this fix - selection for duplicates. Before it, MSCB didn't tick duplicates.

I can't test it with an IT via Testbench API since ComboBox does not differentiate on the amount of selected items if they are considered equal. Also a screenshot comparison test does not seem to work. So hopefully the unit test for checking that the id providers match is enough.

sissbruecker:
Sorry for the late review. It looks like for other scenarios updating the key mapper is handled in the data view implementations:
- [Generic data view](https://github.com/vaadin/flow-components/blob/d09867d2576fa874a566ce0e37aa186ab58b48a3/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/dataview/ComboBoxDataView.java#L97-L102)
- [List data view](https://github.com/vaadin/flow-components/blob/d09867d2576fa874a566ce0e37aa186ab58b48a3/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/dataview/ComboBoxListDataView.java#L100-L105)

As such it would make sense to add this logic to the lazy data view implementation instead. I'll prepare an alternative PR with the respective change.

sissbruecker:
Closing in favor of https://github.com/vaadin/flow-components/pull/6688

PR #6643, opened 2024-09-16
feat: add drag resize support to dashboard

ugur-vaadin:
## Description

Adds drag drop support for dashboard.

Added API:
- `Registration addItemResizeStartListener(ComponentEventListener listener)`: Adds an item resize start listener to the dashboard
- `Registration addItemResizeEndListener(ComponentEventListener listener)`: Adds an item resize end listener to the dashboard

Fixes https://github.com/orgs/vaadin/projects/70/views/1?pane=issue&itemId=79679054

Part of https://github.com/vaadin/platform/issues/6626

## Type of change

Feature

### DIFF
--- a/vaadin-dashboard-flow-parent/vaadin-dashboard-flow-integration-tests/src/test/java/com/vaadin/flow/component/dashboard/tests/DashboardDragResizeIT.java
+++ b/vaadin-dashboard-flow-parent/vaadin-dashboard-flow-integration-tests/src/test/java/com/vaadin/flow/component/dashboard/tests/DashboardDragResizeIT.java
@@ -1,0 +1,157 @@
+/**
+ * Copyright 2000-2024 Vaadin Ltd.
+ *
+ * This program is available under Vaadin Commercial License and Service Terms.
+ *
+ * See {@literal <https://vaadin.com/commercial-license-and-service-terms>} for the full
+ * license.
+ */
+package com.vaadin.flow.component.dashboard.tests;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.openqa.selenium.interactions.Actions;
+
+import com.vaadin.flow.component.dashboard.testbench.DashboardElement;
+import com.vaadin.flow.component.dashboard.testbench.DashboardWidgetElement;
+import com.vaadin.flow.testutil.TestPath;
+import com.vaadin.testbench.TestBenchElement;
+import com.vaadin.tests.AbstractComponentIT;
+
+/**
+ * @author Vaadin Ltd
+ */
+@TestPath("vaadin-dashboard/drag-resize")
+public class DashboardDragResizeIT extends AbstractComponentIT {
+
+    private DashboardElement dashboardElement;
+
+    @Before
+    public void init() {
+        open();
+        dashboardElement = $(DashboardElement.class).waitForFirst();
+    }
+
+    @Test
+    public void enlargeWidgetHorizontally_widgetIsEnlargedCorrectly() {
+        assertWidgetResized(0, 2, 1);
+    }
+
+    @Test
+    public void enlargeWidgetVertically_widgetIsEnlargedCorrectly() {
+        assertWidgetResized(0, 1, 2);
+    }
+
+    @Test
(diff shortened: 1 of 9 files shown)


vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/DashboardItemResizeEndEvent.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /**
+     * Returns the node ID of the resized item
+     *
+     * @return node ID of the resized item
+     */
+    public int getNodeId() {
+        return nodeId;
+    }
```
tomivirkki, line 63: Node ID isn't meaningful for users. It should be a reference to the Widget instance.

ugur-vaadin: Updated to return the resized `Component`. Do you think it makes sense to return a `DashboardWidget` instead? I thought that it would be a breaking change when (if?) the sections get the resizing feature.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow-integration-tests/src/test/java/com/vaadin/flow/component/dashboard/tests/DashboardDragResizeIT.java
```diff
@@ -0,0 +1,157 @@
...
+import com.vaadin.flow.component.dashboard.testbench.DashboardWidgetElement;
+import com.vaadin.flow.testutil.TestPath;
+import com.vaadin.testbench.TestBenchElement;
+import com.vaadin.tests.AbstractComponentIT;
+
+/**
+ * @author Vaadin Ltd
+ */
+@TestPath("vaadin-dashboard/drag-resize")
+public class DashboardDragResizeIT extends AbstractComponentIT {
```
tomivirkki, line 26: Some tests are failing when running locally due to `MoveTargetOutOfBoundsException: move target out of bounds`. `getDriver().manage().window().setSize(WINDOW_SIZE_LARGE);` in init would help.

But I think the problem is that the page just includes too many widgets and the dashboard not having a fixed size makes it responsive which in turn can cause test failures depending on the window size.

Are all the ITs in this file meaningful for testing the added Flow component logic? I'd suggest simplifying the test page and reducing the number of ITs to only what's necessary.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/DashboardItemResizeEndEvent.java
tomivirkki: Only widgets can be resized. Sections always take full width and their height is determined by their child widgets.

ugur-vaadin: Done, returns widgets now.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow-integration-tests/src/test/java/com/vaadin/flow/component/dashboard/tests/DashboardDragResizeIT.java
ugur-vaadin: - Reduced the number of widgets on page
- Set a larger window size
- Simplified the set of
  - enlarge horizontally
  - enlarge vertically
  - shrink horizontally
  - shrink vertically
  - enlarge both horizontally and vertically
  - shrink both horizontally and vertically

  to
  - enlarge both horizontally and vertically

  This can be interpreted as any resize and should be enough to cover the feature.

tomivirkki (approved):

PR #6649, opened 2024-09-17
Avoid creating potentially large empty arrays when clearing grid cache

### LINKED ISSUE #6648: High CPU usage when Grid page size is set to a large value
### Description

Setting the page size of a Grid to a very large value can lead to browser side hangs even if the data provider does not contain many items.

### Expected outcome

I would expect the performance of the Grid to be proportional to the actual number of items in the data provider rather than the potential number of items in a page.

```java
Grid g = new Grid();
... add a data provider with a small number of items
g.setAllRowsVisible(true);
g.setPageSize(Integer.MAX_VALUE / 10);
```

### Steps to reproduce

Add the above snippet to a view

I'm not sure what the requirement is to trigger the code path, but you want to reach line https://github.com/vaadin/flow-components/blob/main/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts#L513 with items being undefined.

This will cause an empty Array to be created of size `pageSize` which then needs to be iterated over. If `pageSize` is some huge value this will cause the browser event loop to hang for a while.

### Environment

Vaadin version(s): 24.4.11
OS: macOS

### Browsers

Issue is not browser related

pepijnve:
## Description

This PR adds a more efficient code path in gridConnector.ts that avoids creating potentially large Arrays in order to clear values in the grid cache. This change depends on https://github.com/vaadin/web-components/pull/7808

Fixes #6648

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
@@ -510,7 +510,11 @@ import { GridFlowSelectionColumn } from "./vaadin-grid-flow-selection-column.js"
             // Update the items in the grid cache or set an array of undefined items
             // to remove the page from the grid cache if there are no corresponding items
             // in the connector cache.
-            gridCache.setPage(page, items || Array.from({ length: grid.pageSize }));
+            if (items) {
+              gridCache.setPage(page, items);
+            } else {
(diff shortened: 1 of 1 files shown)


vursen (changes requested):
Thank you for this contribution. I've requested more details in the ticket to help us reproduce this issue and subsequently create a test case.

vursen:
Closing due to the lack of a reproduction example. Feel free to reopen the ticket if the issue is still relevant.

PR #6655, opened 2024-09-18
fix: change number formatting with comma as decimal separator

### LINKED ISSUE #5731: Spreadsheet:  Incorrect number formatting in formula - locale and precision is lost

DiegoCardoso:
## Description

Change the number formatting configuration for locales that use commas as decimal separators. Before, the fraction part of the numbers in such locales would allow only 3 digits. This change allows more digits in the fractional part.

Fixes #5731

## Type of change

- [X] Bugfix
- [ ] Feature


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/FormulaFormatter.java
```diff
@@ -280,6 +280,7 @@ protected DecimalFormat getDecimalFormat(Locale locale) {
         DecimalFormat instance = (DecimalFormat) DecimalFormat
                 .getInstance(locale);
         instance.setGroupingUsed(false);
+        instance.setMaximumFractionDigits(15);
```
vursen, line 283: If I understand correctly, this setting should match the number of fraction digits specified in `EXCEL_FORMULA_BAR_DECIMAL_FORMAT`. However, that format allows for 16 digits, not 15.

https://github.com/vaadin/flow-components/blob/0a86fa9cfaaf0d4cfd6517ca4489f12537cdae8b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/CellValueManager.java#L81

DiegoCardoso: Good catch. Aligned the number of decimal digits with the one used by the `CellValueManager` class.

vursen (approved):

PR #6669, opened 2024-09-24
fix: remove default login listener when setting form action

mcollovati:
## Description

Setting both login listeners and form action can cause unexpected behaviors because of concurrent processing of form submission processing and login event.
For example, if form submission ends in a session ID change and a redirect to a different page, the UIDL request may fail with a session expiration response, causing the Flow client to reload the page and potentially cancel the ongoing redirection.
In addition, after from submission, the login event would be sent to a dismissed UI.

This change removes the default login listener when setting a form action, and logs a warning when setting both an action and custom login listeners.

References vaadin/flow#12640

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/AbstractLogin.java
+++ b/vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/AbstractLogin.java
@@ -124,7 +124,7 @@ public abstract class AbstractLogin extends Component implements HasEnabled {
                 registration.remove();
                 registration = null;
             }
-            checkActionAttributeUsedWithLoginListeners();
+            warnIfActionAndLoginListenerUsedTogether();
         }
     }
 
@@ -254,7 +254,7 @@ public abstract class AbstractLogin extends Component implements HasEnabled {
             ComponentEventListener<LoginEvent> listener) {
         Registration registration = ComponentUtil.addListener(this,
                 LoginEvent.class, listener);
-        checkActionAttributeUsedWithLoginListeners();
+        warnIfActionAndLoginListenerUsedTogether();
         return registration;
     }
 
@@ -312,7 +312,7 @@ public abstract class AbstractLogin extends Component implements HasEnabled {
         getElement().setProperty(PROP_DISABLED, !enabled);
     }
 
-    private void checkActionAttributeUsedWithLoginListeners() {
+    private void warnIfActionAndLoginListenerUsedTogether() {
         if (getElement().hasProperty(PROP_ACTION)
                 && !getListeners(LoginEvent.class).isEmpty()) {
             LoggerFactory.getLogger(getClass()).warn(
--- a/vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/AbstractLogin.java
+++ b/vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/AbstractLogin.java
@@ -24,6 +24,7 @@ import com.vaadin.flow.component.EventData;
 import com.vaadin.flow.component.HasEnabled;
 import com.vaadin.flow.component.Synchronize;
 import com.vaadin.flow.dom.DisabledUpdateMode;
+import com.vaadin.flow.dom.DomListenerRegistration;
 import com.vaadin.flow.dom.PropertyChangeListener;
 import com.vaadin.flow.internal.JsonSerializer;
 import com.vaadin.flow.shared.Registration;
@@ -40,6 +41,17 @@ import elemental.json.JsonValue;
  * {@link com.vaadin.flow.component.HasEnabled#setEnabled(boolean)} method.
  * Setting error {@link #setError(boolean)} true makes component automatically
  * enabled for the next login attempt.
+ * <p>
+ * </p>
+ * Server side login listener do not work in combination with HTML form
+ * submission configured by setting the {@code action} attribute. The reason is
+ * that form submission, depending on the authentication process result, will
+ * cause a redirection to a different page or to current login view. In both
+ * cases a new Flow UI will be created and the event will potentially be
+ * forwarded to a dismissed UI. In addition, if the HTTP session ID is changed
+ * as a consequence of the authentication process, the server may respond to the
+ * login event with a session expiration error, cause a client resynchronization
+ * that can in turn cancel a potential redirect issued by the form submission.
  *
  * @author Vaadin Ltd
  */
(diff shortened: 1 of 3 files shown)


yuriy-fix:
@mcollovati , are there any further actions planned on this one or is it ready for the review?

vaadin-login-flow-parent/vaadin-login-flow/src/test/java/com/vaadin/flow/component/login/LoginFormTest.java
```diff
@@ -102,4 +103,73 @@ public void showErrorMessage_preservesExistingI18n() {
...
+    public void setAction_nullValue_restoreDefaultListener() {
+        final LoginForm form = new LoginForm();
+        form.setAction("login");
+        form.setError(true);
+
+        ComponentUtil.fireEvent(form, new AbstractLogin.LoginEvent(form, true,
+                "username", "password"));
+        Assert.assertTrue(form.isEnabled());
+        Assert.assertTrue(form.isError());
+    }
```
sissbruecker, line 160: This seems to do the same as the test above in the first step, also the test name doesn't match the test setup. It looks like this one can be removed?

vaadin-login-flow-parent/vaadin-login-flow/src/test/java/com/vaadin/flow/component/login/LoginOverlayTest.java
```diff
@@ -57,4 +60,73 @@ public void showErrorMessage_preservesExistingI18n() {
...
+                true, "username", "password"));
+        Assert.assertFalse("Expected form being disabled by default listener",
+                overlay.isEnabled());
+        Assert.assertFalse(
+                "Expected error status being reset by default listener",
+                overlay.isError());
+    }
+
+    @Test
+    public void setAction_nullValue_restoreDefaultListener() {
```
sissbruecker, line 108: Same here.

vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/AbstractLogin.java
sissbruecker, line 1: Looks like there are some potential changes in behavior:
- The component will now throw at runtime when you configure both action and login listener
- The component does not disable itself anymore on login when using `action`. It sounds like that may or may not have worked previously.

It feels like this should be treated as a behavior altering fix / breaking change, as such I'm not sure if this can be back-ported. Personally I'd be fine with merging this for 24.6, but if we want to ship this in a maintenance release we might need an alternative solution. Looking at the Flow ticket this seems to be a common issue that affects a lot of people, so would be unfortunate if we can only ship this in a new minor.

sissbruecker: Some random ideas:
- Simply ignore adding login listeners as long as an action is set, log an error/warning instead of throwing.
- Maybe handle disabling the component on the client-side when using an `action`

vaadin-login-flow-parent/vaadin-login-flow/src/test/java/com/vaadin/flow/component/login/LoginFormTest.java
mcollovati: Which other test are you referring to?
The test checks that you cannot add a listener if the action is set, and that removing the action you can add a listener again.

mcollovati: But I agree that the test is poorly named :grin:

mcollovati: Ah, wait, is the comment about `setAction_nullValue_restoreDefaultListener`?
I'm talking about `addLoginListener_actionSet_throws`

mcollovati: If it is about `setAction_nullValue_restoreDefaultListener`, you're right. The test can be removed

sissbruecker: Yes, `setAction_nullValue_restoreDefaultListener` seems to do the same as the test above it.

mcollovati: removed

vaadin-login-flow-parent/vaadin-login-flow/src/test/java/com/vaadin/flow/component/login/LoginOverlayTest.java
mcollovati: removed

vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/AbstractLogin.java
mcollovati: I don't have strong opinions about that.
Please, feel free to take over the PR and apply the proper changes

yuriy-fix:
@mcollovati , we will overtake the PR and try to come up with a way of not introducing breaking changes.

vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/AbstractLogin.java
sissbruecker: Changed it so that it only logs a warning for now when you configure both an action and login listeners.

Regarding disabling the component on login, that is already implemented by the web component. Since using an action implies that there will be a redirect with full page reload that creates a new UI, we can ignore the fact that the server-side component is not being disabled anymore.

vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/AbstractLogin.java
```diff
@@ -267,4 +311,12 @@ public ForgotPasswordEvent(AbstractLogin source, boolean fromClient) {
     public void onEnabledStateChanged(boolean enabled) {
         getElement().setProperty(PROP_DISABLED, !enabled);
     }
+
+    private void checkActionAttributeUsedWithLoginListeners() {
```
yuriy-fix, line 315: Nit (optional): This might have a more descriptive name, for example, `warnIfActionAndLoginListenerUsedTogether`

yuriy-fix (approved):

vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/AbstractLogin.java
sissbruecker: Done

Issue #6652, opened 2024-09-17
Remove TabSheetVariant.LUMO_TABS_ICON_ON_TOP


web-padawan:
### Description

Originally reported at the forum: https://vaadin.com/forum/t/tabsheet-theme-variants-seem-to-be-ignored/167106

The `vaadin-tabsheet` web component delegates `theme` attribute to the underlying `vaadin-tabs` which makes most of theme variants just work (e.g. `small`, `minimal`, `equal-width-tabs` etc).

However, the `icon-on-top` variant is expected to be set on individual `vaadin-tab` instances, which isn't handled by `vaadin-tabsheet` and therefore the `TabSheetVariant.LUMO_TABS_ICON_ON_TOP` doesn't work as expected.

UPD: actually, the same value exists in `TabsVariant` since the very old version https://github.com/vaadin/vaadin-tabs-flow/pull/80 and the `TabSheetVariant` was most probably just a copy-paste from there 🙈 
 
### Expected outcome

Expected the theme variant to work.

### Minimal reproducible example

```java
tabSheet.addThemeVariants(TabSheetVariant.LUMO_TABS_ICON_ON_TOP);
```

### Steps to reproduce

Use the TabSheet with several tabs with icons and set the variant.

### Environment

Vaadin version(s): 24.x and older

### Browsers

Issue is not browser related

web-padawan:
Looks like the only way to make this work out of the box is to override most of methods from `HasTheme` in `Tabs` to automatically set `icon-on-top` to all individual tabs when it's added, or remove it from individual tabs when it's removed. 

Alternatively we could also make the web component handle this, but we generally prefer to avoid delegating `theme` to light DOM children especially when they are created from the markup to avoid potential conflicts with user-defined values.

There's already an [icons example](https://vaadin.com/docs/latest/components/tabs#icons) in Tabs documentation that demonstrates how to do this, and same can be applied to `TabSheet` whose `add()` method returns `Tab` instance:

```java
// Set the icon on top
for (Tab tab : new Tab[] { profile, settings, notifications }) {
    tab.addThemeVariants(TabVariant.LUMO_ICON_ON_TOP);
}
```

IMO we should instead deprecate this variant in `TabsVariant` and `TabSheetVariant` instead of trying to get it worked.

PR #6686, opened 2024-10-01
chore: deprecate ICON_ON_TOP Tabs and TabSheet variants

web-padawan:
## Description

Related to #6652

Marked these incorrectly added variants as deprecated with the suggestion to use `TabVariant` instead:

Screenshot 2024-10-01 at 14 16 12

## Type of change

- Deprecation

### DIFF
--- a/vaadin-tabs-flow-parent/vaadin-tabs-flow/src/main/java/com/vaadin/flow/component/tabs/TabSheetVariant.java
+++ b/vaadin-tabs-flow-parent/vaadin-tabs-flow/src/main/java/com/vaadin/flow/component/tabs/TabSheetVariant.java
@@ -22,6 +22,10 @@ import com.vaadin.flow.component.shared.ThemeVariant;
  */
 public enum TabSheetVariant implements ThemeVariant {
     //@formatter:off
+    /**
+     * @deprecated Use {@code TabVariant.LUMO_ICON_ON_TOP} instead.
+     */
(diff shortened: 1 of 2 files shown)


vaadin-tabs-flow-parent/vaadin-tabs-flow/src/main/java/com/vaadin/flow/component/tabs/TabSheetVariant.java
```diff
@@ -22,6 +22,10 @@
  */
 public enum TabSheetVariant implements ThemeVariant {
     //@formatter:off
+    /**
+     * @deprecated Use {@code TabVariant.LUMO_ICON_ON_TOP} instead.
```
sissbruecker, line 26: This is potentially confusing as it could be interpreted to use `TabVariant` together with `TabSheet` or `Tabs`, which doesn't work. Let's add that it should be used on `Tab` instances instead.

web-padawan: Thanks, done.

sissbruecker:
@web-padawan Let's merge this to fix the formatter issues: https://github.com/vaadin/flow-components/pull/6700

sissbruecker (approved):

PR #6706, opened 2024-10-09
feat: add dashboard component

ugur-vaadin:
## Description

This PR merges the `Dashboard` component feature branch `feat/dashboard` to `main`.

## Type of change

Feature


vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/Dashboard.java
```diff
@@ -0,0 +1,1060 @@
...
+/**
+ * @author Vaadin Ltd
+ */
+@Tag("vaadin-dashboard")
+@NpmPackage(value = "@vaadin/polymer-legacy-adapter", version = "24.6.0-alpha2")
+@JsModule("@vaadin/polymer-legacy-adapter/style-modules.js")
+@JsModule("@vaadin/dashboard/src/vaadin-dashboard.js")
+@JsModule("./flow-component-renderer.js")
+@NpmPackage(value = "@vaadin/dashboard", version = "24.6.0-alpha2")
+public class Dashboard extends Component implements HasWidgets, HasSize {
```
sissbruecker, line 49: How would I retrieve the current dashboard config, so that I can persist it somewhere and then later restore it?

In the web component you can just take whatever is in `items`, serialize it to JSON, and then later restore `items` from that JSON. How would I implement something like this in Flow?

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/Dashboard.java
```diff
@@ -0,0 +1,1060 @@
...
+    }
+
+    /**
+     * Adds an item moved listener to this dashboard.
+     *
+     * @param listener
+     *            the listener to add, not <code>null</code>
+     * @return a handle that can be used for removing the listener
+     */
+    public Registration addItemMovedListener(
```
sissbruecker, line 342: How do you move a widget or section programmatically?

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/Dashboard.java
```diff
@@ -0,0 +1,1060 @@
...
+                        widget.getElement().getNode().getId());
+    }
+
+    private void doRemoveAll() {
+        new ArrayList<>(childrenComponents).forEach(child -> {
+            if (child instanceof DashboardSection section) {
+                doRemoveSection(section);
+            } else {
+                doRemoveWidget((DashboardWidget) child);
+            }
```
sissbruecker, line 564: Looks like there is no functional difference between removing sections or items, so a single method that can handle both seems enough.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/DashboardItemMoveModeChangedEvent.java
```diff
@@ -0,0 +1,70 @@
...
+import com.vaadin.flow.component.EventData;
+
+/**
+ * Widget or section move mode state changed event of {@link Dashboard}.
+ *
+ * @author Vaadin Ltd.
+ * @see Dashboard#addItemMoveModeChangedListener(ComponentEventListener)
+ */
+@DomEvent("dashboard-item-move-mode-changed")
+public class DashboardItemMoveModeChangedEvent
```
sissbruecker, line 24: Took me a while to understand what this does. I'm not sure what the use case is?

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/Dashboard.java
ugur-vaadin: Combined both under the name `doRemoveItem`.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/Dashboard.java
ugur-vaadin: Currently, the way to do it would be using add/remove/get methods in combination. Programatically reordering items after initializing them probably is not that common of a use case. Do you have an suggestion for such an API? Pinging @tomivirkki @yuriy-fix @rolfsmeds for visibility.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/Dashboard.java
ugur-vaadin: We have discussed this a lot, and the conclusion was to leave it to the user to choose how to represent their data for serializing/storing. It would require getting the ordered root items through `getChildren` and getting the widgets in sections through those.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/DashboardItemMoveModeChangedEvent.java
ugur-vaadin: Those listeners are mostly there to cover some accessibility requirements. I would expect not many users to use this otherwise.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/Dashboard.java
```diff
@@ -0,0 +1,1060 @@
...
+     * @param section
+     *            the widgets to add, not {@code null}
+     */
+    public void addSection(DashboardSection section) {
+        doAddSection(section);
+        updateClient();
+    }
+
+    @Override
+    public List<DashboardWidget> getWidgets() {
```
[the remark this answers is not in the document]
ugur-vaadin: The `getWidgets` method is mostly there for utility, presuming users would be mostly interested in only the widgets. Making them loop over `getChildren, checking whether they are instances of `DashboardWidget` and then casting every time they wanted to get the widgets would also be cumbersome. 

This structure in general has also been discussed a lot going back and forth. One option was to go with something like a `DashboardItem` (naming this was also an issue) interface, which would have unified the API and structure. That layer would also add another concept that wouldn't be very intuitive to the user. We decided to go with the option that covers the most common use cases. But if you have any suggestions, let's discuss those as well.

Again, pinging @tomivirkki @yuriy-fix @rolfsmeds for visibility.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/DashboardItemMovedEvent.java
```diff
@@ -0,0 +1,84 @@
...
+
+import java.util.List;
+import java.util.Optional;
+
+import com.vaadin.flow.component.Component;
+import com.vaadin.flow.component.ComponentEvent;
+import com.vaadin.flow.component.ComponentEventListener;
+
+/**
+ * Widget or section moved event of {@link Dashboard}.
```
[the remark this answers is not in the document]
ugur-vaadin: Updated the event descriptions.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/Dashboard.java
```diff
@@ -0,0 +1,1060 @@
...
+import com.vaadin.flow.dom.Element;
+import com.vaadin.flow.internal.JsonSerializer;
+import com.vaadin.flow.shared.Registration;
+
+import elemental.json.JsonArray;
+import elemental.json.JsonObject;
+import elemental.json.JsonType;
+
+/**
+ * @author Vaadin Ltd
```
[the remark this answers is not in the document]
ugur-vaadin: Added descriptions to the component classes and `HasWidgets` interface.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/Dashboard.java
sissbruecker: We might want to let someone DX test this and see what they think. Apart from that I'd suggest to add an example of this to the docs if it isn't already planned.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/Dashboard.java
sissbruecker: It looks like you could move a widget, but you can not move a section because there is no way to insert it at an index.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/DashboardItemMoveModeChangedEvent.java
sissbruecker: OK thanks

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/Dashboard.java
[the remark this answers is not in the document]
sissbruecker: If we wan't to keep the API as is then let's at least update JavaDocs to mention:
- `Dashboard.removeAll` also removes sections
- `HasWidgets.addWidgetAtIndex` should be used with `getChildren` to determine the index of an existing widget or section, as `getWidgets` can not be used for that
- `Dashboard.getWidgets` returns nested widgets from sections too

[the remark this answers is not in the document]
ugur-vaadin: Updated JavaDocs for `HasWidgets.addWidgetAtIndex` and `Dashboard.getWidgets`. `Dashboard.removeAll` already mentions that it also removes the sections.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/Dashboard.java
ugur-vaadin: It's still possible but is a little cumbersome and would require a few steps:
- getting the children
- reordering them
- removing all from dashboard
- adding them one by one again (which also needs casting)

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/Dashboard.java
ugur-vaadin: I added a task in the DX tests that require this. I agree that we should add good examples for all these cases to the docs if we keep this API design after the DX tests.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow-integration-tests/pom.xml
```diff
@@ -0,0 +1,162 @@
...
+            <activation>
+                <property>
+                    <name>!skipJetty</name>
+                </property>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.eclipse.jetty</groupId>
+                        <artifactId>jetty-maven-plugin</artifactId>
```
sissbruecker, line 153:
```suggestion
                        <groupId>org.eclipse.jetty.ee10</groupId>
                        <artifactId>jetty-ee10-maven-plugin</artifactId>
```
There was an update to the jetty version in parallel to this PR, this should align it.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/Dashboard.java
sissbruecker: OK, could be added later I guess.

sissbruecker (approved):

Issue #6662, opened 2024-09-20
TreeGrid drag and drop causes complete data refresh/reset each time drag is started


gektron:
### Description

Every time a drag-drop operation is performed a complete data reset is performed - found by looking at what happens when TreeGrid.setDropMode() is called in your code. The problem with the reset is that the contents of the TreeGrid are scrolled back to the top which is a terrible experience for the user when operating on a larger set of data with an active scrollbar.

### Expected outcome

Do not reset the data or at least maintain the scroll position and the selected item.

If I omit calling setDropMode() from the TreeGrid.addDragStartListener() and instead call it once at the point of creating the TreeGrid, then the drag-drop works once but then starts failing so it looks as though it needs calling each time a drag is started.

I’ve tried using the TreeGrid.scrollToIndex() function but couldn’t get it to work (the example in your cookbook doesn’t seem to work for a TreeGrid). Also it wouldn’t really fix the issue because the reset and hence the scroll to top occurs once at the drag start and again at the drag end (TreeGrid.setDropMode() called on both these occasions).

### Minimal reproducible example

Using this code (from your documentation) performs a reset each time.
`
        treeGrid.addDragStartListener(e -> {
            treeGrid.setDropMode(GridDropMode.ON_TOP_OR_BETWEEN);
            dragItem = e.getDraggedItems().get(0);
        });
`
`
        treeGrid.addDragEndListener(e -> {
            treeGrid.setDropMode(null);
            dragItem = null;
        });
`

### Steps to reproduce

Implement the above code with a large data set ensuring the TreeGrid has an active scrollbar.

### Environment

Vaadin version(s): 24
OS: MacOS Sonoma

### Browsers

Firefox, Safari

PR #6712, opened 2024-10-11
fix: do not trigger data communicator reset when no filters

vursen:
The `setDropMode` and `setRowsDraggable` methods were [updated](https://github.com/vaadin/flow-components/pull/6339) to call `dataCommunicator.reset` to ensure drop filters are applied when the drop mode is set dynamically. While this fixed the original issue, it introduced an unintended side effect: an annoying scroll position reset that now happens e.g. when setting the drop mode dynamically on drag start. 

Unfortunately, resolving this new issue completely is impossible without introducing new API in `DataCommunicator`. An API that would allow us to request `refreshItem` for only items that are in the viewport. 

Since not working filters appear to be a more critical issue, it was decided to keep the original change but try to mitigate its impact by preventing the data reset when no filters are set.

Fixes https://github.com/vaadin/flow-components/issues/6662


tomivirkki (approved):

vursen:
I created https://github.com/vaadin/flow-components/issues/6713 in which I described a solution that could resolve this issue.

gektron:
Thank you so much for implementing a fix. What would be the release schedule for this? Will it make 24.5?

vursen:
@gektron Yes, this fix will be included in the upcoming 24.5 release.

Issue #6579, opened 2024-08-27
Select component uses old instance to refresh item


heruan:
### Description

When Select reacts to a `DataChangeEvent` (i.e. `DataProvider.refresh(item)`) the callback does not take into account the new item instance but it just re-render the existing instance.

See in https://github.com/vaadin/flow-components/blob/700058106ddc1e9792016dc98e9d1c49e8410b9a/vaadin-select-flow-parent/vaadin-select-flow/src/main/java/com/vaadin/flow/component/select/Select.java#L1017-L1030

The `updatedItem` is only used to find the corresponding existing `vaadinItem` and then the `updateItem` is invoked with the existing `vaadinItem` instance, so that if anything has changed in `updatedItem` it’s just ignored.

### Expected outcome

Upon `DataChangeEvent` the new data is used to update the item.

### Minimal reproducible example

```java
var select = new Select("Test");
select.setDataProvider(dp);
dp.refreshItem(newItem);
```

### Steps to reproduce

The data provider needs to override the `getId` method so that the `newItem` instance matches an existing item, but the update still uses the old instance.

### Environment

Vaadin version(s): 24.5.0.alpha10

yuriy-fix:
Might need to be verified in e.g. `ComboBox`

MatthewVaadin:
There is possibly another issue with the `onDataChange` method. When a `refreshAll()` is called on the data provider, the select is reset (the `else` branch), which includes clearing the selected value. I would expect the selected value to remain selected if the value still exists. This is what happens with a combo box.

See minimal example here: https://github.com/MatthewVaadin/refresh-select

Steps to reproduce:
1. Select an option in both the select (above) and the combo box (below).
2. Click the "Refresh items" button.
3. Observe that the value is cleared in the select, but not in the combo box.

PR #6714, opened 2024-10-11
fix: allow replacing item instances with `refreshItem`

sissbruecker:
When refreshing a single item in a `Select` data provider, the component currently assumes that the existing item instance has been updated, but does not handle replacing item instances. However the `refreshItem` JavaDoc explicitly mentions that use case:
> Refreshes the given item. This method should be used to inform all DataProviderListeners that an item has been updated or replaced with a new instance.

This change makes `Select` handle replacing item instances as well.

Fixes https://github.com/vaadin/flow-components/issues/6579

### DIFF
--- a/vaadin-select-flow-parent/vaadin-select-flow/src/test/java/com/vaadin/flow/component/select/data/SelectDataViewTest.java
+++ b/vaadin-select-flow-parent/vaadin-select-flow/src/test/java/com/vaadin/flow/component/select/data/SelectDataViewTest.java
@@ -63,10 +63,9 @@ public class SelectDataViewTest extends AbstractComponentDataViewTest {
 
         dataView.setIdentifierProvider(Item::getId);
 
-        first.setValue("changed-1");
         second.setValue("changed-2");
 
-        dataProvider.refreshItem(new Item(1L));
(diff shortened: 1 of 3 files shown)


vaadin-select-flow-parent/vaadin-select-flow/src/test/java/com/vaadin/flow/component/select/data/SelectDataViewTest.java
```diff
@@ -63,10 +63,9 @@ public void setIdentifierProvider_customIdentity_itemRefreshed() {
 
         dataView.setIdentifierProvider(Item::getId);
 
-        first.setValue("changed-1");
         second.setValue("changed-2");
 
-        dataProvider.refreshItem(new Item(1L));
+        dataProvider.refreshItem(new Item(1L, "changed-1"));
```
sissbruecker, line 68: The previous test setup here seems flawed. Passing in a new instance should not update from the previous one. It also seems unlikely that someone would update the existing instance and then, instead of passing that same instance to `refreshItem`, pass in a new one that only matches the identity to get the corresponding item to update. So I updated the setup to cover the new change instead.

vursen (approved):

Issue #6710, opened 2024-10-10
Client side Grid selection is preserved if deselectAll is used while switching between SelectionMode.NONE and SelectionMode.MULTI


mukherjeesudebi:
### Description

After selecting a few items and switching the selection mode from SelectionMode.MULTI to SelectionMode.NONE , calling deselectAll removes all items from selection i.e grid.getSelectedItems().size() is 0, but the client side is still preserved and the rows still remain highlighted and if the selection is toggled back to  SelectionMode.MULTI then the checkboxes appear and they are still checked but grid.getSelectedItems().size() is 0 

The same issue does not happen if the toggling is done between SelectionMode.MULTI to SelectionMode.SINGLE
In this case everything works fine. 

### Expected outcome

The checkboxes should not be checked and neither should the rows be shown as selected when deselectAll is called after toggling the  SelectionMode from MULTI  to NONE and back

### Minimal reproducible example

The view 
```
@Route("grid-toggle-issue")
public class GridToggleIssue extends VerticalLayout {

	public GridToggleIssue() {

		Grid<GridItem> grid = new Grid<>();
		grid.addColumn(GridItem::getVal);
		grid.setItems(List.of(new GridItem("A"), new GridItem("B"), new GridItem("C")));
		grid.setSelectionMode(SelectionMode.NONE);
		
		Button toggleSelectButton = new Button("toggle select");
		toggleSelectButton.addClickListener(e -> {
			if (Grid.SelectionMode.NONE.equals(grid.getSelectionMode())) {
				grid.setSelectionMode(Grid.SelectionMode.MULTI);
				grid.deselectAll();
			} else {
				grid.deselectAll();
				grid.setSelectionMode(Grid.SelectionMode.NONE);
			}
		});

		Button showSelected = new Button("Show selected", event -> {
			Notification.show("Selected items size : " + grid.getSelectedItems().size());
		});

		add(grid, toggleSelectButton, showSelected);
	}
}
```
The GridItem class
```
public class GridItem {
	private String val;

	public GridItem(String val) {
		super();
		this.val = val;
	}

	public String getVal() {
		return val;
	}

	public void setVal(String val) {
		this.val = val;
	}

	@Override
	public int hashCode() {
		return Objects.hash(val);
	}

	@Override
	public boolean equals(Object obj) {
		if (this == obj)
			return true;
		if (obj == null)
			return false;
		if (getClass() != obj.getClass())
			return false;
		GridItem other = (GridItem) obj;
[...]

PR #6720, opened 2024-10-11
fix: clear client-side selection when changing selection mode

sissbruecker:
Fixes https://github.com/vaadin/flow-components/issues/6710


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
```diff
@@ -901,6 +901,7 @@ import { GridFlowSelectionColumn } from "./vaadin-grid-flow-selection-column.js"
           if ((typeof mode === 'string' || mode instanceof String) && validSelectionModes.indexOf(mode) >= 0) {
             selectionMode = mode;
             selectedKeys = {};
+            grid.selectedItems = [];
```
sissbruecker, line 904: Both `AbstractGridMultiSelectionModel` and `AbstractGridSingleSelectionModel` run a `deselectAll` when they are removed from the grid:
https://github.com/vaadin/flow-components/blob/0c2a684c5323276e384c14ce31f85f46c13c55d6/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridMultiSelectionModel.java#L97-L104
https://github.com/vaadin/flow-components/blob/0c2a684c5323276e384c14ce31f85f46c13c55d6/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridSingleSelectionModel.java#L174-L178

That also removes client-side selection when switching between those two. However, when switching to the none selection mode, the connector prevents the selection from being modified: https://github.com/vaadin/flow-components/blob/0c2a684c5323276e384c14ce31f85f46c13c55d6/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts#L140-L143

Rather than add more branches to that method I just reset the selected items when changing the mode. Also decided to keep the `deselectAll` calls, technically the selection model instances are exposed with having only protected visibility and someone might rely on that behavior.

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridChangeSelectionModeIT.java
vursen, line 1: Is there any reason not to use web-test-runner to test this? It seems like a good candidate.

sissbruecker: You mean in addition to the IT or as an alternative?

vursen: As an alternative

sissbruecker: OK, replace the IT with a connector test.

vursen (approved):

PR #6732, opened 2024-10-18
refactor: remove tryCatchWrapper for better stack trace

vursen:
## Description

Why tryCatchWrapper is problematic:

- It suppresses exceptions in favour of console error messages. However, the stack trace that the browser displays with these messages omits the function call that caused the original exception, making it much less helpful
- It introduces an extra layer of code nesting, which has a negative impact on the readability and maintainability
- Its error coverage is still limited e.g. it doesn't catch errors in asynchronous contexts created by the web component, where no connector methods are invoked
- It pollutes the stack trace with its own calls, which can be distracting when debugging
- It must be applied to every function including setTimeout callbacks, which is easy to miss

We have decided that the drawbacks now outweigh the benefits, so we are choosing to remove it.

| Before | After |
|--------|------|
| image | image |

> [!NOTE]
> I recommend reviewing this PR with "Hide whitespaces" enabled.

## Type of change

- [x] Refactor

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
@@ -2,285 +2,276 @@ import { Debouncer } from '@polymer/polymer/lib/utils/debounce.js';
 import { timeOut } from '@polymer/polymer/lib/utils/async.js';
 import { ComboBoxPlaceholder } from '@vaadin/combo-box/src/vaadin-combo-box-placeholder.js';
 
-const tryCatchWrapper = function (callback) {
-  return window.Vaadin.Flow.tryCatchWrapper(callback, 'Vaadin Combo Box');
-};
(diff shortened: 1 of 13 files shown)


sissbruecker (approved):

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -2,285 +2,276 @@ import { Debouncer } from '@polymer/polymer/lib/utils/debounce.js';
...
+
+    for (let i = 0; i < updatedPageCount; i++) {
+      let page = firstPageToSet + i;
+      let slice = items.slice(i * comboBox.pageSize, (i + 1) * comboBox.pageSize);
+
+      cache[page] = slice;
+    }
+  };
+
+  comboBox.$connector.updateData = (items) => {
```
web-padawan, line 191: nit: some functions are changed to arrow functions and others are not. 
Maybe we could align these in a separate PR.

web-padawan (approved):

PR #6743, opened 2024-10-25
feat: add server-side fallback parser

### LINKED ISSUE #2466: Server-side date parsing

vursen:
## Description

The PR adds support for an opt-in server-side parser that DatePicker will use as a fallback when it's unable to parse user input using the i18n parsing formats. The parser is designed as a function that accepts the user-entered string and returns a `Result` object with the parsed date or an error message:

```java
DatePicker datePicker = new DatePicker();
datePicker.setFallbackParser((s) -> {
    if (s.equals("tomorrow")) {
        return Result.ok(LocalDate.now().plusDays(1));
    } else {
        return Result.error("Invalid date format");
    }
});
```

The returned error message always overrides the i18n error message for bad input, even if an empty string is returned. To reuse an i18n message, you should return it explicitly: `Result.error(datePicker.getI18n().getBadInputErrorMessage())`.

Depends on
- https://github.com/vaadin/flow/issues/20365

Fixes #2466 

## Type of change

- [x] Feature


vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
```diff
@@ -658,9 +666,76 @@ private void fireValidationStatusChangeEvent() {
      * @return <code>true</code> if the input element's value is populated,
      *         <code>false</code> otherwise
      */
-    @Synchronize(property = "_hasInputValue", value = "has-input-value-changed")
     protected boolean isInputValuePresent() {
```
vursen, line 669: This getter was made protected only because we needed it to be accessible from DateTimePicker. This PR introduces a change: the getter no longer updates on `has-input-value-changed`. Do you think it's worth mentioning or even restoring the previous behavior?

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
```diff
@@ -257,6 +264,7 @@ private DatePicker(LocalDate initialDate, boolean isInitialValueOptional) {
         addValueChangeListener(e -> validate());
 
         getElement().addEventListener("unparsable-change", event -> {
```
sissbruecker, line 266: I don't think I was familiar with this concept yet. `unparsable-changed` is fired every time the user tries to commit a value, regardless whether the previous value was already unparsable? The event name made it sound like it's only fired when the component state switches from parsable to unparsable and vice versa.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
sissbruecker: Should be fine IMO.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/src/main/java/com/vaadin/flow/component/datepicker/DatePickerFallbackParserPage.java
```diff
@@ -0,0 +1,71 @@
...
+
+import elemental.json.Json;
+import elemental.json.JsonObject;
+
+@Route("vaadin-date-picker/fallback-parser")
+public class DatePickerFallbackParserPage extends Div {
+    public DatePickerFallbackParserPage() {
+        DatePicker datePicker = new DatePicker();
+        datePicker.setFallbackParser((s) -> {
+            if (s.equals("newyear")) {
```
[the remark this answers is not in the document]
vursen: This seems to be a limitation in how DatePicker is currently designed. It assumes that date parsing is done synchronously on the client-side. To highlight the date when entering something like `newyear`, we would need to send a server request on every keystroke, which could lead us to moving the entire parsing logic to the server.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
```diff
@@ -671,30 +746,56 @@ public void setValue(LocalDate value) {
...
+
+            if (result.isError()) {
+                fallbackParserErrorMessage = result.getMessage().orElse(null);
+            } else {
+                fallbackParserErrorMessage = null;
+                newModelValue = result.getOrThrow(IllegalStateException::new);
+                setPresentationValue(newModelValue);
+            }
+
+            isFallbackParserRunning = false;
```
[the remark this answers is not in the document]
vursen: Good idea, thanks! Updated. We could even go further, intercept exceptions, and convert them to logger errors, but I'm not sure what to show to the user, bad input error message from i18n?

vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/src/main/java/com/vaadin/flow/component/datepicker/DatePickerFallbackParserPage.java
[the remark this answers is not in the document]
sissbruecker: My comment was mostly just to point out the obvious, I know there is no easy solution for it. This is something that a developer or user might notice and not understand what the difference is compared to entering some other format. Especially considering that someone could also use this feature to parse a custom date format such as `14.11.2024` instead of using I18N date formats.

Anyway, I don't think it's a blocker for this PR. As discussed we can try to migitate this by not showing a highlight if the date can not be parsed by the client, or at some later point research other solutions if this turns out to be a problem.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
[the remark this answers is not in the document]
sissbruecker: Logging the exception would be a good idea IMO. Could just be a message like "Fallback parser threw an exception" together with the exception that was caught.

For the validation error, maybe just use the I18N bad input message? That seems the most reasonable instead of trying to come up with something new.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
vursen: Note that the event name is `unparsable-change` rather than `unparsable-changed`. The event compliments `change` and, when used together, they allow you to get notified of any input changes, including unparsable ones, as if it was a regular text field.

sissbruecker: OK, thanks

sissbruecker (approved):

PR #6759, opened 2024-10-30
fix: access UI through container element in ComponentRenderer

### LINKED ISSUE #6751: [Regression] JS error occurs when ComponentRenderer is created in a background thread
### Description

The [refactoring](https://github.com/vaadin/flow-components/pull/4148) that updated the ComponentRenderer implementation to be based on LitRenderer introduced a regression: a JS error can now occur if ComponentRenderer is created in a background thread. This happens because the new implementation accesses UI using `UI.getCurrent()` unlike the previous implementation which used a different approach based on `runWhenAttached`. 

### Expected outcome

ComponentRenderer should retrieve the UI instance from the container element.

The example below works without errors in Vaadin 23 but throws a JS error in Vaadin 24, resulting in an empty grid.

```java
@Push // <----- required
public class Application implements AppShellConfigurator {
```

```java
@Route("grid")
    public GridView() {
        Button addGrid = new Button("Add grid", (event) -> {
            final VaadinSession session = VaadinSession.getCurrent();

            new Thread(() -> session.accessSynchronously(this::render)).start();
        });

        add(addGrid);
    }

    public void render() {
        grid.addColumn(new ComponentRenderer<Span, String>(item -> new Span(item)));
        grid.setItems("Item 1", "Item 2", "Item 3", "Item 4", "Item 5");
        add(grid);
    }
}
```

### Steps to reproduce

Click on the button and observe `Cannot read properties of undefined (reading 'getByNodeId')` in the browser console.

### Environment

Vaadin version(s): 24.0 and later

### Browsers

_No response_

vursen:
## Description

Using `UI.getCurrent()` isn't safe because even when the container is attached, it may return `null`. This PR updates the ComponentRenderer implementation to access the UI instance through the container element where it's guaranteed to be available if the container is attached.

Fixes #6751

## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-renderer-flow-parent/vaadin-renderer-flow-integration-tests/src/test/java/com/vaadin/flow/data/renderer/tests/ComponentRendererInNewThreadIT.java
+++ b/vaadin-renderer-flow-parent/vaadin-renderer-flow-integration-tests/src/test/java/com/vaadin/flow/data/renderer/tests/ComponentRendererInNewThreadIT.java
@@ -16,8 +16,10 @@
 package com.vaadin.flow.data.renderer.tests;
 
 import org.junit.Assert;
+import org.junit.Before;
 import org.junit.Test;
 import org.openqa.selenium.By;
+import org.openqa.selenium.WebElement;
 
 import com.vaadin.flow.testutil.TestPath;
 import com.vaadin.tests.AbstractComponentIT;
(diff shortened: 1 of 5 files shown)


vaadin-renderer-flow-parent/vaadin-renderer-flow-integration-tests/src/test/java/com/vaadin/flow/data/renderer/tests/ComponentRendererInNewThreadIT.java
```diff
@@ -16,25 +16,43 @@
...
-    @Test
-    public void componentRendererInNewThread_uiNotNullWhileTemplateExpressionIsCalculated() {
+    @Before
+    public void init() {
         open();
-        findElement(By.id("add-component")).click();
+    }
+
+    @Test
+    public void addComponentRendererBeforeAttach_componentIsRendered() {
```
vursen, line 36: **note:** This test is a modern version of `componentRendererInNewThread_uiNotNullWhileTemplateExpressionIsCalculated` which covered https://github.com/vaadin/flow-components/pull/5534

tomivirkki (approved):

ugur-vaadin (approved):

Issue #4963, opened 2023-04-20
Do not force callers of the LitRenderer constructor to hold a lock on the VaadinSession


mperktold:
### Describe your motivation

In v24, the `LitRenderer` constructor updates a property "__litRendererCount" of the UI. That's only possible while holding the lock on the VaadinSession.

This is unfortunate, and very atypical for a constructor. In our application, we are doing a lot of work in background threads, including creating components. They are then added later via `UI.access` and `@Push`.

### Describe the solution you'd like

Maybe you could consider changing LitRenderer such that it acquires the lock itself, such that the callers do not have to know about this implementation detail?

### Describe alternatives you've considered

Currently, we would need to check all places were renderers are created, either directly or indirectly, and either look whether these places can be called from a context were the lock on the VaadinSession is not hold, or simply acquire the lock everywhere.

Artur-:
But if you are in a background thread, why is `UI.getCurrent()` defined?

Artur-:
You probably should not do that, as many places assume that the framework is the one handling the current UI instance and that it is an indication that a current UI is locked and available

mperktold:
But I *have  to* do that for several things, i.e. opening dialogs from background threads.

Also, the JavaDoc of `UI.setCurrent` explicitly allows to use that method for background threads:

https://github.com/vaadin/flow/blob/4072f6d108d8fdfe5d5bc7d65cf7f8b4c7a0d171/flow-server/src/main/java/com/vaadin/flow/component/UI.java#L297-L299

mperktold:
It's also worth pointing out that this is a regression/breaking change: it worked in Vaadin 23, and stopped working in Vaadin 24.

I understand that breaking changes can happen in major version jumps. Still, it's important to know that this is such a case.

PR #6761, opened 2024-10-30
fix: do not require UI to generate property namespace

vursen:
## Description

Reimplements property namespace generation to work without `UI.getCurrent()` because it can be unavailable when creating an instance of LitRenderer in a background thread.

Fixes https://github.com/vaadin/flow-components/issues/4963

## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/java/com/vaadin/flow/data/renderer/ComponentRenderer.java
+++ b/vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/java/com/vaadin/flow/data/renderer/ComponentRenderer.java
@@ -151,8 +151,8 @@ public class ComponentRenderer<COMPONENT extends Component, SOURCE>
     }
 
     @Override
-    Rendering<SOURCE> render(Element owner, DataKeyMapper<SOURCE> keyMapper,
-            String rendererName, String propertyNamespace) {
+    public Rendering<SOURCE> render(Element owner,
+            DataKeyMapper<SOURCE> keyMapper, String rendererName) {
         this.owner = owner;
         this.container = new Element("div");
         this.container.addAttachListener(event -> {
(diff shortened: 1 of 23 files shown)


vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/java/com/vaadin/flow/data/renderer/LitRenderer.java
```diff
@@ -70,27 +70,13 @@ public class LitRenderer<SOURCE> extends Renderer<SOURCE> {
...
-        if (UI.getCurrent() != null) {
-            // Generate a unique (in scope of the UI) namespace for the renderer
-            // properties.
-            litRendererCount = UI.getCurrent().getElement()
-                    .getProperty("__litRendererCount", 0);
-            UI.getCurrent().getElement().setProperty("__litRendererCount",
-                    litRendererCount + 1);
-
-        }
-        propertyNamespace = "lr_" + litRendererCount + "_";
```
tomivirkki, line 93: Now that the format has changed, we could update the [example](https://github.com/vaadin/flow-components/blob/d4c616c89bdc25fb1366d8af771d6b3353082471/vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/resources/META-INF/resources/frontend/lit-renderer.ts#L79) at `lit-renderer.ts`

vursen: Thanks, done.

vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/java/com/vaadin/flow/data/renderer/ComponentRenderer.java
tomivirkki, line 1: The components are missing on the "Person" column of this example:

Screenshot 2024-10-31 at 15 44 09

vursen: Wow, that's curious.

vursen: I updated the implementation to use UUID instead of nodeId. Nodeid can be -1 if the container has never been attached yet.

tomivirkki (approved):

PR #6787, opened 2024-11-04
feat: implement HasEnabled for Upload

sissbruecker:
Allows to disable the upload component, which disables all buttons, prevents drag and drop of files and focusing elements within the component.

When disabled, the upload component does not accept new uploads, but keeps processing existing ones.

Part of https://github.com/vaadin/web-components/issues/1278
Depends on https://github.com/vaadin/web-components/pull/8090

### DIFF
--- a/vaadin-upload-flow-parent/vaadin-upload-flow/src/test/java/com/vaadin/flow/component/upload/DisabledUploadTest.java
+++ b/vaadin-upload-flow-parent/vaadin-upload-flow/src/test/java/com/vaadin/flow/component/upload/DisabledUploadTest.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2000-2024 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


vaadin-upload-flow-parent/vaadin-upload-flow/src/test/java/com/vaadin/flow/component/upload/DisabledUploadTest.java
```diff
@@ -0,0 +1,34 @@
...
+ * the License.
+ */
+package com.vaadin.flow.component.upload;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class DisabledUploadTest {
+    @Test
+    public void preventsStartingUploads() {
```
ugur-vaadin, line 23: Does it make sense to add an integration test for this? I could not come up with an alternative unit test that avoid using internals.

sissbruecker: I think it's OK as is, the stream variable is the only entry point to start an upload. Do you feel strongly about this?

ugur-vaadin: Not really, LGTM otherwise.

ugur-vaadin (approved):

PR #6792, opened 2024-11-06
refactor: use manualValidation instead of overriding _shouldSetInvalid

vursen:
## Description

Enables manual validation mode for web components instead of overriding their protected `_shouldSetInvalid` method.

Depends on 
- #6795 

## Type of change

- [x] Refactor

### DIFF
--- a/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/ClientValidationUtil.java
+++ b/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/ClientValidationUtil.java
@@ -23,6 +23,7 @@ import com.vaadin.flow.component.HasValidation;
  * to valid as a result of client-side validation.
  *
  * @author Vaadin Ltd
+ * @deprecated since 24.6, use {@code getElement().setProperty("manualValidation", true)} instead.
  */
 public final class ClientValidationUtil {
 
@@ -30,6 +31,9 @@ public final class ClientValidationUtil {
         // utility class should not be instantiated
     }
 
+    /**
+     * @deprecated since 24.6, use {@code getElement().setProperty("manualValidation", true)} instead.
+     */
     public static <C extends Component & HasValidation> void preventWebComponentFromModifyingInvalidState(
             C component) {
         StringBuilder expression = new StringBuilder(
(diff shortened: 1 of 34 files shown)


knoobie:
Suggestion because `ClientValidatedEvent` is kinda recent and probably not much used (hopefully): 

- Add a "compatibility mode" as feature flag which uses the old code - which  a user has to active.. so 24.6 uses the new method and people that need the old way can still return back by adding the flag (which would be removed in 25.?)

e.g.: 

```java
// Pseudo Code
if(attachEvent.getUI().getFeatureFlags(LEGACY_CLIENT_VALIDATION_EVENT_ENABLED, false)) 
  ClientValidationUtil.preventWebComponentFromModifyingInvalidState(this);
else 
  getElement().setProperty("manualValidation", true);
```

vursen:
@knoobie Thanks for the suggestion. I'm afraid this refactoring isn't worth a feature flag since it's just an internal improvement to use public APIs instead of protected ones. I think we can just hold off this change until a later time.

vursen:
I found a way to avoid the breaking change by adding a temporary workaround that fires the validated event on the web component when its validation is in manual mode.

web-padawan (approved):

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/ClientValidationUtil.java
```diff
@@ -23,13 +23,21 @@
  * to valid as a result of client-side validation.
  *
  * @author Vaadin Ltd
+ * @deprecated since 24.6, use
```
sissbruecker, line 26: This would only end up in 24.8, so should we reference that instead?

vursen: Thanks, fixed.

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/ClientValidationUtil.java
```diff
@@ -23,13 +23,21 @@
  * to valid as a result of client-side validation.
  *
  * @author Vaadin Ltd
+ * @deprecated since 24.8, use
+ *             {@code getElement().setProperty("manualValidation", true)}
+ *             instead.
```
mshabarov, line 28: Sorry for commenting on a closed PR, but shouldn't here and in other places the classes have the `@Deprecated` annotation ?

vursen: Good catch, thanks! Here is a fix: https://github.com/vaadin/flow-components/pull/7251

PR #6797, opened 2024-11-07
refactor!: sync value on change event instead of value-changed

vursen:
## Description

The PR configures DatePicker to synchronize its value on `change` event instead of `value-changed`. This helps avoid an extra back request from the client after setting a value programmatically from the server.

> [!WARNING]
> This change is marked as breaking because it may affect applications that set the `value` property directly in their tests, without dispatching a `change` event. However, if they use DatePickerElement’s `setDate` method, they are safe.

## Type of change

- [x] Refactor


vursen:
Closed as not planned.

Issue #6713, opened 2024-10-11
[TreeGrid] Should refresh only visible items instead of triggering full data reset where possible


vursen:
### Description

`setDropMode`, `setDragFilter`, `setPartNameGenerator`, and many other TreeGrid methods call `dataCommunicator.reset` to re-render items, typically to update certain attributes. This operation is effectively equivalent to `refreshAll` and triggers a full reset of the hierarchy, which causes an annoying scroll position reset on the client-side. 

In cases where the grid's hierarchy isn't expected to change – which is true for all the mentioned methods – it should be enough to simply call `refreshItem` for items that are currently visible.

### Expected outcome

The Flow framework should provide an API to retrieve items that are currently in the viewport. With this API, the Flow component could call `refreshItem` for visible items instead of doing a full reset.

### Minimal reproducible example

```java
@Route("tree-grid-bug")
public class TreeGridBugPage extends Div {
    public TreeGridBugPage() {
        TreeGrid<String> treeGrid = new TreeGrid<>();
        treeGrid.addHierarchyColumn(item -> item);
        treeGrid.setRowsDraggable(true);

        treeGrid.setDropFilter(item -> item.startsWith("Child-"));

        TreeData<String> treeData = new TreeData<>();
        for (int i = 0; i < 100; i++) {
            String item = "Root-" + i;
            treeData.addItem(null, item);
            for (int j = 0; j < 10; j++) {
                treeData.addItem(item, "Child-" + i + "-" + j);
            }
        }
        treeGrid.setTreeData(treeData);

        treeGrid.addDragStartListener(e -> {
            treeGrid.setDropMode(GridDropMode.ON_TOP_OR_BETWEEN);
        });
        treeGrid.addDropListener(e -> {
            treeGrid.setDropMode(null);
        });

        add(treeGrid);
    }
}
```

### Steps to reproduce

---

### Environment

Vaadin version(s): 24.5 and earlier
OS: Mac OS

vursen:
- https://github.com/vaadin/flow-components/pull/6712
- https://github.com/vaadin/docs/pull/3786

We should consider reverting these PRs when the issue is fixed.

vursen:
Here is the branch with my attempts to create such an API: 

https://github.com/vaadin/flow/compare/get-requested-range-items-method

However, it currently has a flaw where it can include more items than necessary because DataCommunicator doesn't remove expanded items from the key mapper after they have been visited once. Some additional logic is needed to filter out such items.

[...]

PR #6798, opened 2024-11-07
refactor!: do not generate data for hidden columns

### LINKED ISSUE #6737: Hidden Grid Column should not run code
### Description

Hidden column should not slow down grid

### Expected outcome

Not run code that does nothing

```java
grid.addColumn(this::getRemoteStatus).setHeader("Remote status").setVisible(false);
````

### Steps to reproduce

`getRemoteStatus()` that sleeps

### Environment

24.5.0

### Browsers

Issue is not browser related

### PR DESCRIPTION
**Pull request by ugur-vaadin**
## Description

Value providers, tooltip generators, and part name generators run for every row regardless of column visibility. This adds unnecessary load when these callbacks do expensive work.

Each column's data generators are now wrapped in a `CompositeDataGenerator` that only runs them while the column is visible, and the row-level tooltip and part name generation are skipped for hidden columns too. Showing a previously hidden column refreshes the viewport so its data is generated again (but not reloaded from the data provider).

Depends on
- https://github.com/vaadin/flow/pull/24895
- https://github.com/vaadin/flow-components/pull/9661

> [!WARNING]
> Hidden columns no longer send data to the client side.

Fixes #6737

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
@@ -545,7 +545,7 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
             return this;
         }
 
-        @SuppressWarnings({ "unchecked", "rawtypes" })
+        @SuppressWarnings({ "unchecked" })
         private void setupRenderer(Renderer<T> renderer) {
             this.renderer = Objects.requireNonNull(renderer,
                     "Renderer must not be null.");
@@ -560,8 +560,8 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
                             .getKeyMapper());
 
             rendering.getDataGenerator().ifPresent(dataGenerator -> {
-                rendererRegistrations.add(compositeDataGenerator
-                        .addDataGenerator((DataGenerator) dataGenerator));
+                rendererRegistrations.add(
+                        compositeDataGenerator.addDataGenerator(dataGenerator));
             });
 
             rendererRegistrations.add(rendering.getRegistration());
@@ -1133,7 +1133,7 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
                 editorRenderer = new EditorRenderer<>((Editor) grid.getEditor(),
                         columnInternalId);
                 editorRendererRegistration = compositeDataGenerator
-                        .addDataGenerator((DataGenerator) editorRenderer);
+                        .addDataGenerator(editorRenderer);
             }
 
             editorRenderer.render(getElement(), null);
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
@@ -450,6 +450,13 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
                     super.generateData(item, jsonObject);
                 }
             }
+
+            @Override
+            public void refreshData(T item) {
+                if (Column.this.isVisible()) {
+                    super.refreshData(item);
+                }
+            }
         };
         private Registration compositeDataGeneratorRegistration;
 
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
@@ -417,12 +417,6 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
     /**
      * Server-side component for the {@code <vaadin-grid-column>} element.
      *
-     * <p>
-     * Every added column sends data to the client side regardless of its
-     * visibility state. Don't add a new column at all or use
-     * {@link Grid#removeColumn(Column)} to avoid sending extra data.
-     * </p>
-     *
      * @param <T>
      *            type of the underlying grid this column is compatible with
      */
@@ -449,6 +443,16 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
 
         private SerializableComparator<T> comparator;
 
+        private final CompositeDataGenerator<T> compositeDataGenerator = new CompositeDataGenerator<>() {
+            @Override
+            public void generateData(T item, ObjectNode jsonObject) {
+                if (Column.this.isVisible()) {
+                    super.generateData(item, jsonObject);
+                }
+            }
+        };
+        private Registration compositeDataGeneratorRegistration;
+
         private Renderer<T> renderer;
         private List<Registration> rendererRegistrations = new ArrayList<>();
 
@@ -470,7 +474,18 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
             super(grid);
             this.columnInternalId = columnId;
             comparator = (a, b) -> 0;
+            compositeDataGeneratorRegistration = grid
+                    .addDataGenerator(compositeDataGenerator);
             setupRenderer(renderer);
+        }
+
+        @Override
+        public void setVisible(boolean visible) {
+            boolean refreshViewport = visible && !isVisible();
+            super.setVisible(visible);
+            if (refreshViewport) {
+                getGrid().refreshViewport();
+            }
         }
 
         protected void destroyDataGenerators() {
@@ -482,6 +497,11 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
             if (editorRendererRegistration != null) {
                 editorRendererRegistration.remove();
                 editorRendererRegistration = null;
+            }
+
+            if (compositeDataGeneratorRegistration != null) {
+                compositeDataGeneratorRegistration.remove();
+                compositeDataGeneratorRegistration = null;
             }
         }
 
@@ -540,8 +560,8 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
                             .getKeyMapper());
 
             rendering.getDataGenerator().ifPresent(dataGenerator -> {
-                rendererRegistrations.add(
-                        grid.addDataGenerator((DataGenerator) dataGenerator));
+                rendererRegistrations.add(compositeDataGenerator
+                        .addDataGenerator((DataGenerator) dataGenerator));
             });
 
             rendererRegistrations.add(rendering.getRegistration());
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
Comment (sissbruecker) [inline vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java:1]: When adding a column, it sends two full updates to the client. One based on the old column config, and then another one with the new config. Also calls the data provider twice.

Bildschirmfoto 2024-11-08 um 10 34 10

Comment (ugur-vaadin) [reply vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java:1]: Added a fix and the test for it to handle this case.

File: vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java [hunk 1]
```diff
@@ -1181,15 +1155,54 @@ protected Column<?> getBottomLevelColumn() {
...
+            if (renderer == null) {
+                return;
+            }
+            rendering = renderer.render(getElement(), (KeyMapper<T>) getGrid()
+                    .getDataCommunicator().getKeyMapper());
+            columnDataGeneratorRegistration = rendering.getDataGenerator()
+                    .map(dataGenerator -> grid
+                            .addDataGenerator((DataGenerator) dataGenerator))
+                    .orElse(null);
+            grid.getDataProvider().refreshAll();
```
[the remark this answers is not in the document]
ugur-vaadin: I agree that it would be better to do it that way. However, directly calling in `setVisible` does not work and I'm not sure what would be a good way to refactor the refresh.

tomivirkki:
Would making the data generator run conditionally based on the column's visibility fix the original issue?
Proto: https://github.com/vaadin/flow-components/commit/1a5cb527e733f45369a2d08057847f4a53efa1cc

ugur-vaadin:
> Would making the data generator run conditionally based on the column's visibility fix the original issue? Proto: [1a5cb52](https://github.com/vaadin/flow-components/commit/1a5cb527e733f45369a2d08057847f4a53efa1cc)

Adapted the suggestion. It works nicely and is cleaner.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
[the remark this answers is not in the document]
tomivirkki: In theory, hiding and unhiding columns shouldn't need to result in a data provider call, but only the data generators should need to be run for the already loaded items.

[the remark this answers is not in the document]
vursen: Probably related: https://github.com/vaadin/flow-components/issues/6713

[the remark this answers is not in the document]
ugur-vaadin: Added API to make this optional, and kept the default behaviour the same.

vursen:
DataCommunicator needs a fix so it doesn't trigger DataProvider on `refreshViewport`. HierarchicalDataCommunicator already behaves this way.

I've added the corresponding tests to this PR, and one of them is currently failing because of that issue:

https://github.com/vaadin/flow-components/blob/19790aa5e069db5b9dde084202df4f64196925b8/vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridHiddenColumnRenderingTest.java#L159

https://github.com/vaadin/flow-components/blob/19790aa5e069db5b9dde084202df4f64196925b8/vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridHiddenColumnRenderingTest.java#L171

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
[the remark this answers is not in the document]
vursen: This is now resolved by calling `refreshViewport` instead of `refreshAll`. If both end up scheduled, DataCommunicator's scheduling system guarantees that only `refreshAll` runs, so nothing happens twice.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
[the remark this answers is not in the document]
vursen: I removed the opt-in API and enabled the new behavior by default, since with the new approach, toggling column visibility no longer causes any data provider requests.

vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/java/com/vaadin/flow/component/gridpro/GridPro.java
sissbruecker, line 1: `GridPro` still generates cell editable data for hidden columns when a cell editable provider is set.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
sissbruecker, line 1: There is an edge-case where `setVisible` runs in `DataCommunicator.flush` (e.g. reacting to `ItemCountChangeEvent`). In that case `grid.refreshViewport` doesn't have any effect and it sends empty cells.

I'd consider leaving this as is, but would like to hear your opinion.

sissbruecker, line 1: There is an existing bug where toggling a column group's visibility forces columns to be visible on the client side due to how the web component works. Previously those columns would have data, now they may not.

I would be fine with filing an issue for this for now given that it's an existing problem. Another issue is that hiding columns on the server via the column group does not prevent sending data, though that could be a future improvement.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -449,6 +443,16 @@ public static class Column<T> extends AbstractColumn<Column<T>> {
 
         private SerializableComparator<T> comparator;
 
+        private final CompositeDataGenerator<T> compositeDataGenerator = new CompositeDataGenerator<>() {
+            @Override
+            public void generateData(T item, ObjectNode jsonObject) {
```
sissbruecker, line 448: There is also a `refreshData` which is called through `refreshItem`. So when refreshing items individually it will still run component renderers at least.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -4367,6 +4338,9 @@ private void generateTooltipTextData(T item, ObjectNode jsonObject) {
         }
 
         idToColumnMap.forEach((id, column) -> {
+            if (!column.isVisible()) {
```
sissbruecker, line 4348: There would be an opportunity to register column tooltip / part name generators in the composite one to reuse the visibility check there. Generating the per row and per column tooltips would then need to look something like this:
```java
  private void generateTooltipTextData(T item, ObjectNode jsonObject) {                                                                                                                                                                                                                                                     
      String rowTooltip = tooltipGenerator.apply(item);                                                                                                                                                                                                                                                                     
      if (rowTooltip != null) {                                                                                                                                                                                                                                                                                             
          getOrCreateObject(jsonObject, "gridtooltips").put("row", rowTooltip);                                                                                                                                                                                                                                             
      }                                                                                                                                                                                                                                                                                                                     
  }

  compositeDataGenerator.addDataGenerator((item, jsonObject) -> {                                                                                                                                                                                                                                                           
      String cellTooltip = tooltipGenerator.apply(item);                                                                                                                                                                                                                                                                    
      if (cellTooltip != null) {                                                                                                                                                                                                                                                                                            
          getOrCreateObject(jsonObject, "gridtooltips")                                                                                                                                                                                                                                                                     
                  .put(getInternalId(), cellTooltip);                                                                                                                                                                                                                                                                       
      }                                                                                                                                                                                                                                                                                                                     
  });
```

Not sure if that's better, as it duplicates the property access then.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
vursen: Done: added the check and covered it with a test.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
vursen: I like this approach. Let's just do it in a follow-up PR.

vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/java/com/vaadin/flow/component/gridpro/GridPro.java
vursen: Thanks, fixed.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
vursen: Good catch! Filed a separate issue for this: https://github.com/vaadin/flow-components/issues/9677

web-padawan (approved):

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
vursen: Good question. I'm actually inclined to call this a DataCommunicator bug: the `flushInProgress` flag should probably already be reset by the time events like ItemCountChangeEvent are fired, so that a  flush can be scheduled again. It's worth noting that the same issue can be reproduced with `refreshAll` when calling it from an ItemCountChangeEvent listener, so it's not only specific to `refreshViewport`.

sissbruecker (approved):

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
vursen: - https://github.com/vaadin/flow-components/pull/9679
- https://github.com/vaadin/flow-components/pull/9680

PR #6800, opened 2024-11-08
refactor!(TimePicker): sync value on change event instead of value-changed

vursen:
## Description

For consistency with DatePicker, this PR updates TimePicker to synchronize its value on `change` event instead of `value-changed`.

Related #6797 

## Type of change

- [x] Refactor


vursen:
Transitioning TimePicker to the `change` event turns out to be less trivial than it was with DatePicker. Looks like some tests rely on the web component firing multiple consecutive `value-changed` events in [certain cases](https://github.com/vaadin/web-components/blob/f927f4a5e07c59b0b1963361c030e78b694125b8/packages/time-picker/src/vaadin-time-picker-mixin.js#L490-L495). I'm closing the PR for now.

PR #6814, opened 2024-11-12
test: isolate logger mocking into respective tests

sissbruecker:
https://github.com/vaadin/flow-components/pull/6669 introduced static mocking of `LoggerFactory` in `LoginFormTest`, however that seems to cause flaky tests, such as the failure [here](https://teamcity.vaadin.com/buildConfiguration/VaadinFlowComponents_PullRequestValidation/248881?expandBuildProblemsSection=true&hideProblemsFromDependencies=false&expandBuildTestsSection=true&hideTestsFromDependencies=false) on a PR for `Upload`.

Moving the mocking into the respective test methods [seems to help](https://teamcity.vaadin.com/buildConfiguration/VaadinFlowComponents_PullRequestValidation/248919?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true).

### DIFF
--- a/vaadin-login-flow-parent/vaadin-login-flow/src/test/java/com/vaadin/flow/component/login/LoginFormTest.java
+++ b/vaadin-login-flow-parent/vaadin-login-flow/src/test/java/com/vaadin/flow/component/login/LoginFormTest.java
@@ -17,9 +17,7 @@ package com.vaadin.flow.component.login;
 
 import java.util.concurrent.atomic.AtomicInteger;
 
-import org.junit.After;
 import org.junit.Assert;
-import org.junit.Before;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


ugur-vaadin (approved):
I wonder if the tests in `LoginOverlayTest` are also flaky. It seems to use the same approach.

DiegoCardoso (approved):
Looking at the build log for the failed test, the mock `Logger` could sometimes cause an exception in tests where the `Logger` instance does not need to be checked, so it makes sense to move the mocking to where it's needed.

sissbruecker:
Thanks, also updated `LoginOverlayTest` as it uses the same setup / test cases.

PR #6828, opened 2024-11-18
feat: configure min and max rows in text area

sissbruecker:
Part of https://github.com/vaadin/web-components/issues/1399


vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/TextArea.java
sissbruecker, line 1: Feels a bit strange that one property has a default value and does not allow `null`, while the other one doesn't have a default and does allow `null`. I guess we could change `minRows` to be nullable as well and document that it then uses 2 rows by default, though I'm not sure if that's better than having this inconsistency. Any opinions?

sissbruecker: Probably not very relevant either, I would assume the min/max rows configuration to usually be static.

PR #6850, opened 2024-11-22
fix: restore hover effect on pie charts when using Lumo theme classes

sissbruecker:
Fixes https://github.com/vaadin/web-components/issues/8196

### DIFF
--- a/vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/themes/LumoDarkTheme.java
+++ b/vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/themes/LumoDarkTheme.java
@@ -9,8 +9,6 @@
 package com.vaadin.flow.component.charts.themes;
 
 import com.vaadin.flow.component.charts.model.AbstractDataLabels;
-import com.vaadin.flow.component.charts.model.Hover;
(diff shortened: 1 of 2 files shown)


vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/themes/LumoDarkTheme.java
```diff
@@ -164,10 +162,6 @@ public LumoDarkTheme() {
         setDataLabelsDefaults(getPlotOptions().getPyramid().getDataLabels());
         setDataLabelsDefaults(getPlotOptions().getSpline().getDataLabels());
         setDataLabelsDefaults(getPlotOptions().getWaterfall().getDataLabels());
-
-        States states = new States();
-        states.setHover(new Hover(false));
-        getPlotOptions().getPie().setStates(states);
```
sissbruecker, line 170: This was added initially with https://github.com/vaadin/vaadin-charts-flow/pull/393. From the `Hover` JavaDoc I can not quite understand what the `enabled` state is supposed to do, but apparently disabling it doesn't do anything sensible it then marks all chart slices as inactive on hover. Removing this configuration seems to make the chart behave like any other without a theme.

DiegoCardoso (approved):
LGTM

PR #6859, opened 2024-11-22
test: make integration tests work with new headless mode

sissbruecker:
Flow components validation has been updated to use Chrome 131, which now uses the new headless mode by default. That seems to have the following impact:
- The browser now opens with a rather small window size by default (780x441 according to [screenshots from test failures](https://bender.vaadin.com/repository/download/FlowComponents_Snapshot/620109:id/integration-tests/error-screenshots/run-1/openPopup_firstPageLoaded%5Bany_Chrome_%5D(com.vaadin.flow.component.combobox.test.LazyLoadingIT).png)). That causes some failures where for example a combo box overlay is too small to include expected items, or a dialog can't resize to the expected height.
- Apparently the window now also includes the browser frame, so some tests that already set a specific viewport fail.

To fix this:
- I applied a larger default viewport size in `AbstractComponentIT`. That isn't used by every test case, but was enough for fixing the currently failing tests.
- Applied individual changes to a combo box IT and spreadsheet IT that depend on the viewport size

### DIFF
--- a/vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/themes/LumoDarkTheme.java
+++ b/vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/themes/LumoDarkTheme.java
@@ -9,8 +9,6 @@
 package com.vaadin.flow.component.charts.themes;
 
 import com.vaadin.flow.component.charts.model.AbstractDataLabels;
-import com.vaadin.flow.component.charts.model.Hover;
-import com.vaadin.flow.component.charts.model.States;
 import com.vaadin.flow.component.charts.model.style.AxisStyle;
 import com.vaadin.flow.component.charts.model.style.Color;
 import com.vaadin.flow.component.charts.model.style.FontWeight;
@@ -164,10 +162,6 @@ public class LumoDarkTheme extends Theme {
         setDataLabelsDefaults(getPlotOptions().getPyramid().getDataLabels());
         setDataLabelsDefaults(getPlotOptions().getSpline().getDataLabels());
         setDataLabelsDefaults(getPlotOptions().getWaterfall().getDataLabels());
-
-        States states = new States();
-        states.setHover(new Hover(false));
-        getPlotOptions().getPie().setStates(states);
     }
 
     protected void setDataLabelsDefaults(AbstractDataLabels labels) {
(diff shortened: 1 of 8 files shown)


vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/dataview/AbstractItemCountComboBoxIT.java
```diff
@@ -32,7 +32,7 @@ public abstract class AbstractItemCountComboBoxIT extends AbstractComboBoxIT {
 
     // changing the dimension might get combo box change what it fetches and
     // how many items it shows, so changing this is a bad idea ...
-    private static final Dimension TARGET_SIZE = new Dimension(1000, 900);
+    private static final Dimension TARGET_SIZE = new Dimension(1000, 1100);
```
sissbruecker, line 35: This needs to be larger now, probably because the browser frame takes up some height.

vaadin-flow-components-shared-parent/vaadin-flow-components-test-util/src/main/java/com/vaadin/tests/AbstractComponentIT.java
```diff
@@ -28,22 +23,10 @@ protected int getDeploymentPort() {
     }
 
     @Override
-    protected List<DesiredCapabilities> customizeCapabilities(
```
sissbruecker, line 31: Looks like the override here isn't effective anymore, also not needed.

vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-integration-tests/src/test/java/com/vaadin/flow/component/spreadsheet/test/ContextMenuIT.java
```diff
@@ -34,8 +34,9 @@ public void documentScroll_overlayPosition() {
         // Make sure to use a small enough viewport so that the document needs
         // to be scrolled to see the cell
         getDriver().manage().window().setSize(WINDOW_SIZE_SMALL);
-        // Scroll the document horizontally
-        executeScript("document.documentElement.scrollLeft = 100");
+        // Scroll the document so the cell is visible
```
sissbruecker, line 37: The viewport here has now less height as well, added some scrolling to make the cell visible.

vursen (approved):
I checked a few tests locally and confirmed that they work now.

PR #6864, opened 2024-11-25
fix: make Popover static methods thread-safe

### LINKED ISSUE #6838: popover race condition / thread-unsafe code
### Description

There is a race condition in Popover static functions. If multiple threads call `Popover.setDefaultHideDelay()` at the same time, that can result in multiple calls to `applyConfiguration()` and multiple registrations of UI init listener.

The easiest way is to mark all static `set*()` functions as `synchronized`. Alternatively, you can use `AtomicInteger` for all static fields (don't forget to make them final!) and use `uiInitListenerRegistered.compareAndSet()` in `applyConfiguration()`.

### Expected outcome

The UI Init Listener is registered at most once, regardless of how many threads call Popover static functions at the same time.

Alternatively, tell the programmer to call those static functions in a service init listener (which is called exactly once), or from a Spring singleton - but this is more complex and more demanding on the user.

As is the nature of threading bugs, they are not easy to reproduce. I can create an example program though if you need one 👍 

### Steps to reproduce

1. Create an executor with 10 threads; use CountdownLatch to start them at the same time, then start setting the default Focus Delay in a loop. Alternatively, set default focus delay in a session init listener then start DDOS on the web server, but this is harder.
2. Observe that multiple UI init listeners are registered, because of race condition.

### Environment

Vaadin version(s): 24.5.5+

### Browsers

_No response_

DiegoCardoso:
## Description

Add `synchronized` to `setDefault*` methods in `Popover` to ensure that multiple calls originating from different threads will add only one listener to the UI instance.

Fixes #6838

## Type of change

- [X] Bugfix
- [ ] Feature

### DIFF
--- a/vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
+++ b/vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
@@ -67,7 +67,7 @@ public class Popover extends Component implements HasAriaLabel, HasComponents,
     private static Integer defaultHideDelay;
     private static Integer defaultFocusDelay;
     private static Integer defaultHoverDelay;
-    private static boolean uiInitListenerRegistered = false;
+    static boolean uiInitListenerRegistered = false;
 
     private Component target;
     private Registration targetAttachRegistration;
@@ -111,7 +111,8 @@ public class Popover extends Component implements HasAriaLabel, HasComponents,
      * @param defaultFocusDelay
      *            the default focus delay
      */
-    public static void setDefaultFocusDelay(int defaultFocusDelay) {
+    public synchronized static void setDefaultFocusDelay(
+            int defaultFocusDelay) {
         Popover.defaultFocusDelay = defaultFocusDelay;
         applyConfiguration();
     }
@@ -124,7 +125,7 @@ public class Popover extends Component implements HasAriaLabel, HasComponents,
      * @param defaultHideDelay
      *            the default hide delay
      */
-    public static void setDefaultHideDelay(int defaultHideDelay) {
+    public synchronized static void setDefaultHideDelay(int defaultHideDelay) {
         Popover.defaultHideDelay = defaultHideDelay;
         applyConfiguration();
     }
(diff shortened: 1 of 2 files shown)


vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
```diff
@@ -67,7 +67,7 @@ public class Popover extends Component implements HasAriaLabel, HasComponents,
     private static Integer defaultHideDelay;
     private static Integer defaultFocusDelay;
     private static Integer defaultHoverDelay;
-    private static boolean uiInitListenerRegistered = false;
+    static boolean uiInitListenerRegistered = false;
```
DiegoCardoso, line 70: Changed the visibility here to be able to reset the flag on each test since for the variable being static, its value is preserved in between tests.

ugur-vaadin: Would it make sense to use an `AtomicBoolean` here and update the if check like the following?
```Java
if (uiInitListenerRegistered.compareAndSet(false, true)) {
    VaadinService.getCurrent().addUIInitListener(e -> applyConfigurationForUI(e.getUI()));
}
```

It would get rid of `synchronized` methods, keep the synced block minimal, and still keep the thread safety.

DiegoCardoso: Yeah, this also seems to work and requires fewer changes, so I think it might be better. Changed.

sissbruecker (approved):

ugur-vaadin (approved):

vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
```diff
@@ -67,7 +68,7 @@ public class Popover extends Component implements HasAriaLabel, HasComponents,
     private static Integer defaultHideDelay;
     private static Integer defaultFocusDelay;
     private static Integer defaultHoverDelay;
-    private static boolean uiInitListenerRegistered = false;
+    static AtomicBoolean uiInitListenerRegistered = new AtomicBoolean(false);
```
vursen, line 71: > Alternatively, you can use AtomicInteger for all static fields **(don't forget to make them final!)** and use uiInitListenerRegistered.compareAndSet() in applyConfiguration().

DiegoCardoso: Thanks. Done.

vursen (approved):

PR #6873, opened 2024-11-27
fix: Files map is transient that needs to be restored if null

TatuLund:
The files will be null after de-serialization, which will lead to NPE.


vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/receivers/MultiFileBuffer.java
```diff
@@ -38,6 +38,7 @@
  * For a custom file the constructor {@link AbstractFileBuffer(FileFactory)}
  * should be used.
  */
+@SuppressWarnings("serial")
```
sissbruecker, line 41: Is this needed? I don't see any warnings show up in IntelliJ at least.

sissbruecker: In IntelliJ the suppression itself causes a warning, because it figures that it's not needed. Looks like SonarQube also doesn't find an issue. I've removed the statement for now.

sissbruecker (approved):

PR #6879, opened 2024-12-02
refactor: move disable on click to menu item base

sissbruecker:
Moves the disable on click logic to `MenuItemBase`, which also makes it available for `GridMenuItem`.

### DIFF
--- a/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/internal/DisableOnClickController.java
+++ b/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/internal/DisableOnClickController.java
@@ -18,8 +18,10 @@ package com.vaadin.flow.component.shared.internal;
 import java.io.Serializable;
 import java.util.Objects;
 
-import com.vaadin.flow.component.ClickNotifier;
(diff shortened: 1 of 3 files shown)


vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/internal/DisableOnClickController.java
```diff
@@ -52,13 +53,16 @@ public class DisableOnClickController<C extends Component & HasEnabled & ClickNo
...
-                component.setEnabled(false);
-            }
-        });
+
+        ComponentUtil.addListener(component, ClickEvent.class,
+                (ComponentEventListener) (event -> {
+                    if (isDisableOnClick()) {
+                        component.setEnabled(false);
+                    }
+                }));
```
sissbruecker, line 65: Neither `MenuItemBase` nor `GridMenuItem` has `ClickNotifier`. I figure it's not really required for the controller either, I used `ComponentUtil` for adding the listener instead.

sissbruecker:
@ugur-vaadin We also have no tests for menu bar, and there doesn't seem to be anything grid specific regarding click handling. I'd suggest to skip those.

ugur-vaadin (approved):

PR #6884, opened 2024-12-04
feat: add ClientItemToggleEvent to notify when user toggles an item

vursen:
## Description

The PR adds a new event `ClientItemToggleEvent` that fires when the user selects or deselects an item in the grid. The event provides information about the item, its selected state, and whether <kbd>Shift</kbd> was pressed during selection. The main purpose of this new event is to enable developers to implement features like range selection. The below example demonstrates how to implement range selection with behavior similar to Gmail:

```java
@Route("range-selection")
public class RangeSelectionPage extends Div {
    private String startItem;

    public RangeSelectionPage() {
        grid.addColumn(s -> s);
        grid.setItems("Item 1", "Item 2", "Item 3", "Item 4");
        ((GridMultiSelectionModel<String>) grid.getSelectionModel())
                .addClientItemToggleListener(event -> {
                    String item = event.getItem();

                    startItem = startItem != null ? startItem : item;
                    if (event.isShiftKey()) {
                        Set<String> range = getItemsRange(grid, startItem,
                                item);
                        if (event.isSelected()) {
                            grid.asMultiSelect().select(range);
                        } else {
                            grid.asMultiSelect().deselect(range);
                        }
                    }
                    startItem = item;
                });
        add(grid);
    }

    private Set<String> getItemsRange(Grid<String> grid, String startItem,
            String endItem) {
        var dataView = grid.getListDataView();
        var startIndex = dataView.getItemIndex(startItem).get();
        var endIndex = dataView.getItemIndex(endItem).get();

        return dataView.getItems().skip(Math.min(startIndex, endIndex))
                .limit(Math.abs(startIndex - endIndex) + 1)
                .collect(Collectors.toSet());
    }
}
```

Depends on 
- https://github.com/vaadin/web-components/pull/8231

Part of https://github.com/vaadin/web-components/issues/8142

## Type of change

- [x] Feature


vaadin-grid-flow-parent/vaadin-grid-testbench/src/main/java/com/vaadin/flow/component/grid/testbench/GridElement.java
```diff
@@ -388,7 +388,7 @@ void select(GridTRElement row) {
             CheckboxElement checkbox = wrapElement(cell.getFirstChildElement(),
                     getCommandExecutor()).wrap(CheckboxElement.class);
             if (!checkbox.isChecked()) {
-                checkbox.click();
+                checkbox.getWrappedElement().click();
```
vursen, line 391: In TestBench, the `click()` method is overridden to use the JS `click()` function, which doesn't fully simulate a complete mouse action. We need to use the real simulation to get `mousedown` that sets `_shiftKeyDown`.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/ClientItemToggleEvent.java
```diff
@@ -0,0 +1,81 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component.grid;
+
+import com.vaadin.flow.component.ComponentEvent;
+
+/**
+ * Event fired when the user toggles an item on the client-side.
```
sissbruecker, line 21: Nit: It might not be clear what is toggled (selection, row details, expand children), so could be helpful to add that it toggles the selection state of an item. The methods for adding listeners could be improved similarly. 

We can also mention that:
- This event will always be fired after the selection change event
- What it could be used for: that it provides information on selected / deselected and whether the shift key was pressed, which could be used to implement a range selection feature

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/ClientItemToggleEvent.java
```diff
@@ -0,0 +1,81 @@
...
+     * @return the item that was toggled
+     */
+    public T getItem() {
+        return item;
+    }
+
+    /**
+     * Gets whether the item was selected.
+     *
+     * @return {@code true} if the item was selected, {@code false} otherwise
```
sissbruecker, line 66:
```suggestion
     * @return {@code true} if the item was selected, {@code false} if the item was deselected
```
Might be a bit clearer

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/vaadin-grid-flow-selection-column.js
```diff
@@ -42,6 +49,12 @@ export class GridFlowSelectionColumn extends GridSelectionColumnBaseMixin(GridCo
     }
   }
 
+  /** @private */
+  __shiftKeyDownChanged(newState, oldState) {
+    if (oldState !== undefined) {
+      this.$server.setShiftKeyDown(newState);
```
tomivirkki, line 55: This causes an extra server roundtrip once the user releases the shift key (after doing a range selection)

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/ClientItemToggleEvent.java
vursen: Thanks, good point, updated.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/vaadin-grid-flow-selection-column.js
vursen: Yeah, I happened upon this too, but I assumed that it's not a big deal. One optimization I was considering is to pass the shift key state through `$server.select` / `$server.deselect`, but that would consequently force us to change the signature of public  `selectFromClient` and `deselectFromClient` methods to accept the second parameter:

https://github.com/vaadin/flow-components/blob/bc3fc8b2eef4143bae3db1063c4826b1ea70b3ac/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/GridSelectionModel.java#L34

Another idea I had was to set a flag before calling `selectFromClient`. However, I wasn't sure enough if it's a safe approach:

```diff
  @ClientCallable
- private void select(String key) {
+ private void select(String key, boolean isShiftKeyDown) {
+     getSelectionModel().setShiftKeyDown(isShiftKeyDown);
      findByKey(String.valueOf(key))
              .ifPresent(getSelectionModel()::selectFromClient);
+     getSelectionModel().setShiftKeyDown(false);
  }
```

vursen: > One way to fix it could be to dispatch an event for this instead and use addEventListener(...).debounce(...) on the server.

I'm not sure I get this approach. What should be the debounce condition?

tomivirkki: > I'm not sure I get your suggestion. What should be the debounce condition?

I was thinking of something like
```java
getElement()
    .addEventListener("shift-key-toggle", this::shiftKeyToggle)
    .debounce(Integer.MAX_VALUE);
```

Maybe there's some more elegant way

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/ClientItemToggleEvent.java
sissbruecker: Looks good, thanks

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/vaadin-grid-flow-selection-column.js
vursen: After some thinking, it seems unnecessary to reset the shift key state after the selection and it's only necessary to ensure that it's updated on the server before `selectFromClient` is called.

sissbruecker (approved):

tomivirkki (approved):

PR #6893, opened 2024-12-06
chore: update snapshot version to 24.7

vursen:


sissbruecker:
Let's maybe ask for review from @ZheSun88 so he is in the loop.

We'll also need a 24.6 branch before merging.

vursen:
> We'll also need a 24.6 branch before merging.

It turns out I mistakenly created 24.7 instead :) It's fixed now!

ZheSun88:
look all good.. 
do you still need changes, as it is still a draft?

ZheSun88 (approved):

PR #6896, opened 2024-12-06
docs: use correct Popover delay default values in JavaDoc

### LINKED ISSUE #6889: [Popover] Hover delay server side default not matching client side default?
### Description

The Javadoc for the Popover says, that the hover delay (`getHoverDelay()/setHoverDelay()`) is 0, but the client side default seems to be a higher value (maybe around 100ms or so).

This means, that, as long a no value is set on the server, the default value on the server and on the client are not in sync. Also the javadoc is incorrect. 

I did not test the other delays, but might be worth to check while fixing this issue, if they also are out of sync.

### Expected outcome

Javadoc and Java default value should match the client side value.

```java
Popover popover = new Popover(new Span("Hello World"));
Button button = new Button("Something");

popover.setTarget(button);
popover.setOpenOnHover(true);
// popover.setHoverDelay(0);
add(button);
```

### Steps to reproduce

1. Add the snippet above
2. Open the view.
3. Hover the button > should take some hundred ms before the popover opens.
4. Active the "setHoverDelay(0)" line of code. 
5. Restart the app / hotswap the code and repeat 3). The Popover should now appear immediately.

### Environment

Vaadin version(s): 24.6.0.beta3
OS: Windows 11

### Browsers

Chrome

web-padawan:
## Description

Fixes #6889

Updated defaults to match the change from https://github.com/vaadin/web-components/pull/7682.

## Type of change

- Documentation

### DIFF
--- a/vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
+++ b/vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
@@ -517,7 +517,7 @@ public class Popover extends Component implements HasAriaLabel, HasComponents,
 
     /**
      * The delay in milliseconds before the popover is opened on target focus.
-     * Defaults to {@code 0}.
+     * When not specified, the default {@code 500} milliseconds delay is used.
      *
      * @param focusDelay
      *            the delay in milliseconds
@@ -528,7 +528,7 @@ public class Popover extends Component implements HasAriaLabel, HasComponents,
 
     /**
      * The delay in milliseconds before the popover is opened on target focus.
-     * Defaults to {@code 0}.
+     * When not specified, the default {@code 500} milliseconds delay is used.
      *
      * @return the delay in milliseconds
      */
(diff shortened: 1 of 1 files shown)


vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
```diff
@@ -528,7 +528,7 @@ public void setFocusDelay(int focusDelay) {
 
     /**
      * The delay in milliseconds before the popover is opened on target focus.
-     * Defaults to {@code 0}.
+     * Defaults to {@code 500}.
```
sissbruecker, line 531: This is still somewhat confusing, because `getFocusDelay` returns `0` by default. [In this PR](https://github.com/vaadin/flow-components/pull/6828) we actually return the default value of the web component as well, though I guess that is not consistent everywhere.

web-padawan: Changed to make it clear that `500` is used when the delay is not specified (this is how the web component works). I'm not sure if we should change the `getFocusDelay()` etc as they also return `0` in Tooltip.

vursen (approved):

sissbruecker (approved):

Issue #6885, opened 2024-12-04
[FEATURE] Use MenuConfiguration directly on SideNav  


MarcinVaadin:
### Describe your motivation

While using `@Menu` on Flow project I would like to put all items in SideNav automatically.

### Describe the solution you'd like

I've expected something short and handy like:

```
List<MenuEntry> menuEntries = MenuConfiguration.getMenuEntries();
SideNav sideNav = SideNav.from(menuEntries);
```

or

```
SideNav sideNav = new SideNav();
List<MenuEntry> menuEntries = MenuConfiguration.getMenuEntries();
menuEntries.forEach(sideNav::add);
```

would work.

### Describe alternatives you've considered

Current working approach requires:

```
SideNav sideNav = new SideNav();
List<MenuEntry> menuEntries = MenuConfiguration.getMenuEntries();
menuEntries.stream().map(menuEntry -> new SideNavItem(menuEntry.title(), menuEntry.path())).forEach(sideNav::addItem);
```

which looks boilerplate.

MarcinVaadin:
Also, icon defined as a String in `@Menu` cannot be used directly in SideNavItem. Prefix component is required.

yuriy-fix:
Would you expect it to work on the client side as well?

MarcinVaadin:
> Would you expect it to work on the client side as well?

I'm writing pure Java demo, didn't consider client side.

sissbruecker:
This isn't trivial to implement currently. Creating items with label and path from the menu configuration is straightforward. However you would also expect it to create an icon if one is specified in `@Menu`. When using iconsets, this would require importing `Icon` in `SideNav`. This has the side effect that the Vaadin iconset would be included in the frontend bundle for everyone using side nav, even if no icons are used.

This may become feasible at some point if `Icon` wouldn't automatically include the Vaadin iconset, and if there would be some way for Flow users to explicitly specify which iconsets to include in the bundle.

PR #6957, opened 2024-12-17
feat: create side nav from menu entries

sissbruecker:
Allows to create a side nav instance from a list of `MenuEntry`s which can be aquired from `MenuConfiguration.getMenuEntries`. Also adds a constructor to `SideNavItem` to create an item from a single `MenuEntry`.

- When an entry has a view class, it detects the path from the route annotation of that class and also adds route aliases
- When an entry has no view class (such as client-side routes), it uses the string path of the entry as path
- Automatically sets either an `Icon` or `SvgIcon` as prefix depending on whether the icon string references an iconset icon or an SVG file

Part of https://github.com/vaadin/flow-components/issues/6885

### DIFF
--- a/vaadin-side-nav-flow-parent/vaadin-side-nav-flow/pom.xml
+++ b/vaadin-side-nav-flow-parent/vaadin-side-nav-flow/pom.xml
@@ -22,6 +22,11 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-icons-flow</artifactId>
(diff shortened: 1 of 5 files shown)


tomivirkki (approved):

sephiroth-j:
Hi @sissbruecker, [this PR was mentioned as part of the 25.7.0 release notes](https://github.com/vaadin/flow-components/releases/tag/24.7.0), although the changes have been reverted. Therefore, it should be removed from the release notes again.

sissbruecker:
Thanks for the heads up, I removed that part from the release notes.

Issue #6974, opened 2024-12-20
[Grid] Allow setting tooltip position


sissbruecker:
### Describe your motivation

When using a tooltip generator in `Grid`, tooltips are always placed in the bottom center of a cell. In some case with wide columns you might want to place the tooltip near the start or end of the cell instead.

![image (1)](https://github.com/user-attachments/assets/a245e3b6-a612-42fe-bb8b-86f71b284d5c)

### Describe the solution you'd like

There should be an API for setting the tooltip position.

PR #6975, opened 2024-12-20
feat: configure tooltip position in Grid

sissbruecker:
Adds a `setTooltipPosition` API to configure the position of tooltips in the grid. For now this applies for tooltips on all cells, it would not be possible to configure individual positions per column.

Closes https://github.com/vaadin/flow-components/issues/6974


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -4804,9 +4805,22 @@ public void setTooltipGenerator(
...
     }
 
+    /**
+     * Sets the tooltip position relative to the cell that is being hovered or
+     * focused. The default position is {@link TooltipPosition#BOTTOM}.
+     *
+     * @param position
+     *            the position to set
+     */
+    public void setTooltipPosition(TooltipPosition position) {
```
vursen, line 4815: Would it make sense to add a getter for consistency?

vursen, line 4815: Setting `null` currently results in `java.lang.NullPointerException`.

sissbruecker: Added a getter. To facilitate this, added a method to the enum to get an enum value from the position string, which returns the bottom position by default.

sissbruecker: Done

vursen (approved):

PR #6979, opened 2024-12-20
feat: Card

jouni:


sissbruecker:
Closing in favor of https://github.com/vaadin/flow-components/pull/7002. I did some cleanup based on the changes here, but could not get the `.DS_Store` files removed, so I started from a new commit without that file.

Issue #6962, opened 2024-12-18
Dashboard overrides client-side state


tomivirkki:
### Description

When making quick-paced changes (moving/resizing widgets) to the Flow Dashboard, the client-side component's state gets temporarily overridden by the server-side updates which manifests as jumpiness:

https://github.com/user-attachments/assets/0ec3ea05-0706-4e0f-b346-d45678c5888d

### Expected outcome

Client-side state should not be overridden by server-side updates if it has been changed before a previous server response finishes.

### Minimal reproducible example

https://vaadin.com/docs/latest/components/dashboard#screen-reader-announcements

### Steps to reproduce

1. Open the example
2. Select a widget
3. Quickly move it around with the arrow keys

### Environment

Vaadin version(s): 24.6

### Browsers

Issue is not browser related

PR #6982, opened 2024-12-20
fix: prevent client updates when handling events from client

sissbruecker:
When synchronizing state changes from the client to the server (moving items, resizing items, removing items), the dashboard component currently schedules a client update that overrides whatever state currently exists in the client. That means that any client-side changes that have been made during a server-roundtrip can temporarily be reverted and would only be restored after the following server roundtrip.

Assuming that both the WC and the Flow component handle an update in the same way, there is no need to synchronize the server state back to the client when handling an update from the client. To do so, this change prevents executing `updateClient` when handling updates from the client.

Fixes https://github.com/vaadin/flow-components/issues/6962


vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/Dashboard.java
```diff
@@ -718,8 +728,10 @@ private void handleItemResizedClientEvent(DomEvent e, String idKey,
         DashboardWidget resizedWidget = getWidgets().stream()
                 .filter(child -> nodeId == child.getElement().getNode().getId())
                 .findAny().orElseThrow();
-        resizedWidget.setRowspan(rowspan);
-        resizedWidget.setColspan(colspan);
+        withoutClientUpdate(() -> {
+            resizedWidget.setColspan(colspan);
+            resizedWidget.setRowspan(rowspan);
+        });
```
sissbruecker, line 734: Changing these implicitly triggers a client update.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/Dashboard.java
```diff
@@ -731,14 +743,13 @@ private void initItemRemovedClientEventListener() {
...
             handleItemRemovedClientEvent(e, idKey);
-            updateClient();
         }).addEventData(idKey);
     }
 
     private void handleItemRemovedClientEvent(DomEvent e, String idKey) {
         int nodeId = (int) e.getEventData().getNumber(idKey);
         Component removedItem = getItem(nodeId);
-        removedItem.removeFromParent();
+        withoutClientUpdate(removedItem::removeFromParent);
```
sissbruecker, line 752: Detaching an item implicitly triggers a client update.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/test/java/com/vaadin/flow/component/dashboard/DashboardItemMoveTest.java
```diff
@@ -83,6 +83,18 @@ public void setDashboardNotEditable_moveWidget_orderIsNotUpdated() {
         Assert.assertEquals(expectedRootLevelNodeIds, getRootLevelNodeIds());
     }
 
+    @Test
+    public void moveWidget_noClientUpdate() {
```
sissbruecker, line 87: The test names are somewhat confusing currently, all of them refer to the case of handling a client event, not doing programmatic updates.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/Dashboard.java
```diff
@@ -64,6 +64,9 @@
 @NpmPackage(value = "@vaadin/dashboard", version = "24.7.0-alpha2")
 public class Dashboard extends Component implements HasWidgets, HasSize {
 
+    private static final ThreadLocal<Boolean> suppressClientUpdates = ThreadLocal
```
vursen, line 67: I'm not sure why this can't just be a boolean instance property. Could you explain please?

sissbruecker: Technically it wouldn't be thread-safe otherwise. The component could be processing a request from the client in one thread which sets the boolean flag to `true`. Meanwhile another thread could make modifications to the component, which would then fail to sync to the client. Not sure if it's a big concern, but it seemed the correct way of doing it.

vursen (approved):

PR #6991, opened 2024-12-30
refactor: do not rely on _hasInputValue for detecting bad input

vursen:
## Description

The PR updates number fields to use `_inputElementValue` instead of `_hasInputValue` for detecting and clearing bad input, which aligns them with other field components that allow entering bad input.

Depends on
- https://github.com/vaadin/web-components/pull/8413

## Type of change

- [x] Refactor

### DIFF
--- a/vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/AbstractNumberField.java
+++ b/vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/AbstractNumberField.java
@@ -22,7 +22,6 @@ import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.function.Function;
 
 import com.vaadin.flow.component.AttachEvent;
-import com.vaadin.flow.component.Synchronize;
 import com.vaadin.flow.component.shared.ClientValidationUtil;
 import com.vaadin.flow.component.shared.ValidationUtil;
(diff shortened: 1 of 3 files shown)


vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/AbstractNumberField.java
```diff
@@ -175,6 +174,9 @@ public void setValueChangeMode(ValueChangeMode valueChangeMode) {
         }
 
         super.setValueChangeMode(valueChangeMode);
+
+        getSynchronizationRegistration()
+                .synchronizeProperty("_inputElementValue");
```
sissbruecker, line 179: Spent some time trying to understand everything works together. What I got so far is that this should synchronize `_inputElementValue` whenever `value` changes. However that wouldn't cover cases where `value` doesn't change, for example when having an empty input and then entering an unparsable value. This can be fixed by synchronizing the property for the `unparsable-change` event listener as well. Also looks like the unit tests will not be sufficient to cover this.

vursen: Thanks, it turns out it was caught by existing ITs as well. Fixed by synchronizing the property on `unparsable-change`.

web-padawan (approved):

PR #7010, opened 2025-01-08
chore: Sync eclipse formatting file with Flow

Artur-:


eclipse/VaadinJavaConventions.xml
```diff
@@ -48,7 +48,7 @@
         <setting id="org.eclipse.jdt.core.formatter.blank_lines_before_new_chunk" value="1"/>
         <setting id="org.eclipse.jdt.core.formatter.insert_space_before_binary_operator" value="insert"/>
         <setting id="org.eclipse.jdt.core.formatter.blank_lines_before_package" value="0"/>
-        <setting id="org.eclipse.jdt.core.compiler.source" value="1.5"/>
+        <setting id="org.eclipse.jdt.core.compiler.source" value="17"/>
```
sissbruecker, line 51: Does this work on previous version branches / can we cherry-pick this?

Artur-: I don't know. Seems like Flow has not picked it to the 23.5 branch

Artur-: I would imagine it only affects how the code is formatted, i.e. that records are formatted properly and not what you need to run the formatter

sissbruecker: My concern is that the formatting between branches would be different then, making automated cherry-picks fail more often. Looks like there aren't a lot of differences from this change though, so maybe that's OK.

Artur-: Maybe just try to pick it to all branches and see if it works

Artur-: https://github.com/vaadin/flow-components/pull/7013

Artur-: Looks like it works fine

sissbruecker (approved):

PR #7018, opened 2025-01-08
chore: Sync eclipse formatting file with Flow (CP: 14.12)

sissbruecker:
Cherry-pick of https://github.com/vaadin/flow-components/pull/7010


vursen (approved):

eclipse/VaadinJavaConventions.xml
```diff
@@ -48,7 +48,7 @@
         <setting id="org.eclipse.jdt.core.formatter.blank_lines_before_new_chunk" value="1"/>
         <setting id="org.eclipse.jdt.core.formatter.insert_space_before_binary_operator" value="insert"/>
         <setting id="org.eclipse.jdt.core.formatter.blank_lines_before_package" value="0"/>
-        <setting id="org.eclipse.jdt.core.compiler.source" value="1.5"/>
+        <setting id="org.eclipse.jdt.core.compiler.source" value="17"/>
```
vursen, line 51: Not sure if this is going to cause any issues, but Flow appears to have `1.8` here.

sissbruecker: To be honest I actually didn't check if that really syncs the file with their respective version branch. We only synced the file in the ´main´ branch and then tried to use the same file in all version branches to avoid potential issues with automated cherry-picks. I think it's fine as is, if the formatter at some point produces something that doesn't compile we can fix it.

PR #7022, opened 2025-01-10
chore: automate formatting of POM files with spotless

sissbruecker:
Configure spotless to also format POM files:
- Use consistent indentation and self-closing tags
- Consistently order config sections and properties
- Sort dependencies by `groupId` and `artifactId`

> [!WARNING]
> Order of Maven dependencies matters: https://github.com/Ekryd/sortpom/wiki/SortDependencies#the-parameter-sortdependencies-can-affect-compilation. I could not find any cases where we had some explicit order, and the project seems to compile just fine. I'd suggest to give this a try and revert if we run into issues.


vursen (approved):
If there are any dependencies that need to be added in a specific order, I'd prefer them to be excluded from the formatter explicitly.

PR #7028, opened 2025-01-10
refactor: extract overlay auto add controller

sissbruecker:
This is another attempt at extracting common behavior for automatically adding overlay-based components to the UI. The previous PR (https://github.com/vaadin/flow-components/pull/6620) slightly modified the auto-removal behavior by running it in `beforeClientResponse`.  That in turn caused a regression in `Dialog`, so the previous PR was reverted.

The reason for delaying the auto-close behavior was that:
- Flow always syncs properties first before running event listeners
- That means the `opened` property change listener in `OverlayAutoAddController` runs first and removes the component from the UI
- Event listeners are not run if the element is not in the UI
- In `ConfirmDialog` that can be a problem because clicking confirm / cancel buttons closes the dialog from the client (`opened` property change) and also sends confirm / cancel events which are then not be processed

As an alternative, this PR changes `ConfirmDialog` to:
- Not sync the `opened` property from client to server
- Instead use an `opened-changed` listener to change the `opened` property on the server
- With that other event listeners are still run, even if the `opened-changed` listener removes the component from the UI - the UI check only happens once before running all listeners
- Behavior should still be the same as before, where `opened-changed` also auto-removed the component


vaadin-confirm-dialog-flow-parent/vaadin-confirm-dialog-flow/src/main/java/com/vaadin/flow/component/confirmdialog/ConfirmDialog.java
```diff
@@ -217,21 +212,16 @@ public Optional<String> getAriaDescribedBy() {
...
-            }
-            if (autoAddedToTheUi && !isOpened()) {
-                getElement().removeFromParent();
-                autoAddedToTheUi = false;
-            }
-        });
+        // Initialize auto-add behavior
+        new OverlayAutoAddController<>(this, () -> true);
+
+        addConfirmListener(event -> close());
```
knoobie, line 222: Are those listener added intentionally? Sounds like a breaking change that every confirm click automatically closes the dialog. 

Example: 
- user clicks confirm
- Application does some expensive task and failed 
- Confirm dialog is closed even tho the task failed and user should try to click confirm again to execute the task again

sissbruecker: Currently all of those close the dialog already: https://github.com/vaadin/web-components/blob/e76c4576c2d1875e16681677bbf829637bf48e4f/packages/confirm-dialog/src/vaadin-confirm-dialog-mixin.js#L435-L451

But I was thinking about using a different approach as well, it seems better to stick with the `opened-changed` event than make assumptions what individual actions would do.

knoobie: Oh thanks! Didn't know the web component does that already (I'm using my own confirm dialog based on the normal dialog API) 

Interesting.. another thing to keep in mind when using the official confirm dialog.

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/test/java/com/vaadin/flow/component/shared/internal/OverlayAutoAddControllerTest.java
```diff
@@ -0,0 +1,243 @@
...
+        fakeClientResponse();
+
+        component.setOpened(false);
+
+        Assert.assertEquals(parent.getElement(),
+                component.getElement().getParent());
+    }
+
+    @Test
+    public void close_reopenBeforeClientResponse_notAutoRemoved() {
```
vursen, line 163: **nit:** Based on my understanding, it doesn't auto remove but rather removes and adds it back.

```suggestion
    public void close_reopenBeforeClientResponse_autoRemovedAndAutoAdded() {
```

vaadin-notification-flow-parent/vaadin-notification-flow/src/main/java/com/vaadin/flow/component/notification/Notification.java
```diff
@@ -251,18 +245,8 @@ private void initBaseElementsAndListeners() {
...
-    /**
-     * Removes the notification from its parent if it was added automatically.
-     */
-    private void removeAutoAdded() {
-        if (autoAddedToTheUi && !isOpened()) {
-            autoAddedToTheUi = false;
-            getElement().removeFromParent();
-        }
+        // Initialize auto add behavior
+        new OverlayAutoAddController<>(this);
```
vursen, line 249: This change brings a bonus benefit: it fixes an issue where Notification required an extra server visit to remove the auto-added notification element after calling `close()` on the server.

vursen (approved):

PR #7029, opened 2025-01-13
chore: add shared Vite config

vursen:
## Description

Enabling `useLocalWebComponents` for each component individually can be tedious. This PR introduces a shared Vite config to allow you to add Vite plugins that will apply to all components automatically.

## Type of change

- [x] Internal

### DIFF
--- a/shared/shared-vite-config.ts
+++ b/shared/shared-vite-config.ts
@@ -7,10 +7,10 @@ export const mergeConfigs = (...configs: UserConfigFn[]) => {
 
 export const sharedConfig: UserConfigFn = (env) => ({
   plugins: [
-    // Use local version of web-components, disabled by default
-    // To use this un-comment the lines below and change the path to
-    // the absolute path of your web-components repo's node_modules
-    // folder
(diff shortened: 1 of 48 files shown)


sissbruecker (approved):

shared/shared-vite-config.ts
```diff
@@ -0,0 +1,17 @@
...
+  return configs.reduce((acc, config) => mergeConfig(acc, config));
+};
+
+export const sharedConfig: UserConfigFn = (env) => ({
+  plugins: [
+    // Use local version of web-components, disabled by default.
+    // To use this, uncomment the lines below and change the path
+    // to your local web-components folder if needed (absolute or
+    // relative to this shared config).
+    // DO NOT COMMIT THESE CHANGES!
```
tomivirkki, line 14: We could consider using an environment variable for the path so you wouldn't need to make local changes

vursen: Good idea. Done: https://github.com/vaadin/flow-components/pull/7035

PR #7035, opened 2025-01-13
chore: use .env config to define path to local web components

vursen:
## Description

Use the `.env`  config to define the path to the local web components repository to prevent the path from being accidentally committed to Git.

## Type of change

- [x] Internal

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,6 @@
+
+# To run integration tests with a local version of web components:
+# 1. Clone and install web-components repository
+# 2. Uncomment LOCAL_WEB_COMPONENTS_PATH below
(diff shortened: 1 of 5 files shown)


web-padawan (approved):
Let's also add `.env` file to `.gitignore`.

vursen:
> Let's also add .env file to .gitignore.

Good catch, done.

tomivirkki (approved):

.env.example
tomivirkki, line 1: Should we merge this to `.env.dist`?

tomivirkki: Oh, was looking at the wrong repo, nevermind :)

Issue #7009, opened 2025-01-07
setDisableOnClick sometimes is triggered after a click listener is executed


slargollo:
### Description

After I updated my project from version 24.5.1 to 24.6.0, setDisableOnClick sometimes is triggered after the handler of a click listener is executed and the button keeps disabled when setEnabled(true) is called in the listener.
This was not happening in version 24.5.1.

### Expected outcome

A button with setDisableOnClick(true) should be always enabled when setEnabled(true) is called in a click listener after some operation is executed.

### Minimal reproducible example

In the code bellow, sometimes the buttons never gets enabled again after a few clicks.  

        // Use TextField for standard text input
        TextField textField = new TextField("Your name");
        textField.addClassName("bordered");

        // Button click listeners can be defined as lambda expressions
        Button button = new Button("Say hello", e -> {
            add(new Paragraph(service.greet(textField.getValue())));
            e.getSource().setEnabled(true);
        });

        // Theme variants give you predefined extra styles for components.
        // Example: Primary button has a more prominent look.
        button.addThemeVariants(ButtonVariant.LUMO_PRIMARY);

        button.setDisableOnClick(true);

        // Use custom CSS classes to apply styling. This is defined in
        // styles.css.
        addClassName("centered-content");

        add(textField, button);

### Steps to reproduce

Modify the Hello World Project of version 24.6.0 and add the code bellow. The button will stay disabled after a few clicks.

       Button button = new Button("Say hello", e -> {
            add(new Paragraph(service.greet(textField.getValue())));
            e.getSource().setEnabled(true);
        });

       button.setDisableOnClick(true);

Example:

![Screenshot from 2025-01-07 16-01-12](https://github.com/user-attachments/assets/c99a20ab-1fc0-4a8b-a6e3-3e96d3550487)

### Environment

Vaadin version(s): 24.6.0
OS: Linux 6.8.0-51-generic (Ubuntu 24.04.1 LTS)
Chrome Version 131.0.6778.204 (Official Build) (64-bit)

### Browsers

Chrome, Firefox

slargollo:
This still happens in 24.6.1

sissbruecker:
Unfortunately I couldn't reproduce this. Could you maybe try adding a `Thread.sleep` in the event listener and see if that resolves the issue? I'm mainly trying to figure out if there is a timing issue.

[...]

PR #7045, opened 2025-01-14
fix: replace requestAnimationFrame with global click listener

vursen:
## Description

This PR replaces the requestAnimationFrame approach with a global click listener for disabling components when `disableOnClick` is set to true. This ensures the disabling happens synchronously, avoiding race conditions with server requests, and only after all other click listeners have been executed, allowing the context menu to properly close after a disable-on-click item was clicked.

Fixes https://github.com/vaadin/flow-components/issues/7009

## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/resources/META-INF/resources/frontend/disableOnClickFunctions.js
+++ b/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/resources/META-INF/resources/frontend/disableOnClickFunctions.js
@@ -1,14 +1,6 @@
-function disableOnClickListener({currentTarget: element}) {
-  if (element.disableOnClick) {
-    requestAnimationFrame(() => element.disabled = true);
+document.addEventListener('click', (event) => {
+  const target = event.composedPath().find((node) => node.disableOnClick);
+  if (target) {
+    target.disabled = true;
   }
-}
-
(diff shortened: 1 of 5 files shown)


sissbruecker:
I think the regression was due to a new feature in 24.6, so picking the change to 24.5 wouldn't be necessary?

sissbruecker (approved):

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/resources/META-INF/resources/frontend/disableOnClickFunctions.js
```diff
@@ -1,14 +1,6 @@
-function disableOnClickListener({currentTarget: element}) {
-  if (element.disableOnClick) {
-    requestAnimationFrame(() => element.disabled = true);
+document.addEventListener('click', (event) => {
```
knoobie, line 1: Out of curiosity: wouldn't this be executed for every button I'm going to add on a page.. so I could technically end up with 100 listener that do exactly the same on every click? Could this also create a memory problem with the exact same reasoning because it's not removed - or am I overthinking?

knoobie:
@vursen Do you happen to know a schedule for this to be released in 24.6? I'm not keen to upgrade our applications without this bugfix because we heavily rely on disableOnClick.

PR #7053, opened 2025-01-20
chore: Make Spreadsheet vaadin dependencies provided

TatuLund:
These should be provided in order to avoid conflicts


yuriy-fix:
What is the particular use-case where you are facing conflicts? Is there a need for different version of spreadsheet and other components? If so, why?
Asking primarily because of the `provided` scope for the components, while `flow-html-components` and `flow-server` can be aligned with others.

PR #6815, opened 2024-11-12
feat: implement disable on click for menu item

ugur-vaadin:
## Description

This PR

- implements disable on click behaviour for `MenuItem` with the following API:
  - `void setDisableOnClick(boolean disableOnClick)`
  - `boolean isDisableOnClick()`
- extracts disable on click logic to a controller `DisableOnClickController`

Fixes #1472

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [x] I have read the contribution guide: https://vaadin.com/docs/latest/contributing/overview
- [x] I have added a description following the guideline.
- [x] The issue is created in the corresponding repository and I have referenced it.
- [x] I have added tests to ensure my change is effective and works as intended.
- [x] New and existing tests are passing locally with my change.
- [x] I have performed self-review and corrected misspellings.
- [x] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.

PR #7071, opened 2025-01-28
fix: restore disableOnClick attribute

sissbruecker:
## Description

https://github.com/vaadin/flow-components/pull/6815 extracted the disable on click logic into a controller and as part of that changed the implementation to use a property instead of an attribute. However, customers have been using the attribute for targeting styles (https://github.com/vaadin/web-components/issues/842#issuecomment-1007331620), which now doesn't work anymore. 

This reverts the implementation back to use an attribute.

## Type of change

- Bugfix

### DIFF
--- a/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/internal/DisableOnClickController.java
+++ b/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/internal/DisableOnClickController.java
@@ -77,9 +77,9 @@ public class DisableOnClickController<C extends Component & HasEnabled>
     public void setDisableOnClick(boolean disableOnClick) {
         this.disableOnClick = disableOnClick;
         if (disableOnClick) {
-            component.getElement().setProperty("disableOnClick", "true");
+            component.getElement().setAttribute("disableOnClick", "true");
         } else {
-            component.getElement().removeProperty("disableOnClick");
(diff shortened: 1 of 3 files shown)


vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/internal/DisableOnClickController.java
```diff
@@ -77,9 +77,9 @@ public DisableOnClickController(C component) {
     public void setDisableOnClick(boolean disableOnClick) {
         this.disableOnClick = disableOnClick;
         if (disableOnClick) {
-            component.getElement().setProperty("disableOnClick", "true");
+            component.getElement().setAttribute("disableOnClick", "true");
```
sissbruecker, line 80: Technically could just set an empty string, but `Button` was specifically using a value of `true` before as well.

vursen: As we are reverting to attributes, I would suggest using the name `disableonclick`, which, although is the same, better aligns with the outcome.

sissbruecker: Done

vursen (approved):

PR #7103, opened 2025-02-07
refactor: use Flow element API for setting dimensions

sissbruecker:
## Description

- Replace `executeJs` call with `Element.setProperty` API
- Fix `getWidth` / `getHeight` getters returning wrong value
- Replace ITs with unit tests, as logic is delegated to web component, and detach / reattach scenarios don't need to be tested anymore

## Type of change

- Refactoring


sissbruecker:
The `executeJs` call was previously using a method and as such necessary. In https://github.com/vaadin/flow-components/pull/4470/commits/8f40296b72457c0f56471fb8b3189016b2dd3e40 it was changed to use a property though, so since then it's not necessary anymore.

DiegoCardoso (approved):

web-padawan (approved):

PR #7111, opened 2025-02-11
chore: mark the Material theme as deprecated

jouni:
Related to https://github.com/vaadin/web-components/issues/8580


sissbruecker (approved):

knoobie:
I would suggest to update the annotation so that people get "nifty" build warnings 🙂 if you are already certain that material is going to be removed in 25 - let's be honest and inform people "aggressively"

`@Deprecated(since = "24.7", forRemoval = true) `

jouni:
Thanks, @knoobie, for the suggestion. I admit I don't know what the practice has been with these deprecation annotations. I simply looked copied this from one previous deprecation, which didn't have the `since` or `forRemoval` attributes.

PR #7120, opened 2025-02-16
feat: allow passing HierarchyColumnComponentRenderer to addComponentHierarchyColumn

### LINKED ISSUE #7107: Open the TreeGrid API to allow for instance matching on custom classes
### Describe your motivation

The Grid component allows the user to pass in a Renderer or ValueProvider of a certain subtype when calling "addColumn". That way users can get the Renderer or ValueProvider and match the subtype with the "instanceof"-Operator. 

The TreeGrid-Class has no way of adding a HierarchyColumn by passing in a HierarchyColumnComponentRenderer (it always creates a new one with the given ValueProvider). It would be nice to have an overloaded method which takes the Renderer directly so one can do type matching later on.

This method (line 682 in TreeGrid as of writing):

```
    public <V extends Component> Column<T> addComponentHierarchyColumn(
            ValueProvider<T, V> componentProvider) {
        return addColumn(new HierarchyColumnComponentRenderer<V, T>(
                componentProvider, this).withProperty("children",
                        item -> getDataCommunicator().hasChildren(item)));
    }
```

could turn into the following 2 methods:
```
    public <V extends Component> Column<T> addComponentHierarchyColumn(
            ValueProvider<T, V> componentProvider) {
        return addColumn(new HierarchyColumnComponentRenderer<V, T>(
                componentProvider, this);
    }

    public <V extends Component> Column<T> addComponentHierarchyColumn(
            HierarchyColumnComponentRenderer<V, T> componentRenderer) {
        return addColumn(componentRenderer.withProperty("children",
                        item -> getDataCommunicator().hasChildren(item)));
    }
```

As far as I can tell there would be no impact for existing codebases and it would improve the usability for more advanced use cases.

I have looked at "addHierarchyColumn" aswell the same issue exists with the given ValueProvider. It is wrapped inside another function that (as far as I can tell) ensures that null can't be returned. Maybe upon further inspection one could find a similar solution for said method.

_No response_

### Additional context

_No response_

MikeV2000:
## Description

As mentioned in the issue, unlike with the Grid you can not pass a Renderer into a TreeGrid to add a Column. To enhance the API and make it possible to do type matching on Renderers passed into the method, I have overloaded the "addComponentHierarchyColumn"-method.

Fixes #7107 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [] I have not completed some of the steps above and my pull request can be closed immediately.

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
@@ -663,7 +663,6 @@ public class TreeGrid<T> extends Grid<T>
         return column;
     }
     
-
     /**
      * Adds a new Hierarchy column that shows components.
      * <p>
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
@@ -662,7 +662,7 @@ public class TreeGrid<T> extends Grid<T>
 
         return column;
     }
-    
+
     /**
      * Adds a new Hierarchy column that shows components.
      * <p>
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
@@ -681,7 +681,7 @@ public class TreeGrid<T> extends Grid<T>
      */
     public <V extends Component> Column<T> addComponentHierarchyColumn(
             ValueProvider<T, V> componentProvider) {
-        return addColumn(new HierarchyColumnComponentRenderer<V, T>(
+        return addColumn(new HierarchyColumnComponentRenderer<>(
                 componentProvider, this));
     }
 
@@ -692,8 +692,8 @@ public class TreeGrid<T> extends Grid<T>
      * built in renderers.
      * </p>
      *
-     * @param componentProvider
-     *            the renderer used to create the grid cell structure
+     * @param componentRenderer
+     *            the renderer used to create a component for the given item
      * @param <V>
      *            the component type
      * @return the new column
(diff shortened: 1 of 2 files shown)


yuriy-fix:
Dear @MikeV2000 , would you be able to add a test covering this change? Guidelines for creating tests: https://vaadin.com/docs/latest/contributing/tests. If that is inconvenient for you, we will evaluate if we can proceed on our own and finalize the contribution.

MikeV2000:
> Dear @MikeV2000 , would you be able to add a test covering this change? Guidelines for creating tests: https://vaadin.com/docs/latest/contributing/tests. If that is inconvenient for you, we will evaluate if we can proceed on our own and finalize the contribution.

Hi, I have added a Unit Test testing the expected behaviour. I was not sure about the use of inner classes and interfaces in Test cases so feel free to make any changes to make it fit the "vaadin style" :).

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -682,7 +682,27 @@ public Column<T> addHierarchyColumn(ValueProvider<T, ?> valueProvider) {
...
+     *            the renderer used to create the grid cell structure
+     * @param <V>
+     *            the component type
+     * @return the new column
+     * @see #addColumn(Renderer)
+     * @see #removeColumn(Column)
+     */
+    public <V extends Component> Column<T> addComponentHierarchyColumn(
+            HierarchyColumnComponentRenderer<V, T> componentRenderer) {
+        return addColumn(componentRenderer.withProperty("children",
```
sissbruecker, line 705: I'm not sure if I understood what you want to accomplish by providing the renderer instance. Is it that you want to get the renderer from the column afterwards via `column.getRenderer()` and do some checks based on the result? Can you elaborate a bit in which use cases that would be beneficial? Is there some other mechanism to mark a column so that you can identify it without having to reference the type of the renderer?

MikeV2000: So in our project we have a implemented a custom column based filtering for Grids and TreeGrids. It works by typing a String into a TextField and then using the Renderer associated with the column and TextField to get some sort of String representation of that cell.

For specific cases we don't want any basic representation but a certain "SearchValue". What we do with Grids is we pass in a Renderer that implements the "ISearchable"-Interface, which has a ValueProvider for a search String to be matched with the given filter.

sissbruecker: Mhh OK. There might be other use cases such as creating reusable renderers and the change seems harmless otherwise.

sissbruecker (approved):

sissbruecker:
@MikeV2000 It turns out the change wasn't really necessary. The `withProperty` call that was part of the new overload was not needed as that already happens in the renderer's constructor. With that the implementation just became `addColumn(renderer)`. So you can call this method directly instead. I'm going to revert this change in https://github.com/vaadin/flow-components/pull/7128.

PR #7133, opened 2025-02-20
Added API to configure the filtering debounce timeout

mstahv:
PoC created based on suggestion by @Aleksey2093

Number of different issues related to topic, like #6863

Opening as draft as I can't get the flow-components project to compile.

## Description

Please list all relevant dependencies in this section and provide summary of the change, motivation and context.

Fixes #6863

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [ ] I have read the contribution guide: https://vaadin.com/docs/latest/contributing/overview
- [ ] I have added a description following the guideline.
- [ ] The issue is created in the corresponding repository and I have referenced it.
- [ ] I have added tests to ensure my change is effective and works as intended.
- [ ] New and existing tests are passing locally with my change.
- [ ] I have performed self-review and corrected misspellings.
- [x] I have not completed some of the steps above and my pull request can be closed immediately.

#### Additional for `Feature` type of change

- [ ] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.

Issue #6863, opened 2024-11-25
Vaadin ComboBox Filter timeout


Aleksey2093:
https://github.com/vaadin/flow-components/blob/b2fe10c92ef9812e9e9ca5f045e6e459d0ae4cc7/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js#L92-L99

I have a very long query for DataProvider. It is necessary that the filter does not react immediately when entering and allows the user to enter more characters. 500 milliseconds is very short. You need to be able to change this parameter using "getElement().setProperty(...."

vaadin 23.3.33

web-padawan:
Related to #1031

Aleksey2093:
```java
            let timerValue = 1000
            if (comboBox.filterDebouncerTimer) {
              timerValue = comboBox.filterDebouncerTimer
            }
            this._filterDebouncer = Debouncer.debounce(this._filterDebouncer, timeOut.after(timerValue), () => {
```
Can it be done like this?

mstahv:
Did you @Aleksey2093 try that solution?

mstahv:
To me that would sound like a good enough fix. The "components team" has a tendency to keep "client side API" in sync,  but not sure if there is this kind of lazy loading for the web component 🤷‍♂️ The web component docs mentions lazy loading, but not sure if that API has similar throttling as there is in the Flow version (looks like that logic is in the Flow specific "connector.js") 🤷‍♂️

If you have the patch in your local checkout, I'd just shoot a pull request to get wheels rolling. If you don't, I can try to help as well.

mstahv:
I couldn't resist drafting this as PR. The problem was that the project don't compile for me (related to some other component) 🤷‍♂️

```

[INFO] Scanning for projects...
[ERROR] Some problems were encountered while processing the POMs
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]   
[ERROR]   The project [inherited]:vaadin-flow-components:pom:24.7-SNAPSHOT (/Users/mstahv/git/flow-components/pom.xml) has 2 errors
[...]

PR #7135, opened 2025-02-21
Avoid obsolete filter syncing [2 days]

mstahv:
That appears to be obsolete as the filter is anyways sent to the @ClientCallable method when the filtering should actually be done. After this change there is a lot less XHRs/server-visits happening  aka reduces "chattiness" (although there still is some that I don't really know what they are related to (opened changed and some "confirm" requests).

Opening as draft as somebody involved in the current implementation really should look into this (and couldn't run the tests locally).

See also #6863 and #7133

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyComboBoxFilterIT.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyComboBoxFilterIT.java
@@ -15,10 +15,14 @@
  */
 package com.vaadin.flow.component.combobox.test;
 
-import org.junit.Assert;
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+
 import org.junit.Test;
 import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.ui.ExpectedConditions;
+import org.openqa.selenium.support.ui.WebDriverWait;
 
 import com.vaadin.flow.component.combobox.testbench.ComboBoxElement;
 import com.vaadin.flow.testutil.TestPath;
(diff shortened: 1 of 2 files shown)


mstahv:
One failed test. Which is kind of more a timing issue. In a mode when the popup is configured NOT to be automatically openend, the previous test triggered the filtering more eagerly. Don't know what would be the "correct" behaviour if #6863 gets fixex.

Could be fixed either by adding a wait/timeout to the IT or to keep syncing the filter also via event, but make that lazy. Or changing client side so taht by making open event also contain the filter or so that connector calls some method instead (like it does with regular filtering).

yuriy-fix:
Let's consider combining this PR with #7133 so that the context is available in one place. We should also proceed with adding tests.

@mstahv, would you be able to add tests, or would you like us to handle them? Also, please specify your compilation issues so we can help resolve them.

It might be potentially a small breaking change, since `addPropertyChangeListener` for filter will stop working after removing `@Synchronize`

mstahv:
I'd keep these separatly. This is more like a bugfix or performance improveiment and the other is technically "a feature", so it will probably be require a minor version (and now that 24.7 was just "closed", it will take a looong time).

Would be great if you can tackle the possible new tests.  As I don't work with selenium daily, I don't have good ideas to how to automatically assert extra requests are made.

vursen (commented):
There are failing integration tests with this change. Also, the actual filter logic seems to be more complex than what this PR assumes e.g. the filter value may need to be reset in certain scenarios, see ComboBoxDataController.

mstahv:
@vursen Are you going to work on this further?

vursen:
> Are you going to work on this further?

No, I'm tied up with other tasks right now.

mstahv:
Checked in the project now builds on my local workstataion and tests pass there. Pushed a blind wait that might happen now as the filtering is debounced. Let's see if that helps.

Please explain your concern about ComboBoxDataController. That looks very complex class indeed, but didn't easily see how this could affect that.

vursen:
> Please explain your concern about ComboBoxDataController. That looks very complex class indeed, but didn't easily see how this could affect that.

For example, consider this scenario:

1. `comboBox.setItems(yourDataProvider)`
2. `comboBox.getLazyDataView().getItems()` returns all items (expected)
3. Open, filter, then close the combo box
6. `comboBox.setItems(yourDataProvider)`
7. `comboBox.getLazyDataView().getItems()` returns filtered items because the filter wasn't cleared (regression)

It would help if ComboBoxDataController was the only place handling filters.

mstahv:
Do you have that available somewhere as regression test? I can't see any failing. Or the original thing that the developer/user is trying to do in that scenario? To me it sounds bit odd the filter is ever taken into account with the dataview (although I haven't really understood what is the purpose of the dataview in general either).

The fix is probably easy, but maybe there is some more fundamental isseu in the current implementation 🤷‍♂️

yuriy-fix:
We spent some time investigating the proposed change and identified several cases that are not currently covered. Addressing them would require significant additional work to complete the solution.

For now, we've decided not to proceed with the change and will close the PR. Please feel free to reopen it if you would like to address the review comments.

PR #7141, opened 2025-02-21
chore: format JS and TS files with spotless

sissbruecker:
## Description

Configures spotless to also format JS + TS files using the existing prettier config.

Also reverts https://github.com/vaadin/flow-components/pull/6986. Not strictly necessary, but I wasn't a fan of the fact that I couldn't commit Java files anymore after deleting `node_modules`. Besides, most contributors probably wouldn't think of running `npm install` to install the Git hook, whereas they have to run spotless to get a PR green.

## Type of change

- Internal

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -558,6 +558,18 @@
                             <file>${spotless.licence-header}</file>
                         </licenseHeader>
                     </java>
+                    <typescript>
+                        <includes>
+                            <include>src/**/*.js</include>
+                            <include>src/**/*.ts</include>
+                        </includes>
+                        <excludes>
+                            <exclude>src/**/SpreadsheetSingleScriptTemplate.js</exclude>
+                        </excludes>
+                        <prettier>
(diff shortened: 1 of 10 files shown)


pom.xml
sissbruecker, line 1: Looks like it can't find a Node installation on the CI server. Maybe @ZheSun88 can shed some light on this?

Did take a look at how the Copilot team handles it, but they are using Github actions, so not comparable.

sissbruecker: Also interesting that the formatter check passed for the previous commit.

vursen (approved):
Good idea to add it to spotless!

vursen:
Once finished, let's consider backporting it to previous versions because `lint-staged` was backported there as well.

pom.xml
ZheSun88: seems that is an random issue on that certain agent... 
i have rerun it again.. now it is green..

PR #7148, opened 2025-02-25
feat: Implement colors in custom formatting for vaadin-spreadsheet

### LINKED ISSUE #6835: Vaadin Flow Spreadsheet Conditional Formatting dataFormat color support feature
### Describe your motivation

We have a possibility to use `dataFormat` within Conditional Formatting to change value's color in the cell.
Apache POI support it but Spreadsheet is not.

A good solution which I would to apply is to use `dataFormat` as string value as: `#,##0_);[Red](#,##0)` - which means if value less than zero then apply red color and parentheses.

The only solution for this time is Java API approach (e.g.: SheetConditionalFormatting.java). It's good but with `dataFormat` it will be more laconically.

### Additional context

Link to related issue: https://github.com/vaadin/flow-components/issues/6819

thevaadinman:
## Description

Fixes #6835. Does not introduce new public API. Enables the use of named colors and indexed colors
compatible with Excel: black, white, red, green, blue, yellow, magenta, cyan, as well as Color 1-56.

## Type of change

- [x] Feature

## Checklist


thevaadinman:
SonarCloud Code Analysis result can be ignored, it's hallucinating and complaining about existing code.
Submitted code now looks quite ugly after application of mandatory "beautification" step, which I suggest be dropped or tuned for the future.

vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/org/apache/poi/ss/format/CellFormatPart.java
```diff
@@ -0,0 +1,762 @@
...
+ * See {@literal <https://vaadin.com/commercial-license-and-service-terms>} for the full
+ * license.
+ */
+package org.apache.poi.ss.format;
+
+/* 
+ * The Vaadin CI pipeline requires Vaadin's license header to be first in the file.
+ * While this file is modified, most of it was part of Apache POI, which is licensed
+ * under the Apache License, Version 2.0.
+ */
```
sissbruecker, line 15: Let's exclude this class from the formatter to preserve the license header and not add our own:
https://github.com/vaadin/flow-components/blob/695e072e503425302dba6372df8136b1bea4035d/pom.xml#L547-L550

thevaadinman: I'll also return the file to its original unmangled state

thevaadinman:
Looks like I'll have to shim the custom cell formatting bits. This will take a while.

thevaadinman:
There seems to be a hard-to-resolve issue with getting the reworked CellFormatPart class to load reliably. All tests thus far have worked locally, but something seems very different on the CI cluster.

thevaadinman:
This set of changes adds a field called textColor to CellData and modifies the client side code to propagate the value of that field to where it needs to go, so that the CSS `color` attribute can be set when needed. That field gets set by CustomDataFormatter, which uses the POI CellFormat class to interpret the format string and extract the assigned color value.

In order to properly support the indexed colors used by Excel and Google Sheets, the class `org.apache.poi.ss.format.CellFormatPart` was modified to properly identify and map the color values. This was sufficient to achieve color support in testing, but when run on the CI cluster, the overloaded class was not picked up correctly, and necessitated renaming the customized version of the class to `VCellFormatPart` and including a large chunk of POI code modified to point to that class, while remaining compatible with the rest of POI in use.

In addition to the functional changes and tests, the root POM was modified in order to allow turning the formatter off and on as needed, and to ignore POI code.

This does create the potential to make future POI upgrades more difficult until the corrected functionality of the CellFormatPart class has been successfully submitted to Apache POI upstream, after which the V-prefixed POI classes can be removed and the main code returned to compatibility with mainline.

thevaadinman:
Appeasing SonarCloud Code Analysis is beyond the scope of this feature.

yuriy-fix:
Closing in favour of component factory release.

PR #7163, opened 2025-02-28
refactor: preload items on grid scroll to index

TatuLund:
Back-port of https://github.com/vaadin/flow-components/pull/5459

Make Grid's `scrollToIndex` method preload the items for the target index so another server roundtrip wouldn't be needed for fetching the items.

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
@@ -4036,7 +4036,7 @@ public class Grid<T> extends Component implements HasDataProvider<T>, HasStyle,
         // Preload the items
         setRequestedRange(targetPageStartIndex, preloadedItemsCount);
 
-        // Scroll to the requested index            
+        // Scroll to the requested index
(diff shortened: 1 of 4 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
sissbruecker, line 1: I tested the following: On view initialization, set 1000 items and scroll to index 500.

This doesn't work as intended:
- Server sends items 0..50 to the connector, so it does not preload the requested range from scroll to index
- Grid briefly shows items 0..50
- Then it scrolls to 500
- Connector requests items 450..550
- Round-trip fetches data, grid now renders items around scroll position

That happens both in this PR and on `main`.

I also tested the integration test case. There it prefetches the page containing the newly added item, but not the items before that so it still requires another roundtrip.

What specifically is this trying to solve? Is there a case where this works properly?

sissbruecker: The original PR only handles scrolling to an index outside of the Grid's initialisation, so the scenario above is not covered currently. However the change may still be useful when calling `scrollToIndex` outside of the grid's initialisation.

sissbruecker (approved):

Issue #7162, opened 2025-02-28
[FormLayout] Add Flow API for new autoResponsive mode


vursen:
Implement the following API to support the new autoResponsive mode:

- [ ] `setAutoResponsive(boolean)`
- [ ] `setColumnWidth(String)`
- [ ] `setLabelWidth(String)`
- [ ] `setMaxColumns(int)`
- [ ] `setExpandColumns(boolean)`
- [ ] `setExpandFields(boolean)`
- [ ] `setLabelPosition(LabelPosition)` (or `setLabelsAside(boolean)`?)

Add a new component `FormRow`, similar to `FormItem`.

PR #7173, opened 2025-03-06
feat(FormLayout): add `FormRow` class to `FormLayout`

DiegoCardoso:
## Description

Add a new `FormRow` class that maps to the `<vaadin-form-row>` web component introduced as part of the new `autoResponsive` mode. The class contains a few helper methods to add `FormItem` elements as child of the row.

Part of #7162

## Type of change

- [ ] Bugfix
- [X] Feature

### DIFF
--- a/vaadin-form-layout-flow-parent/vaadin-form-layout-flow/src/main/java/com/vaadin/flow/component/formlayout/FormLayout.java
+++ b/vaadin-form-layout-flow-parent/vaadin-form-layout-flow/src/main/java/com/vaadin/flow/component/formlayout/FormLayout.java
@@ -280,17 +280,13 @@ public class FormLayout extends Component
     }
 
     /**
-     * <p>
      * Server-side component for the {@code <vaadin-form-row>} element. Used to
      * arrange fields into rows inside a {@link FormLayout} when
      * {@code #setAutoResponsive(boolean)} is enabled.
-     * </p>
      * <p>
-     * Each {@code <vaadin-form-row>} always starts on a new row. Fields that
+     * Each {@link FormRow} always starts on a new row. Fields that
      * exceed the available columns wrap to a new row, which then remains
-     * reserved exclusively for the fields of that {@code <vaadin-form-row>}
-     * element. wrap components for display in a {@link FormLayout}.
-     * </p>
+     * reserved exclusively for the fields of that {@link FormRow}.
      *
      * @author Vaadin Ltd
      */
--- a/vaadin-form-layout-flow-parent/vaadin-form-layout-flow/src/main/java/com/vaadin/flow/component/formlayout/FormLayout.java
+++ b/vaadin-form-layout-flow-parent/vaadin-form-layout-flow/src/main/java/com/vaadin/flow/component/formlayout/FormLayout.java
@@ -282,11 +282,26 @@ public class FormLayout extends Component
     /**
      * Server-side component for the {@code <vaadin-form-row>} element. Used to
      * arrange fields into rows inside a {@link FormLayout} when
-     * {@code #setAutoResponsive(boolean)} is enabled.
+     * {@link FormLayout#setAutoResponsive(boolean) auto-responsive mode} is
+     * enabled.
      * <p>
-     * Each {@link FormRow} always starts on a new row. Fields that exceed the
-     * available columns wrap to a new row, which then remains reserved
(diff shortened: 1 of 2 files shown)


vaadin-form-layout-flow-parent/vaadin-form-layout-flow/src/main/java/com/vaadin/flow/component/formlayout/FormLayout.java
```diff
@@ -279,6 +279,64 @@ public void remove(Component... components) {
...
+     * </p>
+     *
+     * @author Vaadin Ltd
+     */
+    @Tag("vaadin-form-row")
+    @NpmPackage(value = "@vaadin/polymer-legacy-adapter", version = "24.7.0-beta1")
+    @JsModule("@vaadin/polymer-legacy-adapter/style-modules.js")
+    @NpmPackage(value = "@vaadin/form-layout", version = "24.7.0-beta1")
+    @JsModule("@vaadin/form-layout/src/vaadin-form-row.js")
+    public static class FormRow extends Component
```
knoobie, line 302: Food for thoughts: How about using `Composite` instead of `Component`? 

In the past.. people often used APIs that were never intended for a flow-component to be used (e.g. Tab(s) and add()) so instead of providing all possible methods from the Component class - how about using Composite so that you can publish only those methods you guys really wanna support on newly added "components"?

DiegoCardoso: That's a good question. I'll see if it makes sense for us in this case, though. `FormRow` is basically a wrapper component that we use to ensure that the first child element is "forced" into a new row. But there could be some API we do not intend to support, so it's worth checking.

vursen: While I agree with the point in general, I'm not sure if FormRow would benefit from this. Since it's mostly just a wrapper with no specific content requirements, as Diego mentioned, it seems natural for it to be a Component that can hold anything.

vaadin-form-layout-flow-parent/vaadin-form-layout-flow/src/main/java/com/vaadin/flow/component/formlayout/FormLayout.java
```diff
@@ -279,6 +279,86 @@ public void remove(Component... components) {
...
+     *
+     * TextArea addressField = new TextArea("Address");
+     * formLayout.setColspan(addressField, 2);
+     * formLayout.addFormRow(addressField);
+     * </pre>
+     *
+     * @author Vaadin Ltd
+     */
+    @Tag("vaadin-form-row")
+    @NpmPackage(value = "@vaadin/polymer-legacy-adapter", version = "24.7.0-beta1")
```
web-padawan, line 309: Should be updated to 24.8

vaadin-form-layout-flow-parent/vaadin-form-layout-flow/src/main/java/com/vaadin/flow/component/formlayout/FormLayout.java
```diff
@@ -279,6 +279,86 @@ public void remove(Component... components) {
...
+    @JsModule("@vaadin/polymer-legacy-adapter/style-modules.js")
+    @NpmPackage(value = "@vaadin/form-layout", version = "24.7.0-beta1")
+    @JsModule("@vaadin/form-layout/src/vaadin-form-row.js")
+    public static class FormRow extends Component
+            implements HasComponents, HasStyle {
```
web-padawan, line 314: nit: the Component already implements HasStyle since https://github.com/vaadin/flow/pull/15742.

web-padawan (commented):
Overall LGTM, let's update the `@NpmPackage` version.

vaadin-form-layout-flow-parent/vaadin-form-layout-flow/src/main/java/com/vaadin/flow/component/formlayout/FormLayout.java
vursen: Removed!

web-padawan (approved):

PR #7175, opened 2025-03-07
fix: use isEmpty in getEmptyValue to satisfy Binder

TatuLund:
Fixes https://github.com/vaadin/web-components/issues/8766


sissbruecker:
Thanks for looking into this. For now, let's go with adding something to the documentation in https://github.com/vaadin/docs/pull/4170, and then see if something comes out of https://github.com/vaadin/flow/issues/21106.

PR #7196, opened 2025-03-13
feat: implement card component flow counterpart

### LINKED ISSUE #7008: Card component API
### Describe your motivation

Add API, and tests, for setting/adding content to the Card component slots.

### Content (default slot)

```java
public void add(Component... content);
// …and all the rest of the HasComponents API
```

### Media

```java
public void setMedia(Component media)
public Component getMedia()
```

### Title

```java
public void setTitle(Component title)
public Component getTitle()
```

Could add a convenience method for a string title which wraps the text in a `Div()`:

```java
public void setTitle(String title)
// What about the getter?
```

### Subtitle

```java
public void setSubtitle(Component subtitle)
public Component getSubtitle()
```

Same as with the `title` slot, could add a convenience method for a string subtitle.

### Header 

```java
public void setHeader(Component header)
public Component getHeader()
```

Note: Setting the header content replaces title and subtitle, if those are set.

### Header Prefix

```java
public void setHeaderPrefix(Component headerPrefix)
public Component getHeaderPrefix()
```

### Header Suffix

```java
public void setHeaderSuffix(Component headerSuffix)
public Component getHeaderSuffix()
```

### Footer

```java
public void addToFooter(Component... footer)
public Component[] getFooterComponents() // Unsure if this is correct
```

### Style Variants

`CardVariant` class with
- Outlined
- Elevated
- Horizontal
- Stretch media
- Cover media

### A11y APIs

```
public void setAriaRole(String role)
public String getAriaRole()

public void setAriaLabel(String label)
public String getAriaLabel()

public void setAriaLabelledBy(String elementId)
public String getAriaLabelledBy()
```

### Functional Interfaces

- `HasComponents`
- `HasStyle`
- `HasSize`

_No response_

### Additional context

_No response_

ugur-vaadin:
## Description

This PR implements the Flow counterpart for the Card component.

Note 1: One difference from the API described in the issue is the getter for ARIA role. In the issue, the getter for the ARIA attributes are written as `public String getAria...()`. However, the already implemented API for ARIA labels in `HasAriaLabel` return `Optional<String>`. Therefore, the getter for the new ARIA role attribute also returns `Optional<String>` in this PR.

Note 2: Convenience APIs that take String parameters are not a part of this initial implementation.

Fixes #7008 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
+++ b/vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
@@ -15,12 +15,26 @@
  */
 package com.vaadin.flow.component.card;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Stream;
+
+import org.slf4j.LoggerFactory;
+
 import com.vaadin.flow.component.Component;
+import com.vaadin.flow.component.HasAriaLabel;
+import com.vaadin.flow.component.HasComponents;
 import com.vaadin.flow.component.HasSize;
 import com.vaadin.flow.component.Tag;
 import com.vaadin.flow.component.dependency.JsModule;
 import com.vaadin.flow.component.dependency.NpmPackage;
 import com.vaadin.flow.component.shared.HasThemeVariant;
+import com.vaadin.flow.component.shared.SlotUtils;
+import com.vaadin.flow.dom.Element;
 
 /**
  * Card is a visual content container for creating a card-based layout.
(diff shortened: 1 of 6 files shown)


vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
```diff
@@ -32,6 +49,358 @@
...
+        var slotElement = new AtomicReference<Element>();
+        SlotUtils.getElementsInSlot(this, slotName).findFirst().ifPresentOrElse(
+                slotElement::set, () -> slotElement.set(initSlot(slotName)));
+        slotElement.get().removeAllChildren();
+        slotElement.get().appendChild(slotContent.getElement());
+    }
+
+    private Element initSlot(String slotName) {
+        var div = new Element("div");
+        SlotUtils.setSlot(this, slotName, div);
```
vursen, line 387: Is there a particular reason for adding an extra `<div>` wrapper for each slot? Could the slot attribute be assigned directly to the component element instead?

ugur-vaadin: Not really. One small benefit is to enable the use of Text component directly. But removing it also makes everything much simpler. Therefore, I removed the use of the wrappers.

vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
```diff
@@ -32,6 +49,358 @@
...
+     */
+    private void checkFeatureFlag() {
+        boolean enabled = featureFlagEnabled
+                || getFeatureFlags().isEnabled(FeatureFlags.CARD_COMPONENT);
+        if (!enabled) {
+            throw new ExperimentalFeatureException();
+        }
+    }
+
+    private void setSlotContent(Component slotContent, String slotName) {
```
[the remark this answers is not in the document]
ugur-vaadin: Done in the PR.

vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
```diff
@@ -32,6 +49,358 @@
 @JsModule("@vaadin/polymer-legacy-adapter/style-modules.js")
 @NpmPackage(value = "@vaadin/card", version = "24.8.0-alpha2")
 @JsModule("@vaadin/card/src/vaadin-card.js")
-public class Card extends Component
-        implements HasSize, HasThemeVariant<CardVariant> {
+public class Card extends Component implements HasSize,
+        HasThemeVariant<CardVariant>, HasComponents, HasAriaLabel {
+
+    static final String MEDIA_SLOT_NAME = "media";
```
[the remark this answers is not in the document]
ugur-vaadin: Done.

vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
sissbruecker: Can you think of some potential issue of putting all of the main content into a wrapper div? I couldn't see anything from a quick glance, but it seems worth double-checking.

vursen: One concern is that wrappers might complicate styling in the future because the `::slotted` selector only works with direct children. Also, if a slot uses `flexbox` or other display models, its children won't automatically inherit it unless the wrapper is given `display: contents`...

vaadin-card-flow-parent/vaadin-card-flow/src/test/java/com/vaadin/flow/component/card/CardTest.java
```diff
@@ -0,0 +1,330 @@
...
+
+    private void setSlotContent_slotContentIsSet(
+            BiConsumer<Card, Component> setter, String slotName) {
+        var slotContent = new Div();
+        setter.accept(card, slotContent);
+        Assert.assertEquals(slotName,
+                slotContent.getElement().getAttribute("slot"));
+    }
+
+    private void slotBasedFieldUpdatedCorrectly(
```
sissbruecker, line 318: The test should probably also check that the component is actually added as a child, and is then removed when setting `null`.

The tests for the main content slot technically also don't test whether components are ever added to the UI or not.

vaadin-card-flow-parent/vaadin-card-flow/src/test/java/com/vaadin/flow/component/card/CardTest.java
```diff
@@ -0,0 +1,330 @@
...
+    }
+
+    @Test
+    public void setAriaRoleNull_ariaRoleUpdated() {
+        card.setAriaRole("custom-role");
+        card.setAriaRole(null);
+        Assert.assertTrue(card.getAriaRole().isEmpty());
+    }
+
+    private void setSlotContent_slotContentIsSet(
```
sissbruecker, line 310:
```suggestion
    private void setSlotContent_slotAttributeIsSet(
```

vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
```diff
@@ -32,6 +48,337 @@
...
+        if (!enabled) {
+            throw new ExperimentalFeatureException();
+        }
+    }
+
+    private void setSlotContent(String slotName, Component slotContent) {
+        if (slotContent == null) {
+            SlotUtils.clearSlot(this, slotName);
+        } else {
+            SlotUtils.setSlot(this, slotName, slotContent);
```
vursen, line 372: `setSlot` already does `clearSlot` and accepts `null` as slot content, so it seems `setSlotContent` could be entirely replaced with `SlotUtils.setSlot(this, slotName, slotContent)`. The `getSlotContent` could then be replaced with direct `SlotUtils.getChildInSlot(this, slotName)`

vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
```diff
@@ -32,6 +48,337 @@
...
+        componentsToAdd.forEach(component -> SlotUtils.addToSlot(this,
+                FOOTER_SLOT_NAME, component));
+    }
+
+    /**
+     * Gets all components added to the card's footer.
+     *
+     * @return an array of footer components
+     */
+    public Component[] getFooterComponents() {
```
vursen, line 224: Should it also return a Stream like `getChildren` does?

```suggestion
    public Stream<Component> getFooterComponents() {
```

vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
vursen: Is there a practical reason to wrap content at the moment though?

sissbruecker: How do you mean? I'm assuming we're wrapping the content because it makes implementing `HasComponents` a lot easier for stuff like `addComponentAtIndex`.

sissbruecker: If we want to avoid a wrapper then `addComponentAtIndex` seems like the only challenge. Everything else could maybe use the default implementation. For `removeAll` it's kind of debatable whether that should remove everything or just children from the content slot.

vaadin-card-flow-parent/vaadin-card-flow/src/test/java/com/vaadin/flow/component/card/CardTest.java
ugur-vaadin: Done.

vaadin-card-flow-parent/vaadin-card-flow/src/test/java/com/vaadin/flow/component/card/CardTest.java
ugur-vaadin: Added assertions that check if the children are attached and have the card as an ancestor.

vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
ugur-vaadin: Updated the PR to add `display: contents` to the content wrapper.

vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
ugur-vaadin: Done.

vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
ugur-vaadin: I'm not sure about this one. We don't seem to return `Stream` except for when we override `getChildren`. But the use of `List<Component>` instead of `Component[]` would also make sense. What do you think? The original signature in the ticket specifies `Component[]`, that's why I added it like that initially.

sissbruecker (approved):

vaadin-card-flow-parent/vaadin-card-flow/src/test/java/com/vaadin/flow/component/card/CardTest.java
```diff
@@ -0,0 +1,388 @@
...
+
+    private boolean isAncestor(Component component, Card probableAncestor) {
+        var parent = component.getParent();
+        while (parent.isPresent()) {
+            if (parent.get().equals(probableAncestor)) {
+                return true;
+            }
+            parent = parent.get().getParent();
+        }
+        return false;
```
vursen, line 386: **nit:**

```suggestion
        return probableAncestor.getElement().getNode().isAncestorOf(component.getElement().getNode());
```

ugur-vaadin: It seems like isAncestorOf is private. I reverted the change. I recall some public alternative exists, but cannot find it.

ugur-vaadin: I will keep it as it is if it's okay.

PR #7214, opened 2025-03-17
chore: Make Spring Data non-optional

### LINKED ISSUE #7213: Add spring-data-commons dependency to ensure compatibility with Eclipse 2024-06 and older
### Describe your motivation

The way we are referencing the `org.springframework.data.domain.Pageable` type in the newly introduced `setItemsPageable` methods is not compatible with Eclipse 2024-06 and older. This has been fixed for Eclipse 2024-09 in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2543.

Any application that uses `Grid` or  `ComboBox` with an affected Eclipse version and no `spring-data-commons` on the classpath will have an unrecoverable compilation error.

We should add a dependency to `spring-data-commons` but excluding its transitive dependencies so that we're not accidentally including the rest of Spring in the application. We should have automated tests that verify this so that we notice if a new version of the dependency introduces new transitive Spring dependencies.

We should document that this dependency can be excluded by users who are unable to include Spring dependencies in their application due to organization policies. Such users can still remain compatible with Eclipse 2024-06 by including an empty `org.springframework.data.domain.Pageable` class as a .java file directly in their application code.

We should also emphasize that this change does not mean that Flow will start requiring Spring but only that we make some targeted optimizations for Spring users and might also do similar optimizations for e.g. Jakarta EE or Quarkus users in the future.

1. We could revert the newly added helpers. We don't want to do this since they are highly useful. In particular, they remove one step that has had very poor discoverability prior to adding an (almost) direct reference to `Pageable`. We also want to keep the door open for adding other similar shorthands in the future, not necessarily only tied to Spring but also for types that are specific to e.g. Jakarta EE or Quarkus.
2. We could omit the dependency and instead document that users of Eclipse 2024-06 or older need to apply a workaround of their choice. We don't want to do this since it has a higher impact of affecting new user who are more likely to give up directly instead of discovering the workaround. In contrast, the explicit inclusion of a dependency provides better discoverability for potential workarounds for those who need a workaround.

### Additional context

Dynamic linking is not a solution for this case since the problem manifests at compile time due to method signatures rather than at run time due to references in method implementations.

Artur-:
Solves compilation issues on old Eclipse versions

Fixes #7213

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/pom.xml
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/pom.xml
@@ -69,12 +69,21 @@
             <artifactId>slf4j-simple</artifactId>
             <scope>test</scope>
         </dependency>
-        <!-- Optional Spring dependencies to be able to provide Spring Data API-->
+        <!-- Spring data dependency to be able to provide Spring Data API-->
         <dependency>
             <groupId>org.springframework.data</groupId>
             <artifactId>spring-data-commons</artifactId>
             <version>${spring-data-commons.version}</version>
-            <optional>true</optional>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.springframework</groupId>
+                    <artifactId>spring-core</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.springframework</groupId>
+                    <artifactId>spring-beans</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
     </dependencies>
     <build>
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/pom.xml
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/pom.xml
@@ -96,12 +96,21 @@
             <artifactId>slf4j-simple</artifactId>
             <scope>test</scope>
         </dependency>
-        <!-- Optional Spring dependencies to be able to provide Spring Data API-->
+        <!-- Spring data dependency to be able to provide Spring Data API-->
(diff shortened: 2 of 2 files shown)


mcollovati:
Would it work to set the dependency as `provided`, to prevent it to being added to the production build when not required?

Artur-:
A provided dependency of Vaadin is not included when building a Vaadin application

mcollovati:
Sure, but do we really need it to be included if you do not use Spring?

Artur-:
As described in the issue, it needs to be included for projects to compile with Eclipse older that 2024-09

mcollovati:
I have memories that if the dependency is `provided` Eclipse compiler will find it anyway. Definitely, it does not work with `optional`.
But perhaps I'm wrong.

Artur-:
> The provided dependencies are available only at compile time and in the test classpath of the project. These dependencies are also not transitive.

If you try to use `VaadinServlet` in https://github.com/vaadin/skeleton-starter-flow you see this
image

mstahv:
👎 Repeating my comment (from the issue) here, I don't like this move. Leaking lot of transitive dependencies (which we are already doing quite a lot) is not nice and because of human nature, certain developers will get additional itching when they see spring dependecies appearing to their builds.

The feature was a nice compromise (much better API for Spring Data users without transitive dependencies), but with this change, done for a fraction of users that can be expected to diminish, I think we would be better off without it or postpone it until we can be sure totally irrelevant number of Eclipse users are evaluating Vaadin.

(and yes, I'm using Spring Data myself in many apps)

PR #7239, opened 2025-03-21
fix: do not set z-index for sub-cells with text

### LINKED ISSUE #7177: Merged cells broken after scrolling
### Description

On a worksheet with some merged cells and some unmerged cells, scrolling down or sideways and then back copies the contents of the merged cell to all cells in the range, displays them as separate cells, but navigates them as merged cells.

Further, sometimes only some merged cells are handled this way, other times multiple or all are.  It seems to depend on the speed and distance scrolled, but probably server latency too.

### Expected outcome

scrolling should not affect display - if no data is changed, display of cells should be consistent when scrolled out of view and back into view.

[Merged_cell_scroll_example.xlsx](https://github.com/user-attachments/files/19134849/Merged_cell_scroll_example.xlsx)

Attaching a minimal example. This can be loaded into start.vaadin.com with a simple Spreadsheet view.  Scroll down even a little, then back up, and note the merged cells behave badly.  The yellow and red ones are the easiest to notice.

### Steps to reproduce

1. Go to start.vaadin.com
2. Create a new default project
3. Add a Spreadsheet Flow view
4. From the view "File" menu choose "import"
5. Upload the example file
6. Select "Load Spreadsheet"
7. scroll via mouse wheel or scrollbar, and note the bad behavior of merged cells.

### Environment

Vaadin version(s): 24.6.6
OS: N/A

### Browsers

Issue is not browser related

ugur-vaadin:
## Description

In the `Cell` class, `z-index` of 1 is automatically applied to cells with text content. However, this should not be the case for sub-cells. The first sub-cell in a merged region still contains the text. Usually, this problem is hidden because initially the sub-cells are added to the DOM before the merged cell. When there are two cells with the same `z-index`, the last added cell is shown on top. However, when the sheet is scrolled, sometimes the sub-cell is removed and the merged cell still stays in the DOM. Scrolling back to the row, the sub-cell is added back to the DOM with the same `z-index` of 1. Since the sub-cell is the last added one in this case, it is shown on top. 

This PR adds a check in the `Cell` class to avoid setting `z-index` for sub-cells with text.

Fixes #7177 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/Cell.java
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/Cell.java
@@ -91,7 +91,12 @@ public class Cell {
             element.setInnerText("");
             element.getStyle().clearZIndex();
         } else {
-            element.getStyle().setZIndex(ZINDEXVALUE);
+            if (sheetWidget.isMergedCell(SheetWidget.toKey(col, row))
+                    && !(this instanceof MergedCell)) {
+                element.getStyle().clearZIndex();
+            } else {
+                element.getStyle().setZIndex(ZINDEXVALUE);
+            }
             if (needsMeasure && getCellWidth() > 0 && sheetWidget
                     .measureValueWidth(cellStyle, value) > getCellWidth()) {
                 element.setInnerText("###");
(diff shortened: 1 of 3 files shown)


yuriy-fix (changes requested):
With this change, the focus outline (+ handle) of the merged cell is broken.
From the provided Merged_cell_scroll_example.xlsx:

Without the change:

- Normal initial state:
initial
- Broken state after scrolling:
current-broken

With the change:

- Merged cell selected:
latest-broken

Notice that the focus outline (+ handle) is not displayed correctly when the merged cell is selected.

ugur-vaadin:
> With this change, the focus outline (+ handle) of the merged cell is broken.

I reverted the initial fix and applied a new one. In the `Cell` class, `z-index` of 1 is automatically applied to cells with text content. However, this should not be the case for sub-cells. The first sub-cell in a merged region still contains the text. Usually, this problem is hidden because initially the sub-cells are added to the DOM before the merged cell. When there are two cells with the same `z-index`, the last added cell is shown on top. However, when the sheet is scrolled, sometimes the sub-cell is removed and the merged cell still stays in the DOM. Scrolling back to the row, the sub-cell is added back to the DOM with the same `z-index` of 1. Since the sub-cell is the last added one in this case, it is shown on top.

The new solution seems much cleaner and appropriate. There are similar checks in the `Cell` class already, changing its style based on whether it is a sub-cell. The focus outline problem is also not present in this version.

yuriy-fix (approved):

PR #7283, opened 2025-04-07
fix: TreeGridElement#getRowCount

TatuLund:
Fixes https://github.com/vaadin/flow-components/issues/7293


TatuLund:
I would recommend to add convenience scrollToEnd method in TreeGridElement using this

```
            executeScript(
                    "arguments[0].scrollToIndex(...Array(10).fill(Infinity));",
                    this);
```

Issue #7296, opened 2025-04-01
Blank Login View after navigating from @AnonymousAllowed to @PermitAll view


KasparScherrer:
### Description of the bug

whenever I navigate _programmatically_ from a `@AnonymousAllowed` view to a `@PermitAll` view (while not being authenticated yet), it redirects to the login view but it is completely blank (white, no content in the body except the outlet and the connection-indicator).

Meanwhile, when I navigate from a `@AnonymousAllowed` to a `@PermitAll` view by typing the url manually, the login view works fine and shows the login form. This difference makes me think the bug lies somewhere in `UI::navigate` maybe?

I can see a similar issue here https://github.com/vaadin/flow/issues/20939 but for me there is no errors being shown, neither in server log nor in browser console.

This is how the LoginView HTML looks when the bug is happening:
![Image](https://github.com/user-attachments/assets/0f1cff1c-89d8-4b13-9c63-33361151f740)

And this is how the LoginView HTML looks when it works properly:
![Image](https://github.com/user-attachments/assets/749837f3-eaee-463f-aa6f-a662ea34a234)

### Expected behavior

The Login Form/Overlay should appear instead of a blank screen

### Minimal reproducible example

Here are some views and configs that represent my setup:
```
@Route(value = "Test")
@AnonymousAllowed
public class TestView extends VerticalLayout {

    public TestView() {
        add(new Button("Navigate to authenticated view", click -> {
            UI.getCurrent().navigate(SomeAuthenticatedView.class);
        }));
    }

}
```

```
@Route(value = "authenticated")
@PermitAll
public class SomeAuthenticatedView extends VerticalLayout {

    public TestView() {
        add(new Span("You are authenticated."));
    }

}
```

```
@Route(value = "login")
@AnonymousAllowed
public class LoginView extends VerticalLayout {

    public LoginView() {
        LoginOverlay loginOverlay = new LoginOverlay();
        loginOverlay.setAction("login");
        loginOverlay.setOpened(true);
        loginOverlay.setI18n(prepareLoginI18N());
    }

}
```

```
@Configuration
@EnableWebSecurity
public class UiSecurityConfiguration extends VaadinWebSecurity {
    public static final String LOGIN_URL = "/login";

    public UiSecurityConfiguration () {
    }

    @Bean
    AuthenticationManager authManager(HttpSecurity httpSecurity, LogoutAuthenticationProvider logoutAuthenticationProvider) throws Exception {
[...]

Issue #7229, opened 2022-09-19
Inconsistent behavior when forwarding to a view vs refresh


OlliTietavainenVaadin:
### Description of the bug

Two example views: `HelloWorldView` and `AboutView`. `AboutView` accepts optional parameters. `HelloWorldView` forwards to `AboutView` with a String parameter. 

```
@Route(value = "hello", layout = MainLayout.class)
public class HelloWorldView extends HorizontalLayout implements BeforeEnterObserver {

    @Override
    public void beforeEnter(BeforeEnterEvent event) {
        event.forwardTo("about/123");
    }
}
```
```
@Route(value = "about", layout = MainLayout.class)
public class AboutView extends VerticalLayout implements HasUrlParameter<String> {

    @Override
    public void setParameter(BeforeEvent event, @OptionalParameter String parameter) {
        Notification.show("Got parameter: " + parameter);
    }
}

```
If I navigate first to `/hello`, the URL gets updated to `/about/123` but no Notification is shown. If I refresh the page with `/about/123` in the URL bar, I get the notification "Got parameter: 123". I would expect the view's parameterized status to be the same if the URL doesn't change.

### Expected behavior

The `setParameter` should fire even after the `forwardTo` with the needed parameter values.

### Minimal reproducible example

[v23-2-start.zip](https://github.com/vaadin/flow/files/9598717/v23-2-start.zip)

### Versions

- Vaadin / Flow version: 23.2.0
- Java version: n/a
- OS version: n/a
- Browser version (if applicable): n/a
- Application Server (if applicable): n/a
- IDE (if applicable):

PR #7304, opened 2025-04-09
fix: preserve overlays after forwarding to view

sissbruecker:
## Description

Fixes https://github.com/vaadin/flow-components/issues/7296
Fixes https://github.com/vaadin/flow-components/issues/7229

## Type of change

- Bugfix


vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/internal/OverlayAutoAddController.java
sissbruecker, line 1: The actual fix is here.

PR #7356, opened 2025-04-28
fix: do not keep column id map in gridpro

### LINKED ISSUE #7322: Grid memory references of removed columns
### Description

We discovered maybe a potential memory-reference issue in the grid (com.vaadin.flow.component.gridpro.GridPro.idToColumnMap). 
The idToColumnMap is Always just adding and reading the entries but never removed.
Even when removeAllColumns is called the references are still kept in the hashMap idToColumnMap.

### Expected outcome

When a column has been removed it no longer should be kepot in the map so that the memopry instance can be cleared by the garbage collection.

1. Having a grid with columns
2. Call removeAllColumns

### Steps to reproduce

Look at the memory references which are hold by the grid.

### Environment

Vaadin version(s): 24.7.2
OS: all

### Browsers

Issue is not browser related

ugur-vaadin:
## Description

The entries in `idToColumnMap` in `GridPro` are not cleared whenever the columns are removed unlike its `Grid` counterpart. It can be fixed by overriding `Grid.removeColumn` and removing the column from the map. However, keeping this map on top of the already existing map in `Grid` seems redundant and prone to errors.

This PR removes this map from `GridPro` entirely, making the `Grid` class the only class responsible for handling this map.

Fixes #7322 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
@@ -3530,13 +3530,14 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
 
     /**
      * Gets a {@link Column} of this grid by its internal id ({@code _flowId}).
+     * Intended only for internal use and can be removed in the future.
      *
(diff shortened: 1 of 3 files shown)


vursen (approved):
LGTM 👍

Issue #7334, opened 2025-04-23
Grid item details closes when the corresponding item is updated


jdewaen:
### Description

Most features in the Grid use `DataProvider#getId` to determine whether to instances of a grid item are to be treated as being the same. This means that updating these items with updated values (while keeping the id the same), in combination with `DataProvider:refreshItem` updates the existing row in the grid without other side effects.

The grid item details feature is an exception to this. It treats items with the same id but different values as different. This means that updating an item also automatically closes the details panel. It also causes a memory leak as the original item is still referenced in the `DetailsManager`'s `detailsVisible` set. Without a "hideAllDetails" function or something similar, it's impossible to clean up that memory leak once it happens.

### Expected outcome

I would expect the item details feature to respect the item ids, just like the other grid features do. Refreshing an item with updated values should keep the details panel open and call `updateComponent` on the details item `ComponentRenderer` with the updated value as well.

### Minimal reproducible example

```java
        // Set up data provider
        List<Item> sourceData = new ArrayList<>();
        sourceData.add(new Item(1, "Before"));
        sourceData.add(new Item(2, "Other item"));
        ListDataProvider<Item> dp = new ListDataProvider<>(sourceData)
        {
            @Override
            public Object getId(Item item)
            {
                return item.id();
            }
        };

        // Set up grid
        Grid<Item> grid = new Grid<>(dp);
        grid.addColumn(Item::id).setHeader("id");
        grid.addColumn(Item::label).setHeader("label");
        grid.setItemDetailsRenderer(new ComponentRenderer<>(() -> new Span("I'm the details!")));
        add(grid);

        // Set up button to alter item
        Button button = new Button("Click me");
        button.addClickListener(event -> {
            Item newItem = new Item(1, "After");
            sourceData.set(0, newItem);
            dp.refreshItem(newItem);
        });
        add(button);
```

Item definition:
```java
    private record Item(int id, String label) {}
```

### Steps to reproduce

1. Add the code above to a view
2. Load the page, click the first item in the grid to open the details
3. Click the button. The item will update, and the details will close

### Environment

Vaadin version(s): 24.7.1
OS: Not relevant

### Browsers

Issue is not browser related

PR #7374, opened 2025-05-01
fix: ensure row details stay opened after refreshing item

vursen:
## Description

Refactors the Grid row details implementation to determine open state using item identity instead of object reference equality. This prevents the row details from closing when an item is refreshed with a new object instance.

Fixes https://github.com/vaadin/flow-components/issues/7334

## Type of change

- [x] Bugfix


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -1264,7 +1265,7 @@ protected Grid<T> getGrid() {
      */
     private class DetailsManager extends AbstractGridExtension<T> {
 
-        private final HashSet<T> detailsVisible = new HashSet<>();
+        private final HashMap<Object, T> detailsVisible = new LinkedHashMap<>();
```
vursen, line 1268: The order of items doesn't seem to be important, so perhaps HashMap would be more appropriate?

```suggestion
        private final HashMap<Object, T> detailsVisible = new HashMap<>();
```

sissbruecker (approved):

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
sissbruecker: Probably best to keep the existing behavior, for example, someone might have tests that rely on the order in which details were opened.

vursen: Well, previously, a HashSet was used, which doesn't guarantee order. However, the selected items, which I used for inspiration, are stored in a LinkedHashMap, which preserves insertion order.

Issue #6529, opened 2024-08-09
Spreadsheet focuses the sheet rather than the component when displaying a cell edit component


WoozyG:
### Description

When a cell with a custom editor is clicked, the custom editor is displayed instead of (or on top of, hiding) the cell static contents.  However, after displaying the editor component, the SheetWidget client-side code sets the focus to the _sheet_, not the editor.

[This line](https://github.com/vaadin/flow-components/blob/47c3d9456a81dbf6b6a83910789992ce08a8de9a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/SheetWidget.java#L5031) is from the initial code commit 10 years ago, per Git-Blame.  To my mind that means there is no institutional knowledge about why that focus choice is there :)  I don't remember my 10 year old code, at least.

This causes custom edit component use to require multiple clicks, and even breaks keyboard navigation, as attempting to edit a cell suddenly jumps focus back to the sheet, requiring navigating to the cell once again before the component can be used.

Further, I noticed this focus shift also interferes with a ComboBox used as an edit component with `autoselect=true` such that once the user finally activates the component the filter contents are no longer selected.

### Expected outcome

Clicking a Spreadsheet cell with a custom editor should display and focus the editor.

### Minimal reproducible example

define a `SpreadsheetComponentFactory` subclass and add it to a Spreadsheet component.  Implement only `getCustomEditorForCell()` and return a simple component, in my case a ComboBox, for all cells or a specific cell.

In the browser, click that cell, and note where the focus ends up after one cell click.

### Steps to reproduce

In the browser, click that cell, and note where the focus ends up after one cell click when the cell has a custom edit component.

### Environment

Vaadin version(s): 24.4.8
OS: N/A

### Browsers

Issue is not browser related

rolfsmeds:
Hi @WoozyG,
It seems to me that the described behavior is pretty much in line with how the default editor behaves: a single click on a cell only focuses the cell, and does not place focus in the editor.
If a single-click _did_ focus the editor, it seems to me that that would make it rather difficult to focus the cell itself, when that is desired.

Perhaps the problem is that the editor component being _revealed_ on click causes a confusing UX that raises expectations of it also getting focused? (Compared to default cells which don't change in appearance when clicked, that is.)

WoozyG:
[...]

PR #7401, opened 2025-05-06
fix: align keyboard/mouse behavior for custom editors

DiegoCardoso:
## Description

Change the keyboard behavior for custom editors to align with how the default editor works:
- Pressing <kbd>F2</kbd>/<kbd>Enter</kbd> moves the focus to the custom editor
- Pressing <kbd>Tab</kbd>/<kbd>Shift</kbd>`+`<kbd>Tab</kbd> moves the focus to next/previous cell
- Pressing any character moves the focus to the editor and sets the pressed character to the editor
- Pressing <kbd>Esc</kbd> in a focused custom editor moves the focus to the cell

Also changes the mouse behavior so that the focus is moved to the custom editor when the user clicks the cell containing a custom editor when the editor is not visible, avoiding the need for a second click to focus the editor.

Part of #6529

## Type of change

- Bugfix


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/SpreadsheetWidget.java
sissbruecker, line 1: > Pressing any character moves the focus to the editor and sets the pressed character to the editor

This doesn't seem to work for me. Pressing a character focuses the editor, but does not enter the character into the editor. I also remember this was tricky to get right for Grid Pro.

sissbruecker, line 1: > Also changes the mouse behavior so that the focus is moved to the custom editor when the user clicks the cell containing a custom editor when the editor is not visible, avoiding the need for a second click to focus the editor.

I have a hard time parsing the requirements from the linked issue, so I'm not sure if that was the desired behavior. From what I understood there were two options discussed, both of which would actually change how custom editors work:
- Show the editors always, even when the cell is not focused. In that case the new behavior makes sense, so that clicking activates the editor.
- Only show the editor once you enter edit mode. In that case use double-click to activate the editor, same as with the default editor.

It doesn't seem to mention how clicks should behave with the current behavior where editors become visible when a cell is focused. IMO, with the current behavior, focusing the custom editor on click is reasonable, but let's double-check if that was intended.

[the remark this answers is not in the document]
DiegoCardoso: Reverted the behavior to the previous state as you pointed out. Now, pressing <kbd>Enter</kbd> in a text field, or <kbd>Space</kbd> in a checkbox won't remove the element from the cell.
no-detach-enter

DiegoCardoso: It works for me, though. Care to share how was the editor you tested configured?
typing-cell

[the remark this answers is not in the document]
DiegoCardoso: It should be working now:
typing-cell

DiegoCardoso: I decide to revert this behavior and make it behaves similar to how the default editor does, where clicking on a cell won't move the focus to the editor, but only after a double-click (with the difference that there won't be need to double-click to start editing, but rather, a second click).

DiegoCardoso (commented):
I addressed the comments, and also removed some dead code/performed some cleanups.

vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/SpreadsheetWidget.java
sissbruecker: Looks like I used the same test setup, but I was using Firefox.

sissbruecker: I double-checked with the latest changes, it works in Chrome and Safari but not in Firefox.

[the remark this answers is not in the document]
sissbruecker: Interesting, this also doesn't work in Firefox.

sissbruecker: Looks like it's a bit incidental, in that the second click focuses the editor only because that's how our components implement focus. For example, if you double-click in some space of the cell that isn't covered by the editor, it does not get focused.

But that would be good enough IMO, it works now like before and there was no complaint about that.

DiegoCardoso: It seems that it was working by accident. The previous call didn't actually work, and what we are seeing in Chrome and Safari seems to be how they work on a focused element receives a `keypress` event and moves the focus to an input element on the same frame. From my tests, the input element captures the event in those browsers, but the same doesn't happen in Firefox.

The same Pen I linked in the other comment has [a test for this behavior](https://codepen.io/diegocardoso/pen/emmxgyP?editors=1010) (it's the second example there).

I made a best effort to try to set the `value` property of the custom editor, but only for Firefox.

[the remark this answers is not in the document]
DiegoCardoso: It seems to be an issue in Firefox where events won't fire for detached elements. I have a recollection of facing something like that in the past. As the element is removed when the user tabs out of it, Firefox skips the `change` event entirely. You can check [this in action here](https://codepen.io/diegocardoso/pen/emmxgyP?editors=1010) (in the first example). I added a delay of an animation frame before removing the element, and it fixes the issue.

vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/SheetEventListener.java
```diff
@@ -102,6 +110,28 @@ public void onBrowserEvent(Event event) {
...
+                        element -> element.getNodeType() == Node.ELEMENT_NODE
+                                && Objects.equals(element.getTagName(), "SLOT")
+                                && element.getAttribute("name")
+                                        .startsWith("custom-editor-"))
+                .findFirst();
+
+        if (result.isPresent()) {
+            return true;
+        }
+
```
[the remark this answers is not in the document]
sissbruecker: @DiegoCardoso This is valid, the second block seems the be a complete duplicate.

vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/SheetWidget.java
```diff
@@ -5007,7 +5023,8 @@ public int getBottomVisibleRowIndex() {
         return lastRowIndex;
     }
```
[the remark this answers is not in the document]
sissbruecker: @DiegoCardoso Do we need to care about API stability here? It looks like this is GWT code that gets compiled to JS, but I can't say for sure

PR #7370, opened 2025-04-30
refactor: optimize message list item text updates

tomivirkki:
## Description

Whenever updating the text of a single `MessageListItem`, `MessageList` currently serializes and sends the full set of items back to the client.

https://github.com/user-attachments/assets/17edb7eb-47fb-4689-9b31-d53408d2346a

This PR optimizes the item text updates so that only the changed text gets sent back to client:

https://github.com/user-attachments/assets/eed63563-ba27-4b1c-bccd-c10ef6b7c2fc

The optimization becomes especially important when creating AI chat -type UIs where the text of the last message gets streamed and we want to efficiently only push the latest token to client on update (instead of the full message history):

https://github.com/user-attachments/assets/ee81532b-4900-42cc-b21e-b50e27213f66

## Type of change

Refactor / performance enhancement

PR #7412, opened 2025-05-08
feat: add message list item and text convenience API

tomivirkki:
## Description

Follow-up for https://github.com/vaadin/flow-components/pull/7370
Related to https://github.com/vaadin/platform/issues/7580

Add convenience API to `MessageList`:
- `MessageList.addItem(messageListItem)`
  - Enables the user to conveniently add new message items to `MessageList`. Currently this requires to call `MessageList.setItems(updatedItems)` with the new item included, which also causes the whole list of message items to get re-sent to the client. With `addItem()`, only the newly added item(s) get sent.
- `MessageListItem.appendText(text)`:
  - Enables the user to conveniently append text to a message list item (instead of having to call `item.setText(item.getText() + text)`

## Type of change

Feature


vaadin-messages-flow-parent/vaadin-messages-flow/src/test/java/com/vaadin/flow/component/messages/tests/MessageListUpdatesTest.java
sissbruecker, line 1: The update logic has become quite complex, and it seems there's a number of test cases that are not covered. Mostly what seems to be missing is verifying that these force a full update:
- Append item, then set items in same roundtrip
- Set text, then set items in same roundtrip
- Same two as above, reversed

That would also involve verifying the actual JS calls that are being made. I'd be fine with adding these as follow-up though.

sissbruecker (approved):

PR #7435, opened 2025-05-13
feat!: show custom editors by default

### LINKED ISSUE #6529: Spreadsheet focuses the sheet rather than the component when displaying a cell edit component

DiegoCardoso:
## Description

This changes the behavior to always show custom editors in cells and adds a new API to change opt-out and restore the previous behavior, where custom cells were only visible when the cell where they were defined was selected. 

Fixes #6529

## Type of change
- Feature


ugur-vaadin (approved):
LGTM overall. Maybe could add some more comments on the new methods.

PR #7453, opened 2025-05-19
chore: upgrade vaadin-parent to 3.0.0

ZheSun88:


ZheSun88:
#7461 #7462 #7463 #7464 #7465 #7466

PR #7454, opened 2025-05-19
fix: avoid error on message list add item

tomivirkki:
## Description

`MessageList` connector currently [maps](https://github.com/vaadin/flow-components/blob/a3cfe40b1d62ffdf93a5e93c47289ddbd04ad6ab/vaadin-messages-flow-parent/vaadin-messages-flow/src/main/resources/META-INF/resources/frontend/messageListConnector.js#L31) the message list items so that the `time` property gets formatted by the given locale. This becomes an issue when the existing mapped items are [re-assigned](https://github.com/vaadin/flow-components/blob/a3cfe40b1d62ffdf93a5e93c47289ddbd04ad6ab/vaadin-messages-flow-parent/vaadin-messages-flow/src/main/resources/META-INF/resources/frontend/messageListConnector.js#L57) and the formatted time value can no longer be [parsed](https://github.com/vaadin/flow-components/blob/a3cfe40b1d62ffdf93a5e93c47289ddbd04ad6ab/vaadin-messages-flow-parent/vaadin-messages-flow/src/main/resources/META-INF/resources/frontend/messageListConnector.js#L31) by `new Date(value)` constructor.

This PR updates the component to only operate on unformatted time values in the message list items.

Reported in https://github.com/marcushellberg/practical-ai-in-java/issues/6

## Type of change

Bugfix

### DIFF
--- a/vaadin-messages-flow-parent/vaadin-messages-flow/src/main/resources/META-INF/resources/frontend/messageListConnector.js
+++ b/vaadin-messages-flow-parent/vaadin-messages-flow/src/main/resources/META-INF/resources/frontend/messageListConnector.js
@@ -26,9 +26,9 @@ window.Vaadin.Flow.messageListConnector = {
       minute: 'numeric'
     });
     list.items = items.map((item) =>
-      item.time
+      item.__unformattedTime
         ? Object.assign(item, {
-            time: formatter.format(new Date(item.time))
+            time: formatter.format(new Date(item.__unformattedTime))
           })
         : item
     );
--- a/vaadin-messages-flow-parent/vaadin-messages-flow/src/main/resources/META-INF/resources/frontend/messageListConnector.js
+++ b/vaadin-messages-flow-parent/vaadin-messages-flow/src/main/resources/META-INF/resources/frontend/messageListConnector.js
@@ -26,9 +26,9 @@ window.Vaadin.Flow.messageListConnector = {
       minute: 'numeric'
     });
     list.items = items.map((item) =>
-      item.__unformattedTime
+      item.time
         ? Object.assign(item, {
-            time: formatter.format(new Date(item.__unformattedTime))
+            time: formatter.format(new Date(item.time))
           })
         : item
     );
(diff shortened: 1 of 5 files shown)


vaadin-messages-flow-parent/vaadin-messages-flow/src/main/resources/META-INF/resources/frontend/messageListConnector.js
```diff
@@ -26,9 +26,9 @@ window.Vaadin.Flow.messageListConnector = {
       minute: 'numeric'
     });
     list.items = items.map((item) =>
-      item.time
+      item.__unformattedTime
         ? Object.assign(item, {
-            time: formatter.format(new Date(item.time))
+            time: formatter.format(new Date(item.__unformattedTime))
```
sissbruecker, line 31: An alternative would be to:
- extract a function for formatting the time on an array of items and call that in both `setItems` and `addItems`
- make `addItems` set the `items` property directly

I think that might be a bit simpler overall? No need to add a new property then, also less data sent to the client.

Another alternative would be to make the connector set a flag on already processed items and then just ignore those when mapping in ´setItems´.

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/resources/META-INF/resources/frontend/messageListConnector.js
```diff
@@ -13,25 +13,34 @@
...
+    minute: 'numeric'
+  });
+
+  return items.map((item) =>
+    item.time
+      ? Object.assign(item, {
+          time: formatter.format(new Date(item.time))
+        })
+      : item
+  );
```
tomivirkki, line 35: Considered changing this to:

```js
return items.map((item) => ({
  ...item,
  time: item.time ? formatter.format(new Date(item.time)) : item.time
}));
```

But maintaining the same item references seems quite intentional so didn't want to touch it

sissbruecker (approved):

PR #7516, opened 2025-05-26
fix: Delegate to original upload handler methods

mshabarov:
## Description

The following upload handler's overrides are ignored, e.g. custom url prefix is not taken into account (as well as other default methods in `UploadHandler`):
```java
        InMemoryUploadHandler inMemoryUploadHandler = new InMemoryUploadHandler(
                (uploadMetadata, bytes) -> {
                    // handle Uploaded data here
                }) {
            @Override
            public String getUrlPostfix() {
                return "custom-url-postfix";
            }

            // other overrides
        };
        upload.setUploadHandler(inMemoryUploadHandler);
```

This change wraps the original upload handler so that the wrapper delegates all methods to original handler, not only the handleUploadRequest.

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
+++ b/vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
@@ -15,6 +15,7 @@
  */
 package com.vaadin.flow.component.upload;
 
+import java.io.IOException;
 import java.io.OutputStream;
 import java.util.ArrayDeque;
 import java.util.Arrays;
@@ -37,6 +38,7 @@ import com.vaadin.flow.component.dependency.JsModule;
 import com.vaadin.flow.component.dependency.NpmPackage;
 import com.vaadin.flow.component.html.Span;
 import com.vaadin.flow.component.shared.SlotUtils;
+import com.vaadin.flow.dom.DisabledUpdateMode;
 import com.vaadin.flow.dom.DomEventListener;
 import com.vaadin.flow.function.SerializableConsumer;
 import com.vaadin.flow.internal.JsonSerializer;
@@ -45,6 +47,8 @@ import com.vaadin.flow.server.NoOutputStreamException;
 import com.vaadin.flow.server.StreamReceiver;
 import com.vaadin.flow.server.StreamResourceRegistry;
 import com.vaadin.flow.server.StreamVariable;
+import com.vaadin.flow.server.VaadinResponse;
+import com.vaadin.flow.server.streams.UploadEvent;
 import com.vaadin.flow.server.streams.UploadHandler;
 import com.vaadin.flow.shared.Registration;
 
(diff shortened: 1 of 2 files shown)


vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -727,14 +731,7 @@ public void setReceiver(Receiver receiver) {
...
     public void setUploadHandler(UploadHandler handler) {
-        UploadHandler newUploadHandler = event -> {
-            try {
-                startUpload();
-                handler.handleUploadRequest(event);
-            } finally {
-                endUpload();
-            }
-        };
+        UploadHandler newUploadHandler = new UploadHandlerWrapper(handler);
```
mshabarov, line 734: The drawbacks of this are:
- once any new method is added into `UploadHandler`, we should modify this code here accordingly.
- delegate for `handleRequest` is skipped, because the default impl uses `this` when it invokes `TransferUtils` and `this` means the original handler.

Any better solutions are welcomed.

caalador: The only idea for nicer usage would be if the handler would fire events that you could add a listener for.
e.g. `handler.addEventListener( eventType -> if (eventType.equals(UploadEvent.START) { startUpload();} ... );`
Then anyone requiring before after etc could just listen for those instead of doing multiple wrappers hiding the actual implementation of the handler.

mshabarov: You are now talking about `TransferProgressListener::onStart` and `TransferProgressListener::onComplete` basically. The obstacle for this approach is that we decided at some point to not expose `addTransferProgressListener` in handlers and, moreover, the `UploadHandler` functional interface has no API/implementation for events/listeners.

caalador: No I meant specifically that the UploadHandler would support events so not tied to the transfer in any way.

caalador: But any which way it is done the need would be to be able to have a way to do something immediately when `TransferUtil.handleUpload` is called and just before it exits. or `UploadHandler.handleRequest` called/exits.

mshabarov:
Waits for https://github.com/vaadin/flow/pull/21554

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -131,6 +135,8 @@ public Upload() {
                     "Upload cannot be performed without a upload handler set. "
                             + "Please firstly set the upload handler implementation with upload.setUploadHandler()");
         });
+        // event listeners are not needed when no handler is set
+        unregisterUploadHandlerEvents();
```
caalador, line 139: Could we have this NO-OP handler as a specific type so that it wouldn't add the event handlers at all?
Would make it simpler as then there would be no need to unregister in different places.

mshabarov: Added a dummy handler and removed registrations.
I think the component doesn't have to clean up these if it switches to a receiver as these won't be invoked anyway.

caalador (approved):

Issue #7538, opened 2025-05-28
Chart XAxis configurations do not work as expected


TatuLund:
### Description

If you try to modify the existing XAxis configuration, the results are mixed. E.g. in this case label rotation is not taking effect when step is modified.

### Expected outcome

All the charts in the view should have labels rotated.

### Minimal reproducible example

```java
@Route(value = "chart-labels", layout = MainLayout.class)
public class ChartView extends Main {

    public ChartView() {
        addClassName(LumoUtility.Padding.MEDIUM);

        add(new H2("No Customizing - good looking"));
        var chart1 = createChart();
        add(chart1);

        add(new H2("Modify Label Steps - broken"));
        var chart2 = createChart();
        var configuration2 = chart2.getConfiguration();
        var xAxis2 = configuration2.getxAxis();
        // Not possible to set rotation
        xAxis2.getLabels().setStep(3);
        xAxis2.getLabels().setRotation(45);
        xAxis2.getLabels().setAlign(HorizontalAlign.RIGHT);
        add(chart2);

        add(new H2("Create a new a XAxis - works"));
        var chart3 = createChart();
        var configuration3 = chart3.getConfiguration();
        var xAxis3 = new XAxis();
        // Do the same config as above
        xAxis3.getLabels().setStep(3);
        xAxis3.getLabels().setRotation(45);
        xAxis3.getLabels().setAlign(HorizontalAlign.RIGHT);
        configuration3.addxAxis(xAxis3);
        add(chart3);

    }

    private Chart createChart() {
        var chart = new Chart();
        chart.setMaxWidth(300, Unit.PIXELS);

        var configuration = chart.getConfiguration();
        configuration.getChart().setType(ChartType.LINE);

        var xAxis = configuration.getxAxis();
        xAxis.setTickInterval(1);

        var series = new DataSeries();
        getChartData().forEach(data -> {
            series.add(new DataSeriesItem(data.year(), data.value()));
        });

        configuration.setSeries(series);
        return chart;
    }

    private List<ChartData> getChartData() {
        return List.of(new ChartData(1995, 13.4), new ChartData(2000, 15.7),
                new ChartData(2005, 21.2), new ChartData(2010, 17.5),
                new ChartData(2015, 11.9));
    }

    record ChartData(int year, double value) {
    }

}
```

### Steps to reproduce

Observe the view, chart in the middle should have labels rotated.

[...]

PR #7544, opened 2025-05-28
fix: serialize rotation as number if possible

sissbruecker:
## Description

`Labels.getRotation` is stored and serialized as a string. Normally that works fine, and Highcharts parses the property as a number. However it doesn't work if you also configure `Labels.setStep` - in this case the property must be a number for rotation to work.

This adds a custom serializer for the property that attempts to serialize as number if a number can be parsed, otherwise it falls back to string serialization.

Fixes https://github.com/vaadin/flow-components/issues/7538

## Type of change

- Bugfix


sissbruecker:
Right, so when using a string value it does not apply that but it may apply auto-rotation. That apparently used exactly the same rotation value I put in the string for testing...

DiegoCardoso (approved):

PR #7563, opened 2025-06-03
fix: do not delay auto-adding Popover by `beforeClientResponse` 

### LINKED ISSUE #7505: Popover not working after displaying a Dialog
### Description of the bug

If i want to display a Popover, but first a Dialog, the Popover does not work after i displayed the Dialog.

### Expected behavior

The Popover should not work while the Dialog is shown (because it is in the background), but before and after displaying the dialog, it should work.

See https://github.com/JohannesRabauer/VaadinPopoverWithDialogExample
If you start the application with `mvn clean spring-boot:run` and close the dialog, the popover is not working.
If you don't show the dialog (comment following line out: https://github.com/JohannesRabauer/VaadinPopoverWithDialogExample/blob/2fe4704c02a7b6f74b8dc0ad06f277e27d4857f4/src/main/java/dev/rabauer/PopoverDialogDemo/PopoverView.java#L28) and then start the application, the popover works as expected.

### Versions

- Vaadin / Flow version: 24.7.4
- Java version: 24
- OS version: Windows 11
- Browser version (if applicable): Chrome and Firefox
- Application Server (if applicable): Spring Boot
- IDE (if applicable): IntelliJ

DiegoCardoso:
## Description

If a popover tries to auto-add itself when another modal element is opened, it may end up being added as a child of said element. That can be problematic in case the target element is not part of the parent modal element and, when the modal is closed, the popover instance is removed with its parent and interacting with the target element won't work as expected.

This fix adds a detach listener to popover and tries to reattach it in case of:
- the popover instance has been auto-added
- the target element is present
- the target element is still attached to the page

Fixes #7505

## Type of change

- Bugfix

### DIFF
--- a/vaadin-popover-flow-parent/vaadin-popover-flow-integration-tests/src/test/java/com/vaadin/flow/component/popover/tests/PopoverIT.java
+++ b/vaadin-popover-flow-parent/vaadin-popover-flow-integration-tests/src/test/java/com/vaadin/flow/component/popover/tests/PopoverIT.java
@@ -135,6 +135,20 @@ public class PopoverIT extends AbstractComponentIT {
         Assert.assertTrue(popover.isOpen());
     }
 
+    @Test
+    public void openDialogAndSetTarget_closeDialog_popoverOpens() {
+        // Clear the target and remove the popover first to ensure the popover
+        // is auto-added when the dialog opens
+        clickElementWithJs("clear-target");
+        clickElementWithJs("remove-popover");
+
+        clickElementWithJs("open-dialog");
+        clickElementWithJs("close-dialog");
+
+        clickTarget();
+        checkPopoverIsOpened();
+    }
+
     private void clickTarget() {
         clickElementWithJs("popover-target");
     }
--- a/vaadin-popover-flow-parent/vaadin-popover-flow-integration-tests/src/test/java/com/vaadin/flow/component/popover/tests/PopoverIT.java
+++ b/vaadin-popover-flow-parent/vaadin-popover-flow-integration-tests/src/test/java/com/vaadin/flow/component/popover/tests/PopoverIT.java
@@ -145,6 +145,8 @@ public class PopoverIT extends AbstractComponentIT {
         clickElementWithJs("open-dialog");
         clickElementWithJs("close-dialog");
 
+        waitForElementNotPresent(By.tagName("vaadin-dialog-overlay"));
+
         clickTarget();
         checkPopoverIsOpened();
     }
--- a/vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
+++ b/vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
@@ -83,7 +83,6 @@ public class Popover extends Component implements HasAriaLabel, HasComponents,
      */
     public Popover() {
         getElement().getNode().addAttachListener(this::attachComponentRenderer);
-        getElement().getNode().addDetachListener(this::dettachListener);
 
         // Workaround for: https://github.com/vaadin/flow/issues/3496
         getElement().setProperty("opened", false);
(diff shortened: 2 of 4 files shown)


vaadin-popover-flow-parent/vaadin-popover-flow-integration-tests/src/test/java/com/vaadin/flow/component/popover/tests/PopoverIT.java
```diff
@@ -135,6 +135,20 @@ public void disableCloseOnEsc_pressEsc_popoverDoesNotClose() {
...
+        // is auto-added when the dialog opens
+        clickElementWithJs("clear-target");
+        clickElementWithJs("remove-popover");
+
+        clickElementWithJs("open-dialog");
+        clickElementWithJs("close-dialog");
+
+        clickTarget();
+        checkPopoverIsOpened();
+    }
```
DiegoCardoso: I tried to avoid adding an IT and use a unit test instead, but with no luck. I tried something like:

```java
    @Test
    public void targetAddedToPage_modalElementOpenedAndClosed_popoverAutoAddedToPage() {
        Popover popover = new Popover();
        Div target = new Div();
        popover.setTarget(target);
        ui.add(target);

        Div modalElement = new Div();
        ui.setChildComponentModal(modalElement, true);
        fakeClientResponse();

        Assert.assertEquals(modalElement.getElement(),
                popover.getElement().getParent());

        ui.remove(modalElement);
        modalElement.getElement().removeAllChildren();
        modalElement.getElement().removeFromTree();
        // popover.getElement().removeFromTree();
        fakeClientResponse();

        Assert.assertEquals(ui.getElement(), popover.getElement().getParent());
    }
```

But the detach listener in the Popover class wasn't being called.

vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
sissbruecker, line 1: I wonder if we can add popovers to the parent element of the target instead. That seems to make more sense than adding them to the current modal.

Otherwise the root issue here seems to be delaying adding to the UI in `beforeClientResponse`. I think that was just added as a copy+paste from an existing overlay component. So maybe removing that delay could be a better option for a fix as it wouldn't require creating a new client-side instance due to the detach.

vaadin-popover-flow-parent/vaadin-popover-flow-integration-tests/src/test/java/com/vaadin/flow/component/popover/tests/PopoverIT.java
```diff
@@ -135,6 +135,20 @@ public void disableCloseOnEsc_pressEsc_popoverDoesNotClose() {
         Assert.assertTrue(popover.isOpen());
     }
 
+    @Test
+    public void openDialogAndSetTarget_closeDialog_popoverOpens() {
```
sissbruecker, line 139: The test also passes on `main` without any changes. Looks like it needs to wait until the dialog is closed.

vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
DiegoCardoso: I refactored it to add the popover element to the target`s parent element.

vaadin-popover-flow-parent/vaadin-popover-flow-integration-tests/src/test/java/com/vaadin/flow/component/popover/tests/PopoverIT.java
DiegoCardoso: Added a wait for the dialog's overlay element to be removed. But now I wonder if I should move to [the unit test](https://github.com/vaadin/flow-components/pull/7563/files/64792e159e9631fdcd47998e5dc290bb3b8b78cf#r2123764396) that seems to work with this new approach.

vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
sissbruecker: Unfortunately this doesn't work with component renderers:
```java
        new ComponentRenderer<>(
                item -> {
                    var span = new Span("Target");
                    Popover popover = new Popover(new Text("Popover content"));
                    popover.setTarget(span);
                    return span;
        });
```

ComponentRenderer puts all generated components into a div that is added as a virtual child to the UI, so it never ends up in the DOM. Let's maybe go with removing `beforeClientResponse` for now?

DiegoCardoso: Done.

sissbruecker (approved):

PR #3277, opened 2022-06-03
fix: disable client-side validation for CustomField

vursen:
## Description

The PR disables the client-side validation for CustomField in order to prevent it from clashing with the binder validation.

Fixes #2636

## Type of change

- [x] Bugfix

## Checklist

- [x] I have read the contribution guide: https://vaadin.com/docs-beta/latest/guide/contributing/overview/
- [x] I have added a description following the guideline.
- [x] The issue is created in the corresponding repository and I have referenced it.
- [x] I have added tests to ensure my change is effective and works as intended.
- [x] New and existing tests are passing locally with my change.
- [x] I have performed self-review and corrected misspellings.

PR #7567, opened 2025-06-03
fix: add manual validation property to custom field

### LINKED ISSUE #7483: Problem with disappearing ValidationError (BlurListener?) when combining DateTimeField+x in CustomField?
### Description

Good Morning!
We debugged a disappearing validation error in a custom field with date picker and are currently assuming it might be a problem with the blur listener.

Please see the classes in the .zip
I downloaded a fresh vaadin spring project from your page and filled it with folder named "minimal" (contains 4 classes). 

### Expected outcome

If the fieldvalidation gives ERROR, then that ERROR should remain until the user changed the fieldvalue.
But instead, the validationerror disappears when the user clicks any other place.
Note: the bug only appears in combination with our compulsory (or "mandatory: value or explicitly null") field. _But:_ We debugged it several hours with 2 people, because we supposed it might be our custom field that is buggy. At least, that field _is_ indeed tricky from time to time...) , but after the analysis we came to the conclusion that the bug might indeed be in the vaadin-framework.
**Please note: We also had the exact same setting but with TextField instead of DateTimePicker. With TextField as CustomField-maincomponent everything worked as expected, the validation error remained the whole time.**

[sleigh-microservices-may.zip](https://github.com/user-attachments/files/20381460/sleigh-microservices-may.zip)

### Steps to reproduce

0) Run application/ go to localhost:8080
1) Klick "Minimal Example"
2) for "Deadline" enter  current-date-plus-one-day (eg. 23.5.25 if today is the 22.5.25), enter some time of your choice
3) for "Reminder" enter current-date-plus-2-days (eg. 24.5.25) plus a time of your choice. 
4) see the validation error "reminder must be earlier than deadline" appear
5) click somewhere else in the browser, some whitespace
6) see the validation error disappear (<- in other words: the bug)
7) click submit
8) see the valdation error appear again

### Environment

Vaadin version(s): 24.7.4
java-version: 17
OS: Linux Mint 21.3 

### Browsers

Chrome

ugur-vaadin:
## Description

This PR sets `manualValidation` property on the `CustomField` in order to keep the validation in sync. Based on the approach in the PR https://github.com/vaadin/flow-components/pull/6792.

Possibly fixes other issues around `CustomField` validation.

Fixes #7483 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


ugur-vaadin:
> I suppose we can now also remove `FieldValidationUtil` from `CustomField` added in #3277. Also it would be good to have some IT that would verify the improvement.

Removed `FieldValidationUtil` and added an IT.

vursen (approved):

PR #7616, opened 2025-06-18
fix: interrupt all uploads

sissbruecker:
## Description

Ensures that all ongoing uploads are interrupted when the application triggers interrupt on the Upload component.

## Type of change

- Bugfix

### DIFF
--- a/vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/main/java/com/vaadin/flow/component/upload/tests/UploadInterruptView.java
+++ b/vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/main/java/com/vaadin/flow/component/upload/tests/UploadInterruptView.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2000-2025 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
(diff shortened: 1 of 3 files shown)


vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/main/java/com/vaadin/flow/component/upload/tests/UploadInterruptView.java
```diff
@@ -0,0 +1,98 @@
...
+        public OutputStream receiveUpload(String fileName, String MIMEType) {
+            OutputStream outputStream = super.receiveUpload(fileName, MIMEType);
+            if (isInterruptableFile(fileName)) {
+                // Also delay the interrupted file to allow other uploads to
+                // start
+                return new SlowOutputStream(outputStream, 500);
+            }
+            {
+                return new SlowOutputStream(outputStream, 1000);
+            }
```
mcollovati: All those braces look a bit confusing to me :smile: . What about something like the following?
```suggestion
            // Also delay the interrupted file to allow other uploads to start
            int delay = isInterruptableFile(fileName) ? 500 : 1000;
            return new SlowOutputStream(outputStream, delay);
```

mcollovati (changes requested):

mcollovati (approved):

PR #7676, opened 2025-07-11
[EXPERIMENT] refactor: support flat hierarchy in TreeGrid

vursen:
> [!WARNING]
> This PR is created for testing purposes only.

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
@@ -24,18 +24,15 @@ import java.util.Objects;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import com.vaadin.flow.component.ClientCallable;
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.ComponentEventListener;
 import com.vaadin.flow.component.ComponentUtil;
 import com.vaadin.flow.component.dependency.JsModule;
 import com.vaadin.flow.component.grid.Grid;
 import com.vaadin.flow.component.grid.GridArrayUpdater;
-import com.vaadin.flow.component.grid.GridArrayUpdater.UpdateQueueData;
 import com.vaadin.flow.component.grid.dataview.GridDataView;
 import com.vaadin.flow.component.grid.dataview.GridLazyDataView;
 import com.vaadin.flow.component.grid.dataview.GridListDataView;
-import com.vaadin.flow.component.internal.AllowInert;
 import com.vaadin.flow.data.binder.PropertyDefinition;
 import com.vaadin.flow.data.provider.BackEndDataProvider;
 import com.vaadin.flow.data.provider.CallbackDataProvider;
@@ -44,29 +41,20 @@ import com.vaadin.flow.data.provider.DataCommunicator;
 import com.vaadin.flow.data.provider.DataProvider;
 import com.vaadin.flow.data.provider.ListDataProvider;
 import com.vaadin.flow.data.provider.hierarchy.HasHierarchicalDataProvider;
-import com.vaadin.flow.data.provider.hierarchy.HierarchicalArrayUpdater.HierarchicalUpdate;
-import com.vaadin.flow.data.provider.hierarchy.HierarchicalDataCommunicator;
 import com.vaadin.flow.data.provider.hierarchy.HierarchicalDataProvider;
 import com.vaadin.flow.data.provider.hierarchy.HierarchicalQuery;
 import com.vaadin.flow.data.provider.hierarchy.TreeData;
 import com.vaadin.flow.data.renderer.ComponentRenderer;
 import com.vaadin.flow.data.renderer.LitRenderer;
 import com.vaadin.flow.data.renderer.Renderer;
-import com.vaadin.flow.dom.DisabledUpdateMode;
 import com.vaadin.flow.dom.Element;
-import com.vaadin.flow.function.SerializableBiFunction;
 import com.vaadin.flow.function.SerializableComparator;
-import com.vaadin.flow.function.SerializableConsumer;
 import com.vaadin.flow.function.SerializablePredicate;
 import com.vaadin.flow.function.SerializableSupplier;
 import com.vaadin.flow.function.ValueProvider;
-import com.vaadin.flow.internal.JsonUtils;
-import com.vaadin.flow.server.VaadinRequest;
 import com.vaadin.flow.shared.Registration;
 
-import elemental.json.JsonArray;
 import elemental.json.JsonObject;
-import elemental.json.JsonValue;
 
 /**
  * Tree Grid is a component for displaying hierarchical tabular data grouped
@@ -80,139 +68,6 @@ import elemental.json.JsonValue;
 @JsModule("@vaadin/grid/src/vaadin-grid-tree-toggle.js")
 public class TreeGrid<T> extends Grid<T>
         implements HasHierarchicalDataProvider<T> {
-
-    private static final class TreeGridUpdateQueue extends UpdateQueue
-            implements HierarchicalUpdate {
-
-        private SerializableConsumer<List<JsonValue>> arrayUpdateListener;
-
-        private TreeGridUpdateQueue(UpdateQueueData data, int size) {
-            super(data, size);
-        }
-
-        public void setArrayUpdateListener(
-                SerializableConsumer<List<JsonValue>> arrayUpdateListener) {
-            this.arrayUpdateListener = arrayUpdateListener;
-        }
-
-        @Override
-        public void set(int start, List<JsonValue> items) {
-            super.set(start, items);
-
-            if (arrayUpdateListener != null) {
-                arrayUpdateListener.accept(items);
-            }
-        }
-
-        @Override
-        public void set(int start, List<JsonValue> items, String parentKey) {
-            enqueue("$connector.set", start,
-                    items.stream().collect(JsonUtils.asArray()), parentKey);
-
-            if (arrayUpdateListener != null) {
-                arrayUpdateListener.accept(items);
-            }
-        }
-
-        @Override
-        public void clear(int start, int length) {
-            if (!getData().getHasExpandedItems().get()) {
-                enqueue("$connector.clearExpanded");
-            }
-            super.clear(start, length);
-        }
-
-        @Override
-        public void clear(int start, int length, String parentKey) {
-            enqueue("$connector.clear", start, length, parentKey);
-        }
-
-        @Override
-        public void commit(int updateId, String parentKey, int levelSize) {
-            enqueue("$connector.confirmParent", updateId, parentKey, levelSize);
-            commit();
-        }
-    }
-
-    private class TreeGridArrayUpdaterImpl implements TreeGridArrayUpdater {
-        // Approximated size of the viewport. Used for eager fetching.
-        private static final int EAGER_FETCH_VIEWPORT_SIZE_ESTIMATE = 40;
-
-        private UpdateQueueData data;
-        private SerializableBiFunction<UpdateQueueData, Integer, UpdateQueue> updateQueueFactory;
-        private int viewportRemaining = 0;
-        private final List<JsonValue> queuedParents = new ArrayList<>();
-        private transient VaadinRequest previousRequest;
-
-        public TreeGridArrayUpdaterImpl(
-                SerializableBiFunction<UpdateQueueData, Integer, UpdateQueue> updateQueueFactory) {
-            this.updateQueueFactory = updateQueueFactory;
-        }
-
-        @Override
-        public TreeGridUpdateQueue startUpdate(int sizeChange) {
-            TreeGridUpdateQueue queue = (TreeGridUpdateQueue) updateQueueFactory
-                    .apply(data, sizeChange);
-
-            if (VaadinRequest.getCurrent() != null
-                    && !VaadinRequest.getCurrent().equals(previousRequest)) {
-                // Reset the viewportRemaining once for a server roundtrip.
-                viewportRemaining = EAGER_FETCH_VIEWPORT_SIZE_ESTIMATE;
-                queuedParents.clear();
-                previousRequest = VaadinRequest.getCurrent();
-            }
-
-            queue.setArrayUpdateListener((items) -> {
-                // Prepend the items to the queue of potential parents.
-                queuedParents.addAll(0, items);
-
-                while (viewportRemaining > 0 && !queuedParents.isEmpty()) {
-                    viewportRemaining--;
-                    JsonObject parent = (JsonObject) queuedParents.remove(0);
-                    T parentItem = getDataCommunicator().getKeyMapper()
-                            .get(parent.getString("key"));
-
-                    if (isExpanded(parentItem)) {
-                        int childLength = Math.max(
-                                EAGER_FETCH_VIEWPORT_SIZE_ESTIMATE,
-                                getPageSize());
-
-                        // There's still room left in the viewport and the item
-                        // is expanded. Set parent requested range for it.
-                        getDataCommunicator().setParentRequestedRange(0,
-                                childLength, parentItem);
-
-                        // Stop iterating the items on this level. The request
-                        // for child items above will end up back in this while
-                        // loop, and to processing any parent siblings that
-                        // might be left in the queue.
-                        break;
-                    }
-
-                }
-            });
-
-            return queue;
-        }
-
-        @Override
-        public void initialize() {
-            initConnector();
-            updateSelectionModeOnClient();
-            getDataCommunicator().setRequestedRange(0, getPageSize());
-        }
-
-        @Override
-        public void setUpdateQueueData(UpdateQueueData data) {
-            this.data = data;
-        }
-
-        @Override
-        public UpdateQueueData getUpdateQueueData() {
-            return data;
-        }
-    }
-
     /**
      * Creates a new {@code TreeGrid} without support for creating columns based
      * on property names. Use an alternative constructor, such as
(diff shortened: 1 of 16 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/treegrid/it/TreeGridHugeTreeIT.java
```diff
@@ -79,6 +80,7 @@ public void collapsed_rows_invalidated_correctly() {
     }
 
     @Test
+    @Ignore
```
vursen, line 83: Currently a known limitation:

> setViewportRange does not currently remove items from Cache and KeyMapper when they end up out of the viewport. However, the previous version also only removed items from KeyMapper while keeping them in HierarchyMapper, so item instances actually continued to remain in memory. I feel this could be optimized separately if needed.

https://github.com/vaadin/flow/pull/22019#issue-3313298808

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/treegrid/it/TreeGridPreloadIT.java
```diff
@@ -94,29 +86,7 @@ public void firstExpanded_shouldPreLoadDataForExpandedChildren() {
     @Test
     public void firstExpanded_shouldOptimizeDataProviderFetches() {
         open(Arrays.asList(0), null, null, null, null);
-        Assert.assertEquals("55", dataProviderFetchCount.getValue());
-    }
-
-    @Test
-    public void firstExpanded_shouldNotHaveDataForExpandedRootItemsOutsideViewportEstimate() {
```
vursen, line 101: The removed tests in this file relied on `setParentRequestedRange`, which is no longer available

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/treegrid/it/TreeGridPreloadIT.java
```diff
@@ -128,16 +98,6 @@ public void firstExpanded_reExpand_shouldHaveItemRecursivelyExpanded() {
         verifyRow(4, "/0/0/1/0/2/0/3/0/4/0");
     }
 
-    @Test
-    public void firstExpanded_reExpandChild_shouldUseCachedDataForExpandedChildren() {
```
vursen, line 132: Expanded children are maintained on the server side now.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/HierarchyColumnComponentRenderer.java
```diff
@@ -58,7 +58,7 @@ protected String getTemplateExpression() {
         var clickListener = "e => requestAnimationFrame(() => { e.defaultPrevented && onClick(e) })";
 
         return "<vaadin-grid-tree-toggle @click=${" + clickListener
-                + "} class=${item.cssClassName} .leaf=${!item.children} .expanded=${model.expanded} .level=${model.level}>"
+                + "} class=${item.cssClassName} .leaf=${!item.children} .expanded=${live(model.expanded)} .level=${model.level}>"
```
vursen: This change prevents a glitch where clicking the tree toggle could leave it in an incorrect state when the same DOM element was reused for a different row. 

Previously, clicking the toggle added the item to the expandedItems array, which triggered a re-render of this Lit template. In the new version, the expanded state of the tree toggle is no longer synchronized with the template on click because there is no expandedItems on the client side. As a result, when the user clicks the toggle, the component updates the expanded state internally, but Lit doesn't re-render because it detects no change when comparing with the previously rendered value. 

Using `live` forces Lit to compare the new value against the actual rendered attribute when re-rendering with a different row, which avoids this issue.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
```diff
@@ -14,45 +14,8 @@ window.Vaadin.Flow.gridConnector.initLazy = (grid) => {
...
-  dataProviderController.isLoadingOriginal = dataProviderController.isLoading;
-  dataProviderController.isLoading = function () {
-    return grid.$connector.hasEnsureSubCacheQueue() || this.isLoadingOriginal();
-  };
-
-  dataProviderController.getItemSubCache = function (item) {
-    return this.getItemContext(item)?.subCache;
-  };
-
   let cache = {};
```
vursen, line 17: Technically, this internal cache is now redundant, and gridConnector could be refactored to add items directly to the web component's cache (dataProviderController). But, I'd prefer to improve that separately to avoid scope creep.

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/treegrid/it/TreeGridRefreshAllIT.java
```diff
@@ -60,8 +60,7 @@ public void expandItems_scroll_refreshAll_loadingStateResolved() {
 
         refreshAllButton.click();
 
-        Assert.assertFalse("TreeGrid was left with pending requests.",
-                getTreeGrid().hasAttribute("loading"));
+        waitUntilNot((driver) -> getTreeGrid().hasAttribute("loading"));
```
vursen, line 63: I'm still not sure what is causing this timing change, but adding `waitUntil` makes the test pass, so I think it's fine considering that it still reaches the "not loading" state.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -326,19 +215,45 @@ public TreeGrid(HierarchicalDataProvider<T, ?> dataProvider) {
...
+                ArrayUpdater arrayUpdater,
+                SerializableConsumer<JsonArray> dataUpdater,
+                StateNode stateNode,
+                SerializableSupplier<ValueProvider<T, String>> uniqueKeyProviderSupplier) {
+            super(dataGenerator, arrayUpdater, dataUpdater, stateNode,
+                    uniqueKeyProviderSupplier);
+        }
+
+        @Override
+        protected List<T> preloadFlatRangeForward(int start, int length) {
```
vursen, line 241: These overrides are here to make these methods accessible from the TreeGrid class without making them public in HierarchicalDataCommunicator.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
```diff
@@ -700,128 +475,38 @@ window.Vaadin.Flow.gridConnector.initLazy = (grid) => {
...
+    currentUpdateClearRange = [firstPage, firstPage + updatedPageCount - 1];
 
     for (let i = 0; i < updatedPageCount; i++) {
       let page = firstPage + i;
-      let items = cache[pkey][page];
-      grid.$connector.doDeselection(items.filter((item) => selectedKeys[item.key]));
-      items.forEach((item) => grid.closeItemDetails(item));
-      delete cache[pkey][page];
-      updateGridCache(page, parentKey);
-      updateGridItemsInDomBasedOnCache(items);
```
vursen, line 714: There is already a call to `grid.__updateVisibleRows()` below which triggers a re-render for all rows, so there is no need to call `updateGridItemsInDomBasedOnCache` here.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -326,20 +214,54 @@ public TreeGrid(HierarchicalDataProvider<T, ?> dataProvider) {
...
+            super(dataGenerator, arrayUpdater, dataUpdater, element.getNode(),
+                    uniqueKeyProviderSupplier);
+            this.element = element;
+        }
+
+        @Override
+        public void reset() {
+            super.reset();
+            if (element != null) {
+                element.callJsFunction("$connector.reset");
```
vursen, line 236: This replaces the `$connector.ensureHierarchy` call which was previously made from the data communicator class. In the future, we could consider adding `ArrayUpdater#reset` for this purpose.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
```diff
@@ -542,87 +356,49 @@ window.Vaadin.Flow.gridConnector.initLazy = (grid) => {
...
 
-      const updatedItems = updateGridCache(page, pkey);
+      const updatedItems = updateGridCache(page);
       if (updatedItems) {
         itemsUpdated(updatedItems);
-        updateGridItemsInDomBasedOnCache(updatedItems);
       }
     }
+
+    grid.__updateVisibleRows();
```
vursen, line 380: `updateGridItemsInDomBasedOnCache` was replaced with `grid.__updateVisibleRows()`

The former only updated existing items in the viewport, matching them by key, but this is no longer enough because the server may now return new items when an item is expanded, and those rows need to be updated as well.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -326,20 +214,54 @@ public TreeGrid(HierarchicalDataProvider<T, ?> dataProvider) {
...
+                GridArrayUpdater arrayUpdater,
                 SerializableSupplier<ValueProvider<T, String>> uniqueKeyProviderSupplier) {
-
-            return new HierarchicalDataCommunicator<>(dataGenerator,
-                    arrayUpdater,
-                    data -> element.callJsFunction(
-                            "$connector.updateHierarchicalData", data),
-                    element.getNode(), uniqueKeyProviderSupplier);
+            return new TreeGridDataCommunicator<>(element, dataGenerator,
+                    arrayUpdater, null, uniqueKeyProviderSupplier);
```
vursen, line 265: The updated data communicator currently ignores dataUpdater:

> 2. setViewportRange currently always sends the full viewport to the client even when some items are already present on the client-side. I plan to address this separately, see https://github.com/vaadin/flow/issues/21989
> 3. refresh(T item) currently also sends the full viewport on every item refresh. I plan to address this separately, see https://github.com/vaadin/flow/issues/21989https://github.com/vaadin/flow/pull/22019#issue-3313298808

https://github.com/vaadin/flow/pull/22019#issue-3313298808

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/treegrid/it/TreeGridRefreshAllIT.java
vursen: I investigated this further, and the timing change seems to be caused by a slight difference in behavior. In the previous implementation, refreshAll didn't trigger the loading of the third level, whereas the new implementation does load the third level because it operates with a flat range. As a result, it needs one additional request to fully fill the grid, which increases the delay here. Personally, it doesn't seem to be a big issue, since both result in the same visual outcome.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -1138,7 +1032,66 @@ public void scrollToIndex(int... indexes) {
...
+     * <p>
+     * This method has package-private visibility to allow testing.
+     *
+     * @param path
+     *            the path to the item to use as the center of the viewport
+     *            range
+     * @param padding
+     *            the number of items to add on each side of the center item
+     */
+    @ClientCallable
```
web-padawan: Do we need `@AllowInert` for this `@ClientCallable` or it's not supposed to be called in the background?

vursen: Good question. Considering that `setViewportRange` has `@AllowInert`, I guess this method should also have it.

web-padawan: It would be good to verify it with `InertTreeGridIT` added in https://github.com/vaadin/flow-components/pull/6465.

sissbruecker: I think it would make sense to add it. The main purpose of the method is to be called through the server itself with `scrollToIndex` which due to the design requires a roundtrip to the client. It also shouldn't cause any harm otherwise, as it essentially does the same thing as loading data which is already allowed when inert.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/treeGridConnector.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  window.Vaadin.Flow.gridConnector.initLazy(grid);
+
+  function getViewportRange() {
+    const renderedRows = grid._getRenderedRows();
+    return [renderedRows[0]?.index ?? 0, renderedRows[renderedRows.length - 1]?.index ?? 0];
+  }
+
+  grid._dataProviderController.__loadCachePage = function (cache, page) {
```
sissbruecker: The other overrides are sort of understandable, but I have no idea what this does. We should maybe leave a comment on why it's necessary.

Now that I think about it, maybe it would also be good to give a brief explanation on why all the hierarchy stuff is overridden from grid: Flow TreeGrid doesn't use the hierarchy features of the web component, instead uses a flat representation of the tree data, hierarchy managed on the server. Otherwise this could easily confuse future contributors who know about the hierarchy features in the web component.

sissbruecker (approved):
Code changes regarding using flattened data seem to make sense, though there are some changes here and there (calculations of page ranges) that are hard to verify with a review. Otherwise did manual testing and could not observe any issues.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
ugur-vaadin: Is there a separate PR or issue to address this? Just want to make sure we don't forget this part.

ugur-vaadin (approved):

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
vursen: Thanks, done.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/treeGridConnector.ts
web-padawan: I think this should eventually use `_shouldLoadCachePage()` added in https://github.com/vaadin/web-components/pull/9984. 
It would be good to add a comment clarifying this indeed.

vursen: The comment is already there: `// TODO: Use _shouldLoadCachePage instead when released`

vursen: > maybe it would also be good to give a brief explanation on why all the hierarchy stuff is overridden from grid

Added a brief explanation of the main points at the start of `treeGridConnector`.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
vursen: Thanks, not yet, I'm planning to create one after this PR is merged.

web-padawan (approved):

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/treegrid/it/TreeGridHugeTreeIT.java
tomivirkki, line 1: Noticed this issue on the `vaadin-grid/treegrid-huge-tree` test page. If you recursively expand the tree grid, scroll to end, and then start scrolling backwards, multiple branches are jumped over once you're close to reaching the first parent item:

https://github.com/user-attachments/assets/a919b6d3-b441-4aa2-91e8-a499172352ce

vursen: This is kind of a known limitation, which also existed in the previous version, didn't it? Items may be skipped because the scroll container can only increase its height at the bottom. I'm going to research separately if the scroll position could be adjusted automatically (and items preloaded in backward direction) when scrolling backward.

tomivirkki: > which also existed in the previous version, didn't it?

I tested the same scenario with `main` revision and didn't notice a jump.

vursen: Yeah, it appears that the main revision preloads items in a breadth-first manner while this new implementation does so depth-first. This allows it to request fewer items but also results in the observed behavior in the mentioned case (3 root items -> 3 children each -> many descendants). However, even having breadth-first traversal in place, items might still be jumped over as soon as TreeGrid has more root items. I'll create a separate issue to explore how we could fix it better.

vursen: Created issue: https://github.com/vaadin/flow-components/issues/7852

PR #7686, opened 2025-07-16
refactor: remove logic associated with $connector.clearExpanded

vursen:
## Description

The `$connector.clearExpanded` method was introduced during the migration of TreeGrid from Vaadin Framework 8 in https://github.com/vaadin/vaadin-grid-flow/pull/241. However, its current purpose is unclear, as there appears to be no scenario in which expanded items are cleared by the data communicator without also invoking `$connector.collapseItems` on the client.

Part of https://github.com/vaadin/flow-components/issues/7688

## Type of change

- [x] Refactor


vursen:
> as their only usage in the codebase is being removed in this PR.

Good question. I was planning to deprecate and remove it in Vaadin 25 as the next step. Any concerns?

DiegoCardoso:
> Good question. I was planning to deprecate and remove it in Vaadin 25 as the next step. Any concerns?

I think it makes sense to deprecate it, if you're not planning to use it anymore.

DiegoCardoso (approved):

PR #7719, opened 2025-07-18
fix: use frozen panel size to calculate the visible rows/cols on scroll

### LINKED ISSUE #7507: Spreadsheets with Freeze Pane not fully rendering all columns/rows upon resizing of Window or rows/columns
### Description

When viewing a spreadsheet with freeze panes, if the window is resized or columns or rows resized; sometimes not enough columns will be rendered to the right during and after scrolling, and sometimes entire rows or columns are not rendered at all between the freeze pane and the unfrozen rows/columns. Sometimes it's so bad that no rows or no columns are rendered at all beyond the freeze pane.

Image
Image
Image

### Expected outcome

I would expect the entire spreadsheet area to be filled with rows/columns and there not to be entirely blank sections.

I was able to reproduce this by downloading and running locally the spreadsheet demo found here:
https://github.com/vaadin/spreadsheet-demo
And by uploading a spreadsheet with several frozen rows and columns.

[TestFreezePanes.xlsx](https://github.com/user-attachments/files/20416270/TestFreezePanes.xlsx)

### Steps to reproduce

1. Run the demo project locally.
2. Upload a spreadsheet with several frozen rows and columns.
3. Resize the browser window in various ways between very small and back to large or in between. Or also resize columns or rows. And scroll around within the non-frozen panes.

### Environment

Vaadin version(s): 24 on this commit of the demo: https://github.com/vaadin/spreadsheet-demo/commit/7a6f28c61eb9775d6eaa2c8a8fde3a525a6c2d1c
OS: Windows 11

### Browsers

Chrome

DiegoCardoso:
## Description

Use the top and left frozen panel size to calculate the number of headers and rows to be rendered when the user scrolls. Currently, the size of the frozen panels are not accounted which causes some of the columns and rows to not be rendered even when there's available space for that.

Fixes #7507

## Type of change

- Bugfix

### DIFF
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-integration-tests/src/main/java/com/vaadin/flow/component/spreadsheet/tests/fixtures/CustomEditorRowFixture.java
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-integration-tests/src/main/java/com/vaadin/flow/component/spreadsheet/tests/fixtures/CustomEditorRowFixture.java
@@ -1,0 +1,90 @@
+/**
+ * Copyright 2000-2025 Vaadin Ltd.
+ *
+ * This program is available under Vaadin Commercial License and Service Terms.
+ *
+ * See {@literal <https://vaadin.com/commercial-license-and-service-terms>} for the full
+ * license.
+ */
+package com.vaadin.flow.component.spreadsheet.tests.fixtures;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.poi.ss.usermodel.Cell;
+import org.apache.poi.ss.usermodel.Sheet;
+
+import com.vaadin.flow.component.Component;
+import com.vaadin.flow.component.button.Button;
+import com.vaadin.flow.component.spreadsheet.Spreadsheet;
+import com.vaadin.flow.component.spreadsheet.SpreadsheetComponentFactory;
+import com.vaadin.flow.component.textfield.TextField;
+
+public class CustomEditorRowFixture implements SpreadsheetFixture {
+
+    static final int COMPONENT_ROW_INDEX = 1;
+
+    @Override
+    public void loadFixture(final Spreadsheet spreadsheet) {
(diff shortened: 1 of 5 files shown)


WoozyG:
One more test that would be helpful is one where a frozen row and/or column is also hidden, to make sure all the math still works.

Some of our spreadsheets hide the first two rows when displaying in the browser as they hold headers already present in the ui, but we unhide them on download so they show in excel.  Many workbooks freeze the top 5 or so rows, including those hidden ones.

DiegoCardoso:
> One more test that would be helpful is one where a frozen row and/or column is also hidden, to make sure all the math still works.

That was a good call. Thanks to your comment, I was able to find another issue with hidden columns/rows in the freeze pane area. The calculation for the left/right shift was not working as expected. Added a fix for that. Thanks!

vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-integration-tests/src/main/java/com/vaadin/flow/component/spreadsheet/tests/fixtures/CustomEditorRowFixture.java
```diff
@@ -0,0 +1,90 @@
...
+        private final Map<Integer, TextField> textFields = new HashMap<>();
+
+        @Override
+        public Component getCustomComponentForCell(Cell cell, int rowIndex,
+                int columnIndex, Spreadsheet spreadsheet, Sheet sheet) {
+            if (rowIndex != 0 || columnIndex != 0) {
+                return null;
+            }
+
+            Button toggleCustomEditorVisibilityButton = new Button(
```
ugur-vaadin, line 46: Nit: We could use `NativeButton` to possibly reduce the dependencies moving forward.

DiegoCardoso: I can address it in a separate PR and change the other fixture at the same time.

PR #7726, opened 2025-07-22
test: extract getting items into helper class

ugur-vaadin:
## Description

Extracts the static method `getItems` for virtual list elements to a helper class.

## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] Test

## Checklist


vaadin-virtual-list-flow-parent/vaadin-virtual-list-flow-integration-tests/src/test/java/com/vaadin/flow/component/virtuallist/tests/VirtualListHelpers.java
```diff
@@ -0,0 +1,53 @@
...
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+import elemental.json.Json;
+import elemental.json.JsonArray;
+import elemental.json.JsonObject;
+
+public class VirtualListHelpers {
+
+    public static JsonArray getItems(WebDriver driver, WebElement element) {
```
web-padawan, line 31: I wonder if this should be part of `VirtualListElement` instead so that it can be used in custom ITs?

ugur-vaadin: I considered that too, but couldn't be sure. What do you think? Would it be worth having an explicit API there?

sissbruecker: I think the API could use some more thought, returning a `JsonArray` is both confusing and inconvenient. Let's keep this internal for now.

PR #7732, opened 2025-07-22
test: update ConfirmDialog ITs to not rely on overlay element

web-padawan:
## Description

Updated ConfirmDialog ITs to not rely on the overlay element in order to make them pass also with native popover.

## Type of change

- Test


vaadin-confirm-dialog-flow-parent/vaadin-confirm-dialog-flow-integration-tests/src/test/java/com/vaadin/flow/component/confirmdialog/tests/StylingIT.java
```diff
@@ -87,7 +87,8 @@ private ConfirmDialogElement getConfirmDialog() {
     }
 
     private TestBenchElement getOverlay() {
-        return ((TestBenchElement) getConfirmDialog().getContext());
+        return ((TestBenchElement) getConfirmDialog()
```
web-padawan, line 90: This is needed because `getContext()` override will be removed from `ConfirmDialogElement`.

vaadin-confirm-dialog-flow-parent/vaadin-confirm-dialog-flow-integration-tests/src/main/java/com/vaadin/flow/component/confirmdialog/tests/EventView.java
```diff
@@ -32,7 +32,8 @@ public EventView() {
         Button toggleCloseOnEscButton = new Button("Toggle close on Esc",
                 e -> dialog.setCloseOnEsc(!dialog.isCloseOnEsc()));
         toggleCloseOnEscButton.setId("toggle-close-on-esc");
+        dialog.add(toggleCloseOnEscButton);
```
web-padawan, line 35: The button is actually only clicked while the dialog is open, so I added it to the dialog itself.

Note: corresponding APIs were deprecated in https://github.com/vaadin/flow-components/pull/7732 but maybe we should un-deprecate those from `HasComponents` after updating to use native popover.

sissbruecker (approved):

PR #7737, opened 2025-07-23
test: update Popover ITs to not rely on the overlay element

web-padawan:
## Description

Updated Popover ITs to not rely on the overlay element in order to make them pass also with native popover.

## Type of change

- Test


vaadin-popover-flow-parent/vaadin-popover-flow-integration-tests/src/test/java/com/vaadin/flow/component/popover/tests/PopoverIT.java
```diff
@@ -138,12 +133,4 @@ public void disableCloseOnEsc_pressEsc_popoverDoesNotClose() {
     private void clickTarget() {
         clickElementWithJs("popover-target");
     }
-
-    private void checkPopoverIsClosed() {
-        waitForElementNotPresent(By.tagName(POPOVER_OVERLAY_TAG));
```
sissbruecker, line 143: Just a note in case we might need some checks like this in the future. We should be able to verify that an overlay is shown as native popover by checking if the overlay element in the shadow root matches the `:popover-open` selector.

PR #7743, opened 2025-07-24
chore: dedupe dependencies when using local web components

sissbruecker:
## Description

Currently dependencies imported both from web-components and somewhere in the Flow integration test sources are loaded / bundled twice, which can result in errors, for example when one of those dependency uses `instanceof` checks. We also load Lit twice at the moment due to that.

This change de-dupes some common dependencies, so that only one version is loaded.

## Type of change

- Internal

### DIFF
--- a/shared/web-components-vite-plugin.ts
+++ b/shared/web-components-vite-plugin.ts
@@ -36,7 +36,7 @@ export function useLocalWebComponents(webComponentsRepoPath: string): PluginOpti
       };
     },
     resolveId(id: string) {
-      if (/^(@vaadin)/.test(id)) {
+      if (id.startsWith('@vaadin')) {
         return this.resolve(path.join(nodeModulesPath, id));
       }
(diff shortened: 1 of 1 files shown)


shared/web-components-vite-plugin.ts
```diff
@@ -22,11 +22,21 @@ export function useLocalWebComponents(webComponentsRepoPath: string): PluginOpti
...
+        },
+        resolve: {
+          dedupe: ['lit', 'lit-html', 'ol'],
         }
       };
     },
-    resolveId(id) {
-      if (/^(@polymer|@vaadin)/.test(id)) {
+    resolveId(id: string) {
+      if (/^(@vaadin)/.test(id)) {
```
vursen:
```suggestion
      if (id.startsWith('@vaadin')) {
```

vursen (approved):

PR #7755, opened 2025-07-28
fix: improve custom editors internal structure to prevent NPE errors

DiegoCardoso:
## Description

Sometimes, the call to `removeCellCustomEditor` can be done to a widget already removed from its parent, which causes a client-side error.

This change tries to address that with some refactoring on the custom editors' internal map.

The refactor aims to replace the keys that currently refer to the cell address with the editor ID itself. That way, in case the same editor is shared by more than one cell, the same `Slot` will be used. With the current logic, more than one `Slot` element would be created for the same editor, which could cause issues like the editor not being shown because it was moved to the other `Slot` parent.

## Type of change

- Bugfix / Refactor


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-integration-tests/src/test/java/com/vaadin/flow/component/spreadsheet/test/CustomEditorIT.java
```diff
@@ -354,6 +354,91 @@ public void customEditorIsVisible_sheetIsChanged_editorsRemoved() {
...
+        editor = maybeEditor.get();
+        var editorC2Id = editor.getId();
+        editor.sendKeys("AnotherValue", Keys.ENTER);
+        clickCell("A1");
+        getCommandExecutor().waitForVaadin();
+        Assert.assertEquals("AnotherValue", getCellValue("C2"));
+
+        Assert.assertNotEquals(editorB2Id, editorC2Id);
+
+        // Test that focusin on cells with custom editors works correctly
```
ugur-vaadin, line 428:
```suggestion
        // Test that focusing on cells with custom editors works correctly
```

vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-integration-tests/src/test/java/com/vaadin/flow/component/spreadsheet/test/CustomEditorIT.java
```diff
@@ -354,6 +354,91 @@ public void customEditorIsVisible_sheetIsChanged_editorsRemoved() {
         }
     }
 
+    @Test
```
ugur-vaadin, line 357: Both the tests seem to pass without the fixes.

DiegoCardoso: I have removed one of the tests and refactored the other one to make it so that it breaks in `main`.

ugur-vaadin (approved):

PR #7771, opened 2025-07-31
Add disableUpdateValueOnChange to the CustomField

### LINKED ISSUE #7108: Custom Field value should be synchronised not only on change
### Describe your motivation

When you're using a custom field the value is automatically updated on `change`.

It might work for certain cases but sometimes it needs to be synchronized on click (on a button) or eagerly.

In that case, the value event might useless or might cause extra communication.

First be able to disable this behaviour (then the developer can manage it entirely) or switch the event to a different one. (I don't know if that would work with the event propagation).

_No response_

### Additional context

_No response_

jcgueriaud1:
## Description

Please list all relevant dependencies in this section and provide summary of the change, motivation and context.

Fixes #7108  (issue)

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/vaadin-custom-field-flow-parent/vaadin-custom-field-flow/src/main/java/com/vaadin/flow/component/customfield/CustomField.java
+++ b/vaadin-custom-field-flow-parent/vaadin-custom-field-flow/src/main/java/com/vaadin/flow/component/customfield/CustomField.java
@@ -31,6 +31,7 @@ import com.vaadin.flow.component.dependency.JsModule;
 import com.vaadin.flow.component.dependency.NpmPackage;
 import com.vaadin.flow.component.shared.HasValidationProperties;
 import com.vaadin.flow.component.shared.InputField;
+import com.vaadin.flow.dom.DomListenerRegistration;
 import com.vaadin.flow.dom.Element;
 
 /**
@@ -53,6 +54,7 @@ public abstract class CustomField<T> extends AbstractField<CustomField<T>, T>
         InputField<AbstractField.ComponentValueChangeEvent<CustomField<T>, T>, T>,
         HasTheme, HasValidationProperties {
 
+    private final DomListenerRegistration updateValueListener;
     /**
      * Default constructor.
      */
@@ -71,7 +73,7 @@ public abstract class CustomField<T> extends AbstractField<CustomField<T>, T>
     public CustomField(T defaultValue) {
         super(defaultValue);
         // Force a value update when the change event generated
-        getElement().addEventListener("change", e -> this.updateValue());
+        updateValueListener = getElement().addEventListener("change", e -> this.updateValue());
         getElement().setProperty("manualValidation", true);
     }
 
@@ -206,5 +208,13 @@ public abstract class CustomField<T> extends AbstractField<CustomField<T>, T>
         getThemeNames().removeAll(
                 Stream.of(variants).map(CustomFieldVariant::getVariantName)
                         .collect(Collectors.toList()));
+    }
+
+    /**
+     * Remove the listener that calls {@link #updateValue()} when an inner field "onchange" event is sent.
+     * That means the {@link #updateValue()}  needs to be called programmatically in the implementation of the CustomField
+     */
+    public void disableUpdateValueOnChange() {
+        updateValueListener.remove();
     }
 }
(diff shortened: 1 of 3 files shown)


vaadin-custom-field-flow-parent/vaadin-custom-field-flow/src/main/java/com/vaadin/flow/component/customfield/CustomField.java
```diff
@@ -207,4 +211,13 @@ public void removeThemeVariants(CustomFieldVariant... variants) {
...
                         .collect(Collectors.toList()));
     }
+
+    /**
+     * Remove the listener that calls {@link #updateValue()} when an inner field
+     * "onchange" event is sent. That means the {@link #updateValue()} needs to
+     * be called programmatically in the implementation of the CustomField
+     */
+    public void disableUpdateValueOnChange() {
+        updateValueListener.remove();
```
thevaadinman: ..so what if I want to re-enable this?

mstahv: Then add it back yourself. This change doesn't make toggling it any more difficult 🤷‍♂️

mstahv:
Very nice improvement, I have missed this feature myself. I'd consider thought adding a constructor that allows you to bypass this "handy" (or "surprising" as I might call it) feature.

vaadin-custom-field-flow-parent/vaadin-custom-field-flow/src/main/java/com/vaadin/flow/component/customfield/CustomField.java
jcgueriaud1: I'm not sure in which case you would need to re enable it.

You could still do this in your code
```
getElement().addEventListener("change",
                e -> this.updateValue());
```

I could technically add a `enableUpdateValueOnChange` that does this
```
if (updateValueListener != null) {
updateValueListener.remove();
updateValueListener = null;
                }
updateValueListener = getElement().addEventListener("change",
                e -> this.updateValue());
```

And also in the disableUpdateValueOnChange
I could add this:
`updateValueListener = null;`

thevaadinman: uh.. it was never toggleable in the first place. It was there by default, this change adds a way to remove it, but once removed I don't see a way to restore default behavior (besides re-creating the object or adding custom logic)

Edit: yes, I know, "who would ever want to do that?". Someone probably would.

jcgueriaud1: I think the main idea of this method is to remove the default behaviour and manage the updateValue in your code, not to toggle it.

web-padawan: Would it make sense to not remove the listener but instead have API that would look make "manual" value updates opt-in e.g. `setManualValueChange(Boolean)` or something similar?

mstahv: 👍 Thus, I'd consider just giving a flag in constructor to skip adding it and nothing else.

sissbruecker: I'd also prefer some API that uses the term "manual". Constructor flag probably makes most sense, I don't see any use-case to make this dynamic.

jcgueriaud1: > Constructor flag probably makes most sense

I didn't do this because that's a breaking change. With a new method that doesn't break anything.

mstahv: Add a new constructor with `boolean manualValueChangeTriggering'. Then keep the old signature as shorthand with false for that. Fully backwards compatible and doesn't add new methods.

sissbruecker: Yes, this would require a new constructor overload.

jcgueriaud1: I've updated the code and added a new constructor with a default value and a manual flag.

mstahv: Excellent, looks good to me now. Tiny little piece of code that is backwards compatible and brings some true value. This will be my default constructuor in the future 😎

mstahv (approved):
Matti approves 😎

PR #7788, opened 2025-08-04
refactor: auto remove overlays on closed event

sissbruecker:
## Description

After refactoring overlay components such as Dialog to keep their overlay in the shadow root the overlay's closing animation will not be shown if the component is immediately removed when it's opened state changes (previously removing the component would still keep the overlay in the document body until the animation was finished). For components that auto-add / auto-remove themselves to / from the UI when they are opened / closed that means that the closing animation is currently skipped.

This changes `OverlayAutoAddController` and affected components (Dialog, ConfirmDialog, Notification, LoginOverlay) to auto-remove on the `closed` event, which is fired after the closing animation has finished, instead of when the opened state changes.

## Type of change

- Refactor

### DIFF
--- a/vaadin-confirm-dialog-flow-parent/vaadin-confirm-dialog-flow-integration-tests/src/test/java/com/vaadin/flow/component/confirmdialog/tests/FeaturesDiyIT.java
+++ b/vaadin-confirm-dialog-flow-parent/vaadin-confirm-dialog-flow-integration-tests/src/test/java/com/vaadin/flow/component/confirmdialog/tests/FeaturesDiyIT.java
@@ -113,7 +113,7 @@ public class FeaturesDiyIT extends AbstractComponentIT {
     }
 
     public void assertDialogClosed() {
-        Assert.assertFalse(getConfirmDialog().isPresent());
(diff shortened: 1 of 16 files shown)


vaadin-notification-flow-parent/vaadin-notification-flow/src/main/java/com/vaadin/flow/component/notification/Notification.java
```diff
@@ -660,6 +662,7 @@ protected void onDetach(DetachEvent detachEvent) {
             // itself when its parent, for example a dialog, gets attached
             // again.
             setOpened(false);
+            autoAddController.remove();
```
sissbruecker, line 665: Notification has custom logic for auto-removing itself whenever it is detached from the UI. This was added to prevent notifications that have been auto-added to a modal dialog to show up again whenever that dialog opens again. This was done by closing the notification, which now doesn't trigger auto-removal as the closed event from the client-side is not fired.

To keep this working I added a method to `OverlayAutoAddController` to allow the component to auto-remove itself when appropriate.

web-padawan (approved):

Issue #7790, opened 2025-08-04
[Popover] Remove usage of component renderer logic in favor of adding children


web-padawan:
### Motivation

In V25, the `vaadin-popover` can accept children thanks to no longer teleporting overlay to the body.
This means we don't need `renderer` anymore (it doesn't do anything else besides adding children).

### Proposed solution

Remove the Flow component renderer logic from the Popover component altogether.

PR #7791, opened 2025-08-04
refactor: remove usage of Flow component renderer from Popover

web-padawan:
## Description

Fixes https://github.com/vaadin/flow-components/issues/7790

This simplifies Popover logic quite a bit as we no longer need to take care of virtual child node IDs.

## Type of change

- Refactor


vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
sissbruecker, line 1: I think we can also remove `vaadin-renderer-flow` from the dependencies in `pom.xml`

web-padawan: Thanks, done.

sissbruecker (approved):

PR #7792, opened 2025-08-04
refactor: auto remove context menu on closed event

sissbruecker:
## Description

Make `ContextMenu` also use `OverlayAutoAddController`, so that it reuses the same logic as other overlay components to auto-remove itself on the `closed` event. This requires adding an `add` method to the controller, as `ContextMenu` has custom logic for opening itself from the client-side.

Related to https://github.com/vaadin/flow-components/pull/7788

## Type of change

- Refactor


web-padawan (approved):

vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/test/java/com/vaadin/flow/component/contextmenu/ContextMenuTest.java
```diff
@@ -246,8 +247,10 @@ public void unregisterOpenedChangeListenerOnEvent() {
             e.unregisterListener();
         });
 
-        contextMenu.getElement().setProperty("opened", true);
-        contextMenu.getElement().setProperty("opened", false);
+        ComponentUtil.fireEvent(contextMenu,
+                new ContextMenuBase.OpenedChangeEvent<>(contextMenu, false));
```
web-padawan, line 251: Should this be `true`?

sissbruecker: The argument indicates whether the event came from the client or not. With the previous setup that should not have been the case. In any case, it also shouldn't matter for this test I think.

Issue #7520, opened 2025-05-27
Refactor ContextMenu TestBench to not expose ContextMenuOverlayElement


web-padawan:
### Describe your motivation

In https://github.com/vaadin/flow-components/pull/4752 the `ContextMenuOverlayElement` was added and it is used like this:

https://github.com/vaadin/flow-components/blob/75ad92775d939889b0e592a8303675c9b6a1f719/vaadin-context-menu-flow-parent/vaadin-context-menu-flow-integration-tests/src/test/java/com/vaadin/flow/component/contextmenu/it/ContextMenuDemoIT.java#L199-L201

This can be problematic in the future as in V25 we'll switch to native `popover` and the `vaadin-context-menu-overlay` element will no longer be teleported to the `<body>` but instead will always stay inside `vaadin-context-menu` (either in shadow or light DOM, that's yet to be discussed separately).

### Describe the solution you'd like

- Consider removing the `ContextMenuOverlayElement` altogether
- Move methods `getMenuItem(String caption)` and `getMenuItems()` to `ContextMenuElement`
- Add `getOverlay()` to the `ContextMenuElement` using `getPropertyElement("_overlayElement")`
  - This can return `TestBenchElement`, see e.g. `DialogElement` - it's there but is private
  - See also https://github.com/vaadin/flow-components/issues/3677
- Optionally, also consider updating the `openByRightClick()` helper to return `getOverlay()` 

https://github.com/vaadin/flow-components/blob/75ad92775d939889b0e592a8303675c9b6a1f719/vaadin-context-menu-flow-parent/vaadin-context-menu-testbench/src/main/java/com/vaadin/flow/component/contextmenu/testbench/ContextMenuElement.java#L43-L45

### Additional context

The API has to be then backported to v23 in https://github.com/vaadin/flow-components/pull/4754

PR #7801, opened 2025-08-06
feat!: context menu test bench improvements

sissbruecker:
## Description

Improves the TestBench API for `ContextMenu`:
- Modified `ContextMenuElement.openByRightClick()` to wait for a menu to open and return it. Throws an exception if no menu opens (behavior change).
- Added `ContextMenuElement.getMenuItems()` to get all items from a menu
- Added `ContextMenuElement.getMenuItem(string)` to find an item by text
- Added `ContextMenuElement.waitUntilClosed()` to wait until a context menu has finished closing. Throws an exception if the menu does not close
- Modified `ContextMenuItemElement.openSubMenu()` to wait for a submenu to open and return it. Throws an exception if no menu opens (behavior change).

Also deprecates `ContextMenuOverlayElement`. The class is now a proxy to `ContextMenuElement` and should thus keep working as before.

Fixes https://github.com/vaadin/flow-components/issues/7520

## Type of change

- Feature

### DIFF
--- a/vaadin-context-menu-flow-parent/vaadin-context-menu-testbench/src/main/java/com/vaadin/flow/component/contextmenu/testbench/ContextMenuOverlayElement.java
+++ b/vaadin-context-menu-flow-parent/vaadin-context-menu-testbench/src/main/java/com/vaadin/flow/component/contextmenu/testbench/ContextMenuOverlayElement.java
@@ -26,27 +26,31 @@ import com.vaadin.testbench.elementsbase.Element;
  * <code>&lt;vaadin-context-menu-overlay&gt;</code> element.
  *
  * @author Vaadin Ltd
- *
+ * @deprecated Use {@link ContextMenuElement} instead.
  */
-@Element("vaadin-context-menu-overlay")
+@Element("vaadin-context-menu")
+@Deprecated(since = "25.0", forRemoval = true)
 public class ContextMenuOverlayElement extends TestBenchElement {
 
     /**
      * Get the first menu item matching the caption.
      *
      * @return Optional menu item.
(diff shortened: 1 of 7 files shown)


vaadin-context-menu-flow-parent/vaadin-context-menu-testbench/src/main/java/com/vaadin/flow/component/contextmenu/testbench/ContextMenuOverlayElement.java
```diff
@@ -26,27 +26,31 @@
  * <code>&lt;vaadin-context-menu-overlay&gt;</code> element.
  *
  * @author Vaadin Ltd
- *
+ * @deprecated Use {@link ContextMenuElement} instead.
  */
-@Element("vaadin-context-menu-overlay")
+@Element("vaadin-context-menu")
```
web-padawan, line 31: I wonder if we should just remove `ContextMenuOverlayElement` completely? It was added quite recently.

web-padawan (approved):

vaadin-context-menu-flow-parent/vaadin-context-menu-testbench/src/main/java/com/vaadin/flow/component/contextmenu/testbench/ContextMenuOverlayElement.java
sissbruecker: That was my intention as well, but since it was so easy to keep it working I thought why not keep it. We probably won't have to touch it again until we remove it.

PR #7824, opened 2025-08-11
chore: try improve flaky spreadsheet context menu IT

sissbruecker:


sissbruecker:
Looks like there is a bigger issue with the theme, so this is not a sufficient fix. Closing.

PR #7840, opened 2025-08-13
fix: cancel request debouncer if all data has been received

vursen:
## Description

Ensures that gridConnector's request debouncer is cancelled if all data has been already received from the server while the debouncer was waiting. This scenario can take place for example when calling `$connector.reset()` on a non-empty grid.

Extracted from #7676 

## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/pom.xml
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/pom.xml
@@ -30,7 +30,7 @@
         <dependency>
             <groupId>com.vaadin</groupId>
             <artifactId>license-checker</artifactId>
-            <version>2.0.0-alpha1</version>
+            <version>2.0.0-alpha2</version>
             <scope>provided</scope>
         </dependency>
         <dependency>


vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/test/grid-connector.test.ts
```diff
@@ -73,7 +82,7 @@ describe('grid connector', () => {
 
     beforeEach(async () => {
       // Use a smaller page size for testing
-      grid.pageSize = 25;
+      grid.pageSize = 10;
```
vursen, line 85: Had to adjust these tests to ensure that at least two pages fit within the viewport. Previously, this wasn't the case, and the tests passed only because the requests created by the clear method weren't cancelled rather than because the rows were actually visible.

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridTooltipIT.java
```diff
@@ -92,6 +92,7 @@ public void columnTooltipOverridesGridTooltip() {
         scrollToElement(grid);
         // set grid tooltip
         clickElementWithJs("set-grid-tooltip-button");
+        flushScrolling(grid);
```
vursen, line 95: This test previously incorrectly passed without `flushScrolling(grid)` because of an extra `setViewport` request that wasn't cancelled.

sissbruecker (approved):

PR #7847, opened 2025-08-15
fix: determine whether a cell is locked on the fly

### LINKED ISSUE #7653: Spreadsheet with custom editor - editor not displayed when navigating to cell with up/down arrows
### Description

in 24.8.0+ Spreadsheet changed custom cell editor behavior to display the editor on various keyboard events, like typing text or tabbing into the cell.

However, while this works for most cases, I found that if you use the up/down arrow keys to navigate between cells, highlight one with a custom editor, and then start typing you get the default editor, not the custom editor.

### Expected outcome

navigating to a cell with a custom editor via the up or down arrow key should behave the same as navigating with the left or right arrow keys, but it does not.

see description.  Create a Spreadsheet implementation with a custom editor for a cell, such as a basic Select.  Set the Spreadsheet to not display editors all the time.
When running, use the arrow keys to navigate in and out of the cell with the editor.  When navigating in, start typing text and notice which editor displays, the default one or the custom one.

### Steps to reproduce

see description.  Create a Spreadsheet implementation with a custom editor for a cell, such as a basic Select.  Set the Spreadsheet to not display editors all the time.
When running, use the arrow keys to navigate in and out of the cell with the editor.  When navigating in, start typing text and notice which editor displays, the default one or the custom one.

### Environment

Vaadin version(s): 24.8.x
OS: N/A

### Browsers

Issue is not browser related

ugur-vaadin:
## Description

This PR removes the extra layer of state management regarding whether the current cell is locked and makes it so that it is retrieved on the fly. It is prone to timing issues and adds complexity without any real benefit.

Fixes #7653 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-integration-tests/src/test/java/com/vaadin/flow/component/spreadsheet/test/LockedCellsWithCustomEditorIT.java
DiegoCardoso, line 1: nit: probably it would be worth adding a test for each direction, just to ensure that it works.

ugur-vaadin: Added horizontal tests as well.

PR #7890, opened 2025-08-25
refactor: simplify connector set and clear

vursen:
## Description

The PR simplifies gridConnector's `set` and `clear` methods and makes them more reliable by avoiding the `delete` operator. Using `delete` creates "holes" in the array and causes deleted items to be skipped by JS iteration methods like `forEach` and `reduce`, which can be unexpected and hard to debug.

## Type of change

- [x] Refactor

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridTestPageIT.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridTestPageIT.java
@@ -487,8 +487,15 @@ public class GridTestPageIT extends AbstractComponentIT {
     public static Map<String, Map<String, ?>> getItems(WebDriver driver,
             WebElement element) {
         Object result = ((JavascriptExecutor) driver).executeScript(
-                "const items = arguments[0]._dataProviderController.rootCache.items;"
-                        + "return items.reduce((obj, item, i) => ({ ...obj, [i]: item }), {});",
+                """
+                        const { items } = arguments[0]._dataProviderController.rootCache;
+                        return items.reduce((obj, item, i) => {
+                            if (item) {
+                                obj[i] = item;
+                            }
+                            return obj;
+                        }, {});
+                        """,
                 element);
 
         return (Map<String, Map<String, ?>>) result;
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
@@ -315,17 +315,19 @@ window.Vaadin.Flow.gridConnector.initLazy = (grid) => {
    * Updates the cache for the given page for grid or tree-grid.
    *
    * @param page index of the page to update
-   * @returns an array of the updated items for the page, or undefined if no items were cached for the page
    */
   const updateGridCache = function (page) {
-    let gridCache = dataProviderController.rootCache;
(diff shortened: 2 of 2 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridTestPageIT.java
```diff
@@ -487,8 +487,15 @@ private String getInnerText(WebElement element) {
     public static Map<String, Map<String, ?>> getItems(WebDriver driver,
             WebElement element) {
         Object result = ((JavascriptExecutor) driver).executeScript(
-                "const items = arguments[0]._dataProviderController.rootCache.items;"
-                        + "return items.reduce((obj, item, i) => ({ ...obj, [i]: item }), {});",
+                """
+                        const { items } = arguments[0]._dataProviderController.rootCache;
+                        return items.reduce((obj, item, i) => {
```
vursen, line 492: The earlier version assumed that `gridConnector` removed items using the `delete` operator in `$connector.clear`. This created "holes" in the array, which caused `reduce` to skip over the deleted entries and not add them to the Java Map here. However, I think that this behavior is unsafe and can lead to unexpected results. I've changed both the test and implementation to clear items by setting them to `undefined`.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
```diff
@@ -477,11 +471,6 @@ window.Vaadin.Flow.gridConnector.initLazy = (grid) => {
         updateGridCache(page);
       }
     }
-    let cacheToClear = dataProviderController.rootCache;
-    const endIndex = index + updatedPageCount * grid.pageSize;
-    for (let itemIndex = index; itemIndex < endIndex; itemIndex++) {
-      delete cacheToClear.items[itemIndex];
```
vursen, line 483: This additional clearing was necessary because `updateGridCache` relied on `rootCache.setPage`, which only updated indexes smaller than the current size. As a result, if the size was reduced before calling `$connector.clear`, the exceeding items weren't removed. I've updated `updateGridCache` to work directly with the `items` array instead of using `setPage`, which allows this logic to be removed.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
```diff
@@ -364,19 +362,15 @@ window.Vaadin.Flow.gridConnector.initLazy = (grid) => {
...
-      if (updatedItems) {
-        itemsUpdated(updatedItems);
-      }
+      const page = firstPage + i;
+      cache[page] = items.slice(i * grid.pageSize, (i + 1) * grid.pageSize);
+      updateGridCache(page);
     }
 
+    itemsUpdated(items);
+    grid.$connector.doSelection(items.filter((item) => item.selected));
```
ugur-vaadin, line 370: Is there a reason for changing the order of updating items and updating selection?

ugur-vaadin (approved):

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
vursen: No, it was unintentional, thanks. Let me fix it.

vursen: Done.

PR #7892, opened 2025-08-25
test: refactor and cleanup DateTimePicker validation tests

ugur-vaadin:
## Description

This PR refactors DateTimePicker validation ITs so that the common logic is extracted and the tests are easier to read.

No related issue.

## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] Test

## Checklist

- [x] I have read the contribution guide: https://vaadin.com/docs/latest/contributing/overview
- [x] I have added a description following the guideline.
- [ ] The issue is created in the corresponding repository and I have referenced it.
- [ ] I have added tests to ensure my change is effective and works as intended.
- [x] New and existing tests are passing locally with my change.
- [x] I have performed self-review and corrected misspellings.

PR #7895, opened 2025-08-25
fix: only validate for incomplete input on change event

### LINKED ISSUE #7881: DateTimePicker does not allow to clear the value
### Description

I have used the DateTimePicker to enter a date and time.
Entering a value is possible.
When I like to delete the date and time it does not work when using TAB to change the selected field.
Not work means that the fields are still invalid in the web browser (marked red) and the empty value is not sent to the server at all. Also it seems that the server does not know that the value is invalid as I am still able to save my form (using a Binder).

Workaround:
1. Clear the date field value
2. Click outside with the mouse
3. Clear the time field value
4. Click outside with the mouse

### Expected outcome

When clearing the date and the time input field I expect that the field is shown in valid state and that an empty value is sent to the server.

The behaviour should be the same as in the documentation: https://vaadin.com/docs/latest/components/date-time-picker. There the field is valid when it is empty. Even more when I clear the date value and I go to time field with TAB the field is not marked red so I can clear the time as well before the intermediate state (date celared and time filld) is validated.

```
@Route("test")
public class Test extends Div {
  public Test() {
    add(new DateTimePicker("Test"));
  }
}
```

### Steps to reproduce

1. Have a DateTimePicker with a valid date entered
2. Put cursor in the date field
3. Clear the date field value
4. Press TAB
5. Clear the time field value
6. Press TAB -> Outcome is that the fiel is invalid but empty value should be valid

### Environment

Vaadin version(s): 24.8.6
OS: Windows

Last known version where this was working: Vaadin 24.6.12

### Browsers

Chrome

ugur-vaadin:
## Description

Description pending.

Depends on https://github.com/vaadin/flow-components/pull/7892

Fixes #7881 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow/src/main/java/com/vaadin/flow/component/datetimepicker/DateTimePicker.java
```diff
@@ -152,7 +153,7 @@ public class DateTimePicker
         }
 
         // Report error if only one of the pickers has a value
-        if (isInputIncomplete()) {
```
ugur-vaadin: This is for when the binder calls the default validator.

vursen:
Would it be possible to use the same approach that's used in DatePicker and TimePicker?

https://github.com/vaadin/flow-components/blob/main/vaadin-time-picker-flow-parent/vaadin-time-picker-flow/src/main/java/com/vaadin/flow/component/timepicker/TimePicker.java#L232-L241

https://github.com/vaadin/flow-components/blob/main/vaadin-time-picker-flow-parent/vaadin-time-picker-flow/src/main/java/com/vaadin/flow/component/timepicker/TimePicker.java#L381-L416

They handle such cases in `setModelValue` instead.

ugur-vaadin:
> Would it be possible to use the same approach that's used in DatePicker and TimePicker?
>
Updated the implementation to remove the flag and use `setModelValue`.

vursen:
I tested this locally and I now believe that the root issue is that DateTimePicker listens to `value-changed` instead of intended `change`. This also makes DateTimePicker validate too early when just switching between pickers. Adding `setSynchronizedEvent("change")` seems to resolve these issues, could you please check?

ugur-vaadin:
> I tested this locally and I now believe that the root issue is that DateTimePicker listens to `value-changed` instead of intended `change`.

Thanks, this is correct. I updated the PR and this greatly simplified the implementation.

vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow/src/main/java/com/vaadin/flow/component/datetimepicker/DateTimePicker.java
```diff
@@ -196,6 +196,7 @@ public class DateTimePicker
      */
     public DateTimePicker() {
         this((LocalDateTime) null);
+        setSynchronizedEvent("change");
```
vursen: This should be set in the [canonical constructor](https://github.com/vaadin/flow-components/blob/main/vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow/src/main/java/com/vaadin/flow/component/datetimepicker/DateTimePicker.java#L236) to make sure all constructor overloads are covered.

ugur-vaadin: Good catch, my mistake. Updated the PR.

vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow-integration-tests/src/test/java/com/vaadin/flow/component/datetimepicker/DateTimePickerIT.java
```diff
@@ -77,6 +78,7 @@ public void selectDateOrTimeSeparately() {
                 message.getText().contains("1987-04-10 14:42"));
 
         picker.setDate(null);
```
vursen, line 80: I wonder if this method should also trigger a `change` event on the DateTimePicker...

ugur-vaadin: My assumption was that the `change` event for `DateTimePicker` will be triggered automatically with the `change` event triggering on `DatePicker`.

vursen: Right, but why do we need to call sendKeys then?

ugur-vaadin: I think it's not incorrect since it is an incomplete value at that point and a change in a single picker should not change the value of the date time picker yet. So it makes sense to use `sendKeys` to blur. But setting a value in one of the inputs (like `setDate`) probably should not necessarily include changing the value of the date time picker.

vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow/src/main/java/com/vaadin/flow/component/datetimepicker/DateTimePicker.java
```diff
@@ -347,6 +348,11 @@ public DateTimePicker(LocalDateTime initialDateTime, Locale locale) {
         setLocale(locale);
     }
 
+    @Override
+    protected final void setSynchronizedEvent(String event) {
```
vursen: Should this method be final though? Why would we not allow it to be overridden? There is no similar thing in other components that use this method, so I'd suggest that we do it separately if it's necessary.

ugur-vaadin: Even though there is no explicit mention of it, I thought this was meant to be more of an internal method so I decided to keep Sonar a little happier. But in any case I agree that it's not really relevant, so I reverted this change.

PR #7896, opened 2025-08-26
refactor: simplify logic that prevents duplicate requests

vursen:
## Description

Checking `lastRequestedRange` to prevent a server fetch request is only relevant when there is already an ongoing request. In all other cases, it isn't needed, since a grid page request always reflects a viewport change or a cleared cache, and such requests should never be blocked.

Part of https://github.com/vaadin/flow/issues/21989

## Type of change

- [x] Refactor


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
```diff
@@ -526,14 +481,9 @@ window.Vaadin.Flow.gridConnector.initLazy = (grid) => {
     // prevent further unnecessary calls.
     if (Object.keys(pendingRequests).length === 0) {
       rootRequestDebouncer?.cancel();
+      lastRequestedRange = [-1, -1];
```
vursen, line 473: Clears `lastRequestedRange` after all fetch requests have been processed.

sissbruecker (approved):

PR #7897, opened 2025-08-26
test: update drilldown tests to work with new DOM structure

DiegoCardoso:
## Description

The DOM structure in Highcharts V12 differs in relation to how the drilldown controls are rendered. That will cause test failures in the ITs, so we need to update them.

Part of https://github.com/vaadin/web-components/issues/8825
Depends on https://github.com/vaadin/web-components/pull/10010


DiegoCardoso:
Closing in favor of #7944

PR #7904, opened 2025-08-26
feat!: add `Breadcrumbs` API to `Drilldown` in charts

DiegoCardoso:
## Description

Add the API related to configuring the breadcrumbs API added to replace the drillup button and mark the API related to `DrillUpButton` as deprecated.

> [!NOTE]
> The correspondent API in Highcharts can be found at: https://api.highcharts.com/highcharts/drilldown.breadcrumbs

Part of https://github.com/vaadin/web-components/issues/8825
Depends on https://github.com/vaadin/web-components/pull/10010
Depends on #7897
Depends on https://github.com/vaadin/flow-components/pull/7951

Fix https://github.com/vaadin/flow-components/issues/7923

## Type of change

- Feature


vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/model/Drilldown.java
```diff
@@ -132,7 +158,9 @@ public DrillUpButton getDrillUpButton() {
      * {@link Lang#setDrillUpText(String)}.
      *
      * @param drillUpButton
+     * @deprecated Since Use {@link #setBreadcrumbs(Breadcrumbs)} instead.
      */
+    @Deprecated(since = "25.0", forRemoval = true)
     public void setDrillUpButton(DrillUpButton drillUpButton) {
```
sissbruecker: How can the deprecated drill up button be used?

vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/model/Breadcrumbs.java
```diff
@@ -0,0 +1,262 @@
...
+    public void setButtonSpacing(Number buttonSpacing) {
+        this.buttonSpacing = buttonSpacing;
+    }
+
+    /**
+     * @return Theme options for the breadcrumb buttons.
+     */
+    public ButtonTheme getButtonTheme() {
+        if (buttonTheme == null) {
+            buttonTheme = new ButtonTheme();
```
sissbruecker, line 53: The default button has a width of `32`, which doesn't work very well here, as it results in the breadcrumbs showing on top of each other.

vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/model/Drilldown.java
DiegoCardoso: From the API, I had the impression that the API still works, but playing around with it, it seems that it doesn't work anymore. They have [a demo page](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/drilldown/drillupbutton/), and the customization is not applied in the running chart. I'll check their repo if there's any issue about it.

DiegoCardoso: I am starting to believe that this API is better off being removed. I wanted to confirm it, but what I've found so far points to the conclusion that, even though the property is still present in the docs, it's no longer used.

vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/model/Breadcrumbs.java
DiegoCardoso: This default width has [been added ages ago](https://github.com/vaadin/charts/commit/1f8bfe88ff8df39a2d1b2606f7425d410709adee#diff-b5b35a0e219aaeb062a331b374df8a0596eb6d9b9ae234af88c6849884c0d32f), and I don't know why. Judging by its tests (which have not been ported to the Flow version, as it seems), the default value is not supposed to be serialized, if I understood it correctly. But it is, so something is not working as intended, per the above link. Nevertheless, I cannot justify having this default value anymore, especially if we end up removing the `DrillUpButton` class.

vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/model/Drilldown.java
sissbruecker: Seems better to remove it if it doesn't work.

vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/model/Breadcrumbs.java
sissbruecker: I'm fine with removing the default, at least for breadcrumbs it seems to handle auto sizing very well.

DiegoCardoso: Removed.

vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/model/Drilldown.java
DiegoCardoso: I think so too. Removed.

vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/model/DrillUpButton.java
sissbruecker, line 1: `Lang` has a `drillUpText` that seems unused now. Can probably also be removed.

sissbruecker (approved):
Would be good to write something about the change in the upgrade guide right away, or at least add a placeholder to expand on later.

vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/model/DrillUpButton.java
DiegoCardoso: Done.

Issue #7666, opened 2025-07-07
[map] Rename borderless theme variant to no-border


web-padawan:
### Describe your motivation

The `borderless` variant will be replaced with `no-border` in https://github.com/vaadin/web-components/pull/9652

### Describe the solution you'd like

Update `MapVariant` to use `NO_BORDER` as e.g. in `CrudVariant` to match the web component.

PR #7912, opened 2025-08-28
refactor: deprecate BORDERLESS map variant, add NO_BORDER

web-padawan:
## Description

Fixes https://github.com/vaadin/flow-components/issues/7666

## Type of change

- Breaking change


vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/MapVariant.java
```diff
@@ -14,7 +14,7 @@
  * Set of theme variants applicable for the {@code vaadin-map} component.
  */
 public enum MapVariant implements ThemeVariant {
-    BORDERLESS("borderless");
+    NO_BORDER("no-border");
```
sissbruecker: If we want to break this, it would be good to deprecate this first?

web-padawan: Restored `BORDERLESS` and marked as deprecated.

PR #7916, opened 2025-08-28
chore: use Lumo theme in individual IT modules

sissbruecker:
## Description

Adds an app shell to each IT module, so that the respective apps use the Lumo theme.

## Type of change

- Internal

### DIFF
--- a/vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow-integration-tests/src/main/java/com/vaadin/flow/component/app/TestAppShell.java
+++ b/vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow-integration-tests/src/main/java/com/vaadin/flow/component/app/TestAppShell.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2000-2025 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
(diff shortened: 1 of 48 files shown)


vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow-integration-tests/src/main/java/com/vaadin/flow/component/app/TestAppShell.java
```diff
@@ -0,0 +1,23 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component.app;
+
+import com.vaadin.flow.component.dependency.CssImport;
+import com.vaadin.flow.component.page.AppShellConfigurator;
+
+@CssImport("@vaadin/vaadin-lumo-styles/lumo.css")
+public class TestAppShell implements AppShellConfigurator {
```
web-padawan: Do we need this in Lumo package as it has `LumoAppShell` already?

sissbruecker: Thanks, missed that. Also removed it from the renderer ITs which don't have the Lumo dependency.

Issue #7920, opened 2025-08-28
Spreadsheet allows pasting to locked cells if there are no cells in the row that has data


mukherjeesudebi:
### Description

If a spreadsheet has a locked sheet with a few rows unlocked or few cells unlocked for editing, then, it should not allow data to be pasted to locked cells. 

But, it does not work  correctly and allows pasting into a cell which is in a row that is locked and does not have any other cell in that row which has data. Once a locked cell has data by pasting, it does not allow pasting again to locked cells. 

Looks like the problem is SpreadsheetHandlerImpl.onPaste() only checks locked status for cells in rows already in the model. 
Similar to the logic in Spreadsheet.isCellLocked() it needs to check if the sheet is locked, if the row is null/undefined, not just if the cell is locked. isCellLocked() should probably take a CellReference instead of a cell, and all callers of it should be adjusted to call it regardless of row existance.

### Expected outcome

If a cell is locked, it should not allow to paste data. 

### Minimal reproducible example

```
@Route("locked-spreadsheet")
public class LockedSpreadsheet extends Div {

	private Button customComponent;
	private TextField customEditor;

	public LockedSpreadsheet() throws IOException {
		InputStream stream = getClass().getResourceAsStream("/testsheets/test.xlsx");
		Spreadsheet spreadsheet = new Spreadsheet(stream);	
		spreadsheet.setHeight("500px");
		add(spreadsheet);
	}
}
```

### Steps to reproduce

1. Add the code to a vaadin 24.8.7 sample
2. Use the attached excel [test.xlsx](https://github.com/user-attachments/files/22023414/test.xlsx) . The sheet is protected and rows 1-5 are unlocked for editing but row 6 onwards they are locked. 
3. Paste some text in row 6 or anywhere below row 6. It gets pasted.
4. Now try pasting again in the same row, it does not allow anymore. 

### Environment

Vaadin version(s):  Tested in 24.8.7

PR #7927, opened 2025-08-29
fix: prevent paste to locked cells

sissbruecker:
## Description

Currently, pasting into a cell for which there is no row created in the model yet works even if the sheet is protected (locked). Another issue is that Spreadsheet considers cells locked if they are not created yet, but the default row or column styles for that cell explicitly unlock cells in that row or column.

However the way this works in Excel is that locking a sheet locks all cells by default, unless either cell, row or column have custom styles for unlocking.

This updates cell locked checks to:
- Always work, regardless whether a row or cell has been created or not
- Consider all cells locked if the sheet is locked
- Consider cells unlocked if:
  - The cell has a custom style that unlocks it
  - The cell is in a row that has a custom style that unlocks it
  - The cell is in a column that has a custom style that unlocks it

Fixes https://github.com/vaadin/flow-components/issues/7920

## Type of change

- Bugfix

### DIFF
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/Spreadsheet.java
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/Spreadsheet.java
@@ -3926,10 +3926,8 @@ public class Spreadsheet extends Component
         Cell cell = row != null ? row.getCell(cellAddress.getColumn()) : null;
 
         // If there is a cell with a custom cell style, return its locked state
-        if (cell != null) {
-            if (cell.getCellStyle().getIndex() != 0) {
-                return cell.getCellStyle().getLocked();
-            }
+        if (cell != null && cell.getCellStyle().getIndex() != 0) {
+            return cell.getCellStyle().getLocked();
         }
 
         // Otherwise inherit locked state from row or column styles


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-integration-tests/src/test/java/com/vaadin/flow/component/spreadsheet/test/LockIT.java
sissbruecker, line 1: The `LockCells` fixture was broken after the fix. It first locked the sheet, and then checked whether selected cells were locked or not. Without the fix it returned unlocked, which was wrong, and then proceeded to add a custom style to lock the cell. I think that was because the old `isCellLocked` implementation relied on the locked columns / rows state, which is not recalculated without calling `reloadActiveSheetColumnRowStyles`.

The new implementation doesn't use this state, so now just locking the sheet correctly reports that the cell is already locked, which caused the fixture to then unlock the cell.

Changed it so that:
- it first locks the sheet, checks that cells are locked
- toggles lock for a range (=unlock)
- checks that unlocked cells can be modified

vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/Spreadsheet.java
```diff
@@ -3908,6 +3907,41 @@ && getLockedRowIndexes()
...
+        if (!isActiveSheetProtected()) {
+            return false;
+        }
+
+        Sheet sheet = getActiveSheet();
+        Row row = sheet.getRow(cellAddress.getRow());
+        Cell cell = row != null ? row.getCell(cellAddress.getColumn()) : null;
+
+        // If there is a cell with a custom cell style, return its locked state
+        if (cell != null) {
```
ugur-vaadin: Nit: We can merge the ifs.

ugur-vaadin:
There is one thing I noticed. When I type something into a cell before loading LockSheet fixture, I can still modify that cell after locking the sheet.

Also when I copy and paste multiple cells multiple times in a quick fashion, sometimes the cells are not actually pasted in the correct place and the focus returns to a previous cell. But this issue seems unrelated.

sissbruecker:
> There is one thing I noticed. When I type something into a cell before loading LockSheet fixture, I can still modify that cell after locking the sheet.
> 
> Also when I copy and paste multiple cells multiple times in a quick fashion, sometimes the cells are not actually pasted in the correct place and the focus returns to a previous cell. But this issue seems unrelated.

I can double check both issues on main. The first one might be `setActiveSheetProtected` not updating some client-side stuff, or there is some flawed assumption in the new logic with how locking works with existing cells.

WoozyG:
> Also when I copy and paste multiple cells multiple times in a quick fashion, sometimes the cells are not actually pasted in the correct place and the focus returns to a previous cell. But this issue seems unrelated.

This may relate to other behavior we are trying to pinpoint so we can potentially file another bug related to rapid cell focus and selection changes.  It does appear there is some ambiguous/asynchronous event ordering or stale state going on somewhere around cell selection/navigation, with or without pasting, although pasting does give excellent obvious feedback for debugging :)

If we can nail it down to a repeatable test case I'll file it and Vaadin can decide if it relates to this or not.

sissbruecker:
> There is one thing I noticed. When I type something into a cell before loading LockSheet fixture, I can still modify that cell after locking the sheet.

@ugur-vaadin Also seems to happen on `main`. It looks like the client-side cell is not aware that it's protected now and the only way to fix it was to call `refreshCells` with the specific cell reference. I would say that's a separate issue, since this one was not about locking the sheet programmatically.

vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/Spreadsheet.java
sissbruecker: Thanks, done

PR #7932, opened 2025-08-29
feat!: add getter and setter for `buttonSpacing` in `ContextButton`; remove `widthAdjust` from `Title`

### LINKED ISSUE #7924: [Charts] Add `buttonSpacing` property in `ContextButton`
### Describe your motivation

In Highcharts 12.0.0, the `Title#widthAdjust` API used for changing the width of the chart's title in case it overlaps with the exporting menu button was removed, and the `exporting.buttons.contextButton.buttonSpacing` property should be used instead.

- Add the new API to the `ContextButton` class: https://api.highcharts.com/highcharts/exporting.buttons.contextButton.buttonSpacing
- Remove the getter and setter to the `widthAdjust` property in the `Title` class, so it will break the build in apps using the API. Since this API now does nothing, it's better to make developers aware of it as soon as possible while migrating to the new version.

DiegoCardoso:
## Description

Add an API for the `buttonSpacing` property to the `ContextButton` class that maps the `exporting.buttons.contextButton.buttonSpacing` in the configuration object API.

Fixes #7924

## Type of change

- Feature


vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/model/ContextButton.java
```diff
@@ -17,6 +17,7 @@
 public class ContextButton extends AbstractConfigurationObject {
 
     private HorizontalAlign align;
+    private int buttonSpacing;
```
sissbruecker: Should this be `Number`? In general it looks like it should be a nullable type to preserve whatever the default is in Highcharts. At least that's how every other property seems to work.

DiegoCardoso: Good catch. It should be `Number` indeed. Changed.

sissbruecker (approved):

PR #7933, opened 2025-08-29
feat: add `timezone` API to Time; deprecate for `useUTC` and `timezoneOffset`

### LINKED ISSUE #7925: [Charts] Mark `useUTC` property in `Time` as deprecated
### Describe your motivation

Since Highcharts 12.0.0, the `useUTC` property in the `Time` class has been marked as deprecated, and it should also be marked as such in the `Time` class.

DiegoCardoso:
## Description

Add a new API for `timezone` property in the `Time` class. This API is not new, but before it required importing `moment.js` to work; since `v11.2.0`, it uses the browser API.

Also, deprecates the `useUTC` and `timezoneOffset` API that is marked as such in the Highcharts API page.

Fixes #7925

### DIFF
--- a/vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/model/Axis.java
+++ b/vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/model/Axis.java
@@ -136,10 +136,25 @@ public abstract class Axis extends AbstractConfigurationObject {
     }
 
     /**
-     * The minimum and maximum value of the axis as Date.
+     * The minimum and maximum value of the axis as Instant.
+     *
+     * @param min
+     *            Minimum value as Instant.
+     * @param max
+     *            Maximum value as Instant.
      *
      * @see #setExtremes(Number, Number)
      */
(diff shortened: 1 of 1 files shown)


ugur-vaadin:
Should we provide alternatives when deprecating this? From what I understand from the API docs, the alternative is using the `timezone` property on `time`:
"Setting useUTC to true is equivalent to setting time.timezone to "UTC". Setting useUTC to false is equivalent to setting time.timezone to undefined."
But we don't seem to have that property on `Time`.

DiegoCardoso:
> Should we provide alternatives when deprecating this? From what I understand from the API docs, the alternative is using the `timezone` property on `time`:
> "Setting useUTC to true is equivalent to setting time.timezone to "UTC". Setting useUTC to false is equivalent to setting time.timezone to undefined."
> But we don't seem to have that property on `Time`.

It makes sense. Checking a bit closer, the reason why `timezone` hasn't been added before is that it required importing `moment.js` to work. But since according [to the docs](https://api.highcharts.com/highcharts/time.timezone), since v11.2.0, it's not required anymore. Added the property and updated the annotation.

ugur-vaadin (approved):
Should we turn this into a feature PR since there is new API now?

PR #7943, opened 2025-09-02
feat: bump Highcharts version in SVG Generator

### LINKED ISSUE #7942: [Charts] Upgrade Highcharts version in the SVG Generator module
### Describe your motivation

The version in the SVG Generator module should match the one used in the web component (and, consequentially in the flow counterpart).
Based on some initial investigation, the version bump should address:
- Upgrade dependencies that can be upgrades (Highcharts, JSDOM, ...)
- Fix build issues related to how the modules have changes since 9.2.2
  - In 9.2.2, the CommonJS modules returns a function where some context object could be provided. That's not the case anymore in V12
- Fix box calculation that leads to wrong placement of text elements
- Update reference SVG files with the new output coming from the generator

-

_No response_

### Additional context

_No response_

DiegoCardoso:
## Description
Upgrades the Highcharts version to match with the web component (12.2.0).

This change also:
- upgrades JSDOM version to its latest major (26.1.0)
- updates how Highcharts modules are imported
- adds workaround to fix a limitation in `getComputedStyles` in JSDOM with relative units
- updates reference SVG files to use the output coming from the updated library

Fixes #7942


vaadin-charts-flow-parent/vaadin-charts-flow-svg-generator/src/test/resources/column-without-title.svg
sissbruecker, line 1: The axis labels are noticably larger now. In a chart rendered in the browser, the axis labels and legends labels seem to have the same size. Here it looks somewhat odd because the axis labels are more prominent.

Maybe the function for the font metrics can be parameterized to pass the base font size, as the old implementation was using 12px as a basis.

vaadin-charts-flow-parent/vaadin-charts-flow-svg-generator/src/test/resources/custom-height.svg
sissbruecker, line 1: In CI this one fails due to a different number in the last decimal place in some random path. Locally I don't have any issues with the test.

vaadin-charts-flow-parent/vaadin-charts-flow-svg-generator/src/test/resources/column-without-title.svg
DiegoCardoso: I agree that they seem a bit larger than expected now. Trying the same configuration in plain Highcharts gives me a somewhat close result:
Screen Shot 2025-09-03 at 15 42 15

The texts are rendered with font-size defined as `0.8em`, which results in `12.8px`. The same seems to be true when I open the new reference file in the browser:
https://raw.githubusercontent.com/vaadin/flow-components/15c3c1e078cef1a01f2951519511fee2b4a6e7bf/vaadin-charts-flow-parent/vaadin-charts-flow-svg-generator/src/test/resources/column-without-title.svg

My understanding is that the font metrics are used to derive the size of the text boxes to allow placing them correctly, not to define the size of the font.

vaadin-charts-flow-parent/vaadin-charts-flow-svg-generator/src/test/resources/custom-height.svg
DiegoCardoso: This has been fixed by this change from @ugur-vaadin https://github.com/vaadin/flow-components/pull/7943/commits/15c3c1e078cef1a01f2951519511fee2b4a6e7bf

vaadin-charts-flow-parent/vaadin-charts-flow-svg-generator/src/test/resources/column-without-title.svg
sissbruecker: OK, maybe I was mislead by the Lumo theme in web components dev pages where labels are quite a bit smaller. Or maybe my eyes are getting old. Anyway, seems fine then.

PR #7958, opened 2025-09-05
chore!: Use jackson for element properties

caalador:
Use jackson for element property
to make it possible to drop
Element#setPropertyJson(String name, JsonValue value)


sissbruecker:
Closing in favor of https://github.com/vaadin/flow-components/pull/7959

PR #7970, opened 2025-09-08
feat!: add `Breadcrumbs` to `PlotOptionsTreemap`

### LINKED ISSUE #7969: [Charts] Add `Breadcrumbs` API to `PlotOptionsTreemap`
### Describe your motivation

The Treemap series has options to customize the breadcrumbs feature that shares the same configuration as the breadcrumbs in the `Drilldown` object. 

Like in the `Drilldown` object, the `Treemap` series used to have a drill-up button as part of its configuration (it's called `traverseUpButton` in the `Treemap`). There's no API currently in the `PlotOptionsTreemap` for the deprecated API, so only the new API needs to be added.

As part of the changes, the `allowDrillToNode` option should be deprecated in favor of a new option called `allowTraversingTree`.

Depends on #7923 

Use the new `Breadcrumbs` class to add a new property in `PlotOptionsTreemap`.

_No response_

### Additional context

_No response_

DiegoCardoso:
## Description

Add the `Breadcrumbs` configuration to the `PlotOptionsTreemap` that is used to configure breadcrumbs when traversing the treemap is enabled. This option has been added to replace another in Highcharts called `traverseUpButton`, which is not present in the `Charts` API.

This change also adds a new property called `allowTraversingTree`, which replaces `allowDrillToNode`, marking the latter as deprecated.

Depends on #7904
Fixes #7969 

## Type of change

- Feature


vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/model/PlotOptionsTreemap.java
```diff
@@ -205,6 +230,23 @@ public void setBorderWidth(Number borderWidth) {
         this.borderWidth = borderWidth;
     }
 
+    /**
+     * {@see #setBreadcrumbs(Breadcrumbs)}
+     */
+    public Breadcrumbs getBreadcrumbs() {
+        return breadcrumbs;
```
ugur-vaadin, line 240: Should this also initialize Breadcrumbs if it is null similar to [Drilldown](https://github.com/vaadin/flow-components/blob/main/vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/model/Drilldown.java#L124-L126)?

DiegoCardoso: Done.

ugur-vaadin (approved):

PR #7971, opened 2025-09-08
refactor: remove custom I18N merging logic

sissbruecker:
## Description

Removes custom logic for merging I18N values from Flow components with web components defaults. That should now be covered by all web components having support for setting partial I18N objects.

## Type of change

- Refactoring


vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/test/java/com/vaadin/flow/component/upload/tests/UploadI18nIT.java
```diff
@@ -52,31 +42,8 @@ public void testFullI18nShouldAffectLabels() {
...
 
-    /**
-     * Verifies that every single translation provided by the UploadI18N
-     * instance is set in the web component.
-     *
-     * Testing internals here, in favour of setting up the web component with
-     * files/event handlers for every possible state.
-     */
     @Test
-    public void testFullI18nShouldOverrideCompleteConfigurationInWebComponentProperty() {
```
sissbruecker, line 63: These tests were checking that the nested I18N structure was properly merged by the `executeJs` call that has now been removed. The equivalent would now be to check the internal `__effectiveI18n` property.

However, since the merging logic is part of the web component, I think it's better to remove this. The web components could use some better test coverage for partial I18n, but that's a separate issue.

ugur-vaadin:
Should this also change similar logic in `AppLayout` and `MultiSelectComboBox`?

sissbruecker:
> Should this also change similar logic in `AppLayout` and `MultiSelectComboBox`?

Thanks, I missed those. Turns out the partial I18n mechanism was missing for MSCB, so I created a PR for this first: https://github.com/vaadin/web-components/pull/10147

The tests for those components will also need to be changed as they make assertions on the `i18n` property. I'll have to think about whether ITs are still feasible or if they should be more basic unit tests instead.

Issue #7988, opened 2025-09-10
[Charts] Remove deprecated methods based on `Date` class


DiegoCardoso:
### Describe your motivation

There are many public API that have been deprecated since the FW8 version, which use `Date` as a parameter type. They have been replaced by similar methods using `Instant` instead. It's fair to assume that they can be safely removed from the public API in the next major version.

It's not possible, however, to remove all usages of the `Date` class in the `vaadin-charts-flow` module, as the `Axis#setExtremes` methods still use them without a replacement using `Instant`. The replacement API could be introduced now, marking the old methods as deprecated to be later removed in the following major (v26).

### Describe the solution you'd like

-

PR #7990, opened 2025-09-11
feat: add `Axis#setExtremes` overloads with `Instant`

DiegoCardoso:
## Description

Add methods overload for `setExtremes` in `Axis` where `Date` is being used with `Instant` params. That is to align with other APIs that have been introduced before using `Instant`. Mark the current methods as deprecated to be removed later.

Part of #7988
## Type of change

- Feature

### DIFF
--- a/scripts/wtr.js
+++ b/scripts/wtr.js
@@ -53,7 +53,7 @@ function runTests() {
       });
 
       // Install dependencies required to run the web-test-runner tests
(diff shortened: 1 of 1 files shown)


DiegoCardoso:
> Does this need to be tagged as a breaking change (`feat!`)?

I wasn't sure if deprecation is eligible to make it a breaking change. If it's not, I can remove it.

sissbruecker:
I would say no, when something gets deprecated you can still use it and it doesn't break anything.

PR #7991, opened 2025-09-11
refactor!: remove deprecated methods using `Date`

### LINKED ISSUE #7988: [Charts] Remove deprecated methods based on `Date` class

DiegoCardoso:
## Description

Such methods have been deprecated since the version for FW8, which makes them safe to be removed in the next major.

Depends on #7990
Fixes #7988

## Type of change
- Refactor

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -478,6 +478,7 @@
                     <version>${flow.version}</version>
                     <configuration>
                         <pnpmEnable>${vaadin.pnpm.enable}</pnpmEnable>
+                        <frontendDirectory>./frontend</frontendDirectory>
                     </configuration>
                     <executions>
                         <execution>
@@ -617,6 +618,9 @@
                         <plugin>
                             <groupId>com.vaadin</groupId>
                             <artifactId>flow-maven-plugin</artifactId>
+                            <configuration>
+                                <frontendDirectory>./frontend</frontendDirectory>
+                            </configuration>
                             <executions>
                                 <execution>
                                     <goals>
--- a/vaadin-accordion-flow-parent/vaadin-accordion-flow-integration-tests/pom.xml
+++ b/vaadin-accordion-flow-parent/vaadin-accordion-flow-integration-tests/pom.xml
@@ -156,6 +156,9 @@
                     <plugin>
                         <groupId>com.vaadin</groupId>
                         <artifactId>flow-maven-plugin</artifactId>
+                        <configuration>
+                            <frontendDirectory>./frontend</frontendDirectory>
+                        </configuration>
                     </plugin>
                 </plugins>
             </build>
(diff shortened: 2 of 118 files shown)


vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/model/ColumnOptions.java
```diff
@@ -567,5 +567,5 @@ public abstract class ColumnOptions extends AbstractPlotOptions {
     /**
      * @see #setPointStart(Number)
      */
-    public abstract void setPointStart(Date date);
+    public abstract void setPointStart(Instant instant);
```
DiegoCardoso, line 570: This is something I am a bit unsure about. All the classes (`PlotOptionsBar`, `PlotOptionsColumn`, `PlotOptionsColumrange`, and `PlotOptionsWaterfall`) that extend this one have deprecated the `setPointStart(Date)` method, but for some reason, it hasn't been here (not even the `setPointStart(Instant)` method has been added). The safest option would be to introduce the `setPointStart(Instant)` method here, marking the other one as deprecated, and undo the removal of the classes extending this one.

vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/model/Axis.java
```diff
@@ -232,8 +208,8 @@ public void setExtremes(Instant minimum, Instant maximum, boolean redraw,
     @Deprecated(since = "25.0", forRemoval = true)
     public void setExtremes(Date minimum, Date maximum, boolean redraw,
             boolean animate) {
-        setMin(minimum);
-        setMax(maximum);
+        setMin(minimum.toInstant());
+        setMax(maximum.toInstant());
```
DiegoCardoso, line 233: Since the `setMin(Date)`/`setMax(Date)` are being removed above, using the `toInstant` method to call the `Instant` setters.

vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/model/ColumnOptions.java
ugur-vaadin: In my opinion it should be fine to deprecate it in v24.9 and remove it in v25.

bwajtr: I totally agree with Ugur

DiegoCardoso: Thanks for your feedback @bwajtr

DiegoCardoso: Created the PR deprecating the method in 24.9: https://github.com/vaadin/flow-components/pull/8003

Issue #7985, opened 2025-09-10
Grid performance has regressed in a few metrics


vursen:
While the TreeGrid refactoring improved over half of the performance metrics in `25.0-SNAPSHOT`, some metrics have regressed relative to `24.9.0-beta3`:

| Metric                                     | Change       |
|:-----------------------------------------------|--------------|
| mixed-selectallverticalscrollframetime-chrome | -74%         |
| mixed-verticalscrollframetime-chrome          | -24%         |

vursen:
| Metric                                     | Change       |
|:-----------------------------------------------|--------------|
| mixed-selectallverticalscrollframetime-chrome | -7%          |
| mixed-verticalscrollframetime-chrome          | -2%         |

PR #7999, opened 2025-09-11
refactor: batch grid row updates

vursen:
## Description

The PR introduces a mechanism for batching row updates in gridConnector, which improves scrolling performance in heavy grids with many selected rows:

| Before | After |
|--------|------|
| Screenshot 2025-09-12 at 13 49 30 | Screenshot 2025-09-12 at 13 25 12 |

Part of #7985 

## Type of change

- [x] Refactor

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -478,6 +478,7 @@
                     <version>${flow.version}</version>
                     <configuration>
                         <pnpmEnable>${vaadin.pnpm.enable}</pnpmEnable>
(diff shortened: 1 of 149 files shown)


vursen:
> Maybe this is something that should be in the web component itself somehow

I agree. It would nice if the web component had a built-in batching mechanism, as it would benefit from that as well. It's indeed a broader topic and it should possibly be considered along with enabling async Lit updates.

PR #8024, opened 2025-09-16
refactor!: use Jackson 3 for chart config serialization

sissbruecker:
## Description

Updates `ChartSerialization` to use Jackson 3.

Since Flow has not been updated to use Jackson 3 yet, this makes Charts temporarily use Jackson 2 and 3 side-by-side:
- Jackson 3 is added as a dependency to charts. Since it uses different Maven coordinates and package names there are no conflicts with Jackson 2. The dependency should be removed again after the Flow change.
- `ChartSerialization` and all the custom config serialization logic is updated to use Jackson 3 API.
- Other usages that use `executeJs` or map event parameters still use Jackson 2 to be compatible with the current Flow version.
- After the Flow change, these other usages will need to be updated as well, but that should require only a few changes.

## Type of change

- Breaking change

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridMultiSelectionModel.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridMultiSelectionModel.java
@@ -101,7 +101,7 @@ public abstract class AbstractGridMultiSelectionModel<T>
     protected void remove() {
         super.remove();
         deselectAll();
-        if (selectionColumn.getElement().getNode().isAttached()) {
+        if (selectionColumn.getParent().map(getGrid()::equals).orElse(false)) {
             getGrid().getElement().removeChild(selectionColumn.getElement());
         }
     }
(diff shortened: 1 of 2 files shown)


vaadin-charts-flow-parent/vaadin-charts-flow/src/test/java/com/vaadin/flow/component/charts/ConfigurationJSONSerializationTest.java
sissbruecker, line 1: Updated several assertions as the properties are now ordered alphabetically.

vaadin-charts-flow-parent/vaadin-charts-flow/src/test/java/com/vaadin/flow/component/charts/ChartsSerializableTest.java
```diff
@@ -17,6 +17,7 @@ public class ChartsSerializableTest extends ClassesSerializableTest {
     protected Stream<String> getExcludedPatterns() {
         return Stream.concat(super.getExcludedPatterns(), Stream.of(
                 ".*\\Serializer(Modifier)?$",
-                "com\\.vaadin\\.flow\\.component\\.charts\\.model\\.serializers\\.BeanSerializationDelegate"));
+                "com\\.vaadin\\.flow\\.component\\.charts\\.model\\.serializers\\.BeanSerializationDelegate",
+                "com\\.vaadin\\.flow\\.component\\.charts\\.model\\.serializers\\.BeanSerializerDelegator"));
```
sissbruecker, line 21: The `BeanSerializerBase` base class of `BeanSerializerDelegator` is not serializable anymore. Seems safe to ignore, as the object mapper instances for charts are not stored in the UI.

Issue #7754, opened 2025-07-28
Load Lumo icons frontend dependency from LumoIcon class


rolfsmeds:
### Describe your motivation

Currently, the `Lumo` class in Flow [loads the jsmodule for Lumo icons](https://github.com/vaadin/flow-components/blob/2e680d7325f3e9e183d42a08a60834151ea8919c/vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/Lumo.java#L32).

As the `Lumo` class will be deprecated, and Lumo will be loaded as a stylesheet instead, Lumo icons will need to be loaded some other way.

### Describe the solution you'd like

Load the Lumo icons jsmodule from the [`LumoIcons`](https://github.com/vaadin/flow-components/blob/main/vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/LumoIcon.java) class instead.

### Additional context

Instead of loading the icons as a jsmodule, a different format could be considered.

PR #8026, opened 2025-09-16
refactor: move Lumo iconset annotation to LumoIcon enum

web-padawan:
## Description

Fixes https://github.com/vaadin/flow-components/issues/7754

Note: I'm not sure if this has a risk of facing the same problem as in https://github.com/vaadin/flow-components/pull/3383#issuecomment-1165251305?
This needs to be tested in production mode, but I'm getting problems when running merged ITs.

## Type of change

- Refactor

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -95,7 +95,7 @@
         <cvdlName/>
         <jakarta.servlet.api.version>6.1.0</jakarta.servlet.api.version>
         <jakarta.web.api.version>11.0.0</jakarta.web.api.version>
-        <spring-data-commons.version>4.0.0-M5</spring-data-commons.version>
+        <spring-data-commons.version>4.0.0-M6</spring-data-commons.version>
 
         <!-- spreadsheet -->
         <poi.version>5.4.1</poi.version>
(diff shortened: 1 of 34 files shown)


vaadin-lumo-theme-flow-parent/vaadin-lumo-theme-flow/src/main/java/com/vaadin/flow/theme/lumo/LumoIcon.java
sissbruecker, line 1: Seems to work in general when doing a production build in a starter:
```
[INFO] Checking if a production mode bundle build is needed
[INFO] Frontend import @vaadin/vaadin-lumo-styles/vaadin-iconset.js is missing from the bundle
[INFO] A production mode bundle build is needed
```

This is assuming that you would use `LumoIcon.ARROW_DOWN.create()` or `LumoIcon.ARROW_DOWN.name()`, which I guess is something the Java compiler can not optimize.

The fact that Flow builds a new prod bundle when using something from Lumo is questionable, I've raised a discussion in Slack.

sissbruecker (approved):

PR #8034, opened 2025-09-17
refactor: use Jackson 3 for map serialization

sissbruecker:
## Description

TBD

## Type of change

- Refactoring

### DIFF
--- a/vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/MapBase.java
+++ b/vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/MapBase.java
@@ -13,6 +13,7 @@ import java.util.LinkedHashSet;
 import java.util.Objects;
 import java.util.Set;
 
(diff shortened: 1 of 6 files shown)


vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/MapBase.java
```diff
@@ -45,12 +47,11 @@
 public abstract class MapBase extends Component
         implements HasSize, HasStyle, HasThemeVariant<MapVariant> {
     private final Configuration configuration;
-    private final MapSerializer serializer;
+    private transient MapSerializer serializer;
```
sissbruecker, line 50: At least `ValueSerializer` and the `StdSerializer` subclass are not serializable anymore with Jackson 3. Since `MapSerializer` has instance-specific state we can not use a single global instance for it. So this makes it transient to exclude it from serialization and creates it on demand in the getter in case it is missing after serialization / deserialization.

PR #8035, opened 2025-09-18
fix: cleanup stream registrations when Map is detached

sissbruecker:
## Description

`Map` uses custom Jackson serializers that automatically register `StreamResource` and `DownloadHandler` instances in the current UI when those are encountered during serialization. However these registrations are never removed when the Map instance is removed from the UI, creating a memory leak.

This changes `MapSerializer` to register a detach listener on the component, which is then used to unregister any stream resources that were created during serialization.

## Type of change

- Bugfix


sissbruecker:
> Not related to this PR, but do we need to deprecate a private class like with [StreamResourceSerializer](https://github.com/vaadin/flow-components/pull/8035/files#diff-845dc3a09e1227032cf9617b14023f17cc3ea0e20c9c8ab78ccaab944f890e10L73)?

No that was unnecessary in this case since it's not public API. Shouldn't be an issue though, or do you think it could cause problems?

ugur-vaadin:
> No that was unnecessary in this case since it's not public API. Shouldn't be an issue though, or do you think it could cause problems?

Not really, I was wondering if there was a reason to add that in the first place.

Issue #3199, opened 2022-05-17
Add cluster feature in the Map Component


jcgueriaud1:
### Describe your motivation

When the number of marker is too high, I want to group them for readability and performance.

### Describe the solution you'd like

There are 3 different openlayers packages:
https://openlayers.org/en/latest/examples/index.html?q=cluster

I would like to be able to configure the component to group the markers.

yuriy-fix:
We should check if OpenLayers does contain the API for achieving that or should we come up with our own. We also should investigate what level of customisation should be available within the API provided, i.e. should the clustering be customisable from the Java side?

PR #8037, opened 2025-09-19
feat: add cluster support to Map

sissbruecker:
## Description

Adds cluster support to the `Map` component.

- Clustering can be enabled on the map's feature layer using `setClusteringEnabled`.
- The layer will then automatically switch to the appropriate cluster source and move all features to the new source.
- Cluster options (distance, min distance) can be configured on the feature layer, or it's source.
- Currently clustering is only supported for point-based features (markers). Supporting polygons or other geometry will require further work in configuring OpenLayers.
- Feature layers have a default style for rendering clusters, currently using an icon.
- A custom style can be configured using `setClusterStyle` on the feature layer. Styles need to be configured on the layer, as there is no entity for clusters themselves, which are generated dynamically on the client.
- While OpenLayers always passes a cluster to the style function, the specific implementation for feature layer will prefer rendering clusters with just a single feature using the feature's own style.

Some other common functionality like clicking on a cluster to zoom in on all the features within it can be done as a follow-up.

References:
- [OL Cluster API](https://openlayers.org/en/latest/apidoc/module-ol_source_Cluster-Cluster.html)
- [OL Cluster example](https://openlayers.org/en/latest/apidoc/module-ol_source_Cluster-Cluster.html)

Fixes https://github.com/vaadin/flow-components/issues/3199

## Type of change

- Feature


vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/layer/FeatureLayer.java
```diff
@@ -24,11 +34,28 @@
...
 
+    private boolean clusteringEnabled = false;
+    private int clusterDistance = 50;
+    private int clusterMinDistance = 50;
+    private Style clusterStyle;
+
     public FeatureLayer() {
-        this.setSource(new VectorSource());
+        setSource(new VectorSource());
+        setClusterStyle(createDefaultClusterStyle());
```
DiegoCardoso, line 53: Would it make sense to create the default cluster style only if cluster is enabled and `clusterStyle` is `null`?

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/layer/FeatureLayer.java
```diff
@@ -57,12 +84,21 @@ public List<Feature> getFeatures() {
...
      *            the feature to be added
+     * @throws IllegalArgumentException
+     *             if clustering is enabled and the feature is not a
+     *             {@link PointBasedFeature}
      */
     public void addFeature(Feature feature) {
+        if (clusteringEnabled && !(feature instanceof PointBasedFeature)) {
+            throw new IllegalArgumentException(
+                    "Only PointBasedFeature types are supported when clustering is enabled");
+        }
```
DiegoCardoso: Isn't it enough to rely on the check inside `ClusterSource#addFeature`?

sissbruecker: Makes sense, removed.

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/layer/FeatureLayer.java
sissbruecker: I figured it keeps the implementation a bit simpler. In most cases there will only be one feature layer, so this also doesn't increase the payload that much.

Another benefit is that this will work instead of throwing a null pointer:
```java
layer.getClusterStyle.setImage(customIcon);
layer.setClusteringEnabled(true);
```

There would be some options to make that work as well without synchronizing the cluster style to the client, but I think it's not neccesary to optimize this right now.

vaadin-map-flow-parent/vaadin-map-flow/src/test/java/com/vaadin/flow/component/map/configuration/layer/FeatureLayerTest.java
```diff
@@ -58,6 +66,130 @@ public void removeFeature() {
...
+        featureLayer.setClusterDistance(50);
+        Assert.assertEquals(50, featureLayer.getClusterDistance());
+
+        featureLayer.setClusteringEnabled(true);
+        ClusterSource clusterSource = (ClusterSource) featureLayer.getSource();
+        Assert.assertEquals(50, clusterSource.getDistance());
+
+        featureLayer.setClusterDistance(100);
+        Assert.assertEquals(100, clusterSource.getDistance());
+    }
```
DiegoCardoso, line 133: It makes more sense to set it to a different value from the default one, IMO:
```suggestion
    public void setClusterDistance() {
        featureLayer.setClusterDistance(10);
        Assert.assertEquals(10, featureLayer.getClusterDistance());

        featureLayer.setClusteringEnabled(true);
        ClusterSource clusterSource = (ClusterSource) featureLayer.getSource();
        Assert.assertEquals(10, clusterSource.getDistance());

        featureLayer.setClusterDistance(100);
        Assert.assertEquals(100, clusterSource.getDistance());
    }
```

DiegoCardoso (approved):
It looks good to me, just one last comment, but I'll mark it as approved.

PR #8041, opened 2025-09-22
feat: add Aura Flow component package with empty class

web-padawan:
## Description

We need to have a class with `@NpmPackage` annotation so that it can be detected by Flow.

## Type of change

- Feature

### DIFF
--- a/vaadin-map-flow-parent/vaadin-map-flow/pom.xml
+++ b/vaadin-map-flow-parent/vaadin-map-flow/pom.xml
@@ -12,6 +12,7 @@
     <description>Vaadin Map</description>
     <properties>
         <osgi.bundle.license>https://vaadin.com/commercial-license-and-service-terms</osgi.bundle.license>
+        <jackson.version>3.0.0-rc9</jackson.version>
     </properties>
     <dependencies>
         <dependency>
@@ -51,6 +52,11 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-simple</artifactId>
             <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>tools.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>${jackson.version}</version>
         </dependency>
     </dependencies>
     <build>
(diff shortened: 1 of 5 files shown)


vaadin-aura-theme-flow-parent/vaadin-aura-theme-flow-integration-tests/pom.xml
sissbruecker, line 1: Is this necessary for the CI build? Bit weird to add an empty module.

vaadin-aura-theme-flow-parent/vaadin-aura-theme-flow/src/main/java/com/vaadin/flow/theme/aura/Aura.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+package com.vaadin.flow.theme.aura;
+
+import com.vaadin.flow.component.dependency.NpmPackage;
+
+/**
+ * Aura theme base class.
+ */
+@NpmPackage(value = "@vaadin/aura", version = "25.0.0-alpha19")
+public class Aura {
```
sissbruecker, line 24: Let's maybe check with the Flow team on whether this approach actually would work. By default this would do nothing unless the class is referenced somewhere.

So there probably needs to be some change in Flow that would install the NPM package regardless. We should at least have an issue for this.

web-padawan: Let's discuss this in the today's theming meeting, if we have time. I can create an issue afterwards.

vaadin-aura-theme-flow-parent/vaadin-aura-theme-flow-integration-tests/pom.xml
web-padawan: Shouldn't be needed, I created it just in case, could probably remove.

vaadin-aura-theme-flow-parent/vaadin-aura-theme-flow/src/main/java/com/vaadin/flow/theme/aura/Aura.java
sissbruecker: As discussed yesterday the approach should be fine as is.

PR #8045, opened 2025-09-22
feat: Add support for accessing multiple grid header rows (#2006)

Artur-:
Extends GridElement to provide access to all header rows, including hierarchical and joined header cells. Previously only the bottom-most header was accessible.

New features:
- GridHeaderRow class to represent individual header rows
- GridHeaderCell class to represent header cells with colspan/rowspan support
- Methods to get all header rows, specific rows by index, and row counts
- Support for detecting joined cells and multi-row spans
- Methods to access first, last, and visible header rows only

Fixes vaadin/testbench#2006

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridMultipleFooterRowsIT.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridMultipleFooterRowsIT.java
@@ -1,0 +1,195 @@
+/*
+ * Copyright 2000-2025 Vaadin Ltd.
+ *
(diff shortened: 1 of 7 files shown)


TatuLund:
There is already PR for the same thing here: https://github.com/vaadin/flow-components/pull/7335

Artur-:
This has tests, the other does not. How should we merge them?

TatuLund:
> This has tests, the other does not. How should we merge them?

Instead of adding new tests from this PR to #7335, one should refactor the existing Header and Footer tests to use the new API, i.e. this one

https://github.com/vaadin/flow-components/blob/main/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/GridHeaderFooterRowIT.java

Unfortunately the old tests are not using TestBench API properly at all, I had a plan to do this in PR when I have slack time.

TatuLund:
I found another test to refactor for #7335 .

PR #8054, opened 2025-09-23
chore: fix platform update script after removing bundles [skip ci]

ZheSun88:


vursen:
Just making sure, this should only apply to v25. In earlier versions, bundles should still be maintained.

ZheSun88:
yes, it is only for main branch.. so will not affect other versions

PR #8058, opened 2025-09-23
feat: Extract theme files during build

Artur-:
- Add Node.js script that reads Aura.java and Lumo.java to extract npm package info
- Downloads, copies and minifies the CSS entry file from the package to target/classes

### DIFF
--- a/vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/CheckboxGroup.java
+++ b/vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/CheckboxGroup.java
@@ -906,7 +906,7 @@ public class CheckboxGroup<T>
         }
         Set<T> set = new HashSet<>();
         for (int i = 0; i < presentation.size(); i++) {
-            set.add(group.keyMapper.get(presentation.get(i).asText()));
+            set.add(group.keyMapper.get(presentation.get(i).asString()));
         }
         return set;
     }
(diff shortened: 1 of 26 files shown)


copy-npm-theme-files.js
```diff
@@ -0,0 +1,163 @@
...
+
+packages.forEach(pkg => {
+    packageJson.dependencies[pkg.name] = pkg.version;
+});
+
+// Add PostCSS and plugins for minification
+packageJson.dependencies['postcss'] = '^8.4.31';
+packageJson.dependencies['postcss-cli'] = '^10.1.0';
+packageJson.dependencies['cssnano'] = '^6.0.1';
+packageJson.dependencies['postcss-import'] = '^15.1.0';
```
web-padawan: We also need `postcss-url` for Aura to make sure fonts are resolved correctly. Example:

https://github.com/vaadin/web-components/blob/89bdb78afae3f25525b20395398db210be5c57a5/dev/rollup.config.js#L35

Artur-: done

web-padawan: Apparently also [`postcss-assets`](https://www.npmjs.com/package/postcss-assets) or something like this to manage fonts.

Artur-: Fixed fonts

copy-npm-theme-files.js
```diff
@@ -0,0 +1,169 @@
...
+        break;
+    }
+}
+
+// Create temporary directory for npm install
+const tempDir = path.join(projectRoot, 'target', 'npm-workspace');
+fs.mkdirSync(tempDir, { recursive: true });
+
+// Create package.json with all dependencies plus PostCSS
+const packageJson = {
```
vursen: Why do we need a separate package? Could we just add `postcss` and the other dependencies to the root `package.json` as dev dependencies and import them directly in this script?

vursen: Nevermind. We anyway need it to download the actual theme npm package.

Artur-:
I thought `lumo.css` was the main entry point to include everything but that's not the case?

web-padawan:
We have two entrypoints - `lumo.css` with all the Lumo stuff including badge, and `utility.css` with utility classes. 
This is mostly done for introducing Tailwind support in the future.

There is also `presets/compact.css` file that would need to be addressed, but can be excluded from the initial version.

sissbruecker:
> We have two entrypoints - `lumo.css` with all the Lumo stuff including badge, and `utility.css` with utility classes. This is mostly done for introducing Tailwind support in the future.
> 
> There is also `presets/compact.css` file that would need to be addressed, but can be excluded from the initial version.

Changed it so that it also processes `utility.css` for Lumo.

web-padawan (approved):

PR #8060, opened 2025-09-23
feat: make Upload target name configurable

patrick-brielmayer:
## Description
This change makes the server-side upload endpoint name of the Upload component configurable instead of hardcoded to "upload".

Motivation and context:
- Some applications need distinct endpoint names per Upload instance (e.g., custom routing, reverse proxies, or security rules bound to specific path segments).
- The static endpoint name limited integration flexibility.

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
+++ b/vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
@@ -100,6 +100,12 @@ public class Upload extends Component implements HasEnabled, HasSize, HasStyle {
     private Receiver receiver;
 
     /**
+     * Represents the name of the target location or identifier where files
+     * or data are to be uploaded. Default value is set to "upload".
+     */
+    private String uploadTargetName = "upload";
+
+    /**
      * Create a new instance of Upload.
      * <p>
      * The upload handler must be set through
@@ -194,6 +200,19 @@ public class Upload extends Component implements HasEnabled, HasSize, HasStyle {
     public Upload(UploadHandler handler) {
         this();
         setUploadHandler(handler);
+    }
+
+    /**
+     * Create a new instance of Upload with the given upload handler and target name.
+     *
+     * @param handler
+     *            upload handler that handles the upload, not {@code null}
+     * @param targetName
+     *            the endpoint name (single path segment) to publish the upload under, not blank
+     */
+    public Upload(UploadHandler handler, String targetName) {
+        this();
+        setUploadHandler(handler, targetName);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -821,6 +859,25 @@ public void clearFileList() {
         getElement().setPropertyJson("files", JacksonUtils.createArrayNode());
     }
 
+    /**
+     * Retrieves the name of the upload target.
+     *
+     * @return the name of the upload target as a String.
+     */
+    public String getUploadTargetName() {
```
sissbruecker: Since the name is only used for configuring the upload handler, I think using a property is unnecessary. It's also potentially confusing because it doesn't work with the deprecated receiver API.

I'd suggest to drop the property and instead move the logic for constructing the `ElementStreamResource` to `setUploadHandler(UploadHandler handler, String name)`, and then let `setUploadHandler(UploadHandler handler)` pass `upload` as default value.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -743,6 +764,23 @@ public String getName() {
...
+     * Set the upload handler to be used for this upload component.
+     * <p>
+     * The given handler defines how uploaded file content is handled on the
+     * server and invoked per each single file to be uploaded. Note! This method
+     * overrides the receiver set by {@link #setReceiver(Receiver)}.
+     *
+     * @param handler
+     *            upload handler to use for file receptions, not {@code null}
+     * @param targetName
+     *            endpoint name (single path segment), not blank
```
sissbruecker: I think the JavaDoc could be improved, as it does not really tell you what this is for.

It should somehow mention that this is used as the last segment in the dynamically generated upload URL, and maybe it should also mention the default somewhere.

sissbruecker (commented):
Thanks for the contribution, left a few comments.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
patrick-brielmayer: Agreed - implemented as suggested

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
patrick-brielmayer: Agreed - implemented as suggested

sissbruecker (approved):

mcollovati:
Be aware that this change may impact defaults used to secure the application, e.g. in the `vaadin-spring` addon

https://github.com/vaadin/flow/blob/cadd1e348899d6833641a49a1064fe34ea315e51/flow-server/src/main/java/com/vaadin/flow/server/HandlerHelper.java#L249-L256

PR #8069, opened 2025-09-24
feat!: add support for server side ModalityMode

tltv:
Introduces `Dialog#setModality(ModalityMode)` with three modality modes: `STRICT`, `VISUAL` and `MODELESS`. Changes a Dialog to have `ModalityMode#VISUAL` by default instead of being strict modal. `LoginOverlay` has `ModalityMode#STRICT` by default instead of being like `ModalityMode#VISUAL`. Deprecates `Dialog#setModal(boolean)` and `Dialog#isModal()`.

Part of: vaadin/flow#22279


vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/LoginOverlay.java
```diff
@@ -63,7 +66,7 @@ public LoginOverlay(LoginI18n i18n) {
     private void init() {
         // Initialize auto-add behavior
         OverlayAutoAddController<LoginOverlay> autoAddController = new OverlayAutoAddController<>(
-                this);
+                this, () -> ModalityMode.STRICT);
```
sissbruecker, line 69: This will only make the overlay modal when it is auto added. When it is added manually to the UI the auto add controller is not used. 

Related to that `testOverlayModalityModeIsStrictByDefault` also passes without this change, as the overlay in the test setup is added manually to the UI.

If we want `LoginOverlay` to be a server-side modal this would require further implementation to make it a modal when opening it, and make it non-modal when closing it.

tltv: I see, there's indeed a test that was supposed to catch this but I see now that test did not take in account that overlay is not removed from the dom when closed. I'll fix it.

tltv: Fixed in same way Dialog does it.

sissbruecker (approved):

PR #8070, opened 2025-09-25
feat: implement scrolltoitem feature for treegrid

ugur-vaadin:
## Description

This PR implements `scrollToItem` functionality for `TreeGrid`.

Part of https://github.com/vaadin/flow-components/issues/8076

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/treegrid/TreeGridScrollTest.java
```diff
@@ -0,0 +1,203 @@
...
+            public HierarchyFormat getHierarchyFormat() {
+                return HierarchyFormat.FLATTENED;
+            }
+
+            @Override
+            public int getDepth(HierarchicalTestBean item) {
+                return item.getDepth();
+            }
+        });
+    }
```
vursen:
```suggestion
        treeGrid.setDataProvider(new TreeDataProvider<>(treeData, HierarchyFormat.FLATTENED));
    }
```

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -1115,21 +1119,91 @@ int setViewportRangeByIndexPath(int[] path, int padding) {
...
+        parents.push(item);
+        var parent = getParent(item);
+        while (parent != null) {
+            parents.push(parent);
+            parent = getParent(parent);
+        }
+        return parents;
+    }
+
+    private int getItemIndex(T item, T parent) {
```
vursen: Suggestion: Move this method to `TreeGridDataCommunicator`

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -1130,21 +1131,69 @@ int setViewportRangeByIndexPath(int[] path, int padding) {
...
+        var indexesToScrollTo = new int[pathFromRoot.size()];
+        indexesToScrollTo[0] = getItemIndex(pathFromRoot.get(0), null);
+        for (var i = 1; i < pathFromRoot.size(); i++) {
+            indexesToScrollTo[i] = getItemIndex(pathFromRoot.get(i),
+                    pathFromRoot.get(i - 1));
+        }
+        scrollToIndex(indexesToScrollTo);
+    }
+
+    private List<T> getPathFromRoot(T item) {
```
vursen: Suggestion:  move this logic and item index calculation from `scrollToNestedIndexes` into `TreeGridDataCommunicator` as a method named something like `getIndexPath(T item)`

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -1130,21 +1131,69 @@ int setViewportRangeByIndexPath(int[] path, int padding) {
...
+                .equals(HierarchicalDataProvider.HierarchyFormat.FLATTENED)) {
+            scrollToFlattenedIndex(itemToScrollTo);
+        } else {
+            scrollToNestedIndexes(itemToScrollTo);
+        }
+    }
+
+    private T getItemToScrollTo(T item) {
+        return getPathFromRoot(item).stream()
+                .filter(itemInPath -> !isExpanded(itemInPath)).findFirst()
```
vursen: `scrollToIndex(int... )` is already designed to stop when it encounters a collapsed item, so this check shouldn't be needed.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
ugur-vaadin: Moved.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
ugur-vaadin: Moved.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
ugur-vaadin: The logic is still required for the flattened providers, but I removed it for the nested providers.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -1130,21 +1161,48 @@ int setViewportRangeByIndexPath(int[] path, int padding) {
     }
 
     /**
-     * TreeGrid does not support scrolling to a given item. Use
-     * {@link #scrollToIndex(int...)} instead.
+     * Scrolls to an item within the tree. Does nothing if the item does not
+     * belong to the tree.
```
ugur-vaadin: When an item is not found, this method now does nothing.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -1130,21 +1161,48 @@ int setViewportRangeByIndexPath(int[] path, int padding) {
...
+                .equals(HierarchicalDataProvider.HierarchyFormat.FLATTENED)) {
+            scrollToFlattenedIndex(item);
+        } else {
+            scrollToNestedIndexes(item);
+        }
+    }
+
+    private void scrollToFlattenedIndex(T item) {
+        var dataCommunicator = (TreeGridDataCommunicator<T>) getDataCommunicator();
+        var itemToScrollTo = dataCommunicator.getPathFromRoot(item).stream()
```
ugur-vaadin: This logic is still required for determining the item to scroll to for flattened data providers.

vursen: I'm actually not sure how it should work in this case. If the item isn't reachable or not present at all, I wonder if it should just not scroll anywhere... could the current behavior be confusing? `scrollToIndex(int...)` feels different because it explicitly takes a full path

ugur-vaadin: That's a good point. In my opinion, even though it accepts a path in `scrollToIndex(int...)`, it can still be argued that the path refers to a single item. In any case, we should also mention how it works in the JavaDoc based on the decision.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -249,6 +250,35 @@ protected List<T> preloadFlatRangeBackward(int start, int length) {
...
+        public List<T> getPathFromRoot(T item) {
+            var dataProvider = TreeGridDataCommunicator.this.getDataProvider();
+            var parents = new LinkedList<T>();
+            parents.push(item);
+            var parent = dataProvider.getParent(item);
+            while (parent != null) {
+                parents.push(parent);
+                parent = dataProvider.getParent(parent);
+            }
+            return parents;
```
vursen:
```suggestion
        private LinkedList<T> getAncenstors(T item) {
            var ancestors = new LinkedList<T>();
            while ((item = getDataProvider().getParent(item)) != null) {
                ancestors.push(item);
            }
            return ancestors;
        }
```

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -249,6 +250,35 @@ protected List<T> preloadFlatRangeBackward(int start, int length) {
...
+
+        public int[] getIndexPath(T item) {
+            var pathFromRoot = getPathFromRoot(item);
+            var indexesToScrollTo = new int[pathFromRoot.size()];
+            indexesToScrollTo[0] = getItemIndex(pathFromRoot.get(0), null);
+            for (var i = 1; i < pathFromRoot.size(); i++) {
+                indexesToScrollTo[i] = getItemIndex(pathFromRoot.get(i),
+                        pathFromRoot.get(i - 1));
+            }
+            return indexesToScrollTo;
```
vursen:
```suggestion
        public List<Integer> getIndexPath(T item) {
            var ancestors = getAncenstors(item);
            if (ancestors.stream().allMatch((ancestor) -> !isExpanded(ancestor))) {
                return null;
            }

            var path = new ArrayList<Integer>();
            if (getDataProvider().getHierarchyFormat()
                    .equals(HierarchyFormat.NESTED)) {
                for (int i = 0; i < ancestors.size(); i++) {
                    path.add(getItemIndex(ancestors.get(i),
                            i == 0 ? null : ancestors.get(i - 1)));
                }
            }
            path.add(getItemIndex(item,
                    path.isEmpty() ? null : ancestors.getLast()));
            return path;
        }
```

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -249,6 +250,35 @@ protected List<T> preloadFlatRangeBackward(int start, int length) {
...
+                parent = dataProvider.getParent(parent);
+            }
+            return parents;
+        }
+
+        public int getItemIndex(T item, T parent) {
+            var query = buildQuery(parent, 0, Integer.MAX_VALUE);
+            return ((HierarchicalDataProvider<T, Object>) TreeGridDataCommunicator.this
+                    .getDataProvider()).getItemIndex(item, query);
+        }
```
vursen:
```suggestion
        public int getItemIndex(T item, T parent) {
            var query = buildQuery(parent, 0, Integer.MAX_VALUE);
            return ((HierarchicalDataProvider<T, Object>) getDataProvider())
                    .getItemIndex(item, query);
        }
```

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
ugur-vaadin: However, I am not sure how `scrollToIndex(int...)` should behave in such situations. Currently, it scrolls to the last reachable item with the path. In the demo page, when `scrollToIndex(int...)` is called with 0-0-200, it scrolls to 0-0-19 which is the last item in that level.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -1130,21 +1160,48 @@ int setViewportRangeByIndexPath(int[] path, int padding) {
...
     public void scrollToItem(T item) {
-        throw new UnsupportedOperationException(
-                "scrollToItem method is not supported in TreeGrid");
+        Objects.requireNonNull(item, "Item to scroll to cannot be null.");
+        if (getDataProvider().getHierarchyFormat()
+                .equals(HierarchicalDataProvider.HierarchyFormat.FLATTENED)) {
+            scrollToFlattenedIndex(item);
+        } else {
+            scrollToNestedIndexes(item);
+        }
```
vursen:
```suggestion
        var indexPath = ((TreeGridDataCommunicator<T>) getDataCommunicator()).getIndexPath(item);
        if (indexPath == null) {
            return;
        }
        if (indexPath.size() == 1) {
            scrollToIndex(indexPath.get(0));
        } else {
            scrollToIndex(indexPath.toArray());
        }
```

vursen (commented):
Left a few ideas for code organization

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
ugur-vaadin: I added this explicitly to specify which `getDataProvider()` is used since both `TreeGrid` and `TreeGridDataCommunicator` has a method with the same name.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
ugur-vaadin: If I understand this correctly, you are proposing we should not scroll to anything if all the ancestors are collapsed. I'm not sure if we should treat the cases differently:

- Item to scroll to is 0-0-0 and all are collapsed
- Item to scroll to is 0-0-0 and only the root item 0 is expanded

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
vursen: Let's extract the TreeGridDataCommunicator into a separate file with package-private visibility then since it has grown enough.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -1130,21 +1160,48 @@ int setViewportRangeByIndexPath(int[] path, int padding) {
...
+     * default. Additionally, {@link TreeDataProvider} implements
+     * {@link HierarchicalDataProvider#getParent(T)} by default.
      *
      * @param item
      *            the item to scroll to
-     * @deprecated
      */
-    @Deprecated
     @Override
     public void scrollToItem(T item) {
```
sissbruecker, line 1150: The issue description has a table for in which cases you need to implement which method. At least for flat data providers, it seems that `getParent` does not need to be implemented. Is that still correct and should we mention that here? If so, we might want to put this table itself into the JavaDoc, as it gives a clearer overview than having the scenarios described in multiple paragraphs.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
ugur-vaadin: Extracted the data communicator.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
ugur-vaadin: Updated the implementation to expand before scrolling to the item. Also now we now do not scroll and throw an `IllegalArgumentException` if the item is not found.

ugur-vaadin: Used your suggestion and moved it to the new `TreeGridDataCommunicator`.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
ugur-vaadin: Applied your suggestion and moved it to the new `TreeGridDataCommunicator`.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
ugur-vaadin: Applied this suggestion and simplified the implementation that resides in `TreeGrid`.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
ugur-vaadin: We decided to expand any collapsed parent on scroll to item. Therefore, we still need the implementation of both the methods for any hierarchy format.

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/treegrid/ScrollToItemTest.java
```diff
@@ -0,0 +1,369 @@
...
+        Assert.assertThrows(UnsupportedOperationException.class, () -> treeGrid
+                .scrollToItem(new CustomTestBean("", 0, 0, null)));
+    }
+
+    @Test
+    public void flattenedProvider_scrollToRootItem_scrollsToCorrectIndex() {
+        setFlattenedProvider();
+        treeGrid.scrollToItem(treeData.getRootItems().get(30));
+        assertScrolledIndex(30);
+    }
```
vursen:
```suggestion
    public void treeDataProvider_flattenedHierarchyFormat_scrollToRootItem_scrollsToCorrectIndex() {
        treeGrid.setDataProvider(new TreeDataProvider<>(treeData, HierarchyFormat.FLATTENED));
        treeGrid.scrollToItem(treeData.getRootItems().get(30));
        assertScrolledIndex(30);
    }
```

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/treegrid/ScrollToItemTest.java
```diff
@@ -0,0 +1,369 @@
...
+    @Test
+    public void nestedProvider_scrollToMissingItem_doesNotScroll() {
+        setNestedProvider();
+        Assert.assertThrows(IllegalArgumentException.class,
+                this::scrollToMissingItem);
+        assertNotScrolled();
+    }
+
+    @Test
+    public void lazyProvider_scrollToMissingItem_doesNotScroll() {
```
vursen: It should be enough to test that scrollToIndex works as expected with TreeDataProvider since its implementation already covers both methods. For other data providers, it should be enough to make sure scrollToIndex throws the proper exception if either method is missing.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGridDataCommunicator.java
```diff
@@ -0,0 +1,154 @@
...
+     * @param item
+     *            the item to get the ancestors for
+     * @return ordered list of ancestors of the given item
+     */
+    private List<T> getAncestors(T item) {
+        var ancestors = new ArrayList<T>();
+        while ((item = getDataProvider().getParent(item)) != null) {
+            ancestors.add(item);
+        }
+        Collections.reverse(ancestors);
```
vursen: Is there a reason to prefer `Collections.reverse` to adding elements to a `LinkedList` using `addFirst`?

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -1130,21 +1091,68 @@ int setViewportRangeByIndexPath(int[] path, int padding) {
...
-     * {@link UnsupportedOperationException}.
+     * In order to be able to use this method, the data provider should
+     * implement {@link HierarchicalDataProvider#getParent(T)} and
+     * {@link HierarchicalDataProvider#getItemIndex(T, HierarchicalQuery)}. The
+     * following table shows which methods have to be explicitly implemented
+     * based on the data provider types.
+     * <table>
+     * <tr>
+     * <th>DataProvider</th>
+     * <th>In-memory</th>
```
vursen:
```suggestion
     * <th>isInMemory()</th>
```

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -1130,21 +1091,68 @@ int setViewportRangeByIndexPath(int[] path, int padding) {
...
+     * <td>HierarchyFormat.FLATTENED</td>
+     * <td>true</td>
+     * <td>not required</td>
+     * <td>not required</td>
+     * </tr>
+     * <tr>
+     * <td>HierarchyFormat.FLATTENED</td>
+     * <td>false</td>
+     * <td>required</td>
+     * <td>not required</td>
```
vursen:
```suggestion
     * <td>required</td>
```

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -1130,21 +1091,68 @@ int setViewportRangeByIndexPath(int[] path, int padding) {
...
+     * <td>HierarchyFormat.NESTED</td>
+     * <td>false</td>
+     * <td>required</td>
+     * <td>required</td>
+     * </tr>
+     * <tr>
+     * <td>HierarchyFormat.FLATTENED</td>
+     * <td>true</td>
+     * <td>not required</td>
+     * <td>not required</td>
```
vursen:
```suggestion
     * <td>required</td>
```

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGridDataCommunicator.java
ugur-vaadin: Not really. It doesn't make sense in terms of the operation itself but it looked slightly simpler to read and we have no performance concerns. I can use `LinkedList` again, no strong opinions.

ugur-vaadin: Okay my bad, pushed the changes. It's much simpler now.

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/treegrid/ScrollToItemTest.java
ugur-vaadin: Removed lazy provider tests except the test that expects an `UnsupportedOperationException`.

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/treegrid/ScrollToItemTest.java
ugur-vaadin: Updated the names of the tests and refactored to explicitly set the providers.

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/treegrid/ScrollToItemTest.java
```diff
@@ -0,0 +1,280 @@
...
+            this.depth = depth;
+        }
+
+        @Override
+        public int getChildCount(
+                HierarchicalQuery<HierarchicalTestBean, Void> query) {
+            var count = query.getParentOptional().flatMap(parent -> Optional
+                    .of(internalHasChildren(parent) ? nodesPerLevel : 0));
+
+            return count.orElse(nodesPerLevel);
```
vursen:
```suggestion
            return 0;
```

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/treegrid/ScrollToItemTest.java
```diff
@@ -0,0 +1,280 @@
...
+        }
+
+        @Override
+        public boolean hasChildren(HierarchicalTestBean item) {
+            return internalHasChildren(item);
```
vursen:
```suggestion
            return false;
```

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/treegrid/ScrollToItemTest.java
```diff
@@ -0,0 +1,280 @@
...
+        }
+
+        private boolean internalHasChildren(HierarchicalTestBean node) {
+            return node.getDepth() < depth;
+        }
+
+        @Override
+        protected Stream<HierarchicalTestBean> fetchChildrenFromBackEnd(
+                HierarchicalQuery<HierarchicalTestBean, Void> query) {
+            var queryDepth = query.getParentOptional().isPresent()
```
vursen:
```suggestion
            return Stream.empty();
```

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -1130,21 +1091,68 @@ int setViewportRangeByIndexPath(int[] path, int padding) {
...
-     * {@link UnsupportedOperationException}.
+     * In order to be able to use this method, the data provider should
+     * implement {@link HierarchicalDataProvider#getParent(T)} and
+     * {@link HierarchicalDataProvider#getItemIndex(T, HierarchicalQuery)}. The
+     * following table shows which methods have to be explicitly implemented
+     * based on the data provider types.
+     * <table>
+     * <tr>
+     * <th>DataProvider</th>
+     * <th>isInMemory()</th>
```
vursen:
```suggestion
     * <th>{@link HierarchicalDataProvider#isInMemory() isInMemory()}</th>
```

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/treegrid/TreeGrid.java
```diff
@@ -1130,21 +1091,68 @@ int setViewportRangeByIndexPath(int[] path, int padding) {
...
-     * indicate that it is not supported. This method will throw an
-     * {@link UnsupportedOperationException}.
+     * In order to be able to use this method, the data provider should
+     * implement {@link HierarchicalDataProvider#getParent(T)} and
+     * {@link HierarchicalDataProvider#getItemIndex(T, HierarchicalQuery)}. The
+     * following table shows which methods have to be explicitly implemented
+     * based on the data provider types.
+     * <table>
+     * <tr>
+     * <th>DataProvider</th>
```
vursen:
```suggestion
     * <th>HierarchicalDataProvider</th>
```

vursen (approved):

PR #8077, opened 2025-09-25
chore: copy theme distribution from NPM theme packages

sissbruecker:
## Description

Changes the existing script to copy distribution files from NPM theme packages instead of building them on the fly with PostCSS.

## Type of change

- Internal


scripts/copy-theme-distribution.js
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const path = require('path');
+const { execSync } = require('child_process');
+
+// Check command line arguments
+if (process.argv.length !== 4) {
+  console.error('Usage: node copy-theme-distribution.js <path-to-theme-class> <npm-theme-package>');
```
web-padawan, line 16: Let's add a comment in the start of the file explaining that this script doesn't have to be run manually, as it runs when running `./scripts/run.sh` -> "one compile" for `vaadin-aura-theme` or `vaadin-lumo-theme`.

web-padawan (approved):

web-padawan (approved):

PR #8081, opened 2025-09-26
feat: cluster click event

sissbruecker:
## Description

Allows listening for clicks on clusters. The event will contain references to all features within the cluster. For clusters with a single feature still fires a feature click event.

Depends on #8037 

## Type of change

- Feature


vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/MapBase.java
```diff
@@ -229,6 +230,19 @@ public Registration addFeatureClickListener(
         return addListener(MapFeatureClickEvent.class, listener);
     }
 
+    /**
```
ugur-vaadin, line 281: Should it also be mentioned in the JavaDoc for `addFeatureClickListener`?

ugur-vaadin (approved):

DiegoCardoso (approved):

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/MapBase.java
sissbruecker: Good point, added some clarification on when each listener should be used.

Issue #3137, opened 2022-05-06
Add "zoomToContent" method or similar


mstahv:
### Describe your motivation

In many cases the developer wants to place a couple of features to the map and then adjust the view port somehow optimally, so that all features fit in and it as detailed zoom level as possible. We should make this to be very easy for developers.

### Describe the solution you'd like

We should provided a zoomToContent method that takes all features (markers, polylines, polygons etc), calculates their extent and optimises the view on the map so that all can become visible.

### Describe alternatives you've considered

Alternatively we should provide API to `View.fit` (https://openlayers.org/en/latest/apidoc/module-ol_View-View.html#fit).

This method will be asked anyway by the users of the add-on at some point, but is not as handy as the proposed zoomToContent without extra parameters. Probably when implementing zoomToContent, `view.fit` can easily be exposed as well.

PR #8082, opened 2025-09-26
feat: zoom to fit the provides features in the viewport

sissbruecker:
## Description

Adds a `zoomToFit` method to the `Map` component that allows to fit the provided features within the viewport. The options allow to configure a padding for the viewport and an animtation duration.

Closes https://github.com/vaadin/flow-components/issues/3137

## Type of change

- Feature

### DIFF
--- a/vaadin-map-flow-parent/vaadin-map-flow-integration-tests/src/main/java/com/vaadin/flow/component/map/ZoomToFitPage.java
+++ b/vaadin-map-flow-parent/vaadin-map-flow-integration-tests/src/main/java/com/vaadin/flow/component/map/ZoomToFitPage.java
@@ -1,0 +1,62 @@
+/**
+ * Copyright 2000-2025 Vaadin Ltd.
+ *
+ * This program is available under Vaadin Commercial License and Service Terms.
+ *
+ * See {@literal <https://vaadin.com/commercial-license-and-service-terms>} for the full
+ * license.
+ */
+package com.vaadin.flow.component.map;
+
+import java.util.List;
+
+import com.vaadin.flow.component.html.Div;
+import com.vaadin.flow.component.html.NativeButton;
+import com.vaadin.flow.component.map.configuration.Coordinate;
+import com.vaadin.flow.component.map.configuration.feature.MarkerFeature;
+import com.vaadin.flow.router.Route;
+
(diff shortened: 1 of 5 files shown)


vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/Map.java
```diff
@@ -321,4 +327,44 @@ public double getZoom() {
...
     }
+
+    /**
+     * Zooms and pans the map to fit the given features in the viewport. Uses a
+     * default padding of 50 pixels and animation duration of 400 milliseconds.
+     *
+     * @param features
+     *            the features to fit in the viewport
+     */
+    public void zoomToFit(List<Feature> features) {
```
sissbruecker, line 338: Ideally this would also be available on the `View` instance of the map. However, none of the configuration objects have any reference to the Map element to trigger JS calls. I'm not sure about adding one at the moment, as technically configuration objects can be shared between multiple map instances (though sharing the view seems extremely unlikely).

Still, I think it's fine to start only with this for now. The new API at least fits with having `setCenter` and `setZoom` methods on the `Map`. If this turns out to be an issue, we can spend more time on making it work through the view instance.

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/Map.java
```diff
@@ -321,4 +327,44 @@ public double getZoom() {
...
+
+        // In order to allow calling this method before the map, or the view
+        // that contains it, is attached to the UI, this needs to be delayed
+        // with beforeClientResponse to ensure that the client-side connector
+        // has been initialized beforehand.
+        getElement().getNode()
+                .runWhenAttached(ui -> ui.beforeClientResponse(this,
+                        context -> getElement().executeJs(
+                                "this.$connector.zoomToFit($0, $1)", featureIds,
+                                options)));
```
sissbruecker, line 368: Could not find a better approach. In some form this needs to run after `onAttach`, which is responsible for creating the connector.

DiegoCardoso (approved):

PR #8088, opened 2025-09-28
fix: Grid column respects custom reversed comparator and improves serialization (#3926)

### LINKED ISSUE #3926: Grid ignores reverse comparator
### Description

I use the [`grid.setComparator(<>)` method](https://github.com/vaadin/flow-components/blob/b46c995ff542d17aaaaf01b60275ad7f21514e44/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java#L712-L717) to supply comparator to the grid. My comparator has custom reverse comparator - I do NOT want the default one. Vaadin ignores my reverse comparator and uses own reverse comparator.

Here is [setComparator method](https://github.com/vaadin/flow-components/blob/b46c995ff542d17aaaaf01b60275ad7f21514e44/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java#L712-L717)  and here is [getComparator method](https://github.com/vaadin/flow-components/blob/b46c995ff542d17aaaaf01b60275ad7f21514e44/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java#L756-L763).

I need the setComparator to do something like this:
````java
        public Column<T> setComparator(Comparator<T> comparator) {
            Objects.requireNonNull(comparator, "Comparator must not be null");
            setSortable(true);
            this.comparator = comparator::compare;
            this.reverseComparator = comparator.reversed()::compare; // <----- new line here
            return this;
        }
````

and the getComparator to do:
````java
        public SerializableComparator<T> getComparator(
                SortDirection sortDirection) {
            Objects.requireNonNull(comparator,
                    "No comparator defined for sorted column.");
            setSortable(true);
            boolean reverse = sortDirection != SortDirection.ASCENDING;
            return reverse ? reverseComparator : comparator; <----- change here
        }
````

Reason: I need to implement "favorites" functionality - a set of rows that is up even if the order is descending. The "natural" way to implement it would be to use reverse comparator for that.

### Expected outcome

If I supply a comparator with custom reverse comparator, vaadin should use that one to to compare rows.

````java
    Grid<Integer> grid2 = new Grid<>();
    Comparator<Integer> comparator = new Comparator<Integer>() {

      @Override
      public int compare(Integer o1, Integer o2) {
        return Comparator
          .comparing(this::isFavorite)
          .thenComparing(getNestedComparator())
          .compare(o1, o2);
      }

      @Override
      public Comparator<Integer> reversed() {
        return Comparator
          .comparing(this::isFavorite)
          .thenComparing(getNestedComparator().reversed());
      }

      private Comparator<Integer> getNestedComparator() {
        return Comparator.naturalOrder();
      }

      private boolean isFavorite(Integer t) {
        return t % 2 == 0;
      }

    };
    Column<Integer> c = grid2.addColumn(i -> String.valueOf(i));
    c.setComparator(comparator);

grid2.setItems(1, 2, 3, 4, 5, 6, 7, 8, 9, 10);
````

### Steps to reproduce

Try to sort the above grid. Even numbers should be always on top.

### Environment

Vaadin version(s): 
OS: 

### Browsers

_No response_

### PR DESCRIPTION
**Pull request by Artur-**
- Store the original Comparator object instead of just method references
- Preserve SerializableComparator instances when provided
- Respect custom reversed() implementations for descending sort

The column now stores the full comparator object, allowing it to:
1. Use custom reversed() implementations when sorting descending
2. Preserve serialization capability when SerializableComparator is provided
3. Only create method references as a fallback for non-serializable comparators

Fixes #3926

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
@@ -445,7 +445,7 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
             return Stream.of(new QuerySortOrder(key, direction));
         };
 
-        private SerializableComparator<T> comparator;
+        private Comparator<T> comparator;
 
         private Registration columnDataGeneratorRegistration;
         private Registration editorDataGeneratorRegistration;
@@ -735,7 +735,7 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
         public Column<T> setComparator(Comparator<T> comparator) {
             Objects.requireNonNull(comparator, "Comparator must not be null");
             setSortable(true);
-            this.comparator = comparator::compare;
+            this.comparator = comparator;
             return this;
         }
 
@@ -782,7 +782,19 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
                     "No comparator defined for sorted column.");
             setSortable(true);
             boolean reverse = sortDirection != SortDirection.ASCENDING;
-            return reverse ? comparator.reversed()::compare : comparator;
+
+            if (reverse) {
+                Comparator<T> reversed = comparator.reversed();
+                if (reversed instanceof SerializableComparator) {
+                    return (SerializableComparator<T>) reversed;
+                }
+                return reversed::compare;
+            }
+
+            if (comparator instanceof SerializableComparator) {
+                return (SerializableComparator<T>) comparator;
+            }
+            return comparator::compare;
         }
 
         /**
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Artur-** [comment]
The API seems to be a lie as there is `setComparator(Comparator<T> comparator)` but `SerializableComparator<T> getComparator( SortDirection sortDirection)` which does not in any way ensure the comparator is serializable

File: vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
Comment (sissbruecker) [inline vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java:1]: It looks like the change breaks serialization:
```
java.lang.RuntimeException: serializeAndDeserializeUi_noExceptionIsThrown[any_Chrome_](com.vaadin.flow.component.grid.it.GridLitRendererSerializationIT)
```

yuriy-fix:
After internal discussion, it seems that the issue couldn't be addressed with this change and would require a breaking one to be fixed.

Artur-:
Which breaking change is that?

PR #8095, opened 2025-09-30
fix: prevent unnecessary viewport update on VirtualList first render [1 day]

Artur-:
The VirtualList connector was initializing lastRequestedRange to [0, 0], causing an unnecessary viewport range update on first render. This update would reload list items, destroying any focused elements and causing focus loss.

Changed initialization to null to properly detect the first request and avoid spurious updates that interfere with focus management in master-detail views.

Fixes the issue where focusing the first item only worked on the second attempt due to the list being unnecessarily reloaded on first render.


Artur-:
I don't think the analysis is correct here and I cannot reproduce the original problem, so let's just close this

PR #8101, opened 2025-09-30
chore!: set Java 21 as baseline

vursen:
## Description

Follow-up to https://github.com/vaadin/flow/pull/22357

## Type of change

- [x] Chore

### DIFF
--- a/vaadin-accordion-flow-parent/vaadin-accordion-flow-integration-tests/src/main/java/com/vaadin/flow/component/app/TestAppShell.java
+++ b/vaadin-accordion-flow-parent/vaadin-accordion-flow-integration-tests/src/main/java/com/vaadin/flow/component/app/TestAppShell.java
@@ -15,9 +15,9 @@
  */
 package com.vaadin.flow.component.app;
 
-import com.vaadin.flow.component.dependency.CssImport;
+import com.vaadin.flow.component.dependency.StyleSheet;
 import com.vaadin.flow.component.page.AppShellConfigurator;
 
-@CssImport("@vaadin/vaadin-lumo-styles/lumo.css")
+@StyleSheet("@vaadin/vaadin-lumo-styles/lumo.css")
 public class TestAppShell implements AppShellConfigurator {
 }
--- a/vaadin-app-layout-flow-parent/vaadin-app-layout-flow-integration-tests/src/main/java/com/vaadin/flow/component/app/TestAppShell.java
+++ b/vaadin-app-layout-flow-parent/vaadin-app-layout-flow-integration-tests/src/main/java/com/vaadin/flow/component/app/TestAppShell.java
@@ -15,9 +15,9 @@
  */
 package com.vaadin.flow.component.app;
 
-import com.vaadin.flow.component.dependency.CssImport;
+import com.vaadin.flow.component.dependency.StyleSheet;
 import com.vaadin.flow.component.page.AppShellConfigurator;
 
-@CssImport("@vaadin/vaadin-lumo-styles/lumo.css")
+@StyleSheet("@vaadin/vaadin-lumo-styles/lumo.css")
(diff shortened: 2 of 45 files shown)


pom.xml
```diff
@@ -71,8 +71,8 @@
     <properties>
         <flow.version>25.0-SNAPSHOT</flow.version>
         <testbench.version>10.0.0-alpha4</testbench.version>
-        <maven.compiler.source>17</maven.compiler.source>
-        <maven.compiler.target>17</maven.compiler.target>
+        <maven.compiler.source>21</maven.compiler.source>
+        <maven.compiler.target>21</maven.compiler.target>
```
mcollovati: Other projects/add-ons use `maven.compiler.release`, that is the first one checked by the `Check project Java version` step in CI builds.

```
        <maven.compiler.release>21</maven.compiler.release>
        <maven.compiler.source>${maven.compiler.release}</maven.compiler.source>
        <maven.compiler.target>${maven.compiler.release}</maven.compiler.target>
```

Probably `maven.compiler.(source|target)` are not needed if `release` is set.

knoobie: Release is enough since Java 9 https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html

mcollovati: Sure thing, thank you for the clarification. My comment was definitely unclear 😁
I wanted to point out that it could be that some CI jobs are still using target or source to identify the jvm to use at build time.

vursen: Thanks, I updated the POM

mcollovati:
Probably needs to set the maven-compiler-plugin version in the pluginManagement section of pom.
CI job is running `maven-compiler-plugin:3.1:compile` but compilation fails, probably because that version is still using `source` and `target` properties. Perhaps, setting it to 3.14.1 will fix the validation error.

knoobie:
100% the case.. that's not even ancient.. that sounds like a pre-historic version 😬

pom.xml
```diff
@@ -104,6 +103,12 @@
     </properties>
     <dependencyManagement>
         <dependencies>
+            <dependency>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.14.1</version>
+                <scope>compile</scope>
+            </dependency>
```
mcollovati: This should be configured in the `pluginManagement` section.

knoobie: Something worth to be added to the vaadin parent?

vursen: Thanks, that worked.

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/AddingColumnsIT.java
```diff
@@ -75,10 +75,10 @@ public void gridRendered_addColumnWithLocalDateTimeRenderer_cellsRendered() {
         // JDK16 adds extra comma after year in en_US
         Assert.assertTrue(
                 TestHelper.stripComments(grid.getCell(0, 0).getInnerHTML())
-                        .matches("January 1, 1980,? 1:20[ \\u00A0]AM"));
+                        .matches("January 1, 1980,? 1:20[ \\u00A0\\u202F]AM"));
```
web-padawan, line 78: nit: could also modify the comment above regarding JDK16 related to this change. Not a blocker though.

vursen: The comment about JDK16 doesn't seem related to this problem – it explains why comma is needed.

PR #8110, opened 2025-10-01
fix: auto-attach Popover based on helper annotation

### LINKED ISSUE #8031: Popover listeners don't work inside a modal dialog
### Description of the bug

When we place a popover inside a modal dialog, the addOpenedChangeListener does not work, and if I place a CheckboxGroup inside the popover, the CheckboxGroup listeners do not work either.

### Expected behavior

Listeners must be triggered correctly within modal dialogs.

```
CheckboxGroup<String> group = new CheckboxGroup<>();
group.addThemeVariants(CheckboxGroupVariant.LUMO_VERTICAL);
group.setItems(List.of("Test 1", "Test 2", "Test 3", "Test 4", "Test 5", "Test 6"));
group.addValueChangeListener(e -> { //Not working when inside a modal dialog
    for (String s : e.getValue())
        System.out.println(s);
});

Button button = new Button();
Popover popover = new Popover(group);
popover.addOpenedChangeListener(e -> {//Not working when inside a modal dialog
    System.out.println("Opened: " + e.isOpened());
});
popover.setTarget(button);

final Dialog dialog = new Dialog();
dialog.setWidth("320px");
dialog.setHeight("320px");
dialog.setModal(true); //if you disable this option it works
dialog.setCloseOnOutsideClick(true);
dialog.add(button);
dialog.open();
```

### Versions

- Vaadin / Flow version: 24.8.6
- Java version: 21.0.7
- OS version: Windows 11
- Browser version (if applicable): Chrome 139.0.7258.139
- Application Server (if applicable): Embedded tomcat from springboot
- IDE (if applicable):IntelliJ

DiegoCardoso:
## Description

Introduces a new internal Java annotation interface to help Popover decide where to attach itself when the component is auto added to the UI. Adds the `@ModalComponent` annotation to the following components:
- Dialog
- ConfirmDialog
- LoginOverlay

When the Popover target's element is inside any of the components above, the component will be placed in the same parent as the target's to allow events to be listened from the Popover content.

Fixes #8031 

## Type of change

- Bugfix


vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
```diff
@@ -729,17 +730,47 @@ private void removeFromUiIfAutoAdded() {
...
                 // Remove the popover from its current state tree
                 getElement().removeFromTree(false);
-                ui.addToModalComponent(this);
+                getModalParentComponent().ifPresentOrElse(modalParent -> {
+                    // Place the popover in the same context as the target
+                    getElement().removeAttribute("slot");
+                    if (target.getElement().hasAttribute("slot")) {
+                        getElement().setAttribute("slot",
+                                target.getElement().getAttribute("slot"));
+                    }
```
DiegoCardoso: Copying the slot value is particularly important for components that don't provide a default `<slot>`, such as `<vaadin-login-overlay>`.

sissbruecker: What if you do this:
```java
var container = new HorizontalLayout();
var target = new Button();
container.add(target);

popover.setTarget(target);
loginOverlay.getFooter().add(container);
```

DiegoCardoso: Hmm, right. That would be a problem in this case.

DiegoCardoso: I updated it to retrieve the slot value from any ancestor until it finds the modal element.

vaadin-popover-flow-parent/vaadin-popover-flow/src/test/java/com/vaadin/flow/component/popover/PopoverAutoAddTest.java
```diff
@@ -189,4 +193,118 @@ private void fakeClientResponse() {
         ui.getInternals().getStateTree().collectChanges(ignore -> {
         });
     }
+
+    @Test
+    public void popoverWithTargetInModalComponent_popoverAttachedToModal() {
```
sissbruecker, line 192: Nit: Let's keep the `fakeClientResponse` helper below the test cases.

sissbruecker, line 192: Let's also add a test case for when the target is added as virtual child, for example:
```java
    @Test
    public void popoverWithTargetAddedAsVirtualChild_popoverAttachedToModal() {
        var modal = new TestModalComponent();
        ui.add(modal);
        var target = new Div();
        modal.getElement().appendVirtualChild(target.getElement());
        var popover = new Popover();
        popover.setTarget(target);
        fakeClientResponse();
        Assert.assertEquals("Popover should be attached to modal", modal,
                popover.getParent().orElse(null));
    }
```

vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
```diff
@@ -729,17 +731,91 @@ private void removeFromUiIfAutoAdded() {
...
+
+    /**
+     * Finds the slot attribute value from the target or its ancestors up to the
+     * modal parent.
+     * 
+     * @param target
+     * @param modalParent
+     * @return the slot attribute value, or {@code null} if not found
+     */
+    private String getSlotValueFromAncestors(Element target,
```
sissbruecker: This seems to work well, but I wonder if it would handle all edge-cases with more complex nesting. 

As an alternative, how about providing a default slot name through the new annotation? That would make the logic also a bit more straight-forward maybe. For LoginOverlay specifically we just have to pick one slot, but I think it doesn't matter which one that is in this use case?

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/internal/ModalComponent.java
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Indicates that a component is a modal component, i.e. a component that blocks
+ * interaction with the rest of the UI when displayed.
+ * <p>
+ * Internal use only. May be renamed or removed in a future release.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface ModalComponent {
```
sissbruecker: Should we make this more specific to the current purpose of the interface? Otherwise it's not clear why this exists without looking at usages. We could name it `ModalRoot` and explain in the JavaDoc that the purpose is for auto-added components such as Popover to figure out where to add themselves to the UI so that interactions from those are not blocked by strict modality.

vaadin-popover-flow-parent/vaadin-popover-flow/src/test/java/com/vaadin/flow/component/popover/PopoverAutoAddTest.java
DiegoCardoso: Done.

DiegoCardoso: Added.

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/internal/ModalComponent.java
DiegoCardoso: I improved the JavaDoc a bit more to explain the purpose of the annotation.

vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
DiegoCardoso: It makes sense. I added a property to the annotation to allow customizing which slot the Popover should be attached to.

vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
```diff
@@ -729,17 +730,73 @@ private void removeFromUiIfAutoAdded() {
...
+                    .isAnnotationPresent(ModalRoot.class)) {
+                return parent;
+            }
+            parent = parentComponent.getParent();
+        }
+        return Optional.empty();
+    }
+
+    /**
+     * Updates the {@code slot} attribute based on the target or its ancestors
```
sissbruecker, line 777: Let's update this to mention that we use the slot name from the annotation now.

vaadin-popover-flow-parent/vaadin-popover-flow/src/test/java/com/vaadin/flow/component/popover/PopoverAutoAddTest.java
```diff
@@ -184,9 +188,142 @@ public void setTarget_openModal_popoverIsAttachedToUi() {
...
+        fakeClientResponse();
+
+        target.removeFromParent();
+        fakeClientResponse();
+        Assert.assertFalse("Popover should be detached",
+                popover.getParent().isPresent());
+    }
+
+    @Test
+    public void popoverWithTargetHavingSlot_popoverInheritsSlotAttribute() {
```
sissbruecker: This one doesn't work anymore. But could be updated to move the target between `TestModalContainerWithSlot` and a different one.

Not sure if the test below has any additional purpose then, it looks like everything is already covered by this one.

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/internal/ModalComponent.java
sissbruecker: Looks good, thanks

vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
DiegoCardoso: Thanks, updated.

vaadin-popover-flow-parent/vaadin-popover-flow/src/test/java/com/vaadin/flow/component/popover/PopoverAutoAddTest.java
DiegoCardoso: Yeah, makes sense to remove the other test. Refactored this one as well.

sissbruecker (approved):

PR #8137, opened 2025-10-09
refactor: simplify accessing features from cluster click event

sissbruecker:
## Description

Currently the cluster click event returns a list of `FeatureEventDetails`, which requires you to map those into actual feature instances. However, since the features can only ever come from the same layer and source, we can just provide a list of features and have the layer and source as separate fields on the event itself.

## Type of change

- Refactor


sissbruecker:
No need to do so far alphas IMO.

PR #8139, opened 2025-10-09
refactor: update theme paths

sissbruecker:
## Description

Update the paths for Lumo and Aura CSS files and provides constants to reference those using `@StyleSheet`.

Also updates IT app shells to use the new path.

## Type of change

- Refactor


mshabarov:
Flow part is ready https://github.com/vaadin/flow/pull/22479.

PR #8148, opened 2025-10-14
refactor: consider a popover as a potential parent for a nested popover

sissbruecker:
## Description

Currently a popover only auto-adds itself to components annotated with `ModalRoot`. With https://github.com/vaadin/web-components/pull/10322 we change the closing logic of the web component so that it specifically checks whether there are nested overlays or not. As such it would make sense to also auto-add a popover to another popover if the target component is within it.

## Type of change

- Refactor

### DIFF
--- a/vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
+++ b/vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
@@ -48,7 +48,7 @@ import tools.jackson.databind.node.ArrayNode;
  * @author Vaadin Ltd.
  */
 @Tag("vaadin-popover")
-@NpmPackage(value = "@vaadin/popover", version = "25.0.0-alpha21")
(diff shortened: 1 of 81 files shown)


vaadin-popover-flow-parent/vaadin-popover-flow/src/main/java/com/vaadin/flow/component/popover/Popover.java
```diff
@@ -754,18 +754,24 @@ private void onTargetAttach(UI ui) {
...
+    private Optional<Component> getParentComponentCandidate() {
         var parent = target.getParent();
         while (parent.isPresent()) {
             var parentComponent = parent.get();
-            if (parentComponent.getClass()
-                    .isAnnotationPresent(ModalRoot.class)) {
+            var isPopover = parentComponent instanceof Popover;
+            var isModal = parentComponent.getClass()
+                    .isAnnotationPresent(ModalRoot.class);
+            if (isPopover || isModal) {
```
sissbruecker, line 774: Chose this approach over adding `ModalRoot` to `Popover`, otherwise it might have required redefining the interface to mean something else. Can maybe reconsider this if we have more use-cases around auto adding something to overlays in general rather than modals specifically.

PR #8162, opened 2025-10-15
fix: check if element exists before initializing grid connector

vursen:
## Description

Fixes the regression from #7734 by checking element existence before initializing the grid connector.

## Type of change

- [x] Bugfix

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/DetachReattachIT.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/DetachReattachIT.java
@@ -220,4 +220,17 @@ public class DetachReattachIT extends AbstractComponentIT {
         sorter = grid.getHeaderCell(0).$("vaadin-grid-sorter").first();
         Assert.assertEquals(direction, sorter.getProperty("direction"));
     }
+
+    @Test
+    public void detach_attachAndDetach_noClientErrors() {
+        open();
+
+        // Detach
+        $("button").id("detach-button").click();
+
+        // Attach and detach in the same round trip
+        $("button").id("attach-and-detach-button").click();
+
+        checkLogsForErrors();
+    }
 }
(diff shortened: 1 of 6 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/test/java/com/vaadin/flow/component/grid/it/DetachReattachIT.java
```diff
@@ -220,4 +220,17 @@ public void sort_detachAndAttach_sortDirectionPreserved() {
         sorter = grid.getHeaderCell(0).$("vaadin-grid-sorter").first();
         Assert.assertEquals(direction, sorter.getProperty("direction"));
     }
+
+    @Test
+    public void detach_attachAndDetach_noClientErrors() {
+        open();
+
+        // Detach
+        $("button").id("detach-button").click();
```
ugur-vaadin, line 229: Nit: we can use `clickElementWithJs("detach-button")` for clicking buttons.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -1746,8 +1746,8 @@ private void handleDataChange(DataChangeEvent<T> dataChangeEvent) {
     protected void initConnector() {
         getUI().orElseThrow(() -> new IllegalStateException(
                 "Connector can only be initialized for an attached Grid"))
-                .getPage()
-                .executeJs("window.Vaadin.Flow.gridConnector.initLazy($0)",
+                .getPage().executeJs(
```
sissbruecker, line 1749: Now I wonder why this uses `getPage` instead of the component instance to execute JS. But can be checked later.

tomivirkki: Some components that also use `getPage` [wrap](https://github.com/vaadin/flow-components/blob/ecbb3b89b20d7aff5057b048d17b6ad0d59c088f/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java#L618-L621) the call in `runWhenAttached`

vursen: I had the same question, and I found that it uses `getPage()` to ensure this runs before any other `executeJs` calls that other methods may have scheduled before the component was attached e.g. `$connector.reset()`. I decided to leave this as is, as refactoring it seemed risky in terms of new regressions.

sissbruecker: Shall we leave a comment then for now: https://github.com/vaadin/flow-components/pull/8163?

vursen: Thanks!

PR #8177, opened 2025-10-22
chore: upgrade spring-data-commons to 4.0.0-RC1

ZheSun88:
To align with spring boot 4.0.0-RC1, which should be released 23rd of October


ZheSun88:
duplicated with #8171

PR #8205, opened 2025-10-30
fix!: Scroll item to view when calling select(int)

TatuLund:
Closes https://github.com/vaadin/flow-components/issues/8236


sissbruecker:
Tagged this as a behavior change, as this could modify the scroll position now whereas previously this method might have worked in single selection if the row was rendered but not visible.

Issue #2886, opened 2022-03-17
[map] Draw multi segment lines on the map


jgoderbauer:
### Describe your motivation

At the moment of writing this in the map component only a "point geometry" and a "Marker-Feature" are supported.

While adding Markers is a good start, more _geo Features_ are required to display more complex data on the map. 
A real life example for a multi line is to display a route to follow. While the route is calculated and optimized somewhere else a provided list of points can be used in the map to display the route.

Currently it is not possible to add a _MultiLineString_ via the flow api.

### Describe the solution you'd like

There should exist a high level, easy to use _MultiLineStringFeature_ for adding a line to the map (or better said: to a FeatureLayer) that is styleable (stroke).

Probably, there should also exist a Geometry for _MultiLineStringFeature_ and at least the ability to set and get the coordinates of all the points from start to end with all points inbetween. Another option would be to append _LineString_ objects to the _MultiLineString_.

### Describe alternatives you've considered

Make the ability to add any _geo Features_ available in flow by passing a _GeoJSON FeatureCollection_ .
This means less configuration possibilites but may be neccessary anyway.

### Additional context

The _MultiLineStringFeature_ is an extension of the _LineStringFeature_: basically it's a list of line strings.

Issue #2885, opened 2022-03-17
[map] Draw single segment lines on the map


jgoderbauer:
### Describe your motivation

At the moment of writing this in the map component only a "point geometry" and a "Marker-Feature" are supported.

While adding Markers is a good start, more _geo Features_ are required to display more complex data on the map. 
A real life example for a simple line could be for example to display the linear distance betweent two points on a map.

Currently it is not possible to add a _LineString_ via the flow api.

### Describe the solution you'd like

There should exist a high level, easy to use _LineStringFeature_ for adding a line to the map (or better said: to a FeatureLayer) that is styleable (stroke).

Probably, there should also exist a Geometry for _LineStrings_ and at least the ability to set and get the coordinates of the start and end point.

### Describe alternatives you've considered

Make the ability to add any _geo Features_ available in flow by passing a _GeoJSON FeatureCollection_ .
This means less configuration possibilites but may be neccessary anyway.

### Additional context

It is worth mentioning that OpenLayers supports also _MultiLineString_ which is not covered in this request.

PR #8208, opened 2025-10-31
feat: show lines on the map

sissbruecker:
## Description

Adds support for showing lines with one or more segments on the map.

Closes https://github.com/vaadin/flow-components/issues/2885
Closes https://github.com/vaadin/flow-components/issues/2886

## Type of change

- Feature

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -230,7 +230,7 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-compiler-plugin</artifactId>
-                    <version>3.14.1</version>
+                    <version>3.15.0</version>
                 </plugin>
                 <plugin>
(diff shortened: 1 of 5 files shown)


ugur-vaadin:
This does not seem to directly add MultiLineString support. Should we keep https://github.com/vaadin/flow-components/issues/2886 open or suggest people using multiple lines to achieve it?

sissbruecker:
I did re-read both tickets today and came to the conclusion that there was a misunderstanding what the purpose of LineString / MultiLineString is. In the ticket for LineString it only mentions single lines with start and end point, while it actually also allows creating paths with multiple segments, which is described as use-case in the ticket for MultiLineString. As such I figured it's fine to close both tickets.

Maybe there would be a use case for adding multiple line strings at once using MultiLineString but that's a different use-case than what's described in the ticket.

Issue #8202, opened 2025-10-29
map scale indicator


bsheely:
### Describe your motivation

Most maps have a scale indicator in one of the lower corners. Without a scale indicator, it's difficult for a user to determine the scale of the map as it zooms in or out. That's especially true as you zoom in or out over the open ocean. 

### Describe the solution you'd like

A simple graphical indicator (e.g., similar to what's shown below) that lets the user know that the distance represented by the indicator represents x number of (miles, km, nautical miles) that distance is on the map. 

It would be nice if the unit of measure was configurable. Obviously, the distance would change as the map is zoomed in or out.

### Additional context

<img width="342" height="54" alt="Image" src="https://github.com/user-attachments/assets/d9e1f85c-480d-4311-b3a5-3d03db40df83" />

DiegoCardoso:
Example of this feature in OpenLayers: https://openlayers.org/en/latest/examples/scale-line.html

bsheely:
Does that mean Vaadin has no interest in making this functionality part of the core map functionality?

DiegoCardoso:
Hi! I was just pointing to the functionality in OpenLayers as an internal note. In fact, there's already a draft PR related to this ticket: https://github.com/vaadin/flow-components/pull/8216

rolfsmeds:
This is planned for Vaadin 25.1 (planned for March 2026).

PR #8216, opened 2025-11-01
feat: configure visible UI controls on the Map

sissbruecker:
## Description

Allows to configure which UI controls (attributions, scale line, zoom buttons) are shown on the map.

Previously the web component provided a default set of controls which included attributions and zoom buttons. This provides a server-side configuration for the same controls and also adds a basic scale line control. For now only the visibility of each control can be configured, other configuration options can be added as a follow-up.

Part of #8202 

## Type of change

- Feature


vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/Configuration.java
```diff
@@ -95,6 +101,57 @@ public void removeLayer(Layer layer) {
...
+     * @param control
+     *            the control to be added
+     */
+    public void addControl(Control control) {
+        Objects.requireNonNull(control);
+
+        controls.add(control);
+        addChild(control);
+        control.addPropertyChangeListener(this.controlPropertyChangeListener);
+    }
```
DiegoCardoso, line 150: One thing I noticed is that if you add a new `Control` through this method of the same type as another one (e.g. `ZoomControl`), then it sort of duplicates that control:

https://github.com/user-attachments/assets/2e5c95cd-8f35-4bc1-ac00-57998d28ac63

I don't know if it would be possible to extend any of the current controls so that the user could replace any of the default ones with another, so it might not even be a problem. But what I was wondering is whether it makes sense or not for the controls list to hold only one control for each type. I might be overthinking it, though, but I wanted to raise this question.

sissbruecker: Probably not a concern at this point, but can't hurt to make this more defensive. Added respective logic to only allow one instance of a control of a certain type.

PR #8239, opened 2025-11-05
Create an actual Badget component

### LINKED ISSUE #5379: PiT 24.2: upgrade spring-petclinic-vaadin-flow to v24
### Describe the bug

The spring project [spring-petclinic-vaadin-flow](https://github.com/spring-petclinic/spring-petclinic-vaadin-flow) is still in version 22.0.0 which is not supported anymore, just changing the version to latest does not work because it needs adjustments in java version, spring libraries, database, java code etc.
Migration could be done in steps, first to 23.x and then to 24.x

### Expected-behavior

app is in latest version

### Reproduction

check version in the pom file https://github.com/spring-petclinic/spring-petclinic-vaadin-flow/blob/master/pom.xml#L24 

### System Info

vaadin 24.x

mstahv:
Fixes #5379

## Type of change

- [ ] Bugfix
- [x] Feature


sissbruecker:
This approach will not work for Aura or component base styles, as we have no badge classes there yet. Let's start with a web component implementation first and then add a Flow wrapper component.

PR #8241, opened 2025-11-06
refactor: replace UI.getCurrent() with UI.getCurrentOrThrow() where UI is required

### LINKED ISSUE #6757: Prefer accessing UI through attached elements over UI.getCurrent()
### Describe your motivation

Components sometimes need access to the UI instance, but there isn't a single standard way how to get that instance. One of the approaches widely used across the codebase is `UI.getCurrent()`. However, it's often neglected that this API isn't thread-safe and may return `null` in some contexts. One such example is when the component instance is created in a background thread without a UI lock:

```java
@Route("grid")
    public GridView() {
        Button addGrid = new Button("Add grid", (event) -> {
            final VaadinSession session = VaadinSession.getCurrent();

            new Thread(() -> session.accessSynchronously(this::render)).start();
        });

        add(addGrid);
    }

    public void render() {
        // UI isn't available in this context
        // ComponentRenderer is a class that relies on UI.getCurrent()
        grid.addColumn(new ComponentRenderer<Span, String>(item -> new Span(item)));
        grid.setItems("Item 1", "Item 2", "Item 3", "Item 4", "Item 5");
        add(grid);
    }
}
```

Then, in ComponentRenderer, a variable is set to an empty string if `UI.getCurrent()` is `null`. This prevents null pointer exceptions but still eventually causes the app to break elsewhere, which isn't the best solution:

https://github.com/vaadin/flow-components/blob/7249a90e014dc9d544b2cda531169c6e13255d5e/vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/java/com/vaadin/flow/data/renderer/ComponentRenderer.java#L146-L149

Here are similar examples from other modules:

https://github.com/vaadin/flow-components/blob/7249a90e014dc9d544b2cda531169c6e13255d5e/vaadin-renderer-flow-parent/vaadin-renderer-flow/src/main/java/com/vaadin/flow/data/renderer/LitRenderer.java#L83-L90

https://github.com/vaadin/flow-components/blob/7249a90e014dc9d544b2cda531169c6e13255d5e/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/Spreadsheet.java#L1181-L1184

https://github.com/vaadin/flow-components/blob/7249a90e014dc9d544b2cda531169c6e13255d5e/vaadin-charts-flow-parent/vaadin-charts-flow/src/main/java/com/vaadin/flow/component/charts/ChartOptions.java#L48-L57

https://github.com/vaadin/flow-components/blob/7249a90e014dc9d544b2cda531169c6e13255d5e/vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java#L1052-L1056

Instead of using `UI.getCurrent()`, methods should access UI through an attached component or element that is available in the current context. This is a thread-safe approach. Here are some ways to do this:

```java
if (isAttached()) {
  UI ui = getUI();

  // do something with UI from the component
}
```

```java
getElement().getNode().runWhenAttached(ui -> {
  // do something with UI once for this element
}); 
```

```java
element.addAttachListener(event -> {
  // attach event doesn't provide a UI reference for elements
  UI ui = ((StateTree) element.getNode().getOwner()).getUI();

  // do something with UI on every attach of this element
});

if (element.isAttached()) {
  UI ui = ((StateTree) element.getNode().getOwner()).getUI();

  // do something with UI if this element is already attached
}
```

However, if no attached element is available and hence the use of `UI.getCurrent()` is unavoidable, the method documentation should state this requirement and the null case should be handled properly with a clear exception like here:

https://github.com/vaadin/flow-components/blob/7249a90e014dc9d544b2cda531169c6e13255d5e/vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java#L762-L772

### Additional context

Originally discovered by @archiecobbs in https://github.com/vaadin/flow/issues/20240#issuecomment-2410952585

Related #6751

Artur-:
Replace UI.getCurrent() with the new UI.getCurrentOrThrow() API in locations where the code already expects a UI context to be present. This includes:

- Code that would NPE without UI (Dialog, Tooltip)
- Code with existing error handling for null UI (OverlayAutoAddController, ChartOptions, Map)
- Code called in UI lifecycle methods where UI is guaranteed to be present (GridPro, Dashboard, etc.)

The new getCurrentOrThrow() API provides better error messages automatically, eliminating the need for custom error handling and making the code cleaner.

Files with intentional fallback behavior (Notification, Popover, TooltipConfiguration, VirtualList, etc.) were kept with UI.getCurrent() and null checks as the fallback is part of their design.

Fixes #6757

### DIFF
--- a/vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/Map.java
+++ b/vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/Map.java
@@ -29,7 +29,6 @@ import com.vaadin.flow.component.map.configuration.source.OSMSource;
 import com.vaadin.flow.component.map.configuration.source.Source;
 import com.vaadin.flow.component.map.configuration.source.VectorSource;
 import com.vaadin.flow.component.map.configuration.source.XYZSource;
-import com.vaadin.flow.component.page.PendingJavaScriptResult;
 import com.vaadin.flow.internal.JacksonUtils;
 
 import tools.jackson.databind.node.ArrayNode;
@@ -112,7 +111,8 @@ public class Map extends MapBase {
      *            the user projection to use for all public facing API
      */
     public static void setUserProjection(String projection) {
-        executeJs("window.Vaadin.Flow.mapConnector.setUserProjection($0)",
+        UI.getCurrentOrThrow().getPage().executeJs(
+                "window.Vaadin.Flow.mapConnector.setUserProjection($0)",
                 projection);
     }
 
@@ -150,14 +150,9 @@ public class Map extends MapBase {
      */
     public static void defineProjection(String projectionName,
             String wksDefinition) {
-        executeJs("window.Vaadin.Flow.mapConnector.defineProjection($0, $1)",
+        UI.getCurrentOrThrow().getPage().executeJs(
+                "window.Vaadin.Flow.mapConnector.defineProjection($0, $1)",
                 projectionName, wksDefinition);
-    }
-
-    private static PendingJavaScriptResult executeJs(String expression,
-            Object... parameters) {
-        return UI.getCurrentOrThrow().getPage().executeJs(expression,
-                parameters);
     }
 
     public Map() {


vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/Map.java
```diff
@@ -111,16 +112,7 @@ public class Map extends MapBase {
...
-        if (ui == null || ui.getPage() == null) {
-            throw new IllegalStateException("UI instance is not available. "
-                    + "It means that you are calling this method "
-                    + "out of a normal workflow where it's always implicitly set. "
-                    + "That may happen if you call the method from the custom thread without "
-                    + "'UI::access' or from tests without proper initialization.");
-        }
-        UI.getCurrent().getPage().executeJs(
-                "window.Vaadin.Flow.mapConnector.setUserProjection($0)",
+        executeJs("window.Vaadin.Flow.mapConnector.setUserProjection($0)",
```
vursen: IMO, the code looks clearer without extracting this into a helper:

```suggestion
        UI.getCurrentOrThrow().getPage().executeJs("window.Vaadin.Flow.mapConnector.setUserProjection($0)",
```

Artur-: Updated

PR #8242, opened 2025-11-06
refactor!: change Upload to use raw instead of multipart requests

Artur-:
Change the Upload component to send files as "raw" requests with the file contents as body, instead of using multipart requests.

### DIFF
--- a/vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/test/java/com/vaadin/flow/component/upload/tests/UploadIT.java
+++ b/vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/test/java/com/vaadin/flow/component/upload/tests/UploadIT.java
@@ -51,7 +51,7 @@ public class UploadIT extends AbstractUploadIT {
 
     @Test
     public void testUploadAnyFile() throws Exception {
-        File tempFile = createTempFile("txt");
+        File tempFile = createTempFile("some file", "txt");
         getUpload().upload(tempFile);
 
         String content = uploadOutput.getText();
@@ -63,20 +63,6 @@ public class UploadIT extends AbstractUploadIT {
                 content.contains(expectedContents));
         Assert.assertTrue("Progress update event was not fired properly",
                 content.contains("PROGRESS:" + tempFile.getName()));
-    }
-
-    @Test
-    public void testUploadFileWithUrlEncodedFileName() throws Exception {
-        File tempFile = createTempFile("religion åk4", "txt");
-        getUpload().upload(tempFile);
-
-        String content = uploadOutput.getText();
-
-        String expectedContents = tempFile.getName() + "text/plain"
-                + getTempFileContents();
-
-        Assert.assertTrue("Upload content does not contain file details",
-                content.contains(expectedContents));
     }
 
     @Test
--- a/vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/test/java/com/vaadin/flow/component/upload/tests/UploadWithHandlerIT.java
+++ b/vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/test/java/com/vaadin/flow/component/upload/tests/UploadWithHandlerIT.java
@@ -52,7 +52,7 @@ public class UploadWithHandlerIT extends AbstractUploadIT {
     @Test
     public void uploadWithHandler_uploadTextFile_fileUploadedAndNoErrorThrown()
             throws Exception {
-        File tempFile = createTempFile("txt");
+        File tempFile = createTempFile("some file", "txt");
         upload.upload(tempFile);
 
         checkLogsForErrors();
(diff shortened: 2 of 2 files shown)


manolo (dismissed):
LGTM

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
sissbruecker, line 1: Did you test the Flow component with the respective web component changes? Do download handler and stream resource variable work out of the box with just switching from multi-part to raw, or does that require changes? Do all the events and constraints work with the change of the default?

Artur-: XHR upload support has apparently been implement a long time ago. https://github.com/vaadin/flow/pull/22661 added support for using X-Filename and Content-Type

sissbruecker: In any case, we should probably release the web component change first to make sure everything passes in CI.

sissbruecker: Locally I get a number of test failures at least, so those could be checked already:

Bildschirmfoto 2025-11-07 um 10 19 56

Otherwise I'll check the WC PR and then release all the changes so that we can verify in CI.

Artur-: Why aren't those tests ran for this PR? Everything looks green

Artur-: Right, it uses the old web component so it happens to work as these changes have no effect

sissbruecker: I updated the buffer and download handler ITs to verify the file name and content type from the metadata.

sissbruecker: We should also add ITs for verifying file contents / metadata using multi-part format.

sissbruecker: WC update is in https://github.com/vaadin/flow-components/pull/8246. I had to restore the `multipart` format for tests to pass (https://github.com/vaadin/flow-components/pull/8246/commits/bbf97f616a22f82dfb30bc3d9929f1b7acea4046), so that needs to be reverted here.

sissbruecker, line 1: All of these would probably be good to keep tested for multipart:
- UploadIT.testUploadAnyFile
- UploadIT.testUploadFileWithUrlEncodedFileName
- UploadIT.testUploadMultipleEventOrder
- UploadIT.testUploadEventOrder
- UploadIT.uploadInvalidFile_fileIsRejected
- UploadIT.uploadFileAndNoErrorThrown
- UploadWithHandlerIT.uploadWithHandler_uploadTextFile_fileUploadedAndNoErrorThrown
- UploadWithHandlerIT.uploadWithHandler_uploadFileWithUrlEncodedFileName_fileNameIsDecoded
- UploadWithHandlerIT.uploadWithHandler_withLargeBinaryFile_fileUploadedAndNoErrorThrown

There's probably some potential for making a saner test structure with the variations of raw/multipart, buffers / download handlers, but that could be done as a follow up.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -78,6 +78,55 @@ public UploadIcon() {
...
     }
 
+    /**
+     * Defines the upload transmission format for files.
+     */
+    public enum UploadFormat {
+        /**
+         * Send files as raw binary data with Content-Type header set to the
+         * file's MIME type. This is the default format, which uses
+         * XMLHttpRequest (XHR) for file transmission.
```
sissbruecker: Both formats use XMLHttpRequest? I'm also not sure why we need to mention the headers, how does that matter for Flow developers?

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -370,6 +419,36 @@ public List<String> getAcceptedFileTypes() {
...
+     * format.
+     * <p>
+     * {@link UploadFormat#MULTIPART} sends files as multipart form data with
+     * Content-Disposition header. This is the traditional format used for HTML
+     * form submissions.
+     *
+     * @param uploadFormat
+     *            the upload format to use, not {@code null}
+     */
+    public void setUploadFormat(UploadFormat uploadFormat) {
```
sissbruecker: You might also want to leave an explanation on why someone would want to switch back to multipart. When is this necessary?

Artur-: The question is very good. I have no explanation. Should we just remove the API?

Artur-: i.e. just drop the whole PR except for some needed test changes

Artur-: Removed all API

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
Artur-: Flow needs another PR in order for this to work

Artur-: This https://github.com/vaadin/flow/pull/22690

sissbruecker (approved):

PR #8253, opened 2025-11-11
fix: Fix Action constructor with Icon and add note about HTML content

TatuLund:


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/java/com/vaadin/flow/component/spreadsheet/framework/Action.java
```diff
@@ -37,29 +40,42 @@ public class Action implements Serializable {
...
      * @param icon
      *            the icon for the new action.
      */
     public Action(String caption, Icon icon) {
-        this.caption = caption;
+        caption = caption == null ? ""
+                : Jsoup.clean(caption, Safelist.relaxed());
+        this.caption = String.format(
+                "<vaadin-icon icon=\"%s\"></vaadin-icon> %s", icon.getIcon(),
+                caption);
```
DiegoCardoso, line 70: The issue with this approach is that it overrides the content created by the developer, which makes `Action#getCaption()` return a different value from the one provided.

DiegoCardoso:
Closing in favor of #8285 and #8303

PR #8263, opened 2025-11-18
chore: bump FW8 version to 8.29.0 in V24

### LINKED ISSUE #8155: Spreadsheet custom context menu does not come up on iOS
### Description

we use a custom context menu for our implementation of Spreadsheet.  It works fine on Windows/MacOS/Android, popping up the context menu with a long-press or right click.  However, on an iPad in Chrome, Firefox, and Safari, it doesn't.

It also works with a long-press in Chrome/Edge/Firefox on Windows using Dev Tools to emulate an iPad.  We only see the issue on actual iPad devices.

You will also still see the "Fill" overlay item, which is no longer replaced by the custom context menu in Flow, as it was in V8.  That's filed separately.

### Expected outcome

long press on an item should open the custom context menu on iOS like it does on Android.

any Vaadin starter app with a spreadsheet and a basic custom context menu, even just a "Hello World" notification on selection item.

### Steps to reproduce

open the app on an iPad, in Chrome or Safari or Firefox. Long-press a cell.  There won't be a popup context menu, but there will be on other platforms.

### Environment

Vaadin version(s): 24.8
OS: iPadOS 17

### Browsers

Firefox, Chrome, Safari on iOS

DiegoCardoso:
## Description

A fix for iPad device detection has been done in Framework 8, which is needed to make context menus in Spreadsheet work on long press. 

> [!NOTE]
> This version has already been updated in V25 by @dependadot.

Fixes #8155

## Type of change

- Bugfix

### DIFF
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/pom.xml
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/pom.xml
@@ -38,7 +38,7 @@
         <!-- we dont put this dependencies in parent poms because they are only
          used in this module just for the compilation Java to JS -->
         <gwt.version>2.9.0</gwt.version>
-        <framework.version>8.27.7</framework.version>
+        <framework.version>8.29.0</framework.version>
     </properties>
 
     <dependencyManagement>


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/pom.xml
```diff
@@ -38,7 +38,7 @@
         <!-- we dont put this dependencies in parent poms because they are only
          used in this module just for the compilation Java to JS -->
         <gwt.version>2.9.0</gwt.version>
-        <framework.version>8.27.7</framework.version>
+        <framework.version>8.29.0</framework.version>
```
tomivirkki, line 41: I wasn't able to reproduce the issue in `24.9` branch using Simulator. I wonder if it only occurs on a physical device?

https://github.com/user-attachments/assets/f57f0abb-2225-489f-a383-1df4c17f04a7

DiegoCardoso: Can you try it on an iPad? It works fine with iPhone, but the issue is with iPads, actually.

DiegoCardoso: (you can test it on simulator as well)

tomivirkki: Ah, I missed it's only an iPad issue. Was able to repro and confirmed this bump fixes the issue.

PR #8271, opened 2025-11-19
refactor!: replace commons-lang3 usage with raw java code

mstahv:
## Description

The only known usage for commons-lang3 these days in Vaadin, so we can get rid of a pretty much obsolete dependency.

Don't know why equals/hashCode has been overridden in the first place for that class, but the re-writtten part ought to be working in pretty much identical manner.

## Type of change

- [x] Bugfix
- [ ] Feature

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/pom.xml
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/pom.xml
@@ -62,21 +62,12 @@
             <artifactId>slf4j-simple</artifactId>
             <scope>test</scope>
         </dependency>
-        <!-- Spring data dependency to be able to provide Spring Data API-->
+        <!-- Optional Spring dependencies to be able to provide Spring Data API-->
         <dependency>
             <groupId>org.springframework.data</groupId>
             <artifactId>spring-data-commons</artifactId>
             <version>${spring-data-commons.version}</version>
-            <exclusions>
-                <exclusion>
-                    <groupId>org.springframework</groupId>
-                    <artifactId>spring-core</artifactId>
-                </exclusion>
-                <exclusion>
-                    <groupId>org.springframework</groupId>
-                    <artifactId>spring-beans</artifactId>
-                </exclusion>
-            </exclusions>
+            <optional>true</optional>
         </dependency>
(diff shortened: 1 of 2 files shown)


yuriy-fix:
We can proceed with the change if `sort_event_not_sent_for_same_data` still passes.
Additionally, we could ask for an insight/review from @caalador based on the initial change done 8 years ago: https://github.com/vaadin/flow-components/commit/920f1099f64e9a8f3b2ca3b7c4457ed0715f868f
 😅

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/GridSortOrder.java
```diff
@@ -91,7 +89,6 @@ public boolean equals(Object obj) {
 
     @Override
     public int hashCode() {
-        return new HashCodeBuilder().append(getSorted()).append(getDirection())
-                .toHashCode();
+        return getSorted().hashCode() + getDirection().hashCode();
```
Legioth: Not blocking, but just pointing out that the typical way in modern Java would be `return Objects.hash(getSorted(), getDirection());`

caalador:
Looking at 8 years ago it is used because it was there (`HashCodeBuilder`)  and probably used as a sort of default for hashCode at the time.

Artur-:
Should we change the tests also that use `org.apache.commons.lang3`

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/GridSortOrder.java
mstahv: 👍 Updated.

mstahv:
> Should we change the tests also that use `org.apache.commons.lang3`

Like generally or is it used here somewhere still? Or did removing the managed version break some other module? I think yes, whenever we work on code where we could replace external library usage with modern JDK/Java features👍 Might even make sense to give Claude a job at some point. But just updating test code manually because we can - maybe not 🤷‍♂️

PR #8287, opened 2025-11-24
fix: add popover to UI as fallback instead of current UI modal

sissbruecker:
## Description

Currently a popover adds itself to the current UI modal component as a fallback when the target component is not within a modal component itself. As shown in https://github.com/vaadin/flow-components/issues/8286 that can be a problem if the target is outside of that modal, in which case removing the current modal will also remove the popover even though the target is still part of the UI.

This changes it so that the popover adds itself to the UI itself instead, so that it will still exist even if the current UI modal is removed.

Fixes https://github.com/vaadin/flow-components/issues/8286

## Type of change

- Bugfix


DiegoCardoso (approved):
LGTM

PR #8289, opened 2025-11-25
fix: refresh editors on relayout complete

### LINKED ISSUE #8274: Spreadsheet scrolling hangs custom editors and loses track of formatting
### Description

Scrolling a spreadsheet with two custom editors in same row, causes formatting inconsistencies and breaks custom editors. The problem appears only when there are two or more custom editors in the same column.  

As shown in the attached video, attempting to open the lower combo box after scrolling, when not initially opened before seems to  hang the fetching of the items. Also, formating is lost in merged cells.

Image

### Expected outcome

- Cell formatting should remain consistent after scrolling.
- Custom editors should open normally after scrolling.
- No JavaScript errors should occur when scrolling.

[spreadsheet-scroll-with-editors-bug.zip](https://github.com/user-attachments/files/23631238/spreadsheet-scroll-with-editors-bug.zip)

### Steps to reproduce

1. Unzip the test project, run the application, open the view and don't open the comboboxes in the view.
2.  Scroll down and then scroll up.
3. Attempt to open a ComboBox that was not opened before scrolling and observe how it hangs while trying to load the items.

### Environment

Vaadin version(s): 24.9.5
OS: Tested on Windows 11

### Browsers

_No response_

ugur-vaadin:
## Description

Custom editors in Spreadsheet cells become non-functional after scrolling the sheet away from and back to the cells containing them. This leads to editors hanging in "loading" state when attempting to interact with them after scrolling. After scrolling, the elements are recreated with new node IDs, but the Spreadsheet still holds references to the old node IDs.

This fix implements a two-part solution:
  - Modifies `Spreadsheet.loadCustomComponents()` to preserve custom editor registrations even when cells scroll out of view
  - Adds `onSheetRelayoutComplete()` to `SheetHandler` interface. `SheetWidget` calls `onSheetRelayoutComplete()` after completing the relayout. `SpreadsheetWidget` implements it to refresh custom editors, storing and using the last `cellKeysToEditorIdMap`.
- Ensures editors are re-attached on scroll back. Skips the editors for cells not currently visible.

Fixes #8274 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/SpreadsheetWidget.java
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/SpreadsheetWidget.java
@@ -473,7 +473,8 @@ public class SpreadsheetWidget extends Composite implements SheetHandler,
             HashMap<String, String> cellKeysToEditorIdMap) {
 
         // Store reference for use in onSheetRelayoutComplete
-        this.lastCellKeysToEditorIdMap = cellKeysToEditorIdMap;
+        this.lastCellKeysToEditorIdMap = cellKeysToEditorIdMap == null ? null
+                : new HashMap<>(cellKeysToEditorIdMap);
 
         if (cellKeysToEditorIdMap == null || customEditorFactory == null) {
             return;


WoozyG:
Does this commit also fix the lost cell formatting on scroll that was also part of the bug report?

ugur-vaadin:
The client-side errors which cause the weird behaviour should be fixed with this PR. I tried the fix with MacOS and Windows 11 using the provided page setup and the spreadsheet, and it seems to work fine. If there are still issues reproducible in other environments, I can take another look and provide further assistance.

Issue #8288, opened 2025-11-24
`Action(String, Icon)` constructor doesn't work in Spreadsheet


DiegoCardoso:
### Description

Creating an `Action` using the `Action(String caption, Icon icon)` constructor doesn't work as expected, as the `Icon` instance is not created in the context menu of the Spreadsheet.

### Expected outcome

The constructor should work as expected.

### Minimal reproducible example

In `ActionFixture`, modify the `getActions` method and make it use the constructor with `Icon`:

```java
 return new Action[] { new Action("Double cell values",
          LumoIcon.ANGLE_UP.create()), };
```

### Steps to reproduce

1. Modify the file mentioned above
2. Run the `Spreadsheet` IT module with `mvn jetty:run`
3. Open http://localhost:8080/vaadin-spreadsheet
4. Click on the "Create new" button
5. Load the "Action" fixture
6. Right-click on any cell

### Environment

Vaadin version(s): 24+

### Browsers

Issue is not browser related

PR #8303, opened 2025-11-27
fix: make `Icon` work when used in `Action`

DiegoCardoso:
## Description

This change fixes the issue when adding an icon to an action in the context menu in Spreadsheet. To make it work, the following modifications were made:
- Add the `Icon` instance passed to the `Action(String, Icon)` as a virtual child to `Spreadsheet`
- Passes the node id of the element to the client side to be fetched
- Modifies the `SpreadsheetAction` class to overwrite the `getHTML` method in order to add a container for the icon, in case the action has one
- Attaches the icon instance to the container
- Creates an event to be dispached when the context menu is closed
- Listens to the event to make sure that any `Icon` instance is safely removed from the client to avoid memory leaks

Fixes https://github.com/vaadin/flow-components/issues/8288

## Type of change

- Bugfix

### DIFF
--- a/vaadin-accordion-flow-parent/vaadin-accordion-flow/src/main/java/com/vaadin/flow/component/accordion/Accordion.java
+++ b/vaadin-accordion-flow-parent/vaadin-accordion-flow/src/main/java/com/vaadin/flow/component/accordion/Accordion.java
@@ -52,7 +52,7 @@ import com.vaadin.flow.shared.Registration;
  * @author Vaadin Ltd
  */
 @Tag("vaadin-accordion")
-@NpmPackage(value = "@vaadin/accordion", version = "25.0.0-beta5")
+@NpmPackage(value = "@vaadin/accordion", version = "25.0.0-beta6")
 @JsModule("@vaadin/accordion/src/vaadin-accordion.js")
 public class Accordion extends Component implements HasSize, HasStyle {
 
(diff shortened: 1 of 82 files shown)


ugur-vaadin (approved):

vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/SpreadsheetAction.java
```diff
@@ -48,4 +54,21 @@ public void execute() {
...
     }
 
+    /**
+     * Overriding the {@link Action#getHTML()} to add a container for the icon
+     * if present.
+     */
+    @Override
+    public String getHTML() {
+        StringBuilder sb = new StringBuilder();
+        sb.append("<div>");
```
tomivirkki: In V8 SpreadSheet, the `v-icon` classname would give the icon 5px right margin, which results in a small gap between the icon and the text. But since `v-icon` probably isn't ideal in this case (now that we're using a `<vaadin-icon>`), we could instead make this wrapper use inline styles `<div style=\"display:flex;align-items:baseline;gap:5px;\">`.

DiegoCardoso: Thanks for the suggestion. I had it in mind to check how it looks in V8, but it slipped my mind.

Here's the before/after:
#### Before
image

#### After
image

It looks better this way.

Issue #8315, opened 2025-12-02
Calling Dialog::removeAll removes header and footer


knoobie:
### Description

Title says it all. Once `Dialog::removeAll`is called the header and footer of a dialog is gone.

### Expected outcome

`Dialog::removeAll` should NOT interfere with the header and footer feature of the dialog.

### Minimal reproducible example

```java
new Button("Open Dialog", e -> {
      AtomicReference<Dialog> reference = new AtomicReference<>();
      var dialog = new Dialog();
      reference.set(dialog);
      dialog.getFooter().add(new Button("Click me after clear", ex -> {
        System.out.println("Was I called? :(");
      }));
      dialog.add(new Button("Click me to clear dialog", event -> {
        reference.get().removeAll();
        reference.get().add(new Paragraph("I was cleared!"));
      }));
      dialog.open();
    })
```

### Steps to reproduce

1. Click "Open Dialog"
2. Click "Click me to clear the dialog"
3. Expectation: I can still click "Click me to clear dialog"

### Environment

Vaadin version(s): 25.0-beta6

Legioth:
`removeAll()` is a general method that is defined to remove _all_ children, i.e. all components that have the same parent. What might be needed is that `Dialog` would have a separate method to remove all content, excluding headers and footers.

knoobie:
Yeah that was implement for the ConfirmDialog, but might be forgotten for the Dialog when it was refactored for V25. In V24 it worked (after it was broken - see linked issue)

sissbruecker:
In v25 we had to change the `vaadin-dialog` web component to slot a `vaadin-dialog-content` element into itself (https://github.com/vaadin/web-components/pull/10367), into which all content is then rendered. `removeAll` apparently results in a special RPC call that just clears `innerHTML` or something similar, so the content element created by the web component is removed and after that all content updates are broken.

So we probably need a special implementation for `removeAll` that just removes children individually.

Legioth:
Please preserve this invariant. Otherwise, things might become quite confusing.

```java
someComponent.removeAll(); 
assertEquals(0, someComponent.getChildren().count());
```

PR #8319, opened 2025-12-03
fix: make Dialog.removeAll preserve slotted content elements

sissbruecker:
## Description

In v25 we had to change the `vaadin-dialog` web component to slot a `vaadin-dialog-content` element into itself (https://github.com/vaadin/web-components/pull/10367), into which all content is then rendered. `removeAll` apparently results in a special RPC call that just clears `innerHTML` or something similar, so the content element created by the web component is removed and after that all content updates are broken.

This overrides `removeAll` to remove all children individually.

Fixes https://github.com/vaadin/flow-components/issues/8315

## Type of change

- Bugfix

### DIFF
--- a/vaadin-accordion-flow-parent/vaadin-accordion-flow/src/main/java/com/vaadin/flow/component/accordion/Accordion.java
+++ b/vaadin-accordion-flow-parent/vaadin-accordion-flow/src/main/java/com/vaadin/flow/component/accordion/Accordion.java
@@ -52,7 +52,7 @@ import com.vaadin.flow.shared.Registration;
  * @author Vaadin Ltd
  */
 @Tag("vaadin-accordion")
-@NpmPackage(value = "@vaadin/accordion", version = "25.0.0-beta6")
+@NpmPackage(value = "@vaadin/accordion", version = "25.0.0-beta7")
 @JsModule("@vaadin/accordion/src/vaadin-accordion.js")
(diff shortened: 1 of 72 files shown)


web-padawan:
Let's wait for https://github.com/vaadin/flow-components/pull/8318 before merging this PR to avoid re-running longer build.

knoobie:
Thanks for the quick fix! You might wanna look at the old test and review or remove it - if it does not create a benefit https://github.com/vaadin/flow-components/blob/d57f23f1edb8b3433c5e36f641770f7b4cbc6522/vaadin-dialog-flow-parent/vaadin-dialog-flow/src/test/java/com/vaadin/flow/component/dialog/DialogChildrenTest.java#L191

sissbruecker:
> Thanks for the quick fix! You might wanna look at the old test and review or remove it - if it does not create a benefit
> 
> https://github.com/vaadin/flow-components/blob/d57f23f1edb8b3433c5e36f641770f7b4cbc6522/vaadin-dialog-flow-parent/vaadin-dialog-flow/src/test/java/com/vaadin/flow/component/dialog/DialogChildrenTest.java#L191

Thanks for the heads-up. I think it's fine to keep the unit test as well, it's quick to run and verifies that the server UI structure is in a proper state. However for this specific bug we additionally need an IT to verify that the client-side element contents are preserved.

PR #8269, opened 2025-11-18
fix: ignore unintentional grid-pro item property changes

tomivirkki:
## Description

Fixes an issue where `ItemPropertyChangedEvent` was incorrectly fired when an editable cell was opened and closed without making any actual changes to the value. This caused unnecessary event handling and potential unintended side effects in applications using Grid Pro.

Fixes https://github.com/vaadin/flow-components/issues/5647

## Type of change

Bugfix

PR #8323, opened 2025-12-04
fix: properly toggle grid-pro updating cell state

tomivirkki:
## Description

Parameter order for the `updatePart` helper in web-components was changed for V25 [here](https://github.com/vaadin/web-components/pull/10481). Turns out we didn't update the order in [gridProConnector](https://github.com/vaadin/flow-components/blob/dd82f2d5829610136129a4b3da46ac3230445c81/vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/resources/META-INF/resources/frontend/gridProConnector.js#L116) to align with this change.

Updating the parameter order reveals a regression from #8269 where the "updating-cell" part name would not get [cleared](https://github.com/vaadin/flow-components/blob/dd82f2d5829610136129a4b3da46ac3230445c81/vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/resources/META-INF/resources/frontend/gridProConnector.js#L121-L125) for unchanged custom editors:

https://github.com/user-attachments/assets/75db24c2-dbb4-4c70-90b0-7ede448ff5ee

To fix the issue, this PR, in addition to updating the parameter order, also moves the `clearUpdatingCell` call before the [value change checks](https://github.com/vaadin/flow-components/pull/8269/files#diff-04265121fef8b2443330d9718cd7ad2bc16c11e3cc8fc161704cf6c9c7fc0e1cR714-R739) added in #8269, to make sure the "updating-cell" part name is properly cleared from cells even when there's no value change.

## Type of change

Bugfix

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -70,7 +70,7 @@
     </scm>
     <properties>
         <flow.version>25.0-SNAPSHOT</flow.version>
-        <testbench.version>10.0.0-beta5</testbench.version>
+        <testbench.version>10.0.0-beta6</testbench.version>
         <maven.compiler.release>21</maven.compiler.release>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/pom.xml
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/pom.xml
@@ -27,7 +27,7 @@
         <dependency>
             <groupId>com.vaadin</groupId>
             <artifactId>license-checker</artifactId>
-            <version>3.0.0-beta1</version>
+            <version>3.0.0-beta2</version>
             <scope>provided</scope>
         </dependency>
         <dependency>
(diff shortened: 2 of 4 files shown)


vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow-integration-tests/src/test/java/com/vaadin/flow/component/gridpro/tests/BasicIT.java
```diff
@@ -432,6 +433,12 @@ public void comboBoxColumn_openAndCloseEditorWithoutChange_valueUnchangedAndNoEv
         Assert.assertEquals(
                 "Cell value should remain unchanged after opening and closing editor without changes",
                 initialValue, cell.getText());
+        // Assert there's no element with 'updating-cell' part
```
ugur-vaadin, line 436: This assertion seems to be valid even without the fix for me.

tomivirkki: It only becomes an issue when the parameter order is fixed in the connector. This test ensures the `clearUpdatingCell` call is also relocated.

ugur-vaadin: Oh, right.

PR #8349, opened 2025-12-09
refactor: remove Card content wrapper

sissbruecker:
## Description

Removes the content wrapper element from `Card`, and instead makes `HasComponents` APIs only operate on components in the default slot.

Closes https://github.com/vaadin/web-components/issues/10612

## Type of change

- Refactor

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,7 @@
         <url>https://github.com/vaadin/flow-components</url>
     </scm>
     <properties>
-        <flow.version>25.0-SNAPSHOT</flow.version>
+        <flow.version>25.1-SNAPSHOT</flow.version>
         <testbench.version>10.0.0</testbench.version>
         <maven.compiler.release>21</maven.compiler.release>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>


vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
```diff
@@ -310,63 +304,13 @@ public Component[] getFooterComponents() {
...
-            }
-        }
-        toRemove.stream().map(Component::getElement)
-                .forEach(contentRoot::removeChild);
-        if (contentRoot != null && contentRoot.getChildCount() == 0) {
-            getElement().removeChild(contentRoot);
-            contentRoot = null;
-        }
+        return super.getChildren().filter(
+                child -> child.getElement().getAttribute("slot") == null);
```
tomivirkki:
```suggestion
        return super.getChildren()
                .filter(child -> !child.getElement().hasAttribute("slot"));
```

vaadin-card-flow-parent/vaadin-card-testbench/src/main/java/com/vaadin/flow/component/card/testbench/CardElement.java
```diff
@@ -35,11 +34,7 @@ public class CardElement extends TestBenchElement {
      * @return the content elements
      */
     public List<TestBenchElement> getContents() {
-        return findElements(By.cssSelector("div:not([slot])")).stream()
-                .findFirst()
-                .map(wrapper -> ((TestBenchElement) wrapper)
-                        .getPropertyElements("children"))
-                .orElse(Collections.emptyList());
+        return $("*").withoutAttribute("slot").all();
```
tomivirkki: This also queries from within the shadow and returns 13 elements on the default CardPage.

sissbruecker: Good catch. Reverted to use `findElements` and updated the assertions to check that the correct content for each slot is returned.

Issue #8350, opened 2025-12-10
BigDecimalField has incorrect label styling


peholmst:
### Description

When using Lumo and Vaadin 25, the label of `BigDecimalField` is not correctly styled. It has normal text size and normal text color.

### Expected outcome

The label should have the same text size and color as the other fields.

### Minimal reproducible example

```java
var discountField = new BigDecimalField("Discount");
```

### Steps to reproduce

Create a new Vaadin 25 project with the Lumo theme, add a `BigDecimalField` with a non-empty label.

### Environment

Vaadin version(s): 25.0.0-beta9
OS: macOS 26.1

### Browsers

Chrome

PR #8351, opened 2025-12-10
fix: ensure Lumo CSS is injected into BigDecimalField

web-padawan:
## Description

Fixes https://github.com/vaadin/flow-components/issues/8350

## Type of change

- Bugfix


vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/resources/META-INF/resources/frontend/vaadin-big-decimal-field.js
```diff
@@ -22,6 +22,10 @@ class BigDecimalField extends TextField {
     return 'vaadin-big-decimal-field';
   }
 
+  static get lumoInjector() {
+    return { is: 'vaadin-text-field' }
```
vursen:
```suggestion
    return { ...super.lumoInjector, is: 'vaadin-text-field' };
```

PR #8359, opened 2025-12-12
Grid: Add support for setting a filtered data provider

peholmst:
Fixes https://github.com/vaadin/flow-components/issues/3842.

This is my outcome from the AI workshop. Claude wrote all the code according to my specifications. This PR:

* Introduces a new set of methods, `setItemsWithFilter` that accept a fetch callback, a fetch/count callback pair, and a backend data provider, respectively, with a non-void filter type.
* Introduces a new `GridLazyFilterDataView` type that makes it possible to set a filter.

This maintains backwards compatibility with the existing methods and types without introducing new generic type parameters.

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
@@ -2844,8 +2844,8 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
      * GridLazyFilterDataView&lt;Person, PersonFilter&gt; dataView = grid
      *         .setItemsWithFilter(query -&gt; {
      *             PersonFilter filter = query.getFilter().orElse(null);
-     *             return personService.fetch(query.getOffset(), query.getLimit(),
-     *                     filter);
+     *             return personService.fetch(query.getOffset(),
+     *                     query.getLimit(), filter);
      *         });
      *
      * // Set a filter - automatically refreshes the grid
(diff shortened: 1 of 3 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
sissbruecker, line 1: I'm trying to figure out what the intended benefit here is, or how much of a benefit compared to adding to the API confusion around `setItems` / `setItemsPageable` / `setDataProvider`.

So for example, how is this approach better than this:
```java
var filterableDataProvider = new CallbackDataProvider<String, String>(query -> Stream.empty(), query -> 0)
        .withConfigurableFilter();

grid.setItems(filterableDataProvider);

filterableDataProvider.setFilter("test");
```

peholmst: The approach you suggested requires a count callback. If I want to use just a fetch callback, I'd have to write:
```java
var filterableDataProvider = new CallbackDataProvider<String, String>(
  query -> Stream.empty(),
  query -> { throw new UnsupportedOperationException(""); }
).withConfigurableFilter();

grid.setItems(filterableDataProvider);
grid.getLazyDataView().setItemCountUnknown();
```

sissbruecker: I feel adding another method for setting a data provider for this use case is not a good trade-off, especially since the method name does not really imply that it's only useful if you don't want to use a count callback. To me it seems that there is a missing data provider implementation that works without a count callback and supports a filter. Having that might also allow auto-configuring the data view / data communicator.

Maybe a documentation example would be a better start.

Issue #8257, opened 2025-11-14
[Dashboard] Confirm dialog before deleting a widget


ShanMerritt:
### Describe your motivation

When editing a dashboard sometimes a user clicks on the widget's X icon accidentally and then they have to recreate the widget which can be frustrating. 

### Describe the solution you'd like

It would be nice to have the possibility to add a confirmation dialog before removing, like the typical "Please confirm. Are you sure you want to remove this widget? Yes/No".

### Describe alternatives you've considered

An alternative would be to reload the dashboard without saving the changes (undo) but that would mean all changes are lost.

yuriy-fix:
@tomivirkki , could you please check if there is a quick workaround for this?

tomivirkki:
Hi @ShanMerritt, here's one possible workaround:

```java
widget.getElement().addEventListener("item-remove", e -> {
    var confirm = new ConfirmDialog();
    confirm.setHeader("Please confirm");
    confirm.setText("Are you sure you want to remove this widget?");
    confirm.setCancelable(true);
    confirm.setConfirmText("Remove");
    confirm.addConfirmListener(event -> {
        dashboard.remove(widget);
    });
    confirm.open();
}).addEventData("event.stopPropagation()");
```

https://github.com/user-attachments/assets/cd395fd9-8b32-43ba-815b-b89deb7c4912

ShanMerritt:
Thanks @tomivirkki @yuriy-fix for the quick workaround, I tried it and it works for me as well.

rolfsmeds:
Good that we have a workaround. Would still be good to have a before-remove event for custom logic.

PR #8360, opened 2025-12-12
feat: dashboard item remove handler

sissbruecker:
## Description

Allows setting an item remove handler in Dashboard, which gives manual control over item removal, for example to show a confirmation dialog.

- Dashboard now always prevents automatic removal on the client-side and handles the removal logic on the server
- Without a remove handler, items are automatically removed (as before)
- With a remove handler items are not automatically removed
- The handler receives an event that provides:
  - The item to be removed
  - Optionally, the section that the item is in
  - A method for removing the item

Example usage with a ConfirmDialog:
```java
dashboard.setItemRemoveHandler(removeEvent -> {
    ConfirmDialog dialog = new ConfirmDialog();
    dialog.addConfirmListener(confirmEvent -> removeEvent.removeItem());
    dialog.open();
});
```

Part of https://github.com/vaadin/flow-components/issues/8257
Depends on https://github.com/vaadin/web-components/pull/10623

## Type of change

- Feature


vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/Dashboard.java
```diff
@@ -430,6 +432,47 @@ public Registration addItemRemovedListener(
...
+     * });
+     * </pre>
+     *
+     * @param handler
+     *            the handler to set, or {@code null} to restore default
+     *            behavior
+     * @see DashboardItemRemoveHandler
+     * @see DashboardItemRemoveEvent
+     */
+    public void setItemRemoveHandler(DashboardItemRemoveHandler handler) {
```
sissbruecker, line 494: We have a similar feature in `Dialog` where `addCloseActionListener` prevents a dialog from automatically closing. However the design and implementation seem overcomplicated (allows multiple listeners, keeps track of listener count). I think just being able to set a handler and prevent the default behavior based on the presence of that handler is more straightforward.

Still providing an event to the handler seems like it could be beneficial. But we could also just pass the item to the handler, and, regardless whether it's a widget or section, removing it would just be `item.removeFromParent()`.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/Dashboard.java
```diff
@@ -754,21 +797,47 @@ private void handleItemResizedClientEvent(DomEvent e, String idKey,
...
-        getElement().addEventListener("dashboard-item-removed", e -> {
+        String sectionKey = "event.detail.section?.id";
+        getElement().addEventListener("dashboard-item-before-remove", e -> {
             if (!isEditable()) {
                 return;
             }
-            handleItemRemovedClientEvent(e, idKey);
-        }).addEventData(idKey);
+            handleItemBeforeRemoveClientEvent(e, idKey, sectionKey);
+        }).preventDefault().addEventData(idKey).addEventData(sectionKey);
```
sissbruecker, line 840: With `preventDefault` this always prevents the web component from removing the item, delegating the logic to the server.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/Dashboard.java
```diff
@@ -786,21 +829,47 @@ private void handleItemResizedClientEvent(DomEvent e, String idKey,
                 getChildren().toList()));
     }
 
-    private void initItemRemovedClientEventListener() {
+    private void initItemBeforeRemoveClientEventListener() {
         String idKey = "event.detail.item.id";
-        getElement().addEventListener("dashboard-item-removed", e -> {
+        String sectionKey = "event.detail.section?.id";
```
DiegoCardoso, line 834: nit: does it make sense to move these two local variables as static final class members? `handleItemBeforeClientEvent` is only called from the listener defined here, so it could get these values from the class scope instead.

DiegoCardoso (approved):
LGTM

One optional suggestion, but other than that, it works as expected.

vaadin-dashboard-flow-parent/vaadin-dashboard-flow/src/main/java/com/vaadin/flow/component/dashboard/Dashboard.java
sissbruecker: There are several other methods that define similar keys inline. Some of those have the same name, but reference different paths. Extracting static constants for only one of these methods might actually lead to more confusion than help. Overall I think that should be done in a separate refactoring to align all usages in the component.

PR #8417, opened 2026-01-05
fix: add locale fallback for MessageList to handle variants

### LINKED ISSUE #8403: MessageList breaks when using Locale Variants [1 DAY]
### Description

The JS of `vaadin-message-list` uses the ISO-Code of the current UI-Locale in `Intl.DateTimeFormat`.
This breaks when using custom Locale variants.

The component does not render at all and the following error is logged in the browser console:
```
Uncaught (in promise) RangeError: Invalid language tag: de-de-x-lvariant-hw
    at new DateTimeFormat (<anonymous>)
    at formatItems (generated-flow-imports-D9DsWzDG.js:196318:21)
    at Object.setItems (generated-flow-imports-D9DsWzDG.js:196336:19)
    at MessageList.eval (eval at zu (FlowClient-DroGq_LA.js:7491:13), <anonymous>:3:68)
```

Cause of the error is the following code:
https://github.com/vaadin/flow-components/blob/d7681ff81b8f608e85e617b517d45260274cf11c/vaadin-messages-flow-parent/vaadin-messages-flow/src/main/resources/META-INF/resources/frontend/messageListConnector.js#L17-L36

> [!NOTE]
>  Context: Our App uses Locale Variants to supply customer-individual Externalization via Resource Bundles

### Expected outcome

Language variants do not cause the rendering of the MessageList to fail

Both of the following behaviors would be fine for our use case:
- Locale Variants are completely ignored by the Message List and the base Locale (`de-DE`) is always used
- The base Locale (`de-DE`) is used as a fallback when `Intl.DateTimeFormat` cannot handle the ISO-Code

```java
package com.marcobsidian.vaadin.views.helloworld;

import java.time.Instant;
import java.util.Locale;

import com.vaadin.flow.component.messages.MessageList;
import com.vaadin.flow.component.messages.MessageListItem;
import com.vaadin.flow.router.RouteAlias;

@PageTitle("Hello World")
@Route(value = "helloworld")
@RouteAlias(value = "")

    // German with customer dependent Locale variant.
    // In prod, we already set this at UI init or when the user changes the app language
    private static final Locale APP_LOCALE = new Locale("de", "DE", "hw");

    public HelloWorldView() {
        UI.getCurrent().setLocale(APP_LOCALE);

        var messageList = new MessageList();
        messageList.setSizeFull();
        messageList.setItems(new MessageListItem("Hello World", Instant.now(), "World Greeter"));
        add(messageList);
    }

}
```

### Steps to reproduce

1. Open http://localhost:8080/helloworld, MessageList does not render
Image

2. Open Chrome DevTools Console
Image

### Environment

Vaadin version(s): 24
OS: Windows 11

### Browsers

Issue is not browser related

Artur-:
MessageList now gracefully handles locale variants (e.g., "de-DE-hw") by implementing a progressive fallback strategy when the full locale string is not valid for Intl.DateTimeFormat:
1. Try full locale (e.g., "de-DE-hw")
2. Fall back to base locale (e.g., "de-DE")
3. Fall back to language only (e.g., "de")
4. Fall back to browser default

This prevents the component from crashing when custom locale variants are used for resource bundle customization.
 Fixes #8403

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,7 +80,7 @@
         <karaf-maven-plugin.version>4.4.9</karaf-maven-plugin.version>
         <maven.surefire.plugin.version>3.5.4</maven.surefire.plugin.version>
         <maven.failsafe.plugin.version>3.5.4</maven.failsafe.plugin.version>
-        <maven.resources.plugin.version>3.3.1</maven.resources.plugin.version>
+        <maven.resources.plugin.version>3.4.0</maven.resources.plugin.version>
         <spotless.plugin.version>3.1.0</spotless.plugin.version>
         <spotless.licence-header>${maven.multiModuleProjectDirectory}/scripts/templates/apache2-license-header.txt</spotless.licence-header>
         <spotless.npmExecutable/>


vaadin-messages-flow-parent/vaadin-messages-flow-integration-tests/src/main/java/com/vaadin/flow/component/messages/tests/MessageListPage.java
```diff
@@ -70,6 +70,9 @@ public MessageListPage() {
 
         addButton("setLocale", () -> UI.getCurrent().setLocale(Locale.ITALIAN));
 
+        addButton("setLocaleVariant",
+                () -> UI.getCurrent().setLocale(new Locale("de", "DE", "hw")));
```
DiegoCardoso: This constructor is deprecated:
```suggestion
                () -> UI.getCurrent().setLocale(Locale.of("de", "DE", "hw")));
```

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/resources/META-INF/resources/frontend/messageListConnector.js
```diff
@@ -15,16 +15,42 @@
...
     minute: 'numeric'
   });
+}
+
+function getFormatter(locale) {
+  // Try creating formatter with progressive fallbacks
+  const fallbackLocales = [
+    locale, // Full locale (e.g., "de-DE-hw")
+    locale?.split('-').slice(0, 2).join('-'), // Base locale without variant (e.g., "de-DE")
+    locale?.split('-')[0] // Language only (e.g., "de")
```
DiegoCardoso: nit: the split could be saved to a variable to avoid doing it twice here.

vaadin-messages-flow-parent/vaadin-messages-flow-integration-tests/src/test/java/com/vaadin/flow/component/messages/tests/MessageListIT.java
```diff
@@ -226,6 +226,35 @@ public void changeLocale_addItem_itemTimesFormatted() {
...
+        clickElementWithJs("setLocaleVariant");
+
+        List<MessageElement> messages = messageList.getMessageElements();
+        Assert.assertEquals("Unexpected items count", 2, messages.size());
+
+        MessageElement msg = messages.get(0);
+        Assert.assertEquals("Unexpected text content", "foo", msg.getText());
+        // Should fallback to German locale formatting
+        Assert.assertTrue("Unexpected time prop", msg.getTime()
+                .matches("[0-9]+\\. [A-Za-z]+ 2021, [0-9]+:[0-9]+"));
```
DiegoCardoso: This fails because the formatted date is defined as `1. Jan. 2021, 12:00`:

Image

The regex could be fixed with:

```suggestion
        Assert.assertTrue("Unexpected time prop", msg.getTime()
                .matches("[0-9]+\\. [A-Za-z\\.]+ 2021, [0-9]+:[0-9]+"));
```

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/resources/META-INF/resources/frontend/messageListConnector.js
Artur-: Done

PR #8470, opened 2026-01-13
chore: use flow BOM

sissbruecker:
## Description

Use the Flow BOM to declare versions for most of the Flow dependencies.

## Type of change

- Internal

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -117,11 +117,6 @@
             </dependency>
             <dependency>
                 <groupId>com.vaadin</groupId>
-                <artifactId>flow-osgi</artifactId>
-                <version>${flow.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>com.vaadin</groupId>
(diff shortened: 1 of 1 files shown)


pom.xml
vursen, line 153: I wonder if we need these too. Couldn't they also come from Flow?

ZheSun88: checking the usage here https://github.com/search?q=repo%3Avaadin%2Fflow-components%20jakarta.jakartaee-web-api&type=code

yes, we need it.

ZheSun88: but it is not suggested to use <scope> in the dependencyManagement, as the scope will be automatically applied to new added dependencies.

sissbruecker: The Flow BOM only declares Flow modules, so any other dependencies need to be declared separately.

vursen (approved):

PR #8474, opened 2026-01-13
chore: cache theme distribution

sissbruecker:
## Description

The Lumo / Aura theme distribution step, which downloads the respective NPM package and copies files to the target resources, is currently run every time you run ITs or start a server.

This adds a caching step that skips the main part of the script if the NPM package version declared in the Lumo class has not changed. Changing the version or running `mvn clean` forces a rebuild.

## Type of change

- Internal


scripts/copy-theme-distribution.js
```diff
@@ -40,12 +40,26 @@ if (!match) {
...
 
-console.log(`Extracting theme distribution for ${npmThemePackage}@${npmThemePackageVersion}`);
-
 // Create temporary directory for npm install
 const tempDir = path.join('target', 'npm-workspace');
 fs.mkdirSync(tempDir, { recursive: true });
 
+// Check if cached and up to date
+const cacheFile = path.join(tempDir, `.theme-cache.json`);
+if (fs.existsSync(cacheFile)) {
```
vursen, line 49: We should check that the bundled theme file is also present in the target folder, I guess.

sissbruecker: How would it not be there if the cache file still exists?

vursen: Ah, I missed that the temp folder is inside the target folder.

vursen (approved):

Issue #8480, opened 2026-01-14
[Slider] Create Slider flow counterpart


vursen:
Initialize `vaadin-slider-flow-parent` module and add a Slider flow component based on the `<vaadin-slider>` web component.

PR #8485, opened 2026-01-14
feat: create Slider component

vursen:
## Description

The PR introduces a Flow counterpart for the `<vaadin-slider>` web component, with the following API:

  **Constructors:**

  1. `Slider()`
  2. `Slider(valueChangeListener)`
  3. `Slider(min, max, value)`
  4. `Slider(min, max, value, valueChangeListener)`
  5. `Slider(min, max, step, value)`
  6. `Slider(min, max, step, value, valueChangeListener)`
  7. `Slider(label)`
  8. `Slider(label, valueChangeListener)`
  9. `Slider(label, min, max, value)`
  10. `Slider(label, min, max, value, valueChangeListener)`
  11. `Slider(label, min, max, step, value)`
  12. `Slider(label, min, max, step, value, valueChangeListener)`

  **Methods:**
  - `setValue(value)`
  - `setValue(value, min, max)`
  - `setValue(value, min, max, step)`
  - `setMin(min)`
  - `getMin()`
  - `setMax(max)`
  - `getMax()`
  - `setStep(step)`
  - `getStep()`

  Fixes https://github.com/vaadin/flow-components/issues/8480

  ## Type of change

  - [x] Feature

### DIFF
--- a/vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/Slider.java
+++ b/vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/Slider.java
@@ -206,8 +206,8 @@ public class Slider extends SliderBase<Slider, Double> {
     }
 
     /**
-     * Constructs a {@code Slider} with the given label, min, max, initial value,
-     * and a value change listener.
+     * Constructs a {@code Slider} with the given label, min, max, initial
+     * value, and a value change listener.
      *
      * @param label
      *            the text to set as the label
--- a/vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/SliderBase.java
+++ b/vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/SliderBase.java
@@ -245,7 +245,8 @@ abstract class SliderBase<TComponent extends SliderBase<TComponent, TValue>, TVa
     }
 
     /**
-     * Validates that the given value is valid for the current min, max and step.
+     * Validates that the given value is valid for the current min, max and
+     * step.
      *
      * @param value
      *            the value to validate


vaadin-slider-flow-parent/vaadin-slider-flow-integration-tests/pom.xml
```diff
@@ -0,0 +1,156 @@
...
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>flow-client</artifactId>
+            <version>${flow.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>flow-html-components</artifactId>
+            <version>${flow.version}</version>
+        </dependency>
```
sissbruecker, line 21: The Flow dependency versions are already declared in the dependency management section of the root POM.

vaadin-slider-flow-parent/pom.xml
sissbruecker, line 1: Some files to add:
- README.md
- LICENSE
- vite.config.ts

vaadin-slider-flow-parent/vaadin-slider-flow-integration-tests/pom.xml
vursen: Thanks, fixed.

vaadin-slider-flow-parent/pom.xml
vursen: Added.

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/Slider.java
```diff
@@ -0,0 +1,187 @@
...
+ * Slider is an input field that allows the user to select a numeric value
+ * within a range by dragging a handle along a track or using arrow keys for
+ * precise input.
+ *
+ * @author Vaadin Ltd.
+ */
+@Tag("vaadin-slider")
+// @NpmPackage(value = "@vaadin/slider", version = "25.1.0-alpha1")
+// @JsModule("@vaadin/slider/src/vaadin-slider.js")
+public class Slider extends SliderBase<Slider, Double> {
```
knoobie, line 37: Thinking out load: I'm wondering if it's better for java developer to either use  

`Slider<T extends Number>` or `IntSlider` / `DoubleSlider`. 

Having to deal with double all the time when I just have a simple 0-5 slider feels awkward. (Same could be said for min/max and step.. might be good to add overloaded methods?)

vursen: Valid point, thank you. We could add an IntegerSlider later

vaadin-slider-flow-parent/vaadin-slider-flow/pom.xml
sissbruecker, line 1: Let's register Mockito as an agent explicitly as is done here: https://github.com/vaadin/flow-components/pull/8478.

vursen: Thanks, done.

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/SliderBase.java
```diff
@@ -0,0 +1,276 @@
...
+     * @param max
+     *            the maximum value
+     * @param value
+     *            the value
+     * @throws IllegalArgumentException
+     *             if min is greater than max
+     * @throws IllegalArgumentException
+     *             if value is not valid for the given range and current step
+     */
+    void setValue(double min, double max, TValue value) {
```
sissbruecker: Feels odd to shift the value argument in the overloads. Was that a conscious decision?

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/Slider.java
```diff
@@ -0,0 +1,417 @@
...
+        double adjustedValue = Math.min(getValue(), max);
+        setValue(adjustedValue);
+    }
+
+    /**
+     * Gets the step value of the slider.
+     *
+     * @return the step value
+     */
+    public double getStep() {
```
sissbruecker, line 342: Should this mention that the step is based off of the min value?

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/Slider.java
```diff
@@ -0,0 +1,417 @@
...
+
+    /**
+     * Gets the empty value of the slider, which is equal to the minimum value.
+     *
+     * @return the empty value
+     * @see #getMin()
+     */
+    @Override
+    public Double getEmptyValue() {
+        return getMin();
```
sissbruecker: There are some issues with this:

- `BindingBuilder.asRequired` uses `HasValue.getEmptyValue` to determine whether a field has a value or not (https://github.com/vaadin/flow/blob/2907dd7ff7f0d61c73fb1826610a4e6f7bfcb7ba/flow-data/src/main/java/com/vaadin/flow/data/binder/Binder.java#L1267-L1273). Adding this validator will result in an invalid field when the value is at the minimum.
- `HasValue.isEmpty` returns `true` when the value is at the minimum.
- `HasValue.getOptionalValue` returns an empty optional when the value is at the minimum.

There are several other usages in the Flow codebase that represent smaller issues. None of those seem critical since the component should never be "empty", there always is a value. So an application wouldn't need to use these APIs intended for handling empty fields. However, it could trip up application-specific code that uses generic logic to construct forms and interact with `HasValue` fields. 

So I wonder if we should make it so that `HasValue.getEmptyValue` always returns something that differs from `HasValue.getValue`. Technically that value could be `null`. The only other component that I could find that always has a value is `Checkbox`. That one uses `false` as empty value, which sounds kind of wrong, but may be necessary for required validation at the moment.

Regarding `clear`, we can just use a custom implementation for that without relying on `HasValue.getEmptyValue`.

vaadin-slider-flow-parent/vaadin-slider-flow/src/test/java/com/vaadin/flow/component/slider/tests/SliderTest.java
sissbruecker, line 1: `setValue(null)` is not covered I think.

vaadin-slider-flow-parent/vaadin-slider-flow/src/test/java/com/vaadin/flow/component/slider/tests/SliderTest.java
```diff
@@ -0,0 +1,361 @@
...
+
+    @Test
+    public void setValueFromClient_valueAboveMax_throws() {
+        Slider slider = new Slider(0, 100, 10, 0);
+
+        Assert.assertThrows(IllegalArgumentException.class, () -> {
+            slider.getElement().getNode().getFeature(ElementPropertyMap.class)
+                    .deferredUpdateFromClient("value", 110.0).run();
+        });
+    }
```
sissbruecker: Why would these throw? What are the cases where this can happen?

If this is just to cover potentially invalid input due to a bug in the web component, or scripting, should this ignore the value instead?

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/SliderBase.java
```diff
@@ -0,0 +1,276 @@
...
+ * Abstract base class for slider components.
+ *
+ * @param <TComponent>
+ *            the component type
+ * @param <TValue>
+ *            the value type
+ *
+ * @author Vaadin Ltd
+ */
+abstract class SliderBase<TComponent extends SliderBase<TComponent, TValue>, TValue>
```
sissbruecker, line 39: `InputField` is not applicable to this component?

vaadin-slider-flow-parent/vaadin-slider-flow/src/test/java/com/vaadin/flow/component/slider/tests/SliderTest.java
vursen: NumberField throws a NumberFormatException if parsing the client-side value fails, so it seemed reasonable for Slider to behave the same way.

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/SliderBase.java
vursen: Thanks, done.

vaadin-slider-flow-parent/vaadin-slider-flow/src/test/java/com/vaadin/flow/component/slider/tests/SliderTest.java
vursen: Thanks, added a test.

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/Slider.java
vursen: Thanks, added a clarification.

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/SliderBase.java
vursen: Good point. I've changed the order now.

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/Slider.java
vursen: Nice catch, thanks for pointing this out. Returning `null` seems reasonable indeed.

vursen: Fixed

vaadin-slider-flow-parent/vaadin-slider-flow/src/test/java/com/vaadin/flow/component/slider/tests/SliderTest.java
vursen: Updated the code to ignore invalid values from client side instead of throwing.

vaadin-slider-flow-parent/vaadin-slider-testbench/src/main/java/com/vaadin/flow/component/slider/testbench/SliderElement.java
```diff
@@ -0,0 +1,63 @@
...
+
+import com.vaadin.testbench.TestBenchElement;
+import com.vaadin.testbench.elementsbase.Element;
+
+/**
+ * A TestBench element representing a <code>&lt;vaadin-slider&gt;</code>
+ * element.
+ */
+@Element("vaadin-slider")
+public class SliderElement extends TestBenchElement {
```
sissbruecker: We probably want a method for setting the value as well.

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/Slider.java
sissbruecker, line 1: There's a somewhat conceptual issue here, where every mouse move while dragging will potentially fire a change event and thus result in a request to the server.

https://github.com/user-attachments/assets/445f048f-f3b6-401f-962d-5186173296f1

I think we should at least support value change modes and highlight this in the documentation. Maybe debounced value updates should even be the default. For now let's add an issue at least, no need to decide / update the PR yet.

vursen: Yeah, thanks for spotting this. The plan was to listen to the change event by default and also support eager value change modes. The fact that the component is currently listening to value-changed is my oversight. Let me fix it.

vaadin-slider-flow-parent/vaadin-slider-testbench/src/main/java/com/vaadin/flow/component/slider/testbench/SliderElement.java
vursen: Yes, I'm planning to improve the TestBench API in a following PR.

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/Slider.java
vursen: Created an issue to support value change modes: https://github.com/vaadin/flow-components/issues/8602

vaadin-slider-flow-parent/vaadin-slider-testbench/src/main/java/com/vaadin/flow/component/slider/testbench/SliderElement.java
vursen: Added a setValue method and also more interfaces.

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/Slider.java
vursen: Configured the component to synchonize on change events. I'll add value change integration tests as part of #8602.

sissbruecker (approved):

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/Slider.java
```diff
@@ -0,0 +1,405 @@
...
+
+/**
+ * Slider is an input field that allows the user to select a numeric value
+ * within a range by dragging a handle along a track or using arrow keys for
+ * precise input.
+ *
+ * @author Vaadin Ltd.
+ */
+@Tag("vaadin-slider")
+@NpmPackage(value = "@vaadin/slider", version = "25.1.0-alpha5")
```
web-padawan: Please update version to alpha6.

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/Slider.java
```diff
@@ -0,0 +1,405 @@
...
+     *            the value change listener
+     */
+    public Slider(double min, double max, double step, double value,
+            ValueChangeListener<? super ComponentValueChangeEvent<Slider, Double>> listener) {
+        this(min, max, step, value);
+        addValueChangeListener(listener);
+    }
+
+    /**
+     * Constructs a {@code Slider} with the given label, range 0-100, and
```
web-padawan: Maybe opinionated, but I would replace "range 0-100" with "min set to 0, max set to 100" to avoid using word "range" that could cause potential confusion with range slider component.

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/Slider.java
```diff
@@ -0,0 +1,405 @@
...
+     *            the value change listener
+     */
+    public Slider(String label,
+            ValueChangeListener<? super ComponentValueChangeEvent<Slider, Double>> listener) {
+        this(listener);
+        setLabel(label);
+    }
+
+    /**
+     * Constructs a {@code Slider} with the given label, range, and initial
```
web-padawan: Let's update to use "given label, min, max, and initial value" here and below.

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/Slider.java
vursen: Done.

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/Slider.java
vursen: Done.

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/Slider.java
vursen: Done.

PR #8498, opened 2026-01-16
feat: add HasThemeVariant::bindThemeVariant

tltv:
Adds bindThemeVariant(TVariantEnum, Signal<Boolean>) shortcut to call HasTheme::bindThemeName from HasThemeVariant with a variant name.

Part of vaadin/flow#23211

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -101,7 +101,7 @@
 
         <!-- spreadsheet -->
         <poi.version>5.5.1</poi.version>
-        <javaparser.version>3.27.1</javaparser.version>
+        <javaparser.version>3.28.0</javaparser.version>
 
         <!-- Additional Surefire argLine - set to -javaagent:${org.mockito:mockito-core:jar} in modules using Mockito -->
         <surefire.argLine/>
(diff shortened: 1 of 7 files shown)


DiegoCardoso (commented):
Let's add some tests to cover this API to `HasThemeVariantTest`.

tltv:
I'll add unit tests but would like to wait for [this PR](https://github.com/vaadin/flow-components/pull/8499) that adds utility to write unit tests for signals and moves MockUI to this module. For now tests uses a copy of MockUI.

DiegoCardoso (approved):

DiegoCardoso:
> I'll add unit tests but would like to wait for [this PR](https://github.com/vaadin/flow-components/pull/8499) that adds utility to write unit tests for signals and moves MockUI to this module. For now tests uses a copy of MockUI.

On a second thought, ideally, that PR could've been split into two, and the `MockUI` refactor could've been done separately.

tltv:
Now there's utility for unit tests but since HasThemeVariant tests are in `vaadin-flow-component-base` and utility is in `vaadin-flow-components-test-util` sibling module and it depends on `vaadin-flow-component-base`, utility is not accessible. This is a problem in maven module structure and probably it would need to be changed a bit to make it work.
I think in general it would make sense that tests in `vaadin-flow-component-base` could use same utilities as all other component specific test modules, but that's a bit out of scope for this change.

tltv:
> @tltv the `MockUI` class has been moved in `main` in #8507, so this PR can use it instead now.

It's actually still not available in base module because it's moved to `vaadin-flow-components-test-util` module which has dependency to `vaadin-flow-component-base` module. I can't add circular dependency to `vaadin-flow-components-test-util`. Also it's under `src/main` but base module's unit tests are under `src/test`.
I tried to move it to `vaadin-flow-component-base`, but it then requires also adding Mockito and introducing test classes in `src/main`. Seemed like safer to just have a copy of MockUI in `src/test` too.

PR #8499, opened 2026-01-16
feat: add constructors for text binding to Button

platosha:
Connected to: vaadin/flow#23193

Depends on #8507 
    
- Add new constructors and methods to support binding text to signals
- Refactor internal text handling using a Text node
- Improve icon slot management and theme attribute updates

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -101,7 +101,7 @@
 
         <!-- spreadsheet -->
         <poi.version>5.5.1</poi.version>
-        <javaparser.version>3.27.1</javaparser.version>
+        <javaparser.version>3.28.0</javaparser.version>
 
         <!-- Additional Surefire argLine - set to -javaagent:${org.mockito:mockito-core:jar} in modules using Mockito -->
         <surefire.argLine/>


DiegoCardoso (commented):
Do we need to add integration tests to check the signals behavior? Wouldn't it be enough to add unit tests instead?

platosha:
> Do we need to add integration tests to check the signals behavior? Wouldn't it be enough to add unit tests instead?

One reason for ITs on my mind is the dynamic Button behavior regarding the icon. Some aspects are challenging to assert: it's quite possible to make the icon accidentally disappear from the DOM upon a text change, and this error condition would be hard to detect from a unit test context.

vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/main/resources/vaadin-featureflags.properties
```diff
@@ -0,0 +1 @@
+com.vaadin.experimental.flowFullstackSignals=true
```
DiegoCardoso: nit: newline at the end of the file is missing.

vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java
```diff
@@ -63,6 +66,14 @@ public class Button extends Component
     private final DisableOnClickController<Button> disableOnClickController = new DisableOnClickController<>(
             this);
 
+    // Explicit text node is used to manage the text content and text signal
+    // bindings of the button separatly from its icon. All `HasText` methods
+    // delegate to this node.
+    private final Text textNode = new Text("");
+    {
+        add(textNode);
+    }
```
DiegoCardoso: This pattern is not common in our codebase, so I wonder if it wouldn't be better to initialize it in the constructor instead to avoid any confusion.

vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/main/java/com/vaadin/flow/component/button/tests/ButtonView.java
```diff
@@ -183,7 +183,8 @@ private void showButtonClickedMessage(ClickEvent<Button> evt) {
 
     private boolean containsChild(Component parent, String tagName) {
         return parent.getElement().getChildren()
-                .anyMatch(element -> element.getTag().equals(tagName));
+                .anyMatch(element -> !element.isTextNode()
+                        && element.getTag().equals(tagName));
```
mshabarov, line 187: This might be a small breaking change for the TestBench tests where the button is involved as the structure seems to change and there is non-zero probability that some element queries would change their behavior. Is this concern valid? I'm not 100% sure as this code snippet is a button testing code and may or may not reflect the real cases.

platosha:
> IMO this PR could've been split to remove the noise from the `MockUI` refactor.
> 
> Added a couple of comments.

See #8507

vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/main/java/com/vaadin/flow/component/button/tests/ButtonView.java
mshabarov: Impact on the Selenium queries is not an issue as this is a text content and Selenium filters these out.
Button add-ons hopefully work with the text and icon through the public methods, so not a concern.
Withdrawing this comment.

platosha: From our F2F discussion we concluded that it's a non-issue. Selenium tests behavior isnt affected, as the DOM structure didn't change. The only touched part is the low-level Java UI tree structure behind the button which now contains an extra TextNode component encapsulating an already existing low-level text node.

mshabarov (approved):

Issue #8431, opened 2026-01-08
Add Flow UploadButton, UploadDropZone, UploadFileListBox components


tomivirkki:
### Describe your motivation

Flow components are needed to use the modular upload web components from Java. These components link to an `UploadManager` to provide a complete upload experience with flexible layouts.

### Describe the solution you'd like

Add three components in `com.vaadin.flow.component.upload`:

### UploadButton

- Wraps `<vaadin-upload-button>` web component
- Implements `HasComponents` for slotted content
- Constructor: `UploadButton()`, `UploadButton(UploadManager manager)`
- Method: `setManager(UploadManager)` - links to manager via JS

### UploadDropZone

- Wraps `<vaadin-upload-drop-zone>` web component
- Implements `HasComponents` for slotted content
- Constructor: `UploadDropZone()`, `UploadDropZone(UploadManager manager)`
- Method: `setManager(UploadManager)` - links to manager via JS

### UploadFileList

- Wraps `<vaadin-upload-file-list>` web component
- Constructor: `UploadFileList()`, `UploadFileList(UploadManager manager)`
- Method: `setManager(UploadManager)` - links to manager via JS
- Automatically displays files from linked manager

All components:

- Set `target` property on web component to link with manager
- Handle attach/detach lifecycle for manager linking
### Additional context

Example usage:

```java
var manager = new UploadManager(owner, handler);

var dropZone = new UploadDropZone(manager);
dropZone.add(new Span("Drop files here"));

var button = new UploadButton(manager);
button.add(new Span("Select Files"));

var fileList = new UploadFileList(manager);

add(dropZone, button, fileList);
```

PR #8508, opened 2026-01-20
feat: add components for UploadManager

tomivirkki:
## Description

Fixes https://github.com/vaadin/flow-components/issues/8431

## Type of change

Feature


vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/UploadButton.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * Creates a new upload button linked to the given manager.
+     *
+     * @param manager
+     *            the upload manager to link to
+     */
+    public UploadButton(UploadManager manager) {
+        setManager(manager);
```
ugur-vaadin: Nit: should we accept null here? It unlinks when a manager was already set but it does not seem to make sense in the constructor.

tomivirkki: Done

PR #8513, opened 2026-01-22
feat: add GridElement methods to read cell contents in bulk

### LINKED ISSUE #1876: Fast Grid Dump
Feature request: I need to test whether the Grid is showing correct data, and for that I need to capture the contents which the Grid is showing, and compare the result against the expected Grid contents, which is just a two-dimensional array of Strings.

Currently what I can do is to call `GridElement.getRows()`, then grab all cells for each row, then call `GridTHTDElement.getText()`. However this generates one browser request per every cell, which is hopelessly slow.

It would be great if `GridElement` offered a faster way to dump the cells into a 2-dimensional array of Strings: perhaps by dumping/calling the DataProvider directly, or by capturing currently shown data from its HTML structure - can be done page-by-page way.

Artur-:
Reading grid data in a TestBench test today means calling `getCell(row, col).getText()`
for every cell, and each call is a separate browser round-trip. For anything larger
than a handful of rows this is very slow, and there is no built-in way to grab the
data the grid is showing as a plain 2D structure to compare against expected values.

This adds three methods to `GridElement` that read cell text in bulk:

- `getVisibleCellContents()` — the rows currently in the viewport, in one round-trip.
- `getCellContents(fromRow, toRow)` — a specific row range, scrolling as needed to
  load the rows.
- `getAllCellContents()` — every row, by scrolling through all pages.

All three return `List<List<String>>` (rows of visible-column text). The text of each
cell matches `GridTHTDElement.getText()`, so existing expected-value comparisons keep
working. Hidden columns are excluded, and rows are always returned in row-index order
(the grid recycles row elements, so DOM order does not match visual order).

The extraction runs in the browser as a single `executeScript` per rendered window
instead of one request per cell, which is the whole point: for a grid with N rows and
M columns this replaces N×M round-trips with 1 (viewport) or roughly one per page
(range/all).

The extraction reads the rendered DOM rather than the server-side data provider, so
the result reflects what the grid actually shows, including what renderers produced.

Tests cover small/medium/large and callback-backed grids, hidden columns, invalid
ranges, parity with `getCell().getText()`, and that the viewport dump returns exactly
the visible rows in order.

Fixes vaadin/testbench#1876


yuriy-fix:
@Artur- , would you like to proceed with the change and mark it ready for review or there is anything you would like to adjust in the PR or have a discussion about?

Artur-:
Updated to make more sense and work better

vaadin-grid-flow-parent/vaadin-grid-testbench/src/main/java/com/vaadin/flow/component/grid/testbench/GridElement.java
```diff
@@ -605,6 +607,153 @@ public List<GridTHTDElement> getCells(int rowIndex) {
...
+    public List<List<String>> getCellContents(int fromRow, int toRow)
+            throws IndexOutOfBoundsException {
+        int rowCount = getRowCount();
+        if (fromRow < 0 || toRow < 0 || fromRow >= rowCount || toRow >= rowCount
+                || fromRow > toRow) {
+            throw new IndexOutOfBoundsException(
+                    "fromRow and toRow: expected to be 0.." + (rowCount - 1)
+                            + " with fromRow <= toRow, but were " + fromRow
+                            + " and " + toRow);
+        }
```
DiegoCardoso, line 694: On an empty grid, `getCellContents(0, 0)` throws
`IndexOutOfBoundsException: ... expected to be 0..-1 with fromRow <= toRow, but were 0 and 0`.
The `0..-1` bound looks broken, and it's inconsistent with `getAllCellContents()`, which returns `[]` for the same grid. Consider special-casing `rowCount == 0` with a clearer message (e.g. "grid is empty").

Issue #8512, opened 2026-01-21
[Spreadsheet] Custom cell editor combobox does not show and is not active on merged cells


Shufflepants:
### Description

After [8317](https://github.com/vaadin/flow-components/issues/8317) it seems there are still issues with merged cells that are causing an issue. When a `ComboBox` editor is on a merged cell, something about the rendering process causes the `ComboBox` to not be visible/active. There is even a split second while the page is loading where the highlighted background of a properly visible `ComboBox` appears in the first cell of a set of merged cells before the merging happens and the `ComboBox` disappears. 
You can see that cell B3 should have a visible `ComboBox` like B18, but it does not.
<img width="518" height="632" alt="Image" src="https://github.com/user-attachments/assets/a6c76ab0-c4e1-4f77-b2f3-d845509037be" />
After a single click on the cell, the combobox becomes visible, but does not open.
<img width="562" height="664" alt="Image" src="https://github.com/user-attachments/assets/c589c41e-bf54-431b-9b2a-668c9296c14e" />

It is as though the `setShowCustomEditorOnFocus` is set to true even though it is not. And `ComboBox` custom editors that are not in merged cells properly show up and only require 1 click to open the drop-down.

### Expected outcome

The drop-down arrow of a `ComboBox` custom editor in a spreadsheet should be visible upon initial load, and there should only be a single click required to open the drop down and have the text field selected.

### Minimal reproducible example

Using the provided example project [scroll-with-editors-formatting-bug.zip](https://github.com/user-attachments/files/23883583/scroll-with-editors-formatting-bug.zip) that was used for [8317](https://github.com/vaadin/flow-components/issues/8317), simply change the `4` on `line 68` to a `2` so that one of the `ComboBox`s is now in the same cell as one of the merged cells. 

### Steps to reproduce

You can reproduce this by putting a `ComboBox` editor in the same cell as an existing merged cell. There are two `ComboBox`s in the screenshots, one on B3 and one on B18, but only the one on B18 shows up without needing to be clicked on first. 

### Environment

Vaadin version(s): 24.9.8
OS: Windows 11

### Browsers

Firefox

PR #8544, opened 2026-01-28
fix: show custom editors for merged cells

sissbruecker:
## Description

Custom editors are added to regular cells before merged regions are applied, which effectively hides the original cell with the editor and shows a merged cell instead. Since `upateMergedRegions` also uses a deferred update, changing the call order of these updates would not have any effect. The fix here is then to apply custom editors again after the deferred update in `upateMergedRegions` has run.

Also fixes `setMergedCellValue` to not remove a custom editor if one is displayed.

Fixes https://github.com/vaadin/flow-components/issues/8512

## Type of change

- Bugfix


sissbruecker:
Should work now.

Issue #8549, opened 2026-01-28
Combobox overwriting value with null with small delayed async calls


mukherjeesudebi:
### Description

If an asysn task return the result quickly, and if a value is set to Combobox and MultiSelectComboBox, then both the  valueChangeListeners are called twice and finally set to null and empty

https://github.com/user-attachments/assets/9f8260f1-0455-4d28-a364-5db0ec5cb6d3

### Expected outcome

The value should be set to what was manually set and not replace it with null value irrespective of the fact that how quickly the async returns value. 

### Minimal reproducible example

```java
import java.util.concurrent.CompletableFuture;

import com.example.TaskWithSecurityContext;
import com.vaadin.flow.component.UI;
import com.vaadin.flow.component.button.Button;
import com.vaadin.flow.component.combobox.ComboBox;
import com.vaadin.flow.component.combobox.MultiSelectComboBox;
import com.vaadin.flow.component.grid.Grid;
import com.vaadin.flow.component.orderedlayout.HorizontalLayout;
import com.vaadin.flow.component.orderedlayout.VerticalLayout;
import com.vaadin.flow.component.select.Select;
import com.vaadin.flow.component.textfield.IntegerField;
import com.vaadin.flow.component.textfield.TextArea;
import com.vaadin.flow.component.textfield.TextField;
import com.vaadin.flow.router.AfterNavigationEvent;
import com.vaadin.flow.router.AfterNavigationObserver;
import com.vaadin.flow.router.QueryParameters;
import com.vaadin.flow.router.Route;

@Route("")
public class ComboBoxAsync extends VerticalLayout implements AfterNavigationObserver {

	final IntegerField waitField;
	final Button runButton;

	final ComboBox<Foo> comboBox;
	final Grid<Foo> grid;
	final MultiSelectComboBox<Foo> multiSelectComboBox;
	final Select<Foo> select;
	final TextArea textArea;
	final TextField textField;

	private Foo foo1 = new Foo("Foo-1");
	private Foo foo2 = new Foo("Foo-2");

	ComboBoxAsync() {
		this.waitField = new IntegerField();
		this.waitField.setValue(0);
		this.waitField.setHelperText(
				"You can input the amount of ms the background task should last for. A value of 0 will run without Thread.sleep()."
						+ "As a result, all fields should display "
						+ "the value \"Foo-2\". Small positive values will cause issues for ComboBox components.");

		this.runButton = new Button("Run to test");
		this.runButton.addClickListener(event -> {
[...]

PR #8555, opened 2026-01-28
fix: initialize combo box value property

sissbruecker:
## Description

When initializing a combo box value asynchronously after the initial response has been sent, the client can send a client-side value change that reverts the value to `null`. This happens due to the initial `value-changed` event fired by Polymer / PolyLitMixin that occurs when initializing the web component's `value` property with an empty string. The Flow client detects that as a property change and syncs the changed value to the server, overwriting the value set by the async task in between. The order of events is:
- Send initial response, `value` property is undefined
- Async task executes on server, changes `value` property
- Client-side receives initial response, web component fires `value-changed`, Flow detects that `value` has changed from undefined to an empty string
- Flow client sends a request to sync `value` to the server
- `value` property on server is set to an empty string, actual component value becomes `null`

This change sets the initial property value to an empty string to match the web component default, which avoids Flow synchronizing the empty value to the server. This matches what we do for several other components / properties to avoid an initial value change event from the client when the property is not explicitly set on the server.

Fixes https://github.com/vaadin/flow-components/issues/8549

## Type of change

- Bugfix

### DIFF
--- a/.claude/skills/cherry-pick/SKILL.md
+++ b/.claude/skills/cherry-pick/SKILL.md
@@ -1,0 +1,53 @@
+---
+name: cherry-pick
+description: Cherry-pick a PR to a target version branch
+argument-hint: <pr-url> <target-branch>
+disable-model-invocation: true
+allowed-tools: Read, Edit, Bash(git:*), Bash(gh:*), Bash(mvn:*)
+---
+
+Cherry-pick the PR with the GitHub URL $0 to the $1 branch.
+This command is only run when automated cherry-picking failed, so expect merge conflicts.
+
+## Steps
+
+1. **Gather PR information** using `gh pr view`:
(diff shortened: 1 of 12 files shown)


sissbruecker:
> Tested the changes, and the issue seems to be gone after them.
> 
> One issue that looks unrelated to this, but I could notice using the tested view provided in the linked issue, is that sometimes the values aren't displayed with some delay values.
> load-issue.mp4
> 
> But I only saw it in Zen (not even in Firefox)

Did you add `@Push` to the application class?

DiegoCardoso:
> Did you add `@Push` to the application class?

Yes, I did.

sissbruecker:
It looks like the sync doesn't happen properly if it affects all components. But that looks more like a Flow issue.

Issue #1531, opened 2021-05-21
Deprecate setRequired(boolean required)


pleku:
This method should not be exposed at all on component API since it is a duplicate of `HasValue::setRequiredIndicatorVisible(boolean requiredIndicatorVisible)`. And both just do the same thing, set the `required` property value (see `HasValueAndElement` interface for the implementation).

We should deprecate the method and say it will be removed later on, and that everyone should use `setRequiredIndicatorVisible` instead.

PR #8568, opened 2026-01-30
feat: Add CheckboxGroup::bindRequired(Signal<Boolean> signal)

mshabarov:
Related-to https://github.com/vaadin/flow/issues/23196

### DIFF
--- a/vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/CheckboxGroup.java
+++ b/vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/CheckboxGroup.java
@@ -73,6 +73,7 @@ import com.vaadin.flow.function.SerializableFunction;
 import com.vaadin.flow.function.SerializablePredicate;
 import com.vaadin.flow.internal.JacksonUtils;
 import com.vaadin.flow.shared.Registration;
+import com.vaadin.signals.Signal;
 
 import tools.jackson.databind.node.ArrayNode;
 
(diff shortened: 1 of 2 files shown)


vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/CheckboxGroup.java
sissbruecker, line 1: `setRequired` only delegates to `setRequiredIndicatorVisible` and does not exist all field components. I also did not understand why the issue only lists the bind API for Combo Box / Checkbox Group, but not for all the other field components that have it.

To the best of my knowledge, `setRequired` is a legacy API that was introduced from generating Flow component APIs from web components, and we don't plan on adding it to other / new components. As such we probably don't want a `bindRequired` API.

sissbruecker: Had a chat in the team about this today. The consensus was also that `setRequired` is a legacy API that should not be necessary anymore. On the other hand, `setRequired` is a much better name than `setRequiredIndicatorVisible`. `setRequiredIndicatorVisible` gives the impression that it only shows the indicator, while it actually is used to trigger the components required validation:
https://github.com/vaadin/flow-components/blob/9f3c7570e856a1cbbe8fb052977b5306ed9176ef/vaadin-text-field-flow-parent/vaadin-text-field-flow/src/main/java/com/vaadin/flow/component/textfield/TextField.java#L101-L105

From what I can see there is no difference between both methods currently, though there is also an old issue where people imply that there is: https://github.com/vaadin/flow-components/issues/1531. Things might have changed since then.

So before we add a bind method for a required property, we should:
- Make sure there is no difference in behavior between the two
- Decide which property we want to use: `setRequired` or `setRequiredIndicatorVisible`
- If we want to use `setRequired`:
  - Move it to `HasValue`, remove it from individual components
  - Deprecate `setRequiredIndicatorVisible`
  - Add `bindRequired` in `HasValue`

mshabarov: Thanks for noticing this 👍 
Saying that, it doesn't make much sense to add bind method to each single component directly.
Better place is `HasValue::bindRequiredIndicatorVisible` or `HasValue::bindRequired` with considerations above.
Let's close this PR.

PR #8583, opened 2026-01-31
refactor: remove custom I18N merging logic

sissbruecker:
## Description

Removes custom logic for merging I18N values from Flow components with web components defaults. That should now be covered by all web components having support for setting partial I18N objects.

Also updated I18N ITs to have the same structure:
- Have a test for custom I18N
- Have a test with empty I18N to verify fallback logic for default I18N works
- Verify I18N is applied to DOM, instead of checking property values

## Type of change

- Refactoring

### DIFF
--- a/vaadin-map-flow-parent/vaadin-map-flow-integration-tests/src/main/java/com/vaadin/flow/component/map/ControlsPage.java
+++ b/vaadin-map-flow-parent/vaadin-map-flow-integration-tests/src/main/java/com/vaadin/flow/component/map/ControlsPage.java
@@ -1,0 +1,47 @@
+/**
+ * Copyright 2000-2025 Vaadin Ltd.
+ *
+ * This program is available under Vaadin Commercial License and Service Terms.
+ *
+ * See {@literal <https://vaadin.com/commercial-license-and-service-terms>} for the full
+ * license.
+ */
+package com.vaadin.flow.component.map;
+
+import com.vaadin.flow.component.html.Div;
+import com.vaadin.flow.component.html.NativeButton;
+import com.vaadin.flow.component.map.configuration.controls.AttributionControl;
+import com.vaadin.flow.component.map.configuration.controls.ScaleLineControl;
+import com.vaadin.flow.component.map.configuration.controls.ZoomControl;
+import com.vaadin.flow.router.Route;
+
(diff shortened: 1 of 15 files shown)


vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/test/java/com/vaadin/flow/component/upload/tests/UploadFileListI18nIT.java
```diff
@@ -15,142 +15,76 @@
...
-        // Verify the partial translation is applied
-        Assert.assertEquals("\"Poista\"", translationMap.get("file.remove"));
-
-        // Verify other file translations still have their default values
-        // (the web component has built-in defaults that are preserved)
-        Assert.assertEquals("\"Retry\"", translationMap.get("file.retry"));
-        Assert.assertEquals("\"Start\"", translationMap.get("file.start"));
-    }
+    public void setEmptyI18n_defaultI18nIsPreserved() throws Exception {
+        clickElementWithJs("set-empty-i18n");
```
DiegoCardoso: I wonder if there should be another test that setting an empty i18n object falls back to the default one after another custom i18n object had been set. This would check the behavior and also ensure that this test doesn't work by accident.

sissbruecker:
> **For the tests setting empty i18n objects**: I wonder if there should be another test that setting an empty i18n object falls back to the default one after another custom i18n object had been set. This would check the behavior and also ensure that this test doesn't work by accident.

Would updating the existing tests to first set a custom I18N be good enough then?

DiegoCardoso:
> Would updating the existing tests to first set a custom I18N be good enough then?

That was my initial thought, actually, so I'd say it is enough.

sissbruecker:
Nice, that actually caught some issues:
- The DatePicker connector still had logic for merging the new object with the existing one, thus preventing a reset. Fixed in [bcf690d](https://github.com/vaadin/flow-components/pull/8583/commits/bcf690dc03e016ac0838382a99d66da1c4dbb98a)
- LoginOverlay overwrites public properties with I18N values, but when setting an empty I18N object it can then not restore them. Created a PR for fixing that: https://github.com/vaadin/web-components/pull/11045

PR #8610, opened 2026-02-04
feat: make scale control configurable

sissbruecker:
## Description

Part of https://github.com/vaadin/flow-components/issues/8202

## Type of change

- Feature

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@ prod.bundle
 # Claude Code
 .claude/*.local.*
 CLAUDE.local.md
+.playwright-cli/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -80,6 +80,8 @@ mvn package jetty:run -Dvaadin.pnpm.enable -Dvaadin.frontend.hotdeploy=true -am 
 - Integration test server can be used for testing pages manually using Playwright MCP, if installed
 - Server needs to be restarted after code changes
 - Integration tests can fail if the 8080 port is already in use. At that point stop and ask the user whether to kill the process using that port. If you started the server yourself and want to run tests against it, add `-DskipJetty` to the integration test command.
+- When waiting for the server to start, use `TaskOutput` with `block=false` to poll the background task output for the message "Frontend compiled successfully" rather than using arbitrary sleep commands.
+- When stopping a server that was started as a background task in the current session, use the `TaskStop` tool with the task ID instead of killing the process directly.
 
 ### Code Quality
 
--- a/pom.xml
+++ b/pom.xml
@@ -72,7 +72,7 @@
     </scm>
     <properties>
         <flow.version>25.1-SNAPSHOT</flow.version>
-        <testbench.version>10.1.0-alpha1</testbench.version>
+        <testbench.version>10.1.0-alpha2</testbench.version>
         <maven.compiler.release>21</maven.compiler.release>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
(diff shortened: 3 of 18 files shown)


vaadin-map-flow-parent/vaadin-map-flow/src/main/resources/META-INF/resources/frontend/vaadin-map/synchronization/controls.js
```diff
@@ -11,10 +11,16 @@ export function synchronizeAttribution(target, source) {
...
+  const options = createOptions({
+    minWidth: source.minWidth,
+    maxWidth: source.maxWidth,
+    units: convertEnumValue(source.units),
+    bar: source.displayMode === 'BAR',
+    steps: source.scaleBarSteps,
+    text: source.scaleBarTextVisible
+  });
+  return new ScaleLine(options);
 }
```
sissbruecker, line 24: The only mutable property on `ScaleLine` is `units`. In order to make the other properties configurable on the server, this recreates the control.

I've considered the following alternatives:
- Use internal APIs to update the other props and force a re-render. Got that to work, but might be prone to breaking on an update.
- Adopt the same API from OpenLayers where you can only configure the other options in the constructor. That would require changing the approach with how controls are exposed: users would need to add / remove controls manually through the Map API. The current approach seems more helpful in terms of DX, as it basically just automates that step. I also think exposing a default set of controls is more helpful than requiring users to add them manually, as you basically only need one of each type.

In order to recreate a config element during a sync I also had to make some tweaks to other files:
- Map now always syncs when any control property is changed. This increases the payload a bit, but it's basically only IDs of referenced elements (layers, controls, view).
- The `synchronizeCollection` helper now also checks whether references are the same, and if not updates a collection
- The lookup for synchronized elements has been changed to allow updating its internal map when putting a different element with the same ID into it

vaadin-map-flow-parent/vaadin-map-flow-integration-tests/src/main/java/com/vaadin/flow/component/map/ScaleControlPage.java
```diff
@@ -0,0 +1,60 @@
+/**
+ * Copyright 2000-2025 Vaadin Ltd.
```
DiegoCardoso: Not only here, but there are some other files with `2000-2025` in the copyright header.
```suggestion
 * Copyright 2000-2026 Vaadin Ltd.
```

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/controls/ScaleControl.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    /**
+     * Sets whether to show the scale as proportion below. This is only used
+     * when {@link #setDisplayMode(DisplayMode)} is set to
+     * {@link DisplayMode#BAR}. Default value is {@code false}.
+     *
+     * @param scaleBarTextVisible
+     *            {@code true} to show text above the scale bar
+     */
```
DiegoCardoso, line 166: The Javadoc first says that the scale as proportion is shown below [the scale bar]. Then the `@param` session says it is to show above the scale bar. The doc in OL says it's above (`Render the text scale above of the scalebar. Only applies when bar is true.`), but running the ITs, I am seeing it below the scale bar 🤔

Image

_same comment is valid to the getter above_

vaadin-map-flow-parent/vaadin-map-flow-integration-tests/src/main/java/com/vaadin/flow/component/map/ScaleControlPage.java
sissbruecker: Fixed.

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/configuration/controls/ScaleControl.java
sissbruecker: Fixed the parameter description.

PR #8619, opened 2026-02-05
feat: Add LoginOverlay::bindOpened method

mshabarov:
Related-to https://github.com/vaadin/flow/issues/23196

### DIFF
--- a/vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/LoginOverlay.java
+++ b/vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/LoginOverlay.java
@@ -21,6 +21,7 @@ import java.util.Objects;
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.HasStyle;
 import com.vaadin.flow.component.ModalityMode;
+import com.vaadin.flow.component.SignalPropertySupport;
 import com.vaadin.flow.component.Synchronize;
 import com.vaadin.flow.component.Tag;
 import com.vaadin.flow.component.UI;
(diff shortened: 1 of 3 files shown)


vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/LoginOverlay.java
```diff
@@ -109,15 +113,41 @@ public boolean isOpened() {
...
+     *            the signal to bind the opened state to, or {@code null} to
+     *            unbind
+     * @see #setOpened(boolean)
+     * @since 25.1
+     */
+    public void bindOpened(Signal<Boolean> signal) {
+        openedSupport.bind(signal == null ? null
+                : signal.map(
+                        opened -> opened == null ? Boolean.FALSE : opened));
+        getElement().bindProperty("opened", signal);
```
tltv, line 148: `bindProperty` may cause some unwanted side effect when "opened" property is synced from the client (see Synchronized annotation on isOpened())? 
It will update the signal value as bindProperty is supposed to do with changes from the client. Signal value change triggers `SignalPropertySupport` action which runs `updateOpened(Boolean)` => this didn't happen before, but now change from the client also calls `updateOpened(Boolean)`. That could be wrong.

vaadin-login-flow-parent/vaadin-login-flow/src/main/java/com/vaadin/flow/component/login/LoginOverlay.java
```diff
@@ -109,15 +113,41 @@ public boolean isOpened() {
...
+     * Signal's value {@code null} is treated as {@code false}.
+     *
+     * @param signal
+     *            the signal to bind the opened state to, or {@code null} to
+     *            unbind
+     * @see #setOpened(boolean)
+     * @since 25.1
+     */
+    public void bindOpened(Signal<Boolean> signal) {
+        openedSupport.bind(signal == null ? null
```
tltv, line 145: Both bind calls makes signal depend on LoginOverlay's attached state. Meaning that it's never actually attached via signal.value(true) call. It won't help to bind property to any other element either since its the overlay's property we are interested to bind to.

Maybe this needs a bit different approach (considering my other comments too).

What if for a starters `SignalPropertySupport` is bind to UI instead?

sissbruecker:
Closing as we decided to not proceed with binding opened for now.

PR #8626, opened 2026-02-05
chore: update FontIconIT to verify glyph content instead of dimensions

sissbruecker:
The vaadin-icon web component removed `!important` from `display: inline-flex` in https://github.com/vaadin/web-components/pull/11056, allowing FontAwesome's `display: inline-block` to take effect. This causes `::before` computed dimensions to be `auto` instead of specific pixel values. Update test to verify the correct glyph code point and font family instead.


vaadin-icons-flow-parent/vaadin-icons-flow-integration-tests/src/test/java/com/vaadin/flow/component/icon/tests/FontIconIT.java
sissbruecker, line 1: Not 100% certain what the purpose of the test was. Thus pinging @tomivirkki who added it, maybe you can say whether this was really just to verify that the Font Awesome icon is in some way applied, rather than measuring exact dimensions.

tomivirkki: Looks like the dimensions still match even though there's no straightforward way to obtain them in the test anymore.

PR #8633, opened 2026-02-06
chore: Update testbench.version to 10.0.3

caalador:


pom.xml
sissbruecker, line 1: This results in 72 test failures. I assume it's a regression from https://github.com/vaadin/testbench/pull/2120.

Common issues include:
- Several tests try to click elements behind a client-side modality curtain / overlay. That doesn't work anymore with the new `click` implementation that uses Selenium actions, as the click now only reaches the overlay but not the element behind it.
- Several tests related to validation are failing. The problem seems to be with subsequent clicks: The first click triggers validation and causes an error message to appear on a field. The error message takes up additional space, moving a button further down. The second click then doesn't click on the button it's supposed to, but probably somewhere above. This is a timing issue, waiting after the first click makes the second click work.

The first issue is arguably a test setup problem, but I would guess that this is not uncommon. The second issue seems rather severe. In any case I'd say this change should not be in a patch, nor in a minor release.

Let's also ping @mcollovati who was involved in the TestBench change.

mcollovati: I have seen similar cases also in the Flow validation.
I think it would be better to revert the change and investigate further on how to properly fix the original issue.

I'll have a deeper look on Monday

mcollovati: Just a note for further discussion: I think the previous implementation based on directly invoking the element `click` javascript function was incorrect and potentially hiding bugs, since it does not precisely emulate what a user does on the web page.

The first example given by @sissbruecker exactly shows how a test could pass, but if you do the same steps manually, you will not be able to get the same result.

It is also worth noting that `click` and `doubleClick` were implemented in two different ways: the first called the JavaScript function, and the latter uses Selenium `Action`

For the second example, we should probably improve the new `click` implementation rather than rely on a workaround that effectively is different from clicking twice on a button.

So, this can be a breaking change, but IMO it is actually fixing a bug. And the fix let us discover that we have some "wrong" tests and that the new implementation is probably still a buggy (see the timing issue in clicking twice)

caalador:
Downgraded the update from 10.0.2 to 10.0.1

pom.xml
mcollovati: ping also @TatuLund as he is the original author of the change and can provide more details

mcollovati: Made a [ticket](https://github.com/vaadin/testbench/issues/2128) on Testbench repo. Let's move the discussion there

TatuLund: > In any case I'd say this change should not be in a patch, nor in a minor release.

I agree. The fixed behavior is correct though. It is just that our tests have been exploiting bugs in the old one. This makes me think that there will be customers who have done the same. Thus this needs to be in a minor.

pom.xml
```diff
@@ -70,7 +70,7 @@
     </scm>
     <properties>
         <flow.version>25.0-SNAPSHOT</flow.version>
-        <testbench.version>10.0.0</testbench.version>
+        <testbench.version>10.0.1</testbench.version>
```
mcollovati:
```suggestion
        <testbench.version>10.0.3</testbench.version>
```

mcollovati (changes requested):

pom.xml
mcollovati: Testbench 10.0.3 reverts the `click` change

PR #8652, opened 2026-02-10
feat: add Dialog::bindOpened(Signal<Boolean>)

tltv:
Fixes: vaadin/flow#23441

### DIFF
--- a/vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
+++ b/vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
@@ -45,10 +45,14 @@ import com.vaadin.flow.dom.Element;
 import com.vaadin.flow.dom.ElementConstants;
 import com.vaadin.flow.dom.ElementDetachEvent;
 import com.vaadin.flow.dom.ElementDetachListener;
+import com.vaadin.flow.dom.ElementEffect;
 import com.vaadin.flow.dom.Style;
 import com.vaadin.flow.function.SerializableFunction;
+import com.vaadin.flow.internal.nodefeature.SignalBindingFeature;
 import com.vaadin.flow.shared.Registration;
+import com.vaadin.flow.signals.BindingActiveException;
 import com.vaadin.flow.signals.Signal;
+import com.vaadin.flow.signals.WritableSignal;
 
 /**
  * A Dialog is a small window that can be used to present information and user
@@ -95,6 +99,8 @@ public class Dialog extends Component implements HasComponents, HasSize,
     private DialogHeader dialogHeader;
     private DialogFooter dialogFooter;
     private ModalityMode modality = ModalityMode.VISUAL;
+
+    private boolean valueSetFromSignal;
 
     /**
      * Creates an empty dialog.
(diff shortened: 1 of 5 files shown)


vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
```diff
@@ -1348,4 +1354,71 @@ public <T, S extends Signal<T>> Registration bindChildren(
...
+     *
+     * @param openedSignal
+     *            the writable signal to bind the opened state to, or
+     *            {@code null} to unbind
+     * @throws com.vaadin.flow.signals.BindingActiveException
+     *             if attempting to bind while a binding is already active
+     * @see #setOpened(boolean)
+     * @see #isOpened()
+     */
+    public void bindOpened(WritableSignal<Boolean> openedSignal) {
```
sissbruecker, line 1386: Do I understand this correctly that every dialog with an opened binding will be referenced through the UI forever unless you explicitly remove the binding (which no one will think of doing)? It seems that could accumulate quite a bit of memory usage over time, for example if users follow some repetitive data entry process that requires switching between views, and where one view always initializes a new dialog. It's not just the dialog but also any nested component and related data (grids). And the signal, and any signals that it is derived, from as well.

Apart from that it seems questionable to completely change the semantics / behavior of individual `bind*` methods. Alternatives could be:
- Throw if the component is not attached. Not great, because it requires calls in a specific order.
- Or if we want to support auto-adding, use an API where the developer needs to provide the "scope" element for the binding, which could then be the view. However the scope element would be unnecessary if the developer attaches the dialog manually anyway.

tltv: That's a good point, yes now it's going to waste resources by registering active binding to UI for each new Dialog instance. It's ok with few dialogs, but definitely not with hundreds.
Also this is not taking into account manually added dialogs, where dialog element itself indeed works best as the owner element and that alone fixes memory concern for those cases. I'll update the PR to do that.

Auto-added dialogs needs the UI as owner, but nothing clears the binding after dialog is removed. If we remove the binding automatically when auto-added dialog is detached, then dialog can't open anymore by the signal value change.

Maybe this should be explained in the javadoc and have a warning that binding stays active through the whole UI lifecycle, if you don't choose the "scope" by adding it manually somewhere.

sissbruecker: From experience I can say that most developers do not manually add dialogs to the UI. As such, the default behavior here is still something that we should not do. No one will read the JavaDoc unless something is not working.

We'll have a meeting tomorrow around component signals integration in general, this would be a good topic to pick up and exchange opinions / ideas.

Legioth: Agreed that we shouldn't automatically attach dialogs to make the binding can work because that will lead to accidental memory leaks.

That basically leaves three options for us:
1. Require giving an explicit lifecycle owner component as an additional parameter to the `bindOpened` method.
2. Make `bindOpened` throw if the dialog isn't already explicitly attached to a parent component (without requiring that the parent component is already attached to the UI).
3. Don't provide a `bindOpened` method at all. Using `ComponentEffect` remains a valid option in application code.

Since there's plenty of other things to work on related to signals, I'd suggest we just skip `bindOpened` for now in all the components that may be automatically attached when opening. We can get back to picking one way or another later when we have have a better understanding of how people want to use the feature.

sissbruecker:
Closing as we decided to not proceed with binding opened for now.

Issue #4497, opened 2023-01-10
Grid.scrollToEnd does not work if called before attaching


mstahv:
### Description

scrollToEnd method fails if called before the component is attached DOM. Quite typical use case: building a log view where grid is initialised in the constructor.

### Expected outcome

scrollToEnd should work

### Minimal reproducible example

```
@Route
public class ScrollToEndView extends VerticalLayout {
    public ScrollToEndView() {
        Grid<String> grid = new ProperGrid();
        grid.addColumn(s -> s);
        grid.setItems(IntStream.range(0,1000).mapToObj(i -> "" + i).collect(Collectors.toList()));
        grid.scrollToEnd();
        add(grid);
    }
}
```

### Steps to reproduce

1. Open view

See that grid is not scrolled to end

### Environment

Vaadin: 23.3.2
Flow: 23.3.2
Java: Eclipse Adoptium 19.0.1
OS: aarch64 Mac OS X 13.0.1
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36
Live reload: Java active (Spring Boot Devtools): Front end active

PR #8664, opened 2026-02-11
fix: make Grid scrollToEnd work when called before attach

vursen:
## Summary

Calling `scrollToEnd` before the Grid component was attached would silently discard the call. Changed the implementation to use `runWhenAttached` to defer the JS execution until attach. This fix applies the same pattern as #8653 (TreeGrid).

Fixes https://github.com/vaadin/flow-components/issues/4497

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridScrollTest.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridScrollTest.java
@@ -19,16 +19,18 @@ import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.stream.IntStream;
 
-import org.junit.Assert;
 import org.junit.After;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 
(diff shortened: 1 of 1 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridScrollTest.java
```diff
@@ -23,14 +23,19 @@
...
 
+    private UI ui;
     private Grid<String> grid;
 
     @Before
     public void setUp() {
+        ui = new MockUI();
         grid = new Grid<>();
         grid.setPageSize(50);
     }
```
vursen, line 44:
```suggestion
    }
    
    @After
    public void tearDown() {
        UI.setCurrent(null);
    }
```

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridScrollTest.java
```diff
@@ -23,14 +23,19 @@
 import org.junit.Before;
```
vursen, line 24:
```suggestion
import org.junit.After;
import org.junit.Before;
```

PR #8666, opened 2026-02-11
fix!: ensure only last Grid scroll invocation is sent to client

### LINKED ISSUE #5880: [Grid] scrollToEnd is always executed after scrollToStart regardless of their call order
### Description

`scrollToEnd` gets always executed after `scrollToStart` regardless of their call order in the code.

### Expected outcome

When `scrollToStart` comes after `scrollToEnd`, I'd expect the grid to scroll to the end and then to the start.

```java
@Route("grid")
    public GridView() {
        grid.addColumn(item -> item);
        grid.setItems(IntStream.range(0, 1000).mapToObj(i -> "Item " + i).collect(Collectors.toList()));

        Button scrollToEndAndStart = new Button("scroll to end and start", event -> {
            grid.scrollToEnd();
            grid.scrollToStart();
        });

        add(grid, scrollToEndAndStart);
    }
}
```

### Steps to reproduce

1. Click "scroll to end and start"
2. See that the grid's scroll position is at the end while I'd expect it at the start

### Environment

Vaadin version(s): 24.4 and earlier
OS: Mac OS

### Browsers

_No response_

vursen:
When multiple Grid scroll methods were called in the same round-trip (e.g. `scrollToEnd()` followed by `scrollToStart()`), `scrollToEnd` was always executed last regardless of call order. This happened because `scrollToEnd` used `beforeClientResponse` to defer its execution, while `scrollToIndex`/`scrollToStart` executed immediately so the deferred `scrollToEnd` always won.

This PR adds a `scheduleScrollExecution` helper in Grid that defers scroll actions via `runWhenAttached` + `beforeClientResponse` and cancels any previously scheduled scroll, so only the last call in a round-trip takes effect. All Grid and TreeGrid scroll methods now use this helper, which eliminates the call order issue.

Also, tests have been reorganized and extended to cover deduplication behavior for both Grid and TreeGrid.

> [!WARNING]
> This change can be considered behavior-altering since it changes the timing of `scrollToIndex` and `scrollToStart`.

Fixes #5880

### DIFF
--- a/vaadin-details-flow-parent/vaadin-details-flow/src/main/java/com/vaadin/flow/component/details/Details.java
+++ b/vaadin-details-flow-parent/vaadin-details-flow/src/main/java/com/vaadin/flow/component/details/Details.java
@@ -16,9 +16,12 @@
 package com.vaadin.flow.component.details;
 
 import java.util.Collection;
+import java.util.List;
+import java.util.Objects;
 import java.util.stream.Stream;
 
 import com.vaadin.flow.component.Component;
+import com.vaadin.flow.component.ComponentEffect;
 import com.vaadin.flow.component.ComponentEvent;
 import com.vaadin.flow.component.ComponentEventListener;
 import com.vaadin.flow.component.HasComponents;
@@ -32,7 +35,11 @@ import com.vaadin.flow.component.html.Span;
 import com.vaadin.flow.component.shared.HasThemeVariant;
 import com.vaadin.flow.component.shared.HasTooltip;
 import com.vaadin.flow.component.shared.SlotUtils;
+import com.vaadin.flow.function.SerializableFunction;
+import com.vaadin.flow.internal.nodefeature.SignalBindingFeature;
 import com.vaadin.flow.shared.Registration;
+import com.vaadin.flow.signals.BindingActiveException;
+import com.vaadin.flow.signals.Signal;
 
 /**
  * Details is an expandable panel for showing and hiding content from the user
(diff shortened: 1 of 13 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -3305,6 +3308,26 @@ private void callJsFunctionBeforeClientResponse(String functionName,
...
+    /**
+     * Schedules a scroll-related action to be executed before the next client
+     * response. Cancels any previously scheduled scroll action, ensuring that
+     * only the last scroll call takes effect.
+     *
+     * @param action
+     *            the scroll action to execute
+     */
+    protected void scheduleScrollExecution(SerializableRunnable action) {
+        getElement().getNode().runWhenAttached(ui -> {
```
tomivirkki, line 3320: Would possibly be cleaner to store the SerializableRunnable action as such here synchronously (`this.pendingScrollAction = action`). Then you won't need to `remove()` any scheduled beforeClientResponse registrations.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -3305,6 +3308,26 @@ private void callJsFunctionBeforeClientResponse(String functionName,
...
+            if (pendingScrollRegistration != null) {
+                pendingScrollRegistration.remove();
+            }
+            pendingScrollRegistration = ui.beforeClientResponse(this, ctx -> {
+                action.run();
```
tomivirkki, line 3325: Grid may no longer be attached at this point. Maybe something like this would work:

```java
if (!grid.isAttached()) {
  scheduleScrollExecution(pending...)
} else {
 ....
}

```

vursen: Does the new implementation actually change this behavior? Wasn't this already how the scroll methods worked?

tomivirkki: I tested locally and my assumption was wrong. The current implementation actually handles this already. LGTM

PR #8667, opened 2026-02-11
feat: upload manager separate MIME type and extension methods

tomivirkki:
## Description

Fixes https://github.com/orgs/vaadin/projects/103/views/3?pane=issue&itemId=156270644

## Type of change

Feature

### DIFF
--- a/vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/UploadManager.java
+++ b/vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/UploadManager.java
@@ -640,7 +640,7 @@ public class UploadManager implements Serializable {
      */
     @Tag("vaadin-upload-manager-connector")
     @JsModule("./vaadin-upload-manager-connector.ts")
-    @NpmPackage(value = "@vaadin/upload", version = "25.1.0-alpha6")
+    @NpmPackage(value = "@vaadin/upload", version = "25.1.0-alpha7")
     static class Connector extends Component {
         @Override
         protected void onAttach(AttachEvent attachEvent) {
--- a/vaadin-accordion-flow-parent/vaadin-accordion-flow/src/main/java/com/vaadin/flow/component/accordion/Accordion.java
+++ b/vaadin-accordion-flow-parent/vaadin-accordion-flow/src/main/java/com/vaadin/flow/component/accordion/Accordion.java
@@ -52,7 +52,7 @@ import com.vaadin.flow.shared.Registration;
  * @author Vaadin Ltd
  */
 @Tag("vaadin-accordion")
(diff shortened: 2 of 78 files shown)


vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/UploadManager.java
```diff
@@ -249,43 +261,215 @@ public long getMaxFileSize() {
...
+    public void setAcceptedMimeTypes(String... mimeTypes) {
+        if (mimeTypes == null || mimeTypes.length == 0) {
+            acceptedMimeTypes = List.of();
+        } else {
+            for (String mimeType : mimeTypes) {
+                if (mimeType == null || mimeType.isBlank()) {
+                    throw new IllegalArgumentException(
+                            "MIME types cannot contain null or blank values");
+                }
+                if (!mimeType.contains("/")) {
```
ugur-vaadin, line 292: Should we add a test for this case as well?

ugur-vaadin:
It looks like the wrapper does not delegate the default methods from `ElementRequestHandler`.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/UploadManager.java
tomivirkki: It should already be covered by test `setAcceptedMimeTypes_withoutSlash_throws`

ugur-vaadin: Missed that one, my bad.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/UploadManager.java
```diff
@@ -249,43 +261,215 @@ public long getMaxFileSize() {
...
+            }
+        }
+        return false;
+    }
+
+    /**
+     * Checks whether an actual MIME type matches a pattern. Supports exact
+     * match and wildcard patterns like {@code "image/*"}.
+     */
+    private static boolean matchesMimeType(String actual, String pattern) {
```
ugur-vaadin, line 470: I wonder if this should match if "actual" mime type contains parameters. Though I'm not sure if that is possible in the framework. What do you think?

tomivirkki: Doesn't hurt to handle this case anyway. Done.

tomivirkki:
> It looks like the wrapper does not delegate the default methods from ElementRequestHandler.

Good catch. Added a test to make sure all the methods are delegated (in case new ones get added later).

PR #8675, opened 2026-02-12
chore: add checkstyle rule for UI.setCurrent [skip ci]

sissbruecker:
Adds the Maven checkstyle plugin to the project to check for code style issues. Currently this only configures a single custom rule which checks that `UI.setCurrent` is used properly in unit tests. The rule ensures that:
- the UI used in the method call is stored in a class field, so that the UI can not be garbage collected during test runs
- The method is not called from a static method, which is potentially invalid as `CurrentInstanceCleaner` from `flow-test-generic` clears the current UI before each test

The checks are currently stricter than they need to be:
- Passing a local variable as argument is also fine as long as `UI.setCurrent` is called in a test method for example.
- Using static methods is also fine as long as they are called within the test method that needs the UI.

However adding support for those cases would add quite a bit of complexity to the check and may lead to some undetected edge-cases. Going with a stricter approach makes things easier, and also doesn't have a large negative impact on the way tests are written. There is also the option to disable the check for a specific scope, which is done in one case in this PR where it would otherwise require a bit more refactoring.

To run the check:
```
mvn install -pl vaadin-flow-components-shared-parent/vaadin-flow-components-checkstyle-rules
mvn checkstyle:check
```

Alternatively, run the script that combines and logs errors from all modules:
```
./scripts/checkstyle.sh
```

CI is configured to run the checkstyle script if it is present in the current branch. When it detects a failure, it adds a comment to the PR with all detected errors, similar to the spotless check:

Bildschirmfoto 2026-02-17 um 09 51 28

### DIFF
--- a/checkstyle/checkstyle.xml
+++ b/checkstyle/checkstyle.xml
@@ -7,7 +7,10 @@
     <property name="severity" value="error"/>
     <property name="fileExtensions" value="java"/>
 
+    <module name="SuppressWarningsFilter"/>
+
     <module name="TreeWalker">
         <module name="com.vaadin.checkstyle.UiSetCurrentCheck"/>
+        <module name="SuppressWarningsHolder"/>
     </module>
 </module>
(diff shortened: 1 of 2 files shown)


vursen (commented):
What about cases like

```java
@Before
public void setUp() {
  new MockUI();
}
```

Shouldn't they also be reported? MockUI sets itself as the current UI internally, and if nothing holds a reference to it, I guess it can be garbage collected as well?

sissbruecker:
Let me try a different approach. Instead of having multiple ways of setting up UIs, and repeating the setup code in every test, we should use a JUnit rule that handles setting up the UI. Then maybe add a checkstyle rule to verify that no test uses `UI.setCurrent` and uses the rule instead.

PR #8683, opened 2026-02-13
refactor: allow setting Slider properties in any order

vursen:
## Summary

- Slider setters no longer throw exceptions or modify values automatically, so users can call them in any order. This is necessary because Signals won't guarantee the order in which properties are updated.
- Instead, added deferred `beforeClientResponse` consistency check in `SliderBase` that warns when properties end up in an incompatible state (min > max, value out of range, value not aligned with step).
- Removed several Slider constructors that accepted multiple double parameters, as they were difficult to read. Users should now use explicit setters such as setMin, setMax, setStep, and setValue, which is better for clarity.

### DIFF
--- a/vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/UploadManager.java
+++ b/vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/UploadManager.java
@@ -94,10 +94,10 @@ public class UploadManager implements Serializable {
      * work.
      *
      * @param owner
-     *            the component that owns this manager. The manager's lifecycle
-     *            is tied to the owner's lifecycle - when the owner is detached
-     *            from the UI, uploads will stop working. The owner is typically
-     *            the view or layout containing the upload UI components.
+     *            The manager's lifecycle is tied to the owner's lifecycle -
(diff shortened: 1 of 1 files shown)


vursen:
The compilation errors are caused by a Signal API change and should be resolved by https://github.com/vaadin/flow-components/pull/8686.

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/SliderBase.java
```diff
@@ -331,4 +251,61 @@ private void applyChangeTimeout() {
...
+                            Invalid configuration: value ({}) is not aligned with step (min={}, max={}, step={}). \
+                            Set only values of the form (min + N * step) to avoid an inconsistent UI state.
+                            """,
+                    value, min, max, step);
+        }
+    }
+
+    abstract boolean isValueAlignedWithStep(TValue value);
+
+    abstract boolean isValueWithinMinMax(TValue value);
```
sissbruecker: It seems like all package-private methods in this class (also see the ones above, plus constructor) could be `protected`.

vursen: Alright, made them protected.

PR #8696, opened 2026-02-16
fix: get upload handler configured state from manager

ugur-vaadin:
## Description

The `target` attribute is also set when the manager is initialized without a handler using the internal `FailFastUploadHandler`. The helper always returns true in that case. This PR fixes this by getting the initialized state from the manager directly via a package-private method.

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -364,6 +364,15 @@
                         </excludes>
                         <skip>${skipUnitTests}</skip>
                     </configuration>
+                    <dependencies>
+                        <!-- Configure JUnit 4 provider explicitly to not depend on Surefire's auto-resolution when JUnit 6 is on the class path. -->
+                        <!-- JUnit 6 is pulled in transitively by modules that use Flow test dependencies -->
+                        <dependency>
+                            <groupId>org.apache.maven.surefire</groupId>
+                            <artifactId>surefire-junit4</artifactId>
+                            <version>${maven.surefire.plugin.version}</version>
+                        </dependency>
+                    </dependencies>
                 </plugin>
                 <plugin>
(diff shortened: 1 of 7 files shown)


vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/UploadManager.java
```diff
@@ -663,6 +669,15 @@ private void checkFeatureFlag(com.vaadin.flow.component.UI ui) {
...
 
+    /**
+     * Returns whether an UploadHandler is explicitly configured.
+     * <p>
+     * Intended only for internal use.
+     */
+    boolean isHandlerExplicitlyConfigured() {
+        return handlerExplicitlyConfigured.get();
+    }
+
```
tomivirkki, line 680: You could avoid the `handlerExplicitlyConfigured` field and the new API with:
```java
if (!(handler instanceof FailFastUploadHandler)) {
    connector.getElement().setAttribute("target", elementStreamResource);
}
```

ugur-vaadin: Will the `FailFastUploadHandler` be still set to throw `IllegalStateException` on upload request else in that case?

tomivirkki: Nope, it wouldn't. Nevermind

tomivirkki (approved):

PR #8723, opened 2026-02-18
refactor: error message string to enum in UploadManager FileRejectedEvent

tomivirkki:
## Description

- `UploadManager.FileRejectedEvent.getErrorMessage()` was misleading because it returned raw client-side error codes like `"fileIsTooBig"` rather than actual human-readable error messages. Replaced the string with a new `FileRejectionReason` enum that accurately represents the fixed set of rejection reasons.
- `getErrorMessage()` is replaced by `getReason()` returning a `FileRejectionReason` enum value (`TOO_MANY_FILES`, `FILE_TOO_LARGE`, `INCORRECT_FILE_TYPE`)

## Type of change

Refactor


tomivirkki:
> I was wondering whether it would be a good idea to have a default unknown error enum as fallback.

Done. It's better than a null.

PR #8731, opened 2026-02-18
fix: prevent 404 for images on locked cells in Spreadsheet

### LINKED ISSUE #8656: Spreadsheet loses images in frozen rows in the first displayed sheet
### Description

The fix for #8428 corrects most use cases:
* workbook sheets without freeze panes
* switching between sheets, with or without freeze panes
* scrolling past the buffered rows and back

However, it still loses track of images:
* when the first displayed sheet in a workbook contains frozen rows, and images are in those frozen rows (i.e. header images)

### Expected outcome

Embedded images always display regardless of which cell they are attached to, whether it is in a frozen pane or not, on initial display and subsequent navigation between and around worksheets in the same workbook.

sample workbook available via direct channels, can't post the easiest one publicly.

### Steps to reproduce

display the attached workbook in the Spreadsheet component, and note images do not initially load as the first sheet has them in a freeze pane.

### Environment

Vaadin version(s): 24.9.10
OS: N/A

### Browsers

Issue is not browser related

DiegoCardoso:
## Summary
- Guard `setResource()` in `addOverlayData()` with a `resources.containsKey()` check to avoid re-registering a stream resource that is already active
- During initial load, two `onSheetScroll` events cause `addOverlayData()` to be called twice for the same overlay. The second `setResource()` call unregisters the first URI, so the client's in-flight request returns 404
- Adds integration test `ImageIT#shouldLoadImageOnLockedCell` that loads a spreadsheet with an image on a locked cell and verifies no 404 errors in browser console

Fixes #8656

## Test plan
- [ ] Run `ImageIT#shouldLoadImageOnLockedCell` — loads `image-locked-cell.xlsx` and asserts no 404 in browser logs
- [ ] Run all `ImageIT` tests — confirm no regressions, especially `shouldUpdateImageURIOnReattach` (scroll-out/scroll-in re-registration path)
- [ ] Manual: start IT server, load `image-locked-cell.xlsx` with CPU/Network throttling, confirm image loads


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-integration-tests/src/test/java/com/vaadin/flow/component/spreadsheet/test/ImageIT.java
```diff
@@ -105,6 +105,18 @@ public void shouldUpdateImageURIOnReattach() {
                 originalSrc, newSrc);
     }
 
+    @Test
+    public void shouldLoadImageOnLockedCell() {
```
sissbruecker, line 109: The test seems to always pass without the fix.

DiegoCardoso: The bug itself seemed a bit inconsistent when I was working on it. The test is a best effort to check it works and prevent any regressions.

sissbruecker (approved):

PR #8733, opened 2026-02-18
feat: add bindDrawerOpened to AppLayout

Artur-:


vaadin-app-layout-flow-parent/vaadin-app-layout-flow/src/main/java/com/vaadin/flow/component/applayout/AppLayout.java
```diff
@@ -145,6 +147,25 @@ public void setDrawerOpened(boolean drawerOpened) {
...
+     * property changes push back to the signal.
+     * <p>
+     * While a signal is bound, any attempt to set the drawer opened state
+     * manually throws {@link com.vaadin.flow.signals.BindingActiveException}.
+     *
+     * @param signal
+     *            the writable signal to bind, not {@code null}
+     * @since 25.1
+     */
+    public void bindDrawerOpened(Signal<Boolean> signal,
```
DiegoCardoso: If I understood it correctly, the two-way binding is not done automatically, right? If so, I feel that the javadoc is misleading. Should it even say "writable signal"? It also doesn't mention the `writeCallback` at all.

Artur-: Updated

PR #8739, opened 2026-02-20
feat: add signal bind methods to crud component

Artur-:


vaadin-crud-flow-parent/vaadin-crud-flow/src/test/java/com/vaadin/flow/component/crud/CrudSignalTest.java
sissbruecker, line 1: Please write proper tests that verify that the executeJs call happens.

PR #8742, opened 2026-02-20
feat: add signal bind methods to grid component

Artur-:


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridSingleSelectionModel.java
```diff
@@ -143,6 +148,45 @@ public Registration addValueChangeListener(
...
+                SignalBindingFeature feature = getElement().getNode()
+                        .getFeature(SignalBindingFeature.class);
+
+                if (feature.hasBinding(SignalBindingFeature.VALUE)) {
+                    throw new BindingActiveException();
+                }
+
+                boolean[] fromSignal = { false };
+
+                Registration effectReg = ElementEffect.bind(getElement(),
```
sissbruecker: How is this cleaned up when you switch to a different selection model?

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridMultiSelectionModel.java
```diff
@@ -294,6 +299,45 @@ public Element getElement() {
...
             }
+
+            @Override
+            public void bindValue(Signal<Set<T>> valueSignal,
+                    SerializableConsumer<Set<T>> writeCallback) {
+                Objects.requireNonNull(valueSignal, "Signal cannot be null");
+                SignalBindingFeature feature = getElement().getNode()
+                        .getFeature(SignalBindingFeature.class);
+
+                if (feature.hasBinding(SignalBindingFeature.VALUE)) {
```
sissbruecker: It seems wrong to bind the "value" of a Grid element to the selection?

It also feels like there is some missing infrastructure here if implementing a single bind method in a component requires 40 lines of code.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridSingleSelectionModel.java
heruan: Good catch, it wasn't. I've added a `removeBinding(VALUE)` call in the `remove()` method of both selection models, so the signal binding (effect + listener) is properly cleaned up when the selection model is replaced via `setSelectionMode()`. Added also tests to verify that after switching modes, the old signal no longer affects the grid and a new binding can be created.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridMultiSelectionModel.java
heruan: > It seems wrong to bind the "value" of a Grid element to the selection?

It's the standard key used by `HasValue.bindValue` implementations, and from the `asSingleSelect()`/`asMultiSelect()` perspective the selection is the grid's value. The key is just an identifier in `SignalBindingFeature` to track the binding.

> It also feels like there is some missing infrastructure here if implementing a single bind method in a component requires 40 lines of code.

I've extracted the shared logic into a package-private `GridSelectionSignalHelper.bindValue()` that both selection models delegate to, which also fixes the SonarQube duplication warning. The implementation follow what's done in `AbstractFieldSupport`, but yes, a default re-usable framework implementation would be better.

heruan:
Also added `getSelectedItemSignal()` and `getSelectedItemsSignal()` convenience methods on `Grid`, `GridSingleSelectionModel`, and `GridMultiSelectionModel`. These create a `ValueSignal` internally, bind it to the selection, so callers can get a signal representing the grid's selection state without managing their own signal and `bindValue()` call.

Legioth:
> Also added `getSelectedItemSignal()` and `getSelectedItemsSignal()` convenience methods

I'm not sure that's a pattern we want to introduce without considering how it relates to similar cases with the value in e.g. a text field. I did some initial exploration of that concept a while ago and I ended up rejecting it because it became weird. Unfortunately, I don't remember exactly what the problem was.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridMultiSelectionModel.java
```diff
@@ -101,6 +104,8 @@ private void insertSelectionColumn(Grid<T> grid,
     @Override
     protected void remove() {
         super.remove();
+        getGrid().getElement().getNode().getFeature(SignalBindingFeature.class)
+                .removeBinding(SignalBindingFeature.VALUE);
```
sissbruecker: `multiSelect_bindValue_switchToSingle_bindingCleanedUp` still passes when I remove this.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridSingleSelectionModel.java
```diff
@@ -176,6 +186,8 @@ public void generateData(T item, ObjectNode jsonObject) {
     @Override
     protected void remove() {
         super.remove();
+        getGrid().getElement().getNode().getFeature(SignalBindingFeature.class)
+                .removeBinding(SignalBindingFeature.VALUE);
```
sissbruecker: `singleSelect_bindValue_switchToMulti_bindingCleanedUp` still passes when I remove this.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/GridSelectionSignalHelper.java
```diff
@@ -0,0 +1,90 @@
...
+            SerializableConsumer<V> writeCallback) {
+        Objects.requireNonNull(valueSignal, "Signal cannot be null");
+        SignalBindingFeature feature = target.getElement().getNode()
+                .getFeature(SignalBindingFeature.class);
+
+        if (feature.hasBinding(SignalBindingFeature.VALUE)) {
+            throw new BindingActiveException();
+        }
+
+        boolean[] fromSignal = { false };
```
sissbruecker, line 79: Tests still pass if I remove this.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -2221,9 +2285,46 @@ public void addColumns(String... propertyNames) {
...
+     * {@link #setColumns(String...)} throws a
+     * {@link com.vaadin.flow.signals.BindingActiveException}.
+     *
+     * @param signal
+     *            the signal providing the list of property names to create
+     *            columns for, not {@code null}
+     * @see #setColumns(String...)
+     * @since 25.1
+     */
+    public void bindColumns(Signal<List<String>> signal) {
```
sissbruecker: I can still use all of these (and many more) to get the signal state out of sync with the grid:
- `addColumn`
- `addColumns`
- `removeColumn`
- `removeAllColumns`

Same for the selection, header and footer text. 

What is the plan here? Should a component really prevent any imperative API from working if a binding is active? Is it a user error and we ignore? Is `setColumns` not working just accidental due to using `SignalPropertySupport`?

heruan: As discussed in Slack, I'm going to revert the `bindColumns` method for now: once a signal is bound, that's the source of truth and imperative methods shouldn't work.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridMultiSelectionModel.java
heruan: Right, tests pass trivially because after mode switch, the old signal writes to the disconnected old model, which never affects the new model regardless of cleanup. Tests now switch back to the original mode and verify the old effect doesn't fire selection events.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/GridSelectionSignalHelper.java
heruan: Correct, the signal's own equality check prevented an actual infinite loop, so existing tests didn't cover the guard. Added now tests that explicitly count calls.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
sissbruecker: There's more:
```java
column.setHeader(new Span("foo"));
grid.getHeaderRows().get(0).getCell(column).setText("foo");
grid.getHeaderRows().get(0).getCell(column).setComponent(new Span("foo"));
column.setFooter(new Span("foo"));
grid.getFooterRows().get(0).getCell(column).setText("foo");
grid.getFooterRows().get(0).getCell(column).setComponent(new Span("foo"));
```

To be honest this all feels a bit rushed. I'd be in favor of dropping everything except the selection maybe.

heruan: > To be honest this all feels a bit rushed. I'd be in favor of dropping everything except the selection maybe.

Dropped everything except the selection binding. Also updated the code to work with latest upstream changes.

heruan:
@sissbruecker could you review this now that is updated with latest Flow changes?

PR #8748, opened 2026-02-20
chore: add JUnit rule to enable feature flags during tests

sissbruecker:
## Description

Adds a JUnit rule that sets up mocks to make feature flag checks pass during tests and applies it in all tests that currently set up their own mocks.

## Type of change

- Internal


vursen (approved):

vursen:
I guess one more candidate for this cleanup could be Logger.

PR #8798, opened 2026-02-27
feat: add shared size theme variants for Avatar

tomivirkki:
## Description

Add shared size theme variants for Avatar

Depends on https://github.com/vaadin/web-components/pull/11218

## Type of change

Feature


tomivirkki:
> We could also add SMALL for button and all field components as it's now supported.

Opened a separate [PR](https://github.com/vaadin/flow-components/pull/8805)

web-padawan (approved):

PR #8803, opened 2026-02-27
fix: Override bind items for combobox

mshabarov:
ComboBox requires filter convertor and thus fails for generic data provider. This PR adds:
- overload for bindItems that takes the converter
- overload for bindItems that uses default items filter (uses items label generator and contains)
- unit tests for both methods
- E2E test.

E2E test uses a custom equality checker for value signal and re-uses the `Person` entity class which equality is based in ID: `ComboBox` uses ID for refreshing items, whereas item signals should use some other equality (first name + last name), because otherwise signals skip calling the effects and value updates if they see equal values.

Related-to https://github.com/vaadin/flow/issues/23659

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -88,4 +88,7 @@ mvn package jetty:run -Dvaadin.pnpm.enable -Dvaadin.frontend.hotdeploy=true -am 
 ```sh
 # Format code
 mvn spotless:apply
+
+# Run checkstyle validation
+mvn checkstyle:check
 ```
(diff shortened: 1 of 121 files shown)


DiegoCardoso (approved):
Adding this test fails when it shouldn't:
```java
    @Test(expected = BindingActiveException.class)
    public void bindItems_thenSetItems_throws() {
        var listSignal = new ValueSignal<>(List.of(new ValueSignal<>("Alice")));

        comboBox.bindItems(listSignal);
        ui.add(comboBox);

        comboBox.setItems("Bob");
    }
```

But, as discussed internally, it should be addressed separately.

PR #8815, opened 2026-03-02
chore: use MockUIRule in date / time picker unit tests

sissbruecker:
Update `DatePicker` / `TimePicker` / `DateTimePicker` unit tests to use `MockUIRule`.


vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow/src/test/java/com/vaadin/flow/component/datetimepicker/HasLabelTest.java
sissbruecker, line 1: This is already covered by a similar test for the `InputField` interface, which also includes `HasLabel`.

PR #8847, opened 2026-03-04
feat: add separate upload accepted extensions and mime types api

### LINKED ISSUE #8779: Vaadin 24 Upload Component setAcceptedFileTypes(...) only validating on client side
### Description of the bug

The file extension is only validated on client side when uploading a file via the vaadin upload component.

### Expected behavior

Validation on client and server side. At least for the first POST request to the server.

1. Create Upload component and setAcceptedFileTypes to pdf.
2. Start interception in a proxy tool such as Burp Suite
3. Select any PDF file and click on Upload
4. Intercept the POST request made to server upload endpoint
5. Replace the file content, file name and file type for example add an HTML code in the request for the POST request body and send the request
6. You will notice that the HTML file will get upload in the server

### Versions

- Vaadin / Flow version: 24.4.13
- Java version: 17

ugur-vaadin:
## Description

This PR:

- Extracts static methods shared between `Upload` and `UploadManager` regarding accepted file types and MIME types to the `UploadHelper` class
- Deprecates `setAcceptedFileTypes` and `getAcceptedFileTypes` for `Upload`
- Introduces new API for `Upload`:
  - `setAcceptedMimeTypes`
  - `getAcceptedMimeTypes`
  - `setAcceptedFileExtensions`
  - `getAcceptedFileExtensions` 

Fixes #8779 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/.claude/skills/migrate-junit6/SKILL.md
+++ b/.claude/skills/migrate-junit6/SKILL.md
@@ -35,6 +35,22 @@ MockUIExtension ui = new MockUIExtension();
 
 Import: `com.vaadin.tests.MockUIRule` -> `com.vaadin.tests.MockUIExtension`
 
+## EnableFeatureFlagRule -> EnableFeatureFlagExtension
+
+```java
+// Before
+@Rule
+public EnableFeatureFlagRule featureFlagRule = new EnableFeatureFlagRule(
+        MyFeatureFlagProvider.MY_FEATURE);
+
+// After
+@RegisterExtension
+EnableFeatureFlagExtension featureFlagExtension = new EnableFeatureFlagExtension(
+        MyFeatureFlagProvider.MY_FEATURE);
+```
+
+Import: `com.vaadin.tests.EnableFeatureFlagRule` -> `com.vaadin.tests.EnableFeatureFlagExtension`
+
 ## Assertions
 
 Replace all `Assert.` with `Assertions.`. Method signatures are the same, except for **message-carrying overloads** where the message moves to the **last** parameter:
@@ -66,6 +82,8 @@ public void throwsException() {
     Assertions.assertThrows(IllegalStateException.class, () -> doSomething());
 }
 ```
+
+Only place the `assertThrows` call around the statement that is expected to throw, not multiple statements.
 
 ## @Ignore -> @Disabled
 
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/pom.xml
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/pom.xml
@@ -25,11 +25,6 @@
     <dependencies>
         <dependency>
             <groupId>com.vaadin</groupId>
-            <artifactId>flow-test-generic</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>com.vaadin</groupId>
             <artifactId>vaadin-flow-components-base</artifactId>
             <version>${project.version}</version>
         </dependency>
@@ -68,11 +63,6 @@
         <dependency>
             <groupId>jakarta.servlet</groupId>
             <artifactId>jakarta.servlet-api</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/test/java/com/vaadin/flow/component/ai/AIComponentsSerializableTest.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/test/java/com/vaadin/flow/component/ai/AIComponentsSerializableTest.java
@@ -21,10 +21,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Stream;
 
-import org.junit.Assert;
-import org.junit.Before;
-import org.junit.Rule;
-import org.junit.Test;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
 import org.mockito.ArgumentCaptor;
 import org.mockito.Mockito;
 
@@ -33,18 +33,18 @@ import com.vaadin.flow.component.ai.common.ChatMessage;
 import com.vaadin.flow.component.ai.orchestrator.AIOrchestrator;
 import com.vaadin.flow.component.ai.provider.LLMProvider;
 import com.vaadin.flow.testutil.ClassesSerializableTest;
-import com.vaadin.tests.EnableFeatureFlagRule;
-import com.vaadin.tests.MockUIRule;
+import com.vaadin.tests.EnableFeatureFlagExtension;
+import com.vaadin.tests.MockUIExtension;
 
 import reactor.core.publisher.Flux;
 
-public class AIComponentsSerializableTest extends ClassesSerializableTest {
+class AIComponentsSerializableTest extends ClassesSerializableTest {
 
-    @Rule
-    public EnableFeatureFlagRule featureFlagRule = new EnableFeatureFlagRule(
+    @RegisterExtension
+    EnableFeatureFlagExtension featureFlagExtension = new EnableFeatureFlagExtension(
             AIComponentsFeatureFlagProvider.AI_COMPONENTS);
-    @Rule
-    public MockUIRule ui = new MockUIRule();
+    @RegisterExtension
+    MockUIExtension ui = new MockUIExtension();
 
     private LLMProvider mockProvider;
 
@@ -61,14 +61,13 @@ public class AIComponentsSerializableTest extends ClassesSerializableTest {
                 "com\\.vaadin\\.flow\\.component\\.ai\\.orchestrator\\.AIOrchestrator\\$Builder"));
     }
 
-    @Before
-    public void setUp() {
+    @BeforeEach
+    void setUp() {
         mockProvider = Mockito.mock(LLMProvider.class);
     }
 
     @Test
-    public void serialization_roundTrip_reconnectRestoresProvider()
-            throws Throwable {
+    void serialization_roundTrip_reconnectRestoresProvider() throws Throwable {
         var orchestrator = AIOrchestrator.builder(mockProvider, null).build();
         var deserialized = serializeAndDeserialize(orchestrator);
 
@@ -84,7 +83,7 @@ public class AIComponentsSerializableTest extends ClassesSerializableTest {
     }
 
     @Test
-    public void serialization_roundTrip_preservesConversationHistory()
+    void serialization_roundTrip_preservesConversationHistory()
             throws Throwable {
         var history = List.of(
                 new ChatMessage(ChatMessage.Role.USER, "Hello", "msg-1",
@@ -97,15 +96,15 @@ public class AIComponentsSerializableTest extends ClassesSerializableTest {
 
         var deserialized = serializeAndDeserialize(orchestrator);
         var restored = deserialized.getHistory();
-        Assert.assertEquals(2, restored.size());
-        Assert.assertEquals("Hello", restored.getFirst().content());
-        Assert.assertEquals("msg-1", restored.getFirst().messageId());
-        Assert.assertEquals("Hi there", restored.get(1).content());
+        Assertions.assertEquals(2, restored.size());
+        Assertions.assertEquals("Hello", restored.getFirst().content());
+        Assertions.assertEquals("msg-1", restored.getFirst().messageId());
+        Assertions.assertEquals("Hi there", restored.get(1).content());
     }
 
     @SuppressWarnings("unchecked")
     @Test
-    public void serialization_roundTrip_reconnectRestoresHistoryOnProvider()
+    void serialization_roundTrip_reconnectRestoresHistoryOnProvider()
             throws Throwable {
         var history = List.of(
                 new ChatMessage(ChatMessage.Role.USER, "Hello", "msg-1", null),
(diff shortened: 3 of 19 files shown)


vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -347,37 +448,64 @@ public boolean isDropAllowed() {
...
-     * File format restrictions are checked only on the client side (browser).
-     * They indicate the hints for users as to what file types to upload. Using
-     * this method won’t restrict the uploaded file’s format on the server side.
-     * If required, it’s the responsibility of the application developer to
-     * implement application-specific restrictions on the server side in one or
-     * more of the Upload component’s event listeners (e.g., in
-     * {@link #addSucceededListener}).
+     * File format restrictions are checked only on the client side (browser)
+     * when this method is used. They indicate the hints for users as to what
+     * file types to upload. If served-side validation is required, use
```
tomivirkki: If I've understood correctly, `setAcceptedFileTypes` now actually does server-side validation also. The old behavior was client-side hints only. So this is breaking behavioral change.

If a user has previously setup their Upload with `setAcceptedFileTypes(".pdf", "image/*")`, they could no longer upload anything

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -347,37 +448,64 @@ public boolean isDropAllowed() {
...
+     *             separate control of MIME types and file extensions with
+     *             server-side validation.
      */
+    @Deprecated(since = "25.2")
     public void setAcceptedFileTypes(String... acceptedFileTypes) {
-        String accepted = "";
-        if (acceptedFileTypes != null) {
-            accepted = String.join(",", acceptedFileTypes);
+        if (acceptedFileTypes == null || acceptedFileTypes.length == 0) {
+            acceptedMimeTypes = List.of();
```
tomivirkki: `List.of()` is used here but `Collections.emptyList()` is used in `setAcceptedMimeTypes` and `setAcceptedFileExtensions`

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -338,6 +346,99 @@ public boolean isDropAllowed() {
...
+     * Sets the accepted MIME types for uploads. Only files matching these MIME
+     * types will be accepted. Wildcard patterns like {@code "image/*"} are
+     * supported.
+     * <p>
+     * MIME types are used both as a client-side hint (to filter the file
+     * picker) and for server-side validation (to reject uploads that don’t
+     * match).
+     * <p>
+     * If both MIME types and file extensions are configured, a file must match
+     * at least one of each (AND logic).
```
tomivirkki, line 363: We should improve this JavaDoc (in UploadManager also). It currently claims it does "server-side" validation for the file type but it only just checks the mime-type string sent from the client. A malicious user could still send an `.exe` file with spoofed `image/png` type. So without `setAcceptedFileExtensions`, it's not any safer than the existing `setAcceptedFileTypes`.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
tomivirkki: typo "served-side"

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -338,6 +346,99 @@ public boolean isDropAllowed() {
...
+     *
+     * @param extensions
+     *            the accepted file extensions, each starting with a dot; or
+     *            {@code null} to clear
+     * @throws IllegalArgumentException
+     *             if any value is null, blank, or does not start with a dot
+     */
+    public void setAcceptedFileExtensions(String... extensions) {
```
tomivirkki, line 450: These only do server-side validation when an UploadHandler is used. When a Receiver is used (deprecated, but still supported), the safety claims of this API no longer holds.

vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/main/java/com/vaadin/flow/component/upload/tests/UploadView.java
```diff
@@ -44,8 +52,8 @@ private void createSimpleUpload() {
         Div fileCount = new Div();
 
         MultiFileMemoryBuffer buffer = new MultiFileMemoryBuffer();
-        Upload upload = new Upload(buffer);
-        upload.setAcceptedFileTypes(".txt");
```
tomivirkki, line 48: This test page no longer covers `setAcceptedFileTypes`. It's depreacted but still supported so we probably shouldn't remove it from existing tests?

vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/main/java/com/vaadin/flow/component/upload/tests/UploadView.java
```diff
@@ -44,8 +52,8 @@ private void createSimpleUpload() {
...
+        upload = new Upload(buffer);
```
tomivirkki, line 55: Since the test page still uses a Receiver, and the server-side validation currently only works with UploadHandler, it indicates that the mime-type [spoofing attack](https://github.com/vaadin/flow-components/blob/f3197c095af5ee16f0f5fe5c01e507f11399795a/vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/test/java/com/vaadin/flow/component/upload/tests/UploadManagerIT.java#L155), which is why the new API is added, isn't covered in the ITs?

ugur-vaadin: The test presses the button with ID `use-upload-handler`, which sets an `UploadHandler` on `Upload`. So it should be covered.

vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/main/java/com/vaadin/flow/component/upload/tests/UploadView.java
ugur-vaadin: Restored.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
ugur-vaadin: Updated JavaDocs to mention this.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
ugur-vaadin: Fixed.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
ugur-vaadin: Updated the JavaDocs in both `Upload` and `UploadManager`.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
ugur-vaadin: Updated to use `List.of()` in all methods.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
ugur-vaadin: Reverted to the original logic. Added tests to prevent regressions.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -347,37 +477,59 @@ public boolean isDropAllowed() {
...
+     *             separate control of MIME types and file extensions with
+     *             server-side validation.
      */
+    @Deprecated(since = "25.2")
     public void setAcceptedFileTypes(String... acceptedFileTypes) {
-        String accepted = "";
-        if (acceptedFileTypes != null) {
-            accepted = String.join(",", acceptedFileTypes);
+        acceptedMimeTypes = List.of();
+        acceptedFileExtensions = List.of();
```
tomivirkki: Instead of silently clearing past configuration, should we also throw (or at least warn) in cases where people might unknowingly disable their server-side validation by mixing use of the new and the old API?

ugur-vaadin: Updated the PR to throw on using the old API after configuring the restrictions with the new API.

tomivirkki: The error message and the description above are a bit confusing. The "silently disabling server-side validation" was a techincal issue with the previous revision, not something that would always happen if you mix the APIs regardless of the impl (we could techincally support both at the same time). I would suggest just mentioning that this combination isn't supported and leaving out technical details.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -347,37 +477,73 @@ public boolean isDropAllowed() {
...
+                            + "setAcceptedMimeTypes or "
+                            + "setAcceptedFileExtensions are configured, as "
+                            + "it would silently disable server-side "
+                            + "validation. Clear them first by calling "
+                            + "setAcceptedMimeTypes(null) and "
+                            + "setAcceptedFileExtensions(null), or use the "
+                            + "separate restriction APIs exclusively.");
+        }
+        acceptedMimeTypes = List.of();
+        acceptedFileExtensions = List.of();
```
tomivirkki: If `acceptedMimeTypes` or `acceptedFileExtensions` are non-empty, the guard above throws. If they're already empty, resetting them to `List.of()` here is a no-op so this is dead code.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
ugur-vaadin: Simplified the error message. Also, added the same check to the new setters to prevent mixing the APIs. Added more unit tests.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
ugur-vaadin: Removed the dead code.

vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/test/java/com/vaadin/flow/component/upload/tests/UploadIT.java
```diff
@@ -207,6 +207,113 @@ public void slowUpload_waitForUpload_pollsUntilUploadFinishes()
...
+
+        getUpload().upload(textFile);
+        assertLogContains("Uploaded: " + textFile.getName());
+    }
+
+    @Test
+    public void setAcceptedFileExtensions_wrongExtension_fileIsRejected()
+            throws Exception {
+        clickElementWithJs("use-upload-handler");
+        clickElementWithJs("clear-accept-ext");
```
tomivirkki: nit: It's a bit confusing (also when manually using the test page), that in order to avoid an exception, you need to click the "clear accepted extension" button before setting accepted extensions, even though there are no accepted extension set initially.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -347,37 +483,71 @@ public boolean isDropAllowed() {
...
     public List<String> getAcceptedFileTypes() {
-        String accepted = getElement().getProperty("accept");
-        if (accepted == null) {
-            return Collections.emptyList();
+        String accept = getElement().getProperty("accept", "");
+        if (accept.isEmpty()) {
+            return List.of();
         }
-        return List.of(accepted.split(","));
+        return List.of(accept.split(","));
```
tomivirkki: Are these changes to the existing logic necessary?

vaadin-upload-flow-parent/vaadin-upload-flow/src/test/java/com/vaadin/flow/component/upload/tests/UploadManagerTest.java
tomivirkki, line 1: There are more tests that should now be moved to the UploadHelperTest, such as `setUploadHandler_wrapperOverridesAllInterfaceMethods`

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -347,37 +483,71 @@ public boolean isDropAllowed() {
...
 
     /**
      * Get the list of accepted file types for upload.
      *
      * @return a list of allowed file types, never <code>null</code>.
+     * @deprecated Use {@link #getAcceptedMimeTypes()} and
+     *             {@link #getAcceptedFileExtensions()} instead.
      */
+    @Deprecated(since = "25.2")
     public List<String> getAcceptedFileTypes() {
```
tomivirkki, line 543: It's unexpected that the deprecated `getAcceptedFileTypes()` returns non-empty value if you use the new API:

```java
var upload = new Upload();
upload.setAcceptedMimeTypes("image/*");
upload.setAcceptedFileExtensions(".pdf");
Assert.assertTrue(upload.getAcceptedFileTypes().isEmpty());
```

vaadin-upload-flow-parent/vaadin-upload-flow-integration-tests/src/test/java/com/vaadin/flow/component/upload/tests/UploadIT.java
ugur-vaadin: Added a button to use the old API explicitly and removed the unconditional use of it along with the clear button clicks.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
ugur-vaadin: Reverted the logic.

vaadin-upload-flow-parent/vaadin-upload-flow/src/test/java/com/vaadin/flow/component/upload/tests/UploadManagerTest.java
ugur-vaadin: Moved the test.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
ugur-vaadin: Updated the old&new getters to throw `IllegalStateException` if old&new API are used both, similar to how it works in the setters.

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -347,39 +492,85 @@ public boolean isDropAllowed() {
...
-            accepted = String.join(",", acceptedFileTypes);
+        if (!acceptedMimeTypes.isEmpty() || !acceptedFileExtensions.isEmpty()) {
+            throw new IllegalStateException(
+                    "Cannot use setAcceptedFileTypes when "
+                            + "setAcceptedMimeTypes or "
+                            + "setAcceptedFileExtensions are configured. "
+                            + "Mixing the deprecated and new APIs is not "
+                            + "supported. Clear them first by calling "
+                            + "setAcceptedMimeTypes(null) and "
+                            + "setAcceptedFileExtensions(null), or use the "
```
tomivirkki: nit: `Clear them first by calling setAcceptedMimeTypes(null) and setAcceptedFileExtensions(null)` somehow feels a bit strange as a suggestion.

It basically prompts you to programmatically clear the values you've previously set using the new API in order to be able to use the deprecated API. If there's a valid use-case for this then ignore this comment.

ugur-vaadin: Removed two instances of the suggestion.

PR #8860, opened 2026-03-06
feat: return SignalBinding from all bind methods

Artur-:
Change all component-level bind methods from void to returning SignalBinding<T>, enabling fluent .onChange() callbacks. This includes bind methods that delegate to super, to SignalPropertySupport.bind(), to Element.bindProperty/bindAttribute, and methods that throw UnsupportedOperationException.


vaadin-avatar-flow-parent/vaadin-avatar-flow/src/main/java/com/vaadin/flow/component/avatar/AvatarGroup.java
sissbruecker, line 1: Should this be picked to 25.1?

tltv: Yes

Issue #8875, opened 2026-03-09
Tab loads indefinitely if page is refreshed during loading


timklge-saxms:
### Description

I have a tabsheet with a couple of tabs on a page. Each tab is instantiated as a route-scoped bean. Each tab takes a while to load (e. g. 2000ms delay in onAttach). When the user repeatedly refreshes the page while the tab is still loading, the tab never finishes loading.

The tab will correctly load once the user switches to a different route, and then navigates back to the original route with the tab. Interestingly, it will return to the infinite loading state when the user refreshes the page again.

https://github.com/user-attachments/assets/558616e2-c099-4a4b-96ac-60bb9b756997

### Expected outcome

The tab loads normally on refresh.

### Minimal reproducible example

```java
@RouteScope
@Component
class Tab1 extends VerticalLayout {
    Tab1() {
        addAttachListener(event -> {
            try {
                Thread.sleep(2000);
            } catch (InterruptedException ex) {
                throw new RuntimeException(ex);
            }
        });

        add(new Text("Content of Tab 1"));
    }
}

@RouteScope
@Component
class Tab2 extends VerticalLayout {
    Tab2() {
        addAttachListener(event -> {
            try {
                Thread.sleep(2000);
            } catch (InterruptedException ex) {
                throw new RuntimeException(ex);
            }
        });

        add(new Text("Content of Tab 2"));
    }
}

@Route("")
@PageTitle("Task List")
@Menu(order = 0, icon = "vaadin:clipboard-check", title = "Task List")
class TaskListView extends VerticalLayout {
    TaskListView(Tab1 tab1, Tab2 tab2) {
        var tabSheet = new TabSheet();
        add(tabSheet);

        try {
            Thread.sleep(1000);
        } catch (InterruptedException ex) {
            throw new RuntimeException(ex);
        }

        tabSheet.add("Tab 1", tab1);
        tabSheet.add("Tab 2", tab2);
    }
}
```

[mre.zip](https://github.com/user-attachments/files/25839923/mre.zip)

### Steps to reproduce

1. Add the snippet above to a view
2. Load the page, but hit refresh while the page is still loading (Refresh repeatedly if it doesn't reproduce immediately)

### Environment

Vaadin version(s): 25.0.5
OS: Ubuntu 24.04

### Browsers

Firefox

yuriy-fix:
The issue was moved to the backlog priority queue for further investigation.

PR #8879, opened 2026-03-09
fix: move tabsheet contents to new instance

sissbruecker:
## Description

When reusing tab contents between multiple UIs, automatically adding the content to the tab sheet fails as the component currently checks if the content already has a parent, and, if so, does nothing.

This change handles reusing contents by checking if the content parent matches the current tab sheet, and if not, moves the content to the new parent.

Fixes https://github.com/vaadin/flow-components/issues/8875

## Type of change

- Bugfix

### DIFF
--- a/vaadin-radio-button-flow-parent/vaadin-radio-button-flow-integration-tests/src/main/java/com/vaadin/flow/component/radiobutton/tests/RadioButtonGroupBindItemsPage.java
+++ b/vaadin-radio-button-flow-parent/vaadin-radio-button-flow-integration-tests/src/main/java/com/vaadin/flow/component/radiobutton/tests/RadioButtonGroupBindItemsPage.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2000-2026 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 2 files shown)


vaadin-tabs-flow-parent/vaadin-tabs-flow/src/main/java/com/vaadin/flow/component/tabs/TabSheet.java
```diff
@@ -385,7 +385,8 @@ private void updateContent() {
 
     private void ensureSelectedTabContentAttached() {
         var content = tabToContent.get(tabs.getSelectedTab());
-        if (content != null && content.getParent() == null) {
+        if (content != null && content.getParent() != getElement()) {
+            content.getNode().removeFromTree(false);
```
DiegoCardoso: This introduces a server-side error when you switch between tabs and refresh the page:

https://github.com/user-attachments/assets/d12f1859-e25c-437d-95f8-8db60fc4463b

The error is:
```
[qtp2086483651-168] WARN com.vaadin.flow.component.internal.UIInternals - Error detaching closed UI 0 
java.lang.IllegalStateException: Unregistered node was not found based on its id. The tree is most likely corrupted.
```

I thought that it could be that this line could be trying to remove an element from a `null` parent, but even with a guard around this line to check for `content.getParent() != null`, the error seems to be thrown still.

sissbruecker: Thanks, should be fixed by using `Element.removeFromTree` instead.

DiegoCardoso: I cannot see this issue anymore, and couldn't reproduce the bug with the MRE provided by the user after the fix.

PR #8892, opened 2026-03-11
chore: use @vaadin/vite-plugin-local-web-components package

vursen:
Replace the local vite plugin with Flow's @vaadin/vite-plugin-local-web-components package.

Depends on 
- https://github.com/vaadin/flow/pull/23829


shared/shared-vite-config.ts
```diff
@@ -1,11 +1,14 @@
-import path from 'path';
+import path from 'node:path';
+// @ts-expect-error this dependency is added by flow-server
 import { UserConfigFn, mergeConfig, loadEnv } from 'vite';
-import { useLocalWebComponents } from './web-components-vite-plugin';
+// @ts-expect-error this dependency is added by flow-build-tools
+import useLocalWebComponents from '@vaadin/vite-plugin-local-web-components';
```
sissbruecker, line 5: Doesn't look like this package is installed automatically when running Jetty / ITs?

PR #8908, opened 2026-03-12
chore: add skill for signal rules

sissbruecker:
## Description

Adds a Claude Code skill that is intended to help with writing or reviewing signal APIs in components. The skill explains fundamental rules that each signal API must follow.

## Type of change

- Internal


sissbruecker:
Makes sense to mention that specifically. Made the section a bit more concrete by mentioning the cases I can think of at this point. Also removed some (`executeJs`, class field) that actually wouldn't violate rule 2, but rather rule 1. Maybe it would make sense to list cases for rule 1 as well, but let's maybe see how this works in practice first.

tomivirkki (approved):

PR #8915, opened 2026-03-12
chore: convert avatar unit tests to JUnit 6

sissbruecker:
Convert Avatar / Avatar Group unit tests to JUnit 6.

### DIFF
--- a/vaadin-app-layout-flow-parent/vaadin-app-layout-flow/pom.xml
+++ b/vaadin-app-layout-flow-parent/vaadin-app-layout-flow/pom.xml
@@ -55,11 +55,6 @@
             <artifactId>jakarta.servlet-api</artifactId>
         </dependency>
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
             <scope>test</scope>
--- a/vaadin-app-layout-flow-parent/vaadin-app-layout-flow/src/test/java/com/vaadin/flow/component/applayout/AppLayoutSerializableTest.java
+++ b/vaadin-app-layout-flow-parent/vaadin-app-layout-flow/src/test/java/com/vaadin/flow/component/applayout/AppLayoutSerializableTest.java
@@ -17,5 +17,5 @@ package com.vaadin.flow.component.applayout;
 
 import com.vaadin.flow.testutil.ClassesSerializableTest;
 
-public class AppLayoutSerializableTest extends ClassesSerializableTest {
+class AppLayoutSerializableTest extends ClassesSerializableTest {
 }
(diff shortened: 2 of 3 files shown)


vaadin-avatar-flow-parent/vaadin-avatar-flow/src/test/java/com/vaadin/flow/component/avatar/tests/AvatarGroupSignalTest.java
```diff
@@ -63,145 +64,155 @@ public void signalConstructor_updatesWhenSignalChanges() {
...
+    @Test
+    void signalConstructor_setItemsWhileBound_throws() {
         var item1Signal = new ValueSignal<>(new AvatarGroupItem("Alice"));
         var listSignal = new ValueSignal<>(List.of(item1Signal));
 
         avatarGroup = new AvatarGroup(listSignal);
         ui.add(avatarGroup);
 
-        avatarGroup.setItems(List.of(new AvatarGroupItem("Bob")));
+        Assertions.assertThrows(BindingActiveException.class, () -> avatarGroup
```
DiegoCardoso, line 83: The sonarcloud comment might not be valid here, since the `BindingActiveException` will never be thrown from the `List.of()` call, but I think it's valid as an instruction to the Claude skill to place only the call that is under test inside the lambda (in this case, to not instantiate the `List` inside the lambda) so we can avoid false positives, e.g., the same exception being raised from another call inside the lambda. It also makes it easier to understand which part of the code is under test as well.

vaadin-avatar-flow-parent/vaadin-avatar-flow/src/test/java/com/vaadin/flow/component/avatar/tests/AvatarGroupTest.java
```diff
@@ -17,137 +17,139 @@
...
-        Assert.assertEquals(avatarGroupItem1.getName(), "foo bar");
+        Assertions.assertEquals(avatarGroupItem1.getName(), "foo bar");
     }
 
     @Test
-    public void setAbbr_getAbbr() {
+    void setAbbr_getAbbr() {
         avatarGroupItem1.setAbbreviation("fb");
-        Assert.assertEquals(avatarGroupItem1.getAbbreviation(), "fb");
+        Assertions.assertEquals(avatarGroupItem1.getAbbreviation(), "fb");
```
DiegoCardoso, line 50: nit: could also add to the skill to swap the expected and actual values.

vaadin-avatar-flow-parent/vaadin-avatar-flow/src/test/java/com/vaadin/flow/component/avatar/tests/AvatarGroupSignalTest.java
sissbruecker: I added some instruction for that, but it did not consider to extract nested expressions. The only improvement was to not include multiple statements (lines) in an assertThrows.

While it might make sense to improve this, let's consider doing this as a cleanup. I feel it's better to sort of limit the scope of what it should do instead of trying to make it fix all Sonar issues along the way.

vaadin-avatar-flow-parent/vaadin-avatar-flow/src/test/java/com/vaadin/flow/component/avatar/tests/AvatarGroupTest.java
sissbruecker: This would be good to clean up, but we can also do this separately. Doing so is basically just running an auto-fix from IntelliJ which does this for the whole codebase in a deterministic manner.

Issue #8890, opened 2026-03-10
Add bindThemeVariants helper method


Legioth:
### Describe your motivation

`HasTheme` has a recently added `bindThemeNames` with a signal that returns a list of theme names to apply.

Usage example, show the button as either `primary` or `warning` depending on some condition :
```java
button.bindThemeNames(() -> List.of(conditionSignal.get() ? "primary" : "warning"));
```

### Describe the solution you'd like

Make the same structure available also for the typed theme variants, e.g. 
```java
button.bindThemeVariants(() -> List.of(conditionSignal.get() ? ButtonVariant.LUMO_PRIMARY : ButtonVariant.LUMO_WARNING));
```

PR #8924, opened 2026-03-16
feat: add bindThemeVariants method for typed theme variant signal binding

sissbruecker:
## Description

Add `HasThemeVariant.bindThemeVariants()` that mirrors `HasTheme.bindThemeNames()` but works with typed theme variant enums instead of raw strings. Introduce an internal `SignalBindingUtil.mapBinding()` helper for mapping between typed and string-based signal bindings, reusable for future component binding APIs.

Fixes https://github.com/vaadin/flow-components/issues/8890

## Type of change

- Feature


vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/internal/SignalBindingUtil.java
```diff
@@ -0,0 +1,78 @@
...
+     *            the source signal value type
+     * @param <U>
+     *            the mapped signal value type used by the underlying binding
+     * @return a {@link SignalBinding} typed to the source type
+     */
+    public static <T, U> SignalBinding<T> mapBinding(Signal<T> source,
+            SerializableFunction<T, U> mapper,
+            SerializableFunction<Signal<U>, SignalBinding<U>> binder) {
+        @SuppressWarnings("unchecked")
+        T[] previous = (T[]) new Object[] { source.peek() };
```
DiegoCardoso, line 67: Should we have some null checks? If the user passes `null` in `source`, for instance.

sissbruecker: Good point, added assertions beforehand.

Issue #8938, opened 2026-03-17
grid.bindItems(SharedListSignal) does not update rows when child value signal changes


Artur-:
### Description

When a Grid is bound to a SharedListSignal via grid.bindItems(), updating a child SharedValueSignal value does not refresh the corresponding grid row. Other signal bindings (e.g. Span.bindText()) react to the same signal change correctly.                                                        

### Expected outcome

It updates

### Minimal reproducible example

```java
public class GridBindingTestView extends VerticalLayout {                                         
                                                                                                  
    public record Item(String name) {                                                             
    }                                                           
    final SharedListSignal<Item> items = new SharedListSignal<>(Item.class);                      

    public GridBindingTestView() {                                                                
        items.insertLast(new Item("original"));                 
                                                                                                  
        var grid = new Grid<>(Item.class, false);
        grid.addColumn(Item::name).setHeader("Name");                                             
        grid.bindItems(items);                                                                    

        var status = new Span();                                                                  
        status.bindText(items.map(list -> {                     
            if (list.isEmpty()) {                                                                 
                return "Signal: (empty)";
            }                                                                                     
            return "Signal: " + list.getFirst().get().name();   
        }));                                                                                      

        var updateButton = new Button("Update", e -> {                                            
            var sig = items.peek().getFirst();                  
            sig.set(new Item("updated"));                                                         
        });                                                                                       
                                                                                                  
        add(status, grid, updateButton);                                                          
    }                                                           
}                                                                                                 
```

### Steps to reproduce

1. Open the view — grid shows "original", span shows "Signal: original"
2. Click "Update"

Expected: Grid shows "updated", span shows "Signal: updated"                              
                                                                                          
Actual: Span updates to "Signal: updated" but grid still shows "original"                 

### Environment

Vaadin version(s): 25.1 beta3

manolo:
does it blocks flow, or should be fixed only in FC? looks like a grid fix

Legioth:
> does it blocks flow

The same problem affects all implementation of `HasDataView.bindItems`. The core of the fix is to add a new `refreshItem` variant that takes an old and a new instance. Components that use `DataCommunicator` should be fixed by the change in Flow whereas some other components might need small direct adjustments.

PR #8940, opened 2026-03-17
fix: Handle item identity change in components without DataCommunicator

Artur-:
Update Select, RadioButtonGroup and CheckboxGroup to update their children based on data provider refreshes where item identity changes. This is the case mainly in combination with `bindItems`.

### DIFF
--- a/vaadin-ai-components-flow-parent/README.md
+++ b/vaadin-ai-components-flow-parent/README.md
@@ -3,9 +3,9 @@
 This project provides interfaces for building AI-powered applications with LLM integration
 and [Vaadin Flow](https://github.com/vaadin/flow).
 
-## Using the component in a Flow application
+## Using the components in a Flow application
 
-To use the component in an application using maven,
+To use the components in an application using maven,
 add the following dependency to your `pom.xml`:
 ```
 <dependency>
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/pom.xml
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/pom.xml
@@ -25,6 +25,11 @@
     <dependencies>
         <dependency>
             <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-charts-flow</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
             <artifactId>vaadin-flow-components-base</artifactId>
             <version>${project.version}</version>
         </dependency>
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/chart/DataConverter.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/chart/DataConverter.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2000-2026 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component.ai.chart;
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.Map;
+
+import com.vaadin.flow.component.charts.model.Series;
+
+/**
+ * Converts tabular data into chart {@link Series} for rendering. The input is a
+ * list of rows, where each row is a column-name-to-value map.
+ * <p>
+ * Implementations may return one or more series depending on the data.
+ * </p>
+ *
+ * @author Vaadin Ltd
+ */
+public interface DataConverter extends Serializable {
+
+    /**
+     * Converts tabular data into one or more chart series.
+     * <p>
+     * Each element in the input list represents a single row, with column names
+     * as keys and column values as values.
+     * </p>
+     *
+     * @param data
+     *            the tabular data to convert, not {@code null}
+     * @return a list of series ready for use in a chart, never {@code null} and
+     *         never empty
+     * @throws NullPointerException
+     *             if data is {@code null}
+     */
+    List<Series> convertToSeries(List<Map<String, Object>> data);
+}
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/test/java/com/vaadin/flow/component/ai/AIComponentsSerializableTest.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/test/java/com/vaadin/flow/component/ai/AIComponentsSerializableTest.java
@@ -53,10 +53,9 @@ class AIComponentsSerializableTest extends ClassesSerializableTest {
         return Stream.concat(super.getExcludedPatterns(), Stream.of(
                 "com\\.vaadin\\.flow\\.component\\.upload\\.receivers\\.TempDirectory\\$LazyHolder",
                 "com\\.vaadin\\.flow\\.component\\.upload\\.receivers\\.TempDirectory",
+                "com\\.vaadin\\.flow\\.component\\.charts\\.model\\.serializers\\..*",
                 "com\\.vaadin\\.flow\\.component\\.ai\\.provider\\..*",
                 "com\\.vaadin\\.flow\\.component\\.ai\\.AIComponentsFeatureFlagProvider",
-                // AIOrchestrator — private constructor, requires LLMProvider
-                "com\\.vaadin\\.flow\\.component\\.ai\\.orchestrator\\.AIOrchestrator",
                 "com\\.vaadin\\.flow\\.component\\.ai\\.orchestrator\\.AIOrchestrator\\$Reconnector",
                 "com\\.vaadin\\.flow\\.component\\.ai\\.orchestrator\\.AIOrchestrator\\$Builder"));
     }
--- a/vaadin-button-flow-parent/vaadin-button-flow/src/test/java/com/vaadin/flow/component/button/tests/AccessibleDisabledButtonTest.java
+++ b/vaadin-button-flow-parent/vaadin-button-flow/src/test/java/com/vaadin/flow/component/button/tests/AccessibleDisabledButtonTest.java
@@ -15,9 +15,9 @@
  */
 package com.vaadin.flow.component.button.tests;
 
-import org.junit.Before;
-import org.junit.Rule;
-import org.junit.Test;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
 import org.mockito.Mockito;
 
 import com.vaadin.experimental.FeatureFlags;
@@ -31,14 +31,14 @@ import com.vaadin.flow.dom.DomEvent;
 import com.vaadin.flow.dom.Element;
 import com.vaadin.flow.internal.JacksonUtils;
 import com.vaadin.flow.internal.nodefeature.ElementListenerMap;
-import com.vaadin.tests.EnableFeatureFlagRule;
-import com.vaadin.tests.MockUIRule;
+import com.vaadin.tests.EnableFeatureFlagExtension;
+import com.vaadin.tests.MockUIExtension;
 
-public class AccessibleDisabledButtonTest {
-    @Rule
-    public MockUIRule ui = new MockUIRule();
-    @Rule
-    public EnableFeatureFlagRule featureFlagRule = new EnableFeatureFlagRule(
+class AccessibleDisabledButtonTest {
+    @RegisterExtension
+    MockUIExtension ui = new MockUIExtension();
+    @RegisterExtension
+    EnableFeatureFlagExtension featureFlagExtension = new EnableFeatureFlagExtension(
             FeatureFlags.ACCESSIBLE_DISABLED_BUTTONS);
 
     private Button button = Mockito.spy(Button.class);
@@ -51,15 +51,15 @@ public class AccessibleDisabledButtonTest {
     private ComponentEventListener mockBlurListener = Mockito
             .mock(ComponentEventListener.class);
 
-    @Before
-    public void setUp() {
+    @BeforeEach
+    void setUp() {
         button.setEnabled(false);
     }
 
     @SuppressWarnings("unchecked")
     @Test
-    public void accessibleButtonsDisabled_focusListenerDisabled() {
-        featureFlagRule.disableFeature();
+    void accessibleButtonsDisabled_focusListenerDisabled() {
+        featureFlagExtension.disableFeature();
 
         button.addFocusListener(mockFocusListener);
 
@@ -71,7 +71,7 @@ public class AccessibleDisabledButtonTest {
 
     @SuppressWarnings("unchecked")
     @Test
-    public void accessibleButtonsEnabled_focusListenerEnabled() {
+    void accessibleButtonsEnabled_focusListenerEnabled() {
         button.addFocusListener(mockFocusListener);
 
         fakeClientDomEvent(button, "focus");
@@ -82,8 +82,8 @@ public class AccessibleDisabledButtonTest {
 
     @SuppressWarnings("unchecked")
     @Test
-    public void accessibleButtonsDisabled_blurListenerDisabled() {
-        featureFlagRule.disableFeature();
+    void accessibleButtonsDisabled_blurListenerDisabled() {
+        featureFlagExtension.disableFeature();
 
         button.addBlurListener(mockBlurListener);
 
@@ -95,7 +95,7 @@ public class AccessibleDisabledButtonTest {
 
     @SuppressWarnings("unchecked")
     @Test
-    public void accessibleButtonsEnabled_blurListenerEnabled() {
+    void accessibleButtonsEnabled_blurListenerEnabled() {
         button.addBlurListener(mockBlurListener);
 
         fakeClientDomEvent(button, "blur");
@@ -105,8 +105,8 @@ public class AccessibleDisabledButtonTest {
     }
 
     @Test
-    public void accessibleButtonsDisabled_focusShortcutDisabled() {
-        featureFlagRule.disableFeature();
+    void accessibleButtonsDisabled_focusShortcutDisabled() {
+        featureFlagExtension.disableFeature();
 
         button.addFocusShortcut(Key.KEY_A);
         ui.add(button);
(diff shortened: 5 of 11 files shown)


Legioth:
Related to https://github.com/vaadin/flow-components/issues/8938

Legioth:
How can these tests pass already before merging the fix in the Flow repo?

Legioth:
Feels like something more would be needed e.g. for  `CheckboxGroup` since `CheckboxGroup.handleDataChange(DataChangeEvent<T>)` calls `keyMapper.refresh(otherItem);` without regard for the new case.

Legioth:
Good. Now the test fails. 🙃

vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/CheckboxGroup.java
```diff
@@ -405,11 +405,25 @@ public void onDiscard(DataChangeEvent<T> dataChangeEvent) {
     private void handleDataChange(DataChangeEvent<T> dataChangeEvent) {
         if (dataChangeEvent instanceof DataChangeEvent.DataRefreshEvent<T> dataRefreshEvent) {
             T otherItem = dataRefreshEvent.getItem();
```
Legioth: Could rename this variable as well to avoid some confusion

vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/test/java/com/vaadin/flow/component/checkbox/CheckboxGroupSignalTest.java
```diff
@@ -92,6 +94,43 @@ public void signalConstructor_setsItemsFromSignal() {
...
+        var listSignal = new ListSignal<String>();
+        listSignal.insertLast("original");
+
+        var group = new CheckboxGroup<String>();
+        group.bindItems(listSignal);
+        ui.add(group);
+
+        KeyMapper<String> keyMapper = getKeyMapper(group);
+        String keyBefore = keyMapper.key("original");
+        Assert.assertTrue(keyMapper.has("original"));
```
Legioth: Previous row adds a mapping if it didn't exist so this would always be true

vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/test/java/com/vaadin/flow/component/checkbox/CheckboxGroupSignalTest.java
```diff
@@ -92,6 +94,43 @@ public void signalConstructor_setsItemsFromSignal() {
...
+        // Update the item signal value
+        listSignal.peek().getFirst().set("updated");
+
+        // The old identity should be gone, replaced by the new one
+        // mapped to the same key
+        Assert.assertFalse(keyMapper.has("original"));
+        Assert.assertTrue(keyMapper.has("updated"));
+        Assert.assertEquals(keyBefore, keyMapper.key("updated"));
+
+        // Verify the component items reflect the update
```
Legioth: Maybe also check that there's an updated `Checkbox` child?

vaadin-radio-button-flow-parent/vaadin-radio-button-flow/src/main/java/com/vaadin/flow/component/radiobutton/RadioButtonGroup.java
```diff
@@ -737,10 +737,29 @@ private void fireSizeEvent() {
     }
 
     private void resetRadioButton(T item) {
```
Legioth: Seems like this private method is no longer used so it should be removed

vaadin-radio-button-flow-parent/vaadin-radio-button-flow/src/main/java/com/vaadin/flow/component/radiobutton/RadioButtonGroup.java
```diff
@@ -737,10 +737,29 @@ private void fireSizeEvent() {
...
+        resetRadioButton(item, item);
+    }
+
+    private void resetRadioButton(T oldItem, T newItem) {
```
Legioth: Let's stick to the (new, old) ordering that is used in corresponding Flow cases

vaadin-radio-button-flow-parent/vaadin-radio-button-flow/src/main/java/com/vaadin/flow/component/radiobutton/RadioButtonGroup.java
```diff
@@ -737,10 +737,29 @@ private void fireSizeEvent() {
...
+        boolean identityChanged = !Objects.equals(getItemId(oldItem),
+                getItemId(newItem));
         getRadioButtons()
                 .filter(radioButton -> getItemId(radioButton.getItem())
-                        .equals(getItemId(item)))
-                .findFirst().ifPresent(this::updateButton);
+                        .equals(getItemId(oldItem)))
+                .findFirst().ifPresent(oldButton -> {
+                    if (identityChanged) {
+                        RadioButton<T> newButton = new RadioButton<>(
```
Legioth: Could maybe reuse existing `createRadioButton` method?

vaadin-radio-button-flow-parent/vaadin-radio-button-flow/src/test/java/com/vaadin/flow/component/radiobutton/RadioButtonGroupSignalTest.java
```diff
@@ -147,6 +149,43 @@ private RadioButtonGroup<String> createRadioButtonGroupWithBoundItems() {
...
+        var listSignal = new ListSignal<String>();
+        listSignal.insertLast("original");
+
+        var group = new RadioButtonGroup<String>();
+        group.bindItems(listSignal);
+        ui.add(group);
+
+        KeyMapper<String> keyMapper = getKeyMapper(group);
+        String keyBefore = keyMapper.key("original");
+        Assert.assertTrue(keyMapper.has("original"));
```
Legioth: This always passes since a mapping was created by the previous line if it didn't already exist

vaadin-radio-button-flow-parent/vaadin-radio-button-flow/src/test/java/com/vaadin/flow/component/radiobutton/RadioButtonGroupSignalTest.java
```diff
@@ -147,6 +149,43 @@ private RadioButtonGroup<String> createRadioButtonGroupWithBoundItems() {
...
+        // mapped to the same key
+        Assert.assertFalse(keyMapper.has("original"));
+        Assert.assertTrue(keyMapper.has("updated"));
+        Assert.assertEquals(keyBefore, keyMapper.key("updated"));
+
+        // Verify the component items reflect the update
+        List<String> items = group.getGenericDataView().getItems().toList();
+        Assert.assertEquals(1, items.size());
+        Assert.assertEquals("updated", items.get(0));
+    }
```
Legioth, line 170: Could maybe also assert that the child elements have been updated

Legioth:
No test for ComboBox?

vaadin-select-flow-parent/vaadin-select-flow/src/test/java/com/vaadin/flow/component/select/SelectSignalTest.java
```diff
@@ -90,6 +92,43 @@ public void signalConstructor_setsItemsFromSignal() {
...
+        var listSignal = new ListSignal<String>();
+        listSignal.insertLast("original");
+
+        var select = new Select<String>();
+        select.bindItems(listSignal);
+        ui.add(select);
+
+        KeyMapper<String> keyMapper = getKeyMapper(select);
+        String keyBefore = keyMapper.key("original");
+        Assert.assertTrue(keyMapper.has("original"));
```
Legioth: Pointless assert since the mapping was created on the previous line if it didn't already exist

vaadin-select-flow-parent/vaadin-select-flow/src/test/java/com/vaadin/flow/component/select/SelectSignalTest.java
```diff
@@ -90,6 +92,43 @@ public void signalConstructor_setsItemsFromSignal() {
...
+        // mapped to the same key
+        Assert.assertFalse(keyMapper.has("original"));
+        Assert.assertTrue(keyMapper.has("updated"));
+        Assert.assertEquals(keyBefore, keyMapper.key("updated"));
+
+        // Verify the component items reflect the update
+        List<String> items = select.getGenericDataView().getItems().toList();
+        Assert.assertEquals(1, items.size());
+        Assert.assertEquals("updated", items.get(0));
+    }
```
Legioth, line 127: Could also assert that the element structure has been updated

vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/test/java/com/vaadin/flow/component/checkbox/CheckboxGroupSignalTest.java
sissbruecker, line 1: Wouldn't it in general make more sense to test whether the generated components are updated instead of testing how the key mapper behaves?

I also don't see the value in testing the generic data view, which just queries the data provider. That doesn't seem to be related to the component logic?

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridSignalTest.java
sissbruecker, line 1: Is this not covered by the Flow change already?

Legioth: The plumbing through `DataCommunicator` isn't really tested there since there's no appropriate test component in the Flow repo

vaadin-select-flow-parent/vaadin-select-flow/src/test/java/com/vaadin/flow/component/select/SelectSignalTest.java
Legioth: This one still tests only through the data view rather than based on the actual elements

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridSignalTest.java
sissbruecker: The test suite for data communicator has tests around the key mapper, so why can't this be tested there as well?

Legioth:
No further remarks from me but still waiting to see what @sissbruecker thinks

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridSignalTest.java
Legioth: Maybe there should be additional tests there as well, but there should also be something resembling an end-to-end test with a real component. We got this issue to begin with because we didn't have any test that used `bindItems` in a realistic way. With this test, we get something that is at least somewhat closer to real usage.

At the same time, I agree that the test should maybe look at something else than what's in the key mapper...

vaadin-select-flow-parent/vaadin-select-flow/src/test/java/com/vaadin/flow/component/select/SelectSignalTest.java
```diff
@@ -90,6 +90,41 @@ public void signalConstructor_setsItemsFromSignal() {
...
+        // Verify the item element was updated with the new label
+        labels = getItemLabels(select);
+        Assert.assertEquals(1, labels.size());
+        Assert.assertEquals("updated", labels.get(0));
+    }
+
+    private List<String> getItemLabels(Select<String> select) {
+        // Items are inside the listBox element (first child of select)
+        var listBoxElement = select.getElement().getChild(0);
+        var labels = new java.util.ArrayList<String>();
```
Legioth: nit: import

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridSignalTest.java
Artur-: https://github.com/vaadin/flow/pull/23923

flow/flow-data/src/test/java/com/vaadin/flow/data/provider/DataCommunicatorTest.java
```diff
@@ -0,0 +1,2206 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class DataCommunicatorTest {
```
Legioth: Good test. Wrong repo

vaadin-grid-flow-parent/vaadin-grid-flow/src/test/java/com/vaadin/flow/component/grid/GridSignalTest.java
```diff
@@ -341,29 +341,25 @@ public void signalConstructor_setsItemsFromSignal() {
...
         grid.bindItems(listSignal);
         ui.add(grid);
 
-        // Force a flush so the key mapper has assigned a key
-        ui.fakeClientCommunication();
-
-        var keyMapper = grid.getDataCommunicator().getKeyMapper();
-        String keyBefore = keyMapper.key("original");
-        Assert.assertTrue(keyMapper.has("original"));
+        List<String> items = grid.getGenericDataView().getItems().toList();
```
Legioth: This does again look like the initial test that passed also without the fix?

vaadin-virtual-list-flow-parent/vaadin-virtual-list-flow/src/test/java/com/vaadin/flow/component/virtuallist/tests/VirtualListSignalTest.java
```diff
@@ -59,29 +59,27 @@ void bindItems_thenSetStream_throws() {
...
+        // Update the item signal value (identity change)
         listSignal.peek().getFirst().set("updated");
 
-        // The old identity should be gone, replaced by the new one
-        // mapped to the same key
-        Assertions.assertFalse(keyMapper.has("original"));
-        Assertions.assertTrue(keyMapper.has("updated"));
-        Assertions.assertEquals(keyBefore, keyMapper.key("updated"));
+        // Verify the items reflect the update
+        items = list.getDataProvider()
```
Legioth: Seems like this test the data provider rather than testing `VirtualList`?

vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/CheckboxGroup.java
sissbruecker, line 1: `ListBox` and `MultiSelectListBox` are not covered.

sissbruecker, line 1: Now that `DataRefreshEvent` can replace items, instead of just updating existing ones, this also impacts item selection. When you select an item and then an event replaces it with a different one, what should happen with the selected item?

In **some** of the affected components we allow configuring a `SelectionPreservationMode` that determines whether items should be removed from the selection when the item is removed from the data provider (`SelectionPreservationMode.PRESERVE_EXISTING`), or if it should be kept (`SelectionPreservationMode.PRESERVE_ALL`). So far that mode was only respected (or needed) when using `refreshAll`. Now that you can replace items, it seems reasonable to assume that this operation would respect that mode as well. Not sure how to handle `SelectionPreservationMode.DISCARD`. It should probably only remove the replaced item instead of all items. In any case, the modes would need some clarification that they also apply when replacing items and what the respective behavior is.

For components that do not allow configuring how to update the selection on data provider changes I would assume the replaced item is also removed from the selection.

sissbruecker: Actually, I'm not sure if components without selection preservation mode clear the selection when using `refreshAll`. Needs to be checked.

Artur-: It would be expected to keep selection when the item is refreshed. Do we want to put this also in the same PR or in a different one?

sissbruecker: From what I understood, this change covers cases where the item's identify has changed. So if you replace `Customer[id=5]` with `Customer[id=7]`. In that case you would not expect that `Customer[id=7]` is selected because `Customer[id=5]` was. Am I missing something?

Artur-: The point is to refresh a value whose identity value has changed but it still refers to the same value. This more typically happens in cases where you do not have any stable id (a record with only a name), or when the initial version did not have an id and got an id when storing in a database

Legioth: I'd say that selection should always be preserved for a single-item refresh regardless of whether that's a refresh that changes identity or not. `SelectionPreservationMode` is about whether an item has been removed but that's not the case when it's still the same logical item even though technical equality has changed.

Would it make sense to tackle selection in a separate PR though?

Artur-: That would be #8951

Legioth: It's easy enough for the application to manually unselect a single item after running `refreshItem(new, old)` if that's the desired semantics.

`SelectionPreservationMode` is for the bulk case where it's not practical for application logic to keep track of each individual item, or in the opposite direction in cases when it's not performant for the framework to check automatically.

mshabarov: > So if you replace Customer[id=5] with Customer[id=7]

This case in my opinion is not the `refreshItem`, better to treat it as a two operations - remove Customer[id=5]  and add a new Customer[id=7]. When you have an ID as an identity that is apart of other data of this item that is usually updated, then the old `refreshItem(item)` can be used without side effects. The problem occurs where all the item's fields are used as an identity and you want to update it, simple case is a Select with just string values.

PR #8951, opened 2026-03-18
fix: preserve selection when a selected item's identity changes

Artur-:
When a selected item's identity changes through a data refresh event,
the selection was lost because the component's value still referenced
the old item. Now after replacing the DOM element, each component checks
if the old item was selected and updates the value to reference the new
item.

Components updated: CheckboxGroup, RadioButtonGroup, ListBox,
MultiSelectListBox, and Select. ListBoxBase adds an onItemIdentityChanged
hook that ListBox and MultiSelectListBox override for their respective
single/multi-select logic.

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
@@ -21,9 +21,6 @@ import java.util.List;
 import java.util.Objects;
 import java.util.stream.Stream;
 
-import com.vaadin.flow.data.provider.DataChangeEvent;
-import com.vaadin.flow.data.provider.DataProvider;
-
 import com.vaadin.flow.component.HasSize;
 import com.vaadin.flow.component.Tag;
 import com.vaadin.flow.component.Unit;
@@ -32,6 +29,7 @@ import com.vaadin.flow.component.dependency.NpmPackage;
 import com.vaadin.flow.component.shared.HasPrefix;
 import com.vaadin.flow.component.shared.HasThemeVariant;
 import com.vaadin.flow.data.binder.Binder;
+import com.vaadin.flow.data.provider.DataChangeEvent;
 import com.vaadin.flow.data.provider.DataCommunicator;
 import com.vaadin.flow.data.provider.DataKeyMapper;
 import com.vaadin.flow.data.provider.DataProvider;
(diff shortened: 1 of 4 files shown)


Legioth:
How about Grid and ComboBox? Do they work automatically or would they instead need changes in Flow related to e.g. `DataCommunicator`?

vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/CheckboxGroup.java
```diff
@@ -404,12 +404,36 @@ public void onDiscard(DataChangeEvent<T> dataChangeEvent) {
...
+                            remove(oldCheckbox);
+                            // Preserve selection
+                            Set<T> currentValue = getValue();
+                            if (currentValue.stream().anyMatch(s -> Objects
+                                    .equals(getItemId(s), oldItemId))) {
+                                Set<T> newValue = currentValue.stream()
+                                        .map(s -> Objects.equals(getItemId(s),
+                                                oldItemId) ? newItem : s)
+                                        .collect(Collectors.toSet());
+                                setValue(newValue);
```
sissbruecker, line 431: In the other PR the argument was that it's the same value. However now we fire a value change event even though the value hasn't changed. From my experience, developers do not appreciate arbitrary value change events because it is often tied to other logic that they only want to run when really necessary.

In general this whole approach with how signals detect changes doesn't seem very compatible with how the components handle identity. The component's selection / value relies on the identify being stable. Signals require the identity to change. Personally I'm not confident about this change, and it's hard for me to say whether this is correct and if there might be some other case that we have overlooked.

Artur-: This isn't really related to signals at all. 

The value change question is good though - should you get a value change when per definition the value does not change?

Artur-: On the other hand, you would need the value change event if you use the value for anything at all..

Legioth: Silly use case: a list of words (i.e. strings) and then you select your favorites. My selection is `["this", "is", "my", "favorite"]` and that's also reflected in another part of the UI. Now, someone reminds us that the true imperialist spelling is "favourite", so I edit that one row in the list. In that case, I would certainly want the other part of my UI to be updated to also show the updated spelling. That can only happen if there's an event telling that the selection value has changed.

PR #8956, opened 2026-03-18
chore: migrate spreadsheet tests junit6

DiegoCardoso:
## Description

Convert `Spreadsheet` unit tests to JUnit 6.

### DIFF
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow-integration-tests/src/main/java/com/vaadin/flow/component/ai/tests/AIOrchestratorPage.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow-integration-tests/src/main/java/com/vaadin/flow/component/ai/tests/AIOrchestratorPage.java
@@ -15,11 +15,13 @@
  */
 package com.vaadin.flow.component.ai.tests;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
 import com.vaadin.flow.component.ai.common.AIAttachment;
+import com.vaadin.flow.component.ai.common.ChatMessage;
 import com.vaadin.flow.component.ai.orchestrator.AIOrchestrator;
 import com.vaadin.flow.component.ai.provider.LLMProvider;
 import com.vaadin.flow.component.html.Div;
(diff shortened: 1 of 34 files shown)


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/pom.xml
```diff
@@ -125,6 +125,12 @@
             <artifactId>poi-scratchpad</artifactId>
             <version>${poi.version}</version>
         </dependency>
+        <dependency>
```
sissbruecker: Maybe this should just be grouped with all the other dependencies in the Surefire plugin configuration. After migrating everything we still have to figure out how to consolidate the dependency setup.

DiegoCardoso: I could add it there, but it still seems to be needed to add this dependency here as well, according to my tests. It would only omit the `<version>` here in that case.

sissbruecker: It looks like `AllChartTestsSuite` is dead code. It's not picked by up `mvn test` and the test classes it selects are run individually outside of it.

So let's drop that, at which point the dependency it also unnecessary.

DiegoCardoso: Removed the `AllChartTestsSuite` class and reverted the change in the `pom.xml` file.

vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/test/java/com/vaadin/flow/component/spreadsheet/tests/charts/ChartTestBase.java
```diff
@@ -193,13 +195,10 @@ protected void assertDashStyle(List<Series> seriesList,
                         .getDashStyle();
             }
 
-            System.out.println("Dash style is " + actualDashStyle);
```
DiegoCardoso, line 196: Removed this line that was polluting the logs.

PR #8969, opened 2026-03-19
chore: migrate select unit tests to JUnit 6

DiegoCardoso:
## Description

Convert `Select` unit tests to JUnit 6.


vaadin-select-flow-parent/vaadin-select-flow/src/test/java/com/vaadin/flow/component/select/SelectTest.java
```diff
@@ -694,11 +703,12 @@ public void elementHasValue_wrapIntoField_doesNotThrow() {
 
         Mockito.when(instantiator.createComponent(Select.class))
                 .thenAnswer(invocation -> new Select());
-        Select field = Component.from(element, Select.class);
+        Assertions.assertDoesNotThrow(
+                () -> Component.from(element, Select.class));
```
DiegoCardoso: This change is a bit debatable. I noticed SonarCloud reporting tests without an explicit assertion, and I found out that there's an `assertDoesNotThrow` assertion that could be used to prevent this report. For that reason, I added a session in the skill to handle such cases. It shouldn't change the code under test, it just adds an explicit assertion to what we are supposed to test.

sissbruecker: The scenario described in the skill seems like the only reasonable one, but we also have tests that basically only test that something compiles / some API is present. For example: https://github.com/vaadin/flow-components/blob/0991ff6e96bcd68f7df000a1163ca7c97d9e8141/vaadin-side-nav-flow-parent/vaadin-side-nav-flow/src/test/java/com/vaadin/flow/component/sidenav/tests/SideNavItemTest.java#L62-L75

These tests should be improved, just as the ones that want to assert that something doesn't throw. But I think we should do that separately. Reviewing the existing PRs is already challenging with the large number of changes, and it would be easy to miss cases where it now adds a new assertion in tests where it would later be confusing. I think it's also better to let the AI do one thing at a time for large scale tasks, though the complexity here is probably not too high yet.

So I'd prefer to do that separately. We can let Sonar run over the whole codebase and identify tests that do not have assertions. Then try come up with a prompt that specifically identifies cases where not throwing is asserted. Or triage the list manually beforehand and give it a refined list of test cases to fix.

sissbruecker: With the prompt, it might also trip up in cases where a test calls some helper from a base class / util class, which would then rely on the AI actually digging through all code paths to see if there is an assertion somewhere. Using a deterministic tool like Sonar is probably a better fit for this.

DiegoCardoso: Yeah, makes sense. Let's roll back this change to reduce any cognitive load from the migrations.

PR #8978, opened 2026-03-20
refactor: remove redundant cache from grid connector

vursen:
## Summary
- Remove the separate `cache` object from `gridConnector.ts` and use the grid's built-in `dataProviderController.rootCache` directly
- Eliminate redundant cache management functions (`updateGridCache`, `itemToCacheLocation`) that duplicated state already tracked by the grid's data provider controller
- Simplify cache lookups and pending request resolution to work directly with `rootCache.items` and `rootCache.pendingRequests`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,7 +84,7 @@
         <maven.surefire.plugin.version>3.5.5</maven.surefire.plugin.version>
         <maven.failsafe.plugin.version>3.5.5</maven.failsafe.plugin.version>
         <maven.resources.plugin.version>3.5.0</maven.resources.plugin.version>
-        <spotless.plugin.version>3.3.0</spotless.plugin.version>
+        <spotless.plugin.version>3.4.0</spotless.plugin.version>
         <checkstyle.version>13.3.0</checkstyle.version>
         <maven.checkstyle.plugin.version>3.6.0</maven.checkstyle.plugin.version>
         <spotless.licence-header>${maven.multiModuleProjectDirectory}/scripts/templates/apache2-license-header.txt</spotless.licence-header>
@@ -103,7 +103,7 @@
         <spring-data-commons.version>4.0.4</spring-data-commons.version>
         <reactor-core.version>3.8.4</reactor-core.version>
         <langchain4j.version>1.12.2</langchain4j.version>
-        <spring-ai.version>2.0.0-M2</spring-ai.version>
+        <spring-ai.version>2.0.0-M3</spring-ai.version>
         <junit.jupiter.version>6.0.3</junit.jupiter.version>
 
         <!-- spreadsheet -->
--- a/vaadin-accordion-flow-parent/vaadin-accordion-flow/src/main/java/com/vaadin/flow/component/accordion/Accordion.java
+++ b/vaadin-accordion-flow-parent/vaadin-accordion-flow/src/main/java/com/vaadin/flow/component/accordion/Accordion.java
@@ -52,7 +52,7 @@ import com.vaadin.flow.shared.Registration;
  * @author Vaadin Ltd
  */
 @Tag("vaadin-accordion")
-@NpmPackage(value = "@vaadin/accordion", version = "25.1.0-rc1")
+@NpmPackage(value = "@vaadin/accordion", version = "25.2.0-alpha1")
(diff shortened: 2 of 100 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
```diff
@@ -433,15 +375,16 @@ window.Vaadin.Flow.gridConnector.initLazy = (grid) => {
...
       const lastAvailablePage = grid.size ? Math.ceil(grid.size / grid.pageSize) - 1 : 0;
       // It's possible that the lastRequestedRange includes a page that's beyond lastAvailablePage if the grid's size got reduced during an ongoing data request
       const lastRequestedRangeEnd = Math.min(lastRequestedRange[1], lastAvailablePage);
       // Resolve if we have data or if we don't expect to get data
-      if (cache[page]) {
+      const startIndex = page * grid.pageSize;
+      if (rootCache.items[startIndex] !== undefined) {
         // Cached data is available, resolve the callback
-        callback(cache[page]);
+        callback([]);
```
sissbruecker, line 387: Why does this not return the actual page contents?

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
```diff
@@ -205,11 +201,7 @@ window.Vaadin.Flow.gridConnector.initLazy = (grid) => {
       return;
     }
 
-    if (cache[page]) {
-      callback(cache[page]);
-    } else {
-      grid.$connector.debounceRootRequest(page);
-    }
+    grid.$connector.debounceRootRequest(params.page);
```
sissbruecker, line 204: Why are these equivalent?

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
```diff
@@ -305,40 +297,12 @@ window.Vaadin.Flow.gridConnector.initLazy = (grid) => {
...
-  /**
-   * Updates the cache for the given page for grid or tree-grid.
-   *
-   * @param page index of the page to update
-   */
-  const updateGridCache = function (page) {
-    const { rootCache } = dataProviderController;
-
-    // Force update unless there's a callback waiting.
-    if (cache[page] && rootCache.pendingRequests[page]) {
```
sissbruecker, line 317: Why is this not needed anymore?

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
vursen: It's no longer necessary because the updated `updateGridCache` method now adds data to the cache regardless if there is a page pending.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
vursen: Previously, when data update was initiated by the server (e.g. refreshItem), the grid connector manually updated the web component's cache. On the other hand, if data arrived as a response to a client request (e.g. setViewportRange), this check ensured it reached the cache through the web component's page callback. The new approach is that cache updates are now always done by the grid connector regardless of where they come from

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
vursen: Because the web component never requests pages that are already in its cache. It's just that its cache could previously get out of sync with the grid connector's cache and this check served as a safeguard in that case.

PR #9007, opened 2026-03-25
chore: Update testbench to 8.3.2 (23.6)

mcollovati:


pom.xml
sissbruecker, line 1: Should we pick this to 23.7? It's also using 8.3.1 at the moment.

mcollovati: I guess so, but let's ask @ZheSun88

sissbruecker: Replaced usages of `attributeToBe` that caused some tests to fail. I assume there was a behavior change in some newer Selenium version.

ZheSun88: yes, also 23.7

PR #9027, opened 2026-03-26
feat: add grid ai controller

ugur-vaadin:
## Description

Adds `GridAIController` that implements `AIController`.

Constructors:
- `GridAIController(Grid<Map<String, Object>>, DatabaseProvider)`

APIs:
- `String getSystemPrompt()`.

Features:
- Default system prompt
- Grid tools support
- Empty state message
- Built-in renderer support
- End-aligned columns for number columns
- Resizable, sortable, and auto-width by default
- Column grouping support
- Header formatting

Support not included in this PR:
- Custom renderers
- State management and persistence

Fixes https://github.com/orgs/vaadin/projects/103/views/1?filterQuery=&pane=issue&itemId=165824311

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/grid/GridAIController.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/grid/GridAIController.java
@@ -41,11 +41,6 @@ import com.vaadin.flow.component.grid.Grid;
  * <li>Column grouping from dot-separated aliases</li>
  * </ul>
  * <p>
- * Grid state ({@link GridEntry}) is stored directly on the {@link Grid}
- * instance via {@link GridEntry#getOrCreate(Grid, String)}, so there is no
- * separate registry to maintain.
- * </p>
- * <p>
  * State changes requested by the LLM through the update tools are deferred and
  * applied in {@link #onRequestCompleted()}, avoiding partial state and multiple
  * redraws during a multi-tool LLM turn.


vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/grid/GridAIController.java
```diff
@@ -0,0 +1,380 @@
...
+            LOGGER.info("onRequestCompleted: applying query: {}", query);
+            applyQuery(query);
+            currentQuery.set(query);
+            LOGGER.info("onRequestCompleted: grid updated successfully");
+        } catch (Exception e) {
+            LOGGER.error("Error updating grid", e);
+        }
+    }
+
+    // --- Grid rendering ---
```
tomivirkki: Would be nice to move the Grid rendering logic below to a separate [renderer class](https://github.com/vaadin/flow-components/blob/proto/ai-dashboard-dev/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/grid/GridRenderer.java) so that a future DashboardAIController (for example) could re-use it.

Makes sense to still have it tested via the controller API, as it is now.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/grid/GridAIController.java
```diff
@@ -0,0 +1,380 @@
...
+
+    private void applyQuery(String query) {
+        grid.getUI().ifPresentOrElse(ui -> ui.access(() -> {
```
tomivirkki: Could we use `runWhenAttached` to avoid the error?

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/grid/GridAIController.java
ugur-vaadin: Extracted the renderer.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/grid/GridAIController.java
ugur-vaadin: Updated to use `runWhenAttached`.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/grid/GridAIController.java
```diff
@@ -0,0 +1,179 @@
...
+    private static final Logger LOGGER = LoggerFactory
+            .getLogger(GridAIController.class);
+
+    private static final String DEFAULT_GRID_ID = "grid";
+
+    private final Grid<Map<String, Object>> grid;
+    private final DatabaseProvider databaseProvider;
+    private final GridRenderer gridRenderer;
+
+    private final AtomicReference<String> currentQuery = new AtomicReference<>();
```
tomivirkki: If you store the grid's state in the controller, it will get lost on serialize/deserialize because controllers are transient in orchestrator.
Would be better to attach the state to the Grid instance with `ComponentUtil.setData`

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/grid/GridAIController.java
```diff
@@ -0,0 +1,179 @@
...
+
+    @Override
+    public void onRequestCompleted() {
+        var query = pendingQuery.getAndSet(null);
+        if (query == null) {
+            LOGGER.debug("onRequestCompleted: no pending query");
+            return;
+        }
+        LOGGER.info("onRequestCompleted: applying query: {}", query);
+        grid.getElement().getNode().runWhenAttached(ui -> ui.access(() -> {
```
tomivirkki, line 155: Why not do this also inside the renderer?

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/grid/GridRenderer.java
```diff
@@ -0,0 +1,247 @@
...
+                        var firstValue = countResult.getFirst().values()
+                                .iterator().next();
+                        return firstValue instanceof Number n ? n.intValue()
+                                : 0;
+                    }
+                    return 0;
+                }, row -> row);
+
+        grid.setItems(dataProvider);
+    }
```
tomivirkki: Seems that `setupLazyDataProvider` (including the utilities it uses) is untested.

ugur-vaadin: Added tests.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/grid/GridAIController.java
ugur-vaadin: Moved inside the renderer.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/grid/GridAIController.java
ugur-vaadin: Updated to use `ComponentUtil.setData` in order to set the query state.

tomivirkki: Not sure why the grid referene was moved to the renderer...now it's required to create a separate renderer for each grid insteance. Instead, the renderGrid method could accept the grid and read the state from it using ComponentUtil

ugur-vaadin: I thought it would be more natural this way because I assumed we would usually have a 1-1 grid-database relation. I might be wrong though. It is also pretty lightweight. However, I guess it might be better to align it with the ChartRenderer approach. What do you think?

tomivirkki: Could the renderer be passed both (in the method call), the grid and the db provider (from the controller)? Ah, maybe it also makes sense to also pass the query so renderer wouldn't need to know about how the state is stored?

ugur-vaadin: Made the renderer stateless. Added a `GridEntry` class similar to the `ChartEntry` to handle state.

tomivirkki: Looks good, made the same change to Chart renderer

ugur-vaadin: Removed `applyPendingState`. `GridRenderer` now only exposes `renderGrid`. Since it is not concerned with any state, I moved `runWhenAttached` logic back to the controller to have better control over the flow.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/grid/GridAIController.java
```diff
@@ -0,0 +1,172 @@
...
+ * <li>Built-in renderers for dates and numbers</li>
+ * <li>Right-alignment for numeric columns</li>
+ * <li>Resizable, sortable columns with auto-width</li>
+ * <li>Column grouping from dot-separated aliases</li>
+ * </ul>
+ * <p>
+ * Grid state ({@link GridEntry}) is stored directly on the {@link Grid}
+ * instance via {@link GridEntry#getOrCreate(Grid, String)}, so there is no
+ * separate registry to maintain.
+ * </p>
```
tomivirkki: `GridEntry` in an internal technical detail, never exposed to public API so it shouldn't be described in the JavaDocs.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/test/java/com/vaadin/flow/component/ai/grid/GridEntryTest.java
tomivirkki, line 1: Since `GridEntry` is an internal technical detail, maybe we shouldn't have test suite for it? Ideally we should be able to refactor the state architecture of the controllers later without breaking the tests. As long as the `GridEntry` is fully covered via the controller-level public API in `GridAIControllerTest` (seem to be close to it currently), this separate file shouldn't be necessary.

ugur-vaadin: Removed the tests.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/grid/GridAIController.java
ugur-vaadin: Removed the mention.

PR #9033, opened 2026-03-27
refactor: extract test for common signal behaviors

sissbruecker:
## Description

- Added `generateBindingTests` to `AbstractSignalsTest`, generating a common suite of JUnit 5 dynamic tests for a signal binding on a component property
  - Covers synchronization while attached, applying the initial value while detached, no synchronization while detached, resynchronization after attach, and `BindingActiveException` on manual set and rebind
- Added test cases that were not covered before: binding a `null` signal throws `NullPointerException`, and the signal's initial value is applied while the component is detached
- Migrated `DatePickerSignalTest` to the new helper for `bindMin`, `bindMax` and `bindInitialPosition`
- Added public `setUp` and `tearDown` methods to `MockUIExtension` so each dynamic test can set up a fresh mock UI, since dynamic tests do not run lifecycle callbacks

## Type of change

- Refactor

> [!NOTE]
> Other components' signal tests (e.g. `TimePickerSignalTest`, `NumberFieldSignalTest`, `IntegerFieldSignalTest`) still contain the duplicated binding tests and can be migrated to the helper in follow-up PRs.
>
> The attach -> bind -> detach -> no change test case was dropped as unnecessary, the attach / detach behavior lives in the Flow signal implementation anyway.


vaadin-flow-components-shared-parent/vaadin-flow-components-test-util/src/main/java/com/vaadin/tests/AbstractSignalsTest.java
```diff
@@ -25,4 +41,170 @@
...
+    }
+
+    /**
+     * Creates a dynamic test that sets up and tears down a mock UI around the
+     * test executable, since JUnit 5 dynamic tests do not support lifecycle
+     * callbacks.
+     */
+    private DynamicTest createTest(String name, Executable test) {
+        return dynamicTest(name, () -> {
+            ui.setUp();
```
vursen, line 202: This could be moved to beforeEach/afterEach. From what I found online, they should work with dynamic tests too

sissbruecker: I gave it a try, those would only run around `bindMin` for example, but not for each dynamic test.

vursen (approved):

Issue #8863, opened 2026-03-06
Slider only supports Double values


mstahv:
### Description

The new Slider component (25.1) has its value type fixed to Double. In many real-world use cases, the APIs being bound to use Integer, Long, or other Number types, requiring manual conversion or custom Binder converters.

SliderBase<TComponent, TValue> is already well-designed for this — it accepts converter functions between the web component's Double representation and an arbitrary model type. However, SliderBase is currently package-private, so it cannot be extended outside com.vaadin.flow.component.slider without resorting to split-package tricks (which break in modular runtimes like Quarkus).

For my test app I vibed a following helper (and copied SliderBase etc to my custom package), but a solution with class variants for other types of Number's might be better (I have used that approach previously for regular number fields).

```java 
/**
 * A slider component that supports arbitrary {@link Number} types as its value.
 * <p>
 * While the standard {@link com.vaadin.flow.component.slider.Slider} is fixed
 * to {@link Double} values, this component can work with {@link Integer},
 * {@link Long}, or any other {@link Number} subtype, making it easier to bind
 * to existing APIs that use specific number types.
 * <p>
 * For standard JDK number types, just pass the class:
 *
 * <pre>
 * var intSlider = new NumberSlider&lt;&gt;(Integer.class);
 * var longSlider = new NumberSlider&lt;&gt;("Offset", Long.class, 0, 1000);
 * </pre>
 *
 * @param <T>
 *            the number type for the slider value
 */
@Tag("vaadin-slider")
@NpmPackage(value = "@vaadin/slider", version = "25.1.0-beta1")
@JsModule("@vaadin/slider/src/vaadin-slider.js")
public class NumberSlider<T extends Number>
        extends SliderBase<NumberSlider<T>, T> implements HasAriaLabel {

    @SuppressWarnings("unchecked")
    private static final Map<Class<? extends Number>, SerializableFunction<Double, ?>> CONVERTERS = Map.of(
            Integer.class, (SerializableFunction<Double, Integer>) Double::intValue,
            Long.class, (SerializableFunction<Double, Long>) Double::longValue,
            Float.class, (SerializableFunction<Double, Float>) Double::floatValue,
            Short.class, (SerializableFunction<Double, Short>) Double::shortValue,
            Byte.class, (SerializableFunction<Double, Byte>) Double::byteValue,
            Double.class, (SerializableFunction<Double, Double>) d -> d);

    private final SerializableFunction<Double, T> fromDouble;

    /**
[...]

Issue #8799, opened 2026-02-27
Flow components signal APIs with broken / inconsistent behavior


sissbruecker:
These are the results of an audit, checking existing Flow components one-way signal binding methods across three categories of issues:
  
1. **Invalid one-way binding:** The component internally modifies the bound state (e.g., in event handlers), causing `BindingActiveException` at runtime — the binding simply cannot work.  
2. **Unguarded imperative API:** An imperative API can modify the bound state without throwing `BindingActiveException`, allowing silent desynchronization with the signal.  
3. **Missing side effect:** The imperative setter performs side effects beyond setting the property that are not replicated when the signal updates the value.

## Summary

| #   | Finding                                                                                                      | Category                 |
| --- | ------------------------------------------------------------------------------------------------------------ | ------------------------ |
| 1   | `Details.setSummary(Component)` bypasses `bindSummaryText` guard                                             | Unguarded imperative API |
| 2   | `Notification.add(Component)` conflicts with `bindText` without guard                                        | Unguarded imperative API |
| 3   | `AbstractLogin` — inherited `bindEnabled` conflicts with internal `setEnabled` calls                         | Invalid one-way binding  |
| 4   | `AbstractNumberField.bindMin` does not set `minSetByUser` flag                                               | Missing side effect      |
| 5   | `SliderBase.bindMin/Max/Step` miss `schedulePropertyConsistencyCheck`                                        | Missing side effect      |
| 6   | `Button` with `disableOnClick` — `DisableOnClickController` calls `setEnabled(false)` on click               | Invalid one-way binding  |
| 7   | `Dialog.bindVisible` does not trigger `applyModality()`                                                      | Missing side effect      |
| 8   | `RadioButtonGroup.bindReadOnly` does not call `refreshButtons()`                                             | Missing side effect      |
| 9   | `DateTimePicker.bindReadOnly` does not propagate to internal pickers                                         | Missing side effect      |
| 10  | `Dashboard`/`DashboardWidget`/`DashboardSection` — `bindVisible` bypasses `setVisible` throw                 | Unguarded imperative API |
| 11  | `ListBox`/`MultiSelectListBox` — `bindRequiredIndicatorVisible` bypasses `setRequiredIndicatorVisible` throw | Unguarded imperative API |
| 12  | `MenuItemBase` — `add(Component)` unguarded when `bindText` is active                                        | Unguarded imperative API |

### Finding 1: `Details.setSummary(Component)` bypasses `bindSummaryText` guard  
[...]

PR #9108, opened 2026-04-13
refactor!: extract `NumberSlider` / `NumberRangeSlider`

sissbruecker:
## Description

As a preparation for introducing an integer variant of Slider and Range Slider, this extracts common logic for single value / range value sliders into `NumberSlider` / `NumberRangeSlider`.

This moves the min, max and step properties to `SliderBase` using a generic number type. Likewise, the start and end values of range slider values now also use a generic type to facilitate internal logic. As a side effect, all of these now require passing a double literal (`10.0`, `10d`) as auto-converting from an integer literal to a boxed double is not supported by Java.

Part of https://github.com/vaadin/flow-components/issues/8863

## Type of change

- Breaking change

### DIFF
--- a/vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/NumberRangeSlider.java
+++ b/vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/NumberRangeSlider.java
@@ -60,7 +60,7 @@ abstract class NumberRangeSlider<TComponent extends NumberRangeSlider<TComponent
      *            a function to convert a value of the slider's numeric type to
      *            double
      */
-    protected NumberRangeSlider(TNumber min, TNumber max,
+    NumberRangeSlider(TNumber min, TNumber max,
             SerializableBiFunction<TNumber, TNumber, TValue> rangeFactory,
             SerializableFunction<Double, TNumber> fromDouble,
             SerializableFunction<TNumber, Double> toDouble) {
--- a/vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/NumberRangeSlider.java
+++ b/vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/NumberRangeSlider.java
@@ -1,0 +1,174 @@
+/*
+ * Copyright 2000-2026 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component.slider;
+
+import java.util.Optional;
+
+import com.vaadin.flow.function.SerializableFunction;
+
+import tools.jackson.databind.node.ArrayNode;
+
+/**
+ * Base class for sliders that allow selecting a range of numeric values within
+ * configured bounds.
+ * 
+ * @param <TComponent>
+ *            the type of the slider component
+ * @param <TValue>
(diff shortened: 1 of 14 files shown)


vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/SliderBase.java
```diff
@@ -198,7 +232,7 @@ void setStepDouble(double step) {
      *      SerializableConsumer)
      * @since 25.1
      */
-    public SignalBinding<Double> bindMin(Signal<Double> signal) {
+    public SignalBinding<TNumber> bindMin(Signal<TNumber> signal) {
         return getElement().bindProperty("min", signal, null);
     }
```
sissbruecker, line 236: The bind methods work differently than the imperative setters. The setter converts from the number type to double and stores a double in the property. The getter reverses that.

The bind methods store whatever the number type is directly in the property. Theoretically we could compute a mapped signal, but then binding that to the property would return a `SignalBinding<Double>` instead of `SignalBinding<TNumber>`. It would be possible to add more logic to return a wrapping signal binding that does the conversion back, but I'm not sure if all that is worth it. I tried binding an integer signal to the property and then calling `getProperty` with the double signature, which seems to work fine. Flow seems to convert between values under the hood. So I'm in favor of just leaving this as is for now. If this turns out to be incorrect, we can still add the mapping logic without having to break the API.

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/NumberRangeSlider.java
```diff
@@ -0,0 +1,180 @@
...
+     * @see #getMin()
+     * @see #getMax()
+     */
+    @Override
+    public void clear() {
+        setValue(rangeFactory.apply(getMin(), getMax()));
+    }
+
+    @Override
+    protected boolean hasValidValue() {
```
vursen: I think hasValidValue could be moved to BaseSlider as well since that class already has all the required types as generic parameters.

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/NumberSlider.java
```diff
@@ -0,0 +1,154 @@
...
+     * @param max
+     *            the maximum value
+     * @param fromDouble
+     *            a function to convert a double value to the slider's value
+     *            type
+     * @param toDouble
+     *            a function to convert a value of the slider's value type to
+     *            double
+     */
+    protected NumberSlider(TValue min, TValue max,
```
vursen: Should this constructor also be public, same as the one in NumberRangeSlider?

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/NumberRangeSlider.java
sissbruecker: Good catch, done.

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/NumberSlider.java
sissbruecker: There isn't really any point in having the constructor public, as you can (typically) only extend `Slider` and have to call its constructors instead. So this was just an oversight on my part.

This brings up a good point though. Having a visibility modifier on the constructor is potentially misleading. Also some of the fields shared between the slider variants were `protected` which exposes them to custom subclasses. So I made the constructors and (shared) fields package-private instead.

vursen (approved):

vaadin-slider-flow-parent/vaadin-slider-flow/src/main/java/com/vaadin/flow/component/slider/SliderBase.java
vursen: I've just noticed that signal methods also don't schedule consistency checks, which I apparentely overlooked when I was adding them.

sissbruecker: It was discussed in https://github.com/vaadin/flow-components/issues/8799, and deemed as not a high priority issue. I can try to take a look at it later on if I don't forget.

PR #9113, opened 2026-04-14
fix: prevent infinite focus loop and stale selection in custom editors

### LINKED ISSUE #9036: [Spreadsheet] ComboBox inconsistent text selection and other odd behavior
### Description

Ultimately the issue that I want fixed that is the problem in our application is that text autoselect in a comboBox in a spreadsheet does not always work. But in the example code provided, there are also a number of other weird behaviors that may or may not be related.

Specifically, it seems that sometimes after `comboBox.setValue()` is called within `onCustomEditorDisplayed()`, the text that gets set does not get selected.

### Expected outcome

Any time `comboBox.setValue()` is called within `onCustomEditorDisplayed()` the new text that is set gets selected if autoSelect is turned on.

```java
package com.hpd.demo.views;

import com.vaadin.flow.component.Component;
import com.vaadin.flow.component.combobox.ComboBox;
import com.vaadin.flow.component.spreadsheet.Spreadsheet;
import com.vaadin.flow.component.spreadsheet.SpreadsheetComponentFactory;
import org.apache.poi.ss.usermodel.Cell;
import org.apache.poi.ss.usermodel.Sheet;

@Route("spreadsheet-combo")
public class SpreadsheetComboBoxView extends VerticalLayout {

    public SpreadsheetComboBoxView() {
        setSizeFull();

        // 1. Create the Spreadsheet
        Spreadsheet spreadsheet = new Spreadsheet();

        // 2. Add some label text in A2 so we know where to look
        spreadsheet.createCell(1, 0, "Select Fruit:"); // Row 1 (2nd row), Col 0 (A)
        spreadsheet.refreshCells(spreadsheet.getCell(1, 0));

        // 3. Define the Component Factory
        spreadsheet.setSpreadsheetComponentFactory(new SpreadsheetComponentFactory() {

            @Override
            public Component getCustomComponentForCell(Cell cell, int rowIndex, int columnIndex,
                                                       Spreadsheet spreadsheet, Sheet sheet) {


                return null;
            }

            @Override
            public Component getCustomEditorForCell(Cell cell, int rowIndex, int columnIndex,
                                                    Spreadsheet spreadsheet, Sheet sheet) {
                if (rowIndex == 1 && columnIndex >= 1 && columnIndex <= 4) {
                    return createComboBoxForCell(cell, spreadsheet);
                }
                return null;
            }

            @Override
            public void onCustomEditorDisplayed(Cell cell, int rowIndex, int columnIndex,
                                                Spreadsheet spreadsheet, Sheet sheet, Component customEditor) {

                if (customEditor instanceof ComboBox<?> comboBox) {
                    //((ComboBox<String>) comboBox).setValue(null);
                    try {
                        Thread.sleep(1000);
                    } catch (InterruptedException e) {
                        throw new RuntimeException(e);
                    }
                    ((ComboBox<String>) comboBox).setValue("Value Reset " + columnIndex );
                }

                System.out.println("onCustomEditorDisplayed " + columnIndex);
            }
        });

        add(spreadsheet);
    }

    private Component createComboBoxForCell(Cell cell, Spreadsheet spreadsheet) {
        comboBox.setItems("Apple", "Banana", "Cherry");

        comboBox.setWidthFull();

        if (cell != null) {
            comboBox.setValue(cell.getStringCellValue());
        }

            if (cell != null) {
                cell.setCellValue(e.getValue());
                spreadsheet.refreshCells(cell);
            }
        });

//        comboBox.getElement().executeJs(
//                "$0.addEventListener('click', (e) => $0.inputElement.select());", comboBox.getElement());
        comboBox.getElement().setAttribute("autoselect", true);
        comboBox.addAttachListener(ev -> {
            // comboBox.getElement().executeJs(...)
        });
        return comboBox;
    }
}
```

### Steps to reproduce

I'm not 100% sure of the exact steps to produce the behavior. I'm able to produce some odd behavior where the text in the comboBoxes will get wiped out with clicks on other parts of the sheet, or when resizing. When one of the elements in the drop down is selected, click elsewhere to deselect the cell, and then comeback and click the comboBox again, the `onCustomEditorDisplayed()` method will setValue() the text to the new value, but that new value won't get selected.

In the attached video, I had managed to get it into some state through clicking around on the sheet, the comboboxes and selecting from the drop downs where the autoselect stops working even for the `Value Reset #` text values where the text did not change.

https://github.com/user-attachments/assets/8d76914d-2588-4beb-b0ba-1e6f6acfd4aa

### Environment

Vaadin version(s): 24.10.0
OS: Windows

### Browsers

Firefox

DiegoCardoso:
## Summary
- Fix infinite `cellSelected` feedback loop triggered by programmatic focus changes (e.g. `inputElement.select()`) in `onCustomEditorDisplayed` callbacks
- Prevent delayed server responses from hijacking the cell selection when the user has already navigated away, including via keyboard navigation through cells with custom editors
- Add server-side reentrancy guard to prevent recursive `onCustomEditorDisplayed` calls when callback code calls `refreshCells()`

## Root cause
`CustomEditorEventListener.ONFOCUS` unconditionally sent `cellSelected` to the server and updated the client-side selection on every focus event — including programmatic focus from `inputElement.select()` called in `onCustomEditorDisplayed`. When the callback had non-trivial processing time, the delayed response would steal focus and selection from whatever cell the user had since navigated to, and could trigger an infinite client→server→client loop.

## Approach
**Client-side** (`CustomEditorEventListener`): track `mousedown`/`touchstart` events to distinguish user-initiated focus from programmatic focus. Only update the selection and send `cellSelected` for user-initiated focus. For programmatic focus on a cell the user has already left, return focus to the sheet. Keyboard navigation is unaffected because `SelectionHandler` already handles selection updates and `cellSelected` independently.

**Server-side** (`Spreadsheet`): reentrancy guard (`insideCustomEditorCallback`) on `loadCustomEditorOnSelectedCell()` to prevent recursive `onCustomEditorDisplayed` calls when user code in the callback calls `refreshCells()`.

Fixes #9036

## Test plan
- [ ] `CustomEditorIT#comboBoxEditorWithSelect_clickCell_callbackFiredOnce` — verifies single click triggers exactly 1 callback (no loop)
- [ ] `CustomEditorIT#comboBoxEditorWithSelect_switchBetweenCells_noInfiniteLoop` — rapidly clicks through editor cells to a non-editor cell, verifies selection stays on the final cell and callback count is bounded
- [ ] All 26 existing `CustomEditorIT` tests pass (keyboard navigation, Tab, Shift+Tab, F2, Enter, ESC, shared editors, frozen panes, always-visible mode)

### DIFF
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/CustomEditorEventListener.java
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/CustomEditorEventListener.java
@@ -64,8 +64,7 @@ public class CustomEditorEventListener implements EventListener {
                 break;
             }
             break;
-        case Event.ONMOUSEDOWN:
-        case Event.ONTOUCHSTART:
+        case Event.ONMOUSEDOWN, Event.ONTOUCHSTART:
             userInitiatedFocus = true;
             break;
         case Event.ONFOCUS:
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/CustomEditorEventListener.java
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/CustomEditorEventListener.java
@@ -30,13 +30,14 @@ public class CustomEditorEventListener implements EventListener {
     private Slot slot;
     private String cellAddress;
     private SpreadsheetWidget widget;
+    private boolean userInitiatedFocus;
 
     public void init(Slot slot, String cellAddress) {
         this.slot = slot;
         this.cellAddress = cellAddress;
         Event.setEventListener(slot.getAssignedElement(), this);
-        DOM.sinkEvents(slot.getAssignedElement(),
-                Event.ONKEYDOWN | Event.FOCUSEVENTS);
+        DOM.sinkEvents(slot.getAssignedElement(), Event.ONKEYDOWN
+                | Event.FOCUSEVENTS | Event.ONMOUSEDOWN | Event.ONTOUCHSTART);
     }
 
     public void setCellAddress(String cellAddress) {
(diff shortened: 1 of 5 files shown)


sissbruecker (approved):
Doesn't seem to work with keyboard selection and navigating with the keyboard into a cell with an editor breaks the custom selection logic. But it's better than before with fixing the infinite loops, so approved.

Edit: For cell navigation with the keyboard the focus probably shouldn't move into the editor as that would be quite disruptive. So even less of an issue.

PR #9130, opened 2026-04-16
chore: update workflow to work with external contributors [skip ci]

ZheSun88:
- use `pull_request_target` instead of `pull_request`, so the secrets can be used for PR coming from forked repo
- added the `environment` checking which asks team to approval before running PRs from external contributors

### DIFF
--- a/.github/workflows/checkstyle-check.yml
+++ b/.github/workflows/checkstyle-check.yml
@@ -1,9 +1,7 @@
 name: Checkstyle Check
 
 on:
-  pull_request_target:
-    types: [ opened, synchronize, reopened ]
-    branches: [ main ]
+  pull_request:
 
 permissions:
   contents: read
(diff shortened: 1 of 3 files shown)


.github/workflows/checkstyle-check.yml
sissbruecker, line 1: For checkstyle and the formatter `pull_request` would be enough since those don't need any secrets.

.github/workflows/sonar-scan.yml
```diff
@@ -1,13 +1,16 @@
 name: Sonar Scan
 
 on:
-  pull_request:
+  pull_request_target:
+    types: [ opened, synchronize, reopened ]
+    branches: [ main ]
```
sissbruecker, line 6: Do we need to filter the branch here? We probably also want to run this in other branches?

ZheSun88: yes, this is still needed, without this branch filter, all other branches will use the workflow from main branch, that is the requirement from `pull_request_target`

.github/workflows/checkstyle-check.yml
ZheSun88: updated

sissbruecker (approved):

PR #9147, opened 2026-04-20
refactor: extract viewport range computation in grid connector

vursen:
## Description

Refactor the page-request logic in `gridConnector.ts` so the viewport range is computed in one place and can be reused by both `gridConnector` and `treeGridConnector`:

- Extract `getViewportRange` helper on `grid.$connector` from the inline logic previously in `fetchPage`, and reuse it from `treeGridConnector.ts` in place of its local `getViewportRange`.
- Inline `fetchPage` into the renamed `requestPage` (formerly `debounceRootRequest`), and move the debouncer call directly into `dataProvider` since `debounceRootRequest` had no other callers.
- Rename `rootRequestDelay`/`rootRequestDebouncer` to `requestDebouncerDelay`/`requestDebouncer` since they apply to all requests, not just root.

## Type of change

- Refactor

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/treeGridConnector.ts
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/treeGridConnector.ts
@@ -56,7 +56,7 @@ window.Vaadin.Flow.treeGridConnector.initLazy = function (grid) {
       return;
     }
 
-    const [start, end] = this.$connector.getViewportRange();
+    const [start, end] = grid.$connector.getViewportRange();
     const padding = Math.floor((end - start) * 1.5);
     const flatIndex = await grid.$server.setViewportRangeByIndexPath(indexes, padding);
     grid._scrollToFlatIndex(flatIndex);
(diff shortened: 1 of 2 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/treeGridConnector.ts
```diff
@@ -56,7 +56,7 @@ window.Vaadin.Flow.treeGridConnector.initLazy = function (grid) {
       return;
     }
 
-    const [start, end] = getViewportRange();
+    const [start, end] = this.$connector.getViewportRange();
```
sissbruecker: Nit: Grid is accessed with `this` here, and with `grid` in other places of this function. Should work the same, but might be good to align to prevent confusion on what the difference is.

sissbruecker (approved):

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/treeGridConnector.ts
vursen:
```suggestion
    const [start, end] = grid.$connector.getViewportRange();
```

PR #9166, opened 2026-04-22
docs: add examples on how to retrieve schemas

ugur-vaadin:
## Description

Adds examples on how to retrieve schemas for

- H2
- MySQL
- PostgreSQL

Fixes https://github.com/orgs/vaadin/projects/103/views/1?filterQuery=&pane=issue&itemId=178097610

## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] Docs

## Checklist

### DIFF
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/provider/DatabaseProvider.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/provider/DatabaseProvider.java
@@ -74,14 +74,19 @@ import java.util.Map;
  * public String getSchema() {
  *     try (var connection = readOnlyDataSource.getConnection()) {
  *         var meta = connection.getMetaData();
+ *         var catalog = connection.getCatalog();
+ *         var schemaName = connection.getSchema();
  *         var schema = new StringBuilder();
- *         try (var tables = meta.getTables(null, null, "%",
+ *         try (var tables = meta.getTables(catalog, schemaName, "%",
  *                 new String[] { "TABLE" })) {
  *             while (tables.next()) {
  *                 var table = tables.getString("TABLE_NAME");
- *                 // Append columns via meta.getColumns
- *                 // Append primary keys via meta.getPrimaryKeys
- *                 // Append foreign keys via meta.getImportedKeys
+ *                 // Append columns via meta.getColumns(catalog, schemaName,
+ *                 // table, "%")
+ *                 // Append primary keys via meta.getPrimaryKeys(catalog,
+ *                 // schemaName, table)
+ *                 // Append foreign keys via meta.getImportedKeys(catalog,
+ *                 // schemaName, table)
  *             }
  *         }
  *         return schema.toString();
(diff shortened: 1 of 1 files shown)


vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/provider/DatabaseProvider.java
```diff
@@ -60,6 +60,55 @@
...
  *
+ * <p>
+ * <b>Dynamic schema retrieval:</b> Hand-writing the schema string works for
+ * small fixed schemas, but most databases expose introspection commands you can
+ * run on the read-only connection to build the description at runtime. Pick the
+ * idiom that matches your database:
+ * </p>
+ * <ul>
+ * <li><b>H2 (2.x, including in-memory):</b> run
+ * {@code SCRIPT NODATA NOSETTINGS} — returns the DDL needed to recreate the
```
tomivirkki: The above statement "databases expose introspection commands you can run on the read-only connection" conflicts with executing `SCRIPT`, which requires admin rights in H2. The output of `SCRIPT NODATA NOSETTINGS` also includes user, role, sequence, and grant statements — account metadata we'd want to exclude from the LLM prompt.

The Postgres example also generates a schema without primary/foreign keys, so joins are left for the LLM to guess.

Maybe instead of documenting vendor-specific examples, we could suggest using `java.sql.DatabaseMetaData` (`getTables` / `getColumns` / `getPrimaryKeys` / `getImportedKeys`), which works consistently across H2, MySQL, and Postgres?

ugur-vaadin: Updated to use `DatabaseMetaData`. The example points to the use of `getColumns`, `getPrimaryKeys`, and `getImportedKeys` rather than fully implementing them in order to keep it brief.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/provider/DatabaseProvider.java
```diff
@@ -60,6 +60,37 @@
...
+ * names:
+ * </p>
+ *
+ * <pre>
+ * &#064;Override
+ * public String getSchema() {
+ *     try (var connection = readOnlyDataSource.getConnection()) {
+ *         var meta = connection.getMetaData();
+ *         var schema = new StringBuilder();
+ *         try (var tables = meta.getTables(null, null, "%",
```
tomivirkki: Could you verify that `getTables(null, null, "%", …)` returns consistent, expected results across H2, MySQL, and PostgreSQL? The catalog and schemaPattern arguments may be interpreted differently by each driver, and since the snippet is pitched as working consistently across all three, it'd be good to confirm that holds in practice before readers copy it.

ugur-vaadin: Updated the example to use the actual catalog and schema so that it works consistently throughout all 3. Tested all 3 and it seems fine.

tomivirkki (approved):

PR #9167, opened 2026-04-22
docs: note chart tool schema provider compatibility

tomivirkki:
## Summary

- Replace `oneOf` with `anyOf` in chart tool schemas (`{c:HEIGHT}`, `{c:TITLE}`, `{c:SUBTITLE}`, `{c:Y_AXIS}`). The alternatives differ only by `type` in every site, so the two are semantically equivalent here — but `anyOf` is more portable form
- Add a *Provider compatibility* paragraph to `ChartAIController`'s Javadoc noting the chart tools remain incompatible with OpenAI's strict tool-calling mode (schemas use optional properties, which strict mode disallows). Strict mode is off by default in LangChain4J and Spring AI, so only users who explicitly opt in with `strictTools(true)` are affected.

### DIFF
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/chart/ChartAIController.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/chart/ChartAIController.java
@@ -85,14 +85,13 @@ import com.vaadin.flow.shared.Registration;
  * {@link #getState()} after each successful AI request.
  * </p>
  * <p>
- * <b>Provider compatibility:</b> The chart tools use {@code anyOf} unions and
- * optional properties in their parameter schemas, and are therefore
- * incompatible with OpenAI's strict tool-calling mode (used when the OpenAI
- * tool definition is sent with {@code "strict": true}). Strict tool calling is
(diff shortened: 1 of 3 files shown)


vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/chart/ChartAIController.java
```diff
@@ -84,6 +84,16 @@
  * to be notified when the chart state changes, for example to persist
  * {@link #getState()} after each successful AI request.
  * </p>
+ * <p>
+ * <b>Provider compatibility:</b> The chart tools use {@code anyOf} unions and
```
ugur-vaadin: "anyOf" seems to be [supported](https://developers.openai.com/api/docs/guides/structured-outputs) in OpenAI's strict tool-calling mode, at least under some conditions. I'm not sure if it makes sense to clarify it or it would be overkill. What do you think?

tomivirkki: Good catch, dropped the mention of `anyOf`

ugur-vaadin (approved):

PR #9169, opened 2026-04-22
fix: Change client selectAll update to use LinkedHashSet

### LINKED ISSUE #9170: Grid select all checkbox does not preserve data provider order
### Description

Grid select all checkbox does not preserve data provider order in all cases

### Expected outcome

I would expect the order to be same as the order of the data provider.

For reference Grid intentionally uses LinkedHashMap internally so that order of the individually selected items is preserved. Hence this would be logically also here.

Trivial

### Steps to reproduce

Trivial

### Environment

Vaadin version(s): Vaadin 25

### Browsers

_No response_

TatuLund:
Fixes https://github.com/vaadin/flow-components/issues/9170

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridMultiSelectionModel.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridMultiSelectionModel.java
@@ -233,7 +233,7 @@ public abstract class AbstractGridMultiSelectionModel<T>
                 (Set<T>) getGrid().getDataCommunicator().getDataProvider()
                         .fetch(getGrid().getDataCommunicator().buildQuery(0,
                                 Integer.MAX_VALUE))
-                        .collect(Collectors.toSet()),
+                        .collect(Collectors.toCollection(LinkedHashSet::new)),
                 Collections.emptySet());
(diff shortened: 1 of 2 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/AbstractGridMultiSelectionModel.java
sissbruecker, line 1: Consider this test:
```java
    @Test
    void selectAll_withPreSelection_preservesDataProviderOrder() {
        selectionModel.select(PERSON_C);
        selectionModel.selectAll();

        var value = selectionModel.getSelectedItems().stream().toList();

        assertEquals(PERSON_A, value.get(0));
        assertEquals(PERSON_B, value.get(1));
        assertEquals(PERSON_C, value.get(2));
    }
```

It fails with:
```
org.opentest4j.AssertionFailedError: 
Expected :a(1)
Actual   :c(3)
```

Expected?

TatuLund: Yes, and it should fail. Test selects C first and the rest after that, so selection order should be C, A, B. I.e. it follows the order of the user actions.

sissbruecker (approved):

PR #9189, opened 2026-04-24
refactor: use column references instead of ID lookup

sissbruecker:
`GridColumnElement` is currently not an actual `TestBenchElement`, but rather "virtual" element that allows referencing columns in the grid via IDs that are generated per column by `GridElement` whenever columns are accessed. The JavaDoc mentions something about polyfilled browsers, but it seems this this is obsolete and using column references works fine now.

This refactors `GridColumnElement` to be a `TestBenchElement`, removes the old infrastructure around ID handling and simplifies several methods that work with columns. Includes the same change for Grid Pro TestBench helpers.


vaadin-grid-flow-parent/vaadin-grid-testbench/src/main/java/com/vaadin/flow/component/grid/testbench/GridColumnElement.java
```diff
@@ -19,28 +19,9 @@
...
- * element. This is not a TestBenchElement as polyfilled browsers are not
- * capable of finding it or handling it as a web element.
+ * element.
  */
-public class GridColumnElement {
-
-    private GridElement grid;
-    private Long __generatedId;
-
-    public GridColumnElement(Long __generatedId, GridElement grid) {
```
sissbruecker, line 30: Decided to drop the constructor. It's public, but I have a hard time seeing how someone might have directly used this.

web-padawan (approved):

PR #9194, opened 2026-04-24
feat: add focusSelectedItem flag to combo-box

DiegoCardoso:
## Summary

Re-enables an opt-in version of the dropdown behavior removed in [web-components#6055](https://github.com/vaadin/web-components/pull/6055): when the dropdown opens, the overlay scrolls to and visually focuses the currently-selected item. Useful for long lists where the user wants to see/confirm their selection on open.

Adds `ComboBox.setFocusSelectedItem(boolean)` / `isFocusSelectedItem()` plus an `@ClientCallable resolveSelectedItemIndex()` that resolves the authoritative index for the current value via `ItemIndexProvider` (lazy) or the list data view (in-memory). The connector listens for the dropdown-open event and scrolls to the resolved index.

Fixes [vaadin/web-components#6061](https://github.com/vaadin/web-components/issues/6061).

## Depends on

- **#9220** — `fix(combo-box): preserve committed pages in connector range management`. This PR is stacked on top; base will be retargeted to `main` once #9220 merges. Without #9220, the `scrollToIndex` to a far/unloaded page leaves the dropdown stuck loading.
- **[vaadin/web-components#11552](https://github.com/vaadin/web-components/pull/11552)** — adds the `comboBox.scrollToIndex(index)` method the connector calls. Must land in `@vaadin/combo-box` before this PR can merge.

## Notes for reviewers

A summary comment with the deeper rationale (token-based cancellation, close-time `KeyMapper` re-key handling, microtask polling for the connector flag) is posted as a separate PR comment to keep the diff readable.

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
@@ -327,9 +327,7 @@ window.Vaadin.Flow.comboBoxConnector.initLazy = (comboBox) => {
   // Prevent setting the custom value as the 'value'-prop automatically
   comboBox.addEventListener('custom-value-set', (e) => e.preventDefault());
 
-  // Feature flag for the focus-selected-item behavior. Set from the Java side
-  // via `comboBox.$connector.setFocusSelectedItem(value)` so it doesn't need
-  // to live as an element property the web component would otherwise ignore.
+  // Set from the Java side via `comboBox.$connector.setFocusSelectedItem(value)`.
   let focusSelectedItemEnabled = false;
   comboBox.$connector.setFocusSelectedItem = (value) => {
     focusSelectedItemEnabled = !!value;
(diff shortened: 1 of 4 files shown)


vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -280,6 +280,98 @@ window.Vaadin.Flow.comboBoxConnector.initLazy = (comboBox) => {
...
+      } else {
+        invoke();
+      }
+    });
+  };
+
+  comboBox.addEventListener('vaadin-combo-box-dropdown-opened', resolveFocusSelectedItem);
+  // `_onOpened` doesn't re-run on filter change, so the web component can't
+  // re-arm its own `__shouldFocusSelectedItem` latch — re-resolve ourselves.
+  comboBox.addEventListener('filter-changed', () => {
```
sissbruecker: I think I would prefer if it would not do that. If I enter a new filter, I likely want to select a different item. Based on the filtered results I then want to scan the list from top to bottom. Then it doesn't sound helpful that it scrolls down somewhere to the current selection.

DiegoCardoso:
**Reviewer notes — context for the trimmed comments**

The latest commit (`8576f92e7d`) shortens a few inline comments that read more like commit-message material than code documentation. Posting the historical context here so reviewers have it without bloating the source files.

### `comboBoxConnector.js` — `focusSelectedItemEnabled` flag storage

The flag lives in a closure variable in the connector rather than as a property on the web-component element because the web-component does not declare or react to a \`focusSelectedItem\` property — Flow is the only consumer of this signal, so an element property would just be ignored by the WC and would expose a non-functional surface to non-Flow callers. \`comboBox.\$connector.setFocusSelectedItem(value)\` keeps the surface explicit and Flow-scoped.

### `comboBoxConnector.js` — `opened-changed` reset arm: KeyMapper rationale

The full version of the (now shortened) close-time-reset comment, for context:

> Arm a DataCommunicator reset when the dropdown closes with focusSelectedItem enabled. On the next open, the feature scrolls to the selected item's position — the same viewport range the server last sent — and the server would no-op the fetch, leaving pending callbacks unresolved. Forcing a reset on the next setViewportRange RPC ensures the server re-sends data for the requested range.
>
> The reset also re-keys items via the data communicator's normal passivate-and-unregister flow — items outside the post-reset active range have their KeyMapper entries removed and get fresh keys when re-fetched. Cached pages in the connector and the data-provider controller would then hold items with stale keys that no longer match the freshly-keyed selectedItem, breaking selection highlighting (and leaving them as targets for the focusSelectedItem RPC's scrollToIndex when it sees a real item at the resolved index but with a stale key). Wipe all client-side cache state on close so the next open populates from the post-reset server data with current keys.

The trimmed version keeps the load-bearing parts (server no-ops on unchanged range; KeyMapper re-keying invalidates cache) and drops the narrative buildup. The reasoning chain is preserved here.

### `ComboBox.java` — `setFocusSelectedItem` microtask polling

Why the polling approach instead of `runBeforeClientResponse` ordering:

> The connector's `setFocusSelectedItem` becomes available only once `initLazy` has run. Both this call (the JS in `setFocusSelectedItem`) and `initLazy` are queued via `executeJs`, but on the first attach there's no guarantee `initLazy` runs first. Polling via microtask until `\$connector` is ready is the simplest way to make ordering immaterial.

The trimmed code-comment now says only the load-bearing part: poll because attach order isn't guaranteed.

DiegoCardoso:
Closing in favor of #9239

PR #9195, opened 2026-04-24
fix: handle stream ends with no finish reasons

ugur-vaadin:
## Description

In some cases, the response terminates without a `finish_reason`. One case it happens is a request going over the context limit, and the result is treated like an empty response rather than an error. This is reproducible using LM Studio with a local model and setting the context length less than 4k for a chart generation query.

This PR reports any response without a `finish_reason` as an exception. This allows the developer to see the server logs for the details, and the user to see a generic error message that they can report.

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/test/java/com/vaadin/flow/component/ai/provider/SpringAILLMProviderTest.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/test/java/com/vaadin/flow/component/ai/provider/SpringAILLMProviderTest.java
@@ -28,12 +28,16 @@ import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.NullAndEmptySource;
+import org.junit.jupiter.params.provider.ValueSource;
 import org.mockito.ArgumentCaptor;
 import org.mockito.Mockito;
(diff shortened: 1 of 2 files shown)


vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/test/java/com/vaadin/flow/component/ai/provider/SpringAILLMProviderTest.java
```diff
@@ -803,15 +807,194 @@ void setHistory_withEmptyAttachmentMap_behavesLikeTextOnly() {
...
+        var request = createSimpleRequest("Hello");
+        Mockito.when(mockChatModel.stream(Mockito.any(Prompt.class)))
+                .thenReturn(Flux.empty());
+
+        Assertions.assertThrows(IllegalStateException.class,
+                () -> provider.stream(request).collectList().block());
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = { "STOP", "TOOL_CALLS", "CONTENT_FILTER" })
```
tomivirkki: nit: The code doesn't inspect the actual finish_reason value so "STOP", "TOOL_CALLS", and "CONTENT_FILTER" all hit the exact same code path. Three runs, same assertion, same branches.

tomivirkki (approved):

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/test/java/com/vaadin/flow/component/ai/provider/SpringAILLMProviderTest.java
ugur-vaadin: Removed the unnecessary parametrization.

PR #9212, opened 2026-04-28
fix: detect silent abort after tool calls

ugur-vaadin:
## Description

Follow-up to https://github.com/vaadin/flow-components/pull/9195.

That previous fix catches a streaming response that ends without any `finish_reason`. But it misses a related case: a tool-using exchange where the first roundtrip succeeds (emitting tool calls and a `finish_reason`) but the follow-up roundtrip silently aborts. The check is sticky: once a real terminal chunk has been observed, trailing metadata-only chunks (e.g. OpenAI's streamUsage trailer with empty choices and usage stats) cannot flip the gate back.

This PR widens the check to track whether any chunk in the stream represented a terminal model state — a chunk that has a `finish_reason` and no pending tool calls. Pending tool calls mean Spring AI is going to make another roundtrip, so a chunk with them is intermediate even if it has a `finish_reason`.   

The signal is a logged warning rather than a thrown error. The heuristic cannot enumerate every provider quirk across vendors and configurations, so a false positive must not surface as a user-visible failure on a valid response. Developers retain observability via logs; if responses appear truncated, the warning is the signal.

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/provider/SpringAILLMProvider.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/provider/SpringAILLMProvider.java
@@ -201,36 +201,44 @@ public class SpringAILLMProvider implements LLMProvider {
 
     /**
      * Passes the stream through unchanged, raising an
-     * {@link IllegalStateException} on completion if no chunk carried a
-     * finish_reason.
+     * {@link IllegalStateException} on completion if the most recent chunk did
+     * not represent a terminal model state.
      * <p>
-     * A compliant OpenAI-style streaming response terminates with a chunk whose
-     * finish_reason is set; any termination reason that legitimately yields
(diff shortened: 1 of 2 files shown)


vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/provider/SpringAILLMProvider.java
tomivirkki, line 1: This change causes a regression when OpenAI's `streamUsage(true)` is used: it emits a final empty-choices usage event after the finish_reason chunk. The old "any chunk seen" gate passed; the new "last chunk wins" gate fails it.

Image

Throwing an exception (and basically failing a possibly valid response) also feels excess. If we're not 100% certain it's going to work in all possible setups, warning would be a safer choice.

ugur-vaadin: Switched back to "any chunk seen" approach and switched to warning. Added regression tests.

tomivirkki (approved):

PR #9215, opened 2026-04-29
fix: fire response complete on empty responses

ugur-vaadin:
## Description

A turn that produces no visible text — typically a tool-only turn where the model emits tool calls and stops without commentary — is still a successful exchange. But the response-complete lifecycle hooks were skipped: both the application-facing `ResponseCompleteListener.onResponseComplete` and the controller-facing `AIController.onResponseComplete` were gated behind a non-empty-text check. Controllers that flush their pending UI state in `onResponseComplete` (e.g. `ChartAIController` applying its pending query) never render on these turns; application listeners miss the completion entirely.

This PR moves both hook invocations out of the non-empty-text gate so every successful exchange fires them, with the firing cadence unchanged (still once per user prompt, not per tool call or round-trip). The `conversationHistory` add stays gated to avoid polluting history with empty assistant entries.

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/test/java/com/vaadin/flow/component/ai/orchestrator/AIOrchestratorTest.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/test/java/com/vaadin/flow/component/ai/orchestrator/AIOrchestratorTest.java
@@ -1575,29 +1575,6 @@ class AIOrchestratorTest {
     }
 
     @Test
-    void responseCompleteListener_afterEmptyResponse_firesWithEmptyText() {
-        var mockMessage = createMockMessage();
-        Mockito.when(mockMessageList.addMessage(Mockito.anyString(),
-                Mockito.anyString(), Mockito.anyList()))
-                .thenReturn(mockMessage);
-        Mockito.when(
-                mockProvider.stream(Mockito.any(LLMProvider.LLMRequest.class)))
-                .thenReturn(Flux.empty());
-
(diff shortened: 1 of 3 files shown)


vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/test/java/com/vaadin/flow/component/ai/orchestrator/AIOrchestratorTest.java
```diff
@@ -1481,7 +1540,7 @@ void responseCompleteListener_afterStreamError_doesNotFire() {
     }
 
     @Test
-    void responseCompleteListener_afterEmptyResponse_doesNotFire() {
+    void responseCompleteListener_afterEmptyResponse_firesWithEmptyText() {
```
tomivirkki: Isn't this now a duplicate of "prompt_emptyResponse_firesUserListenerWithEmptyText"?

tomivirkki (approved):

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/test/java/com/vaadin/flow/component/ai/orchestrator/AIOrchestratorTest.java
ugur-vaadin: Removed the test "responseCompleteListener_afterEmptyResponse_firesWithEmptyText".

PR #9237, opened 2026-05-06
chore: fix compilation after flow DnD updates

ZheSun88:
fix flow-components test after vaadin/flow#23187


mcollovati:
It looks like https://github.com/vaadin/flow/pull/23187 can potentially be a breaking change for add-ons.
We got a complaint time ago when we added a parameter to an event constructor, without preserving the old behaviour.
Probably nobody is firing DnD events programmatically, but I wonder if we should re-introduce the old constructor or mark the change a breaking on in the release notes.

/cc @mstahv

vursen (approved):

PR #9238, opened 2026-05-07
refactor: request data based on visible range rather than by page

DiegoCardoso:
The combo-box connector tracked pending page-load callbacks in a private `pageCallbacks` map, separate from the data-provider controller's own `rootCache.pendingRequests`. The two diverged in lifetime: `pageCallbacks` accumulated entries until cleared explicitly, while the controller's map shrank as responses arrived.

A `scrollToIndex` call (or any programmatic jump) landing on a page non-contiguous with `pageCallbacks` was classified by the connector's "active range" check as a stale gap and cleared instead of issuing the fetch — lazy callbacks never resolved and `loading` stayed `true`.

The connector now reads `pendingRequests` directly from the controller (single source of truth), and the request path is rewritten to mirror `gridConnector`:

- `requestPage` derives the fetch range from the virtualizer's viewport plus a same-sized buffer (~2 pages of prefetch), with `lastRequestedRange` deduplicating back-to-back requests.
- Filter changes go through a 500 ms debouncer on `comboBox._filterDebouncer`. The filter guard in `$connector.set` / `$connector.confirm` drops stale-filter responses immediately so old data never briefly paints.
- `$connector.clear` stamps placeholders into `filteredItems` so the virtualizer re-fetches on scroll-back instead of rendering stale items (necessary for `ComponentRenderer` items, whose `nodeId` references are invalidated server-side when items leave the active range).

> [!WARNING]
> The old overflow-eviction heuristic (`active range * pageSize > 500`) no longer fires — the shared `pendingRequests` only holds in-flight pages. Loaded pages stay cached for the lifetime of the active range; the server still drops items outside that range, which the connector propagates to `filteredItems` as placeholders so the virtualizer re-fetches on scroll-back. Memory grows with how far the user has scrolled (~50 items per page), matching Grid's behavior.

Tests:
- `combo-box-connector.test.ts`: new `pending requests` describe block (5 tests) covering the shared map, the post-debounce callback, and the stale-filter response guard. `shared.ts` extended with the matching connector-side typings.
- IT suites (`LazyLoadingIT`, `FilteringIT`, `ItemCountUnknownComboBoxIT`, `ComboBoxClientSideDataRangeIT`) updated to match the new single-request-per-scroll fetch sequence and tighter active range.
- `AbstractItemCountComboBoxIT.verifyFetchForUndefinedItemCountCallback`: wraps the log-element lookup in a `try/catch` so a missing log produces a clear failure message instead of `NoSuchElementException`.

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
@@ -413,7 +413,7 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
     @Test
     // https://github.com/vaadin/vaadin-combo-box-flow/issues/227
     // https://github.com/vaadin/vaadin-combo-box-flow/issues/232
-    public void setComponentRenderer_scrollDown_itemsRendered() {
+    public void setComponentRenderer_scrollDown_scrollUp_itemsRendered() {
         clickButton("component-renderer");
         beanBox.openPopup();
 
@@ -421,9 +421,8 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
         waitUntilTextInContent(beanBox, "<h4>Person 300</h4>");
 
         scrollToItem(beanBox, 0);
-
-        // Items rendered at the top with the new component renderer.
         waitUntilTextInContent(beanBox, "<h4>Person 0</h4>");
+
         assertComponentRendered(beanBox, "<h4>Person 0</h4>");
         assertComponentRendered(beanBox, "<h4>Person 4</h4>");
         assertComponentRendered(beanBox, "<h4>Person 7</h4>");
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
@@ -154,7 +154,7 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
         stringBox.openPopup();
         scrollToItem(stringBox, 60);
         assertLoadedItemsCount(
-                "No new items should be loaded for disabled ComboBox.", 100,
+                "No new items should be loaded for disabled ComboBox.", 50,
                 stringBox);
     }
 
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
@@ -417,15 +417,16 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
         clickButton("component-renderer");
         beanBox.openPopup();
 
+        scrollToItem(beanBox, 300);
+        waitUntilTextInContent(beanBox, "<h4>Person 300</h4>");
+
+        scrollToItem(beanBox, 0);
+
         // Items rendered at the top with the new component renderer.
         waitUntilTextInContent(beanBox, "<h4>Person 0</h4>");
         assertComponentRendered(beanBox, "<h4>Person 0</h4>");
-
-        // Items rendered at a far-scrolled page with the new component
-        // renderer (regression guard for vaadin-combo-box-flow#227 / #232).
-        scrollToItem(beanBox, 300);
-        waitUntilTextInContent(beanBox, "<h4>Person 300</h4>");
-        assertComponentRendered(beanBox, "<h4>Person 300</h4>");
+        assertComponentRendered(beanBox, "<h4>Person 4</h4>");
+        assertComponentRendered(beanBox, "<h4>Person 7</h4>");
     }
 
     @Test
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
@@ -154,7 +154,7 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
         stringBox.openPopup();
         scrollToItem(stringBox, 60);
         assertLoadedItemsCount(
-                "No new items should be loaded for disabled ComboBox.", 50,
+                "No new items should be loaded for disabled ComboBox.", 100,
                 stringBox);
     }
 
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
@@ -413,19 +413,19 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
     @Test
     // https://github.com/vaadin/vaadin-combo-box-flow/issues/227
     // https://github.com/vaadin/vaadin-combo-box-flow/issues/232
-    public void setComponentRenderer_scrollDown_itemsRendered() {
+    public void setComponentRenderer_scrollDown_scrollUp_itemsRendered() {
         clickButton("component-renderer");
         beanBox.openPopup();
 
-        // Items rendered at the top with the new component renderer.
-        waitUntilTextInContent(beanBox, "<h4>Person 0</h4>");
-        assertComponentRendered(beanBox, "<h4>Person 0</h4>");
-
-        // Items rendered at a far-scrolled page with the new component
-        // renderer (regression guard for vaadin-combo-box-flow#227 / #232).
         scrollToItem(beanBox, 300);
         waitUntilTextInContent(beanBox, "<h4>Person 300</h4>");
-        assertComponentRendered(beanBox, "<h4>Person 300</h4>");
+
+        scrollToItem(beanBox, 0);
+        waitUntilTextInContent(beanBox, "<h4>Person 0</h4>");
+
+        assertComponentRendered(beanBox, "<h4>Person 0</h4>");
+        assertComponentRendered(beanBox, "<h4>Person 4</h4>");
+        assertComponentRendered(beanBox, "<h4>Person 7</h4>");
     }
 
     @Test
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
@@ -497,7 +497,7 @@ public class LazyLoadingIT extends AbstractComboBoxIT {
                 disabledLazyLoadingBox);
 
         scrollToItem(disabledLazyLoadingBox, 100);
-        assertLoadedItemsCount("Scrolling down should load further pages", 100,
+        assertLoadedItemsCount("Scrolling down should load further pages", 50,
                 disabledLazyLoadingBox);
         assertRendered(disabledLazyLoadingBox, "99");
     }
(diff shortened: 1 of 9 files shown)


vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/test/shared.ts
```diff
@@ -17,6 +17,11 @@ export type FlowComboBox = ComboBox & {
   $server: ComboBoxServer;
   _filterTimeout: number;
   _filterDebouncer: unknown;
+  __dataProviderController: {
```
vursen:
```suggestion
  __dataProviderController: DataProviderController
```

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -44,11 +43,12 @@ window.Vaadin.Flow.comboBoxConnector.initLazy = (comboBox) => {
...
 
-  const clearPageCallbacks = (pages = Object.keys(pageCallbacks)) => {
+  const clearPageCallbacks = (pages = Object.keys(getPendingRequests())) => {
     // Flush and empty the existing requests
+    const pendingRequests = getPendingRequests();
     pages.forEach((page) => {
-      pageCallbacks[page]([], comboBox.size);
-      delete pageCallbacks[page];
+      pendingRequests[page]([], comboBox.size);
+      delete pendingRequests[page];
```
vursen: Should be no need to delete it since it's already done by DataProviderController inside the callback.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -234,12 +234,13 @@ window.Vaadin.Flow.comboBoxConnector.initLazy = (comboBox) => {
...
+    const pendingRequests = getPendingRequests();
+    let activePages = Object.getOwnPropertyNames(pendingRequests);
     for (let i = 0; i < activePages.length; i++) {
       let page = activePages[i];
 
       if (cache[page]) {
-        commitPage(page, pageCallbacks[page]);
+        commitPage(page, pendingRequests[page]);
       }
     }
```
vursen:
```suggestion
    Object.entries(getPendingRequests()).forEach(([page, callback]) => {
       if (cache[page]) {
         commitPage(page, callback);
       }
    });
}
```

vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/dataview/ItemCountUnknownComboBoxIT.java
```diff
@@ -36,24 +36,28 @@ public void undefinedItemCount_defaultPageSizeEvenToDatasetItemCount_scrollingTo
         doScroll(45, getDefaultInitialItemCount(), "Callback Item 45");
 
         // trigger next page fetch and item count buffer increase
-        doScroll(150, 400, "Callback Item 150", RangeLog.of(2, 100, 150));
+        doScroll(150, 400, "Callback Item 150", RangeLog.of(1, 100, 150),
```
vursen: Could you explain why these tests had to be updated? What's causing that?

vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/test/shared.ts
DiegoCardoso: Done.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
DiegoCardoso: Done.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
DiegoCardoso: Done.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/dataview/ItemCountUnknownComboBoxIT.java
DiegoCardoso: I reduced some of the changes in the tests. Locally it works. There's an issue with `mvn verify` locally that makes the styles not be correctly loaded in every run, which seems to cause some confusion in the logs.

Now, some changes had to be made in some tests, because now pages are not discarded as part of this refactor. For instance, in the `undefinedItemCount_defaultPageSizeEvenToDatasetItemCount_scrollingToEnd` test, scrolling back does not check for the logs anymore, as the page is not re-requested.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/LazyLoadingIT.java
```diff
@@ -413,19 +413,19 @@ public void autoOpenDisabled_setValue_valueChanged() {
...
-
         assertComponentRendered(beanBox, "<h4>Person 0</h4>");
-        assertComponentRendered(beanBox, "<h4>Person 4</h4>");
-        assertComponentRendered(beanBox, "<h4>Person 7</h4>");
+
+        // Items rendered at a far-scrolled page with the new component
+        // renderer (regression guard for vaadin-combo-box-flow#227 / #232).
+        scrollToItem(beanBox, 300);
+        waitUntilTextInContent(beanBox, "<h4>Person 300</h4>");
+        assertComponentRendered(beanBox, "<h4>Person 300</h4>");
```
DiegoCardoso: This actually hides a bug. Investigating...

DiegoCardoso: When a renderer is used, the item can hold a different label, even though the `item` property is properly set to the actual item.

https://github.com/user-attachments/assets/805f6db3-7b56-413d-acab-a33bcf8957e8

The rendered label says "Person 120" while the `item` is set to "Person 0"

Image

vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/dataview/AbstractItemCountComboBoxIT.java
```diff
@@ -104,10 +105,16 @@ protected void verifyFetchForUndefinedItemCountCallback(
...
+            try {
+                WebElement log = findElement(By.id("log-" + index));
+                Assert.assertEquals("Invalid range for index " + index,
+                        index + ":" + rangeLog.getRange().toString(),
+                        log.getText());
+            } catch (NoSuchElementException e) {
+                Assert.fail("Log element not found for index " + index
+                        + ", expected range: "
+                        + rangeLog.getRange().toString());
+            }
```
DiegoCardoso, line 117: This is purely to give a better feedback in case the `log-{N}` element is not found.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/test/combo-box-connector.test.ts
```diff
@@ -17,6 +17,58 @@ describe('combo-box connector', () => {
...
+    it('should be populated when the controller loads a page', () => {
+      comboBox.__dataProviderController.loadFirstPage();
+
+      expect(comboBox.__dataProviderController.rootCache.pendingRequests[0]).to.be.a('function');
+    });
+
+    it('should be cleared by $connector.confirm when items are cached', () => {
+      comboBox.__dataProviderController.loadFirstPage();
+
+      comboBox.$connector.set(0, [{ key: '1', label: 'one' }], '');
```
vursen, line 36: The `ComboBoxConnector` type is currently missing the `set` and `confirm` methods

vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/test/combo-box-connector.test.ts
```diff
@@ -17,6 +17,58 @@ describe('combo-box connector', () => {
     expect(comboBox.$connector).to.equal(connector);
   });
 
+  describe('pending requests', () => {
```
vursen, line 20:
```suggestion
  describe('pending requests', () => {
    let dataProviderController;
    
    beforeEach(() => {
      dataProviderController = comboBox.__dataProviderController;
    });
```

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -44,119 +43,102 @@ window.Vaadin.Flow.comboBoxConnector.initLazy = (comboBox) => {
...
-        params.filter = '';
       }
+      // First fetch: ignore the typed filter so we get the full dataset
+      params = { ...params, filter: '' };
     }
 
-    const filterChanged = params.filter !== lastFilter;
-    if (filterChanged) {
-      cache = {};
+    if (params.filter !== lastFilter) {
```
vursen, line 60: Can we reuse `getLastFilterSentToServer` here?

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
```diff
@@ -44,119 +43,102 @@ window.Vaadin.Flow.comboBoxConnector.initLazy = (comboBox) => {
...
+    }
+    return [indices[0], indices[indices.length - 1]];
+  };
+
+  comboBox.$connector.requestPage = function (page, filter) {
+    let viewportRange = comboBox.$connector.getViewportRange();
+    const buffer = viewportRange[1] - viewportRange[0];
+    const sizeLimit = Number.isFinite(comboBox.size) ? comboBox.size : Number.POSITIVE_INFINITY;
+    viewportRange[0] = Math.max(viewportRange[0] - buffer, 0);
+    viewportRange[1] = Math.min(viewportRange[1] + buffer, sizeLimit);
```
vursen: A small bug from Grid's logic got copied along here:

```suggestion
    viewportRange[1] = Math.min(viewportRange[1] + buffer, sizeLimit - 1);
```

vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/test/combo-box-connector.test.ts
DiegoCardoso: Applied in edff246e25.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
DiegoCardoso: Fixed in 5cbef70def.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/test/combo-box-connector.test.ts
DiegoCardoso: Fixed in edff246e25.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/resources/META-INF/resources/frontend/comboBoxConnector.js
DiegoCardoso: Tried the change locally — the two trackers are intentionally separate, so reusing `getLastFilterSentToServer()` here regresses the `$connector.set` / `$connector.confirm` stale-response guards (lines 151, 210).

`lastFilter` updates synchronously when `dataProvider` observes a new filter; `lastFilterSentToServer` only updates when the debounced request actually fires. If we drop `lastFilter`, a response for filter "a" arriving after the user has typed "b" (but before "b" has been debounce-sent) passes the guard with `lastFilterSentToServer === "a"`, and the stale items get painted before the new filter is sent — the very thing bullet 3 of the PR description was written to prevent.

Added a test covering this in ad176a9880, so the contract is now explicit. Happy to rename `lastFilter` → `lastSeenFilter` (or similar) if the naming is what's confusing.

vursen (approved):

PR #9239, opened 2026-05-07
feat: add setFocusSelectedItem to ComboBox

DiegoCardoso:
Restores the pre-https://github.com/vaadin/web-components/pull/6055 behavior of focusing the currently selected item when the combo-box dropdown opens, behind an opt-in flag.

```java
```

For lazy data providers, an `ItemIndexProvider` must be configured so the selected item's index can be resolved against the current sorting; otherwise opening the dropdown throws `UnsupportedOperationException`. In-memory data works out of the box.

Part of vaadin/web-components#6061

---

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/ComboBoxFocusSelectedItemIT.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/ComboBoxFocusSelectedItemIT.java
@@ -77,17 +77,6 @@ public class ComboBoxFocusSelectedItemIT extends AbstractComboBoxIT {
     }
 
     @Test
-    public void lazyWithProvider_filterThenClear_open_scrollsToSelected() {
-        ComboBoxElement combo = $(ComboBoxElement.class)
-                .id("lazy-with-provider");
-        combo.setFilter("5");
-        assertLoadingStateResolved(combo);
-        combo.setFilter("");
-        assertLoadingStateResolved(combo);
-        assertOverlayContains(combo, "Item 5000");
-    }
-
-    @Test
     public void lazyToggleOff_runtimeToggleOn_open_scrollsToSelected() {
         clickButton("lazy-toggle-off-toggle-on");
         openAndAssertContains("lazy-toggle-off", "Item 5000");
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/ComboBoxFocusSelectedItemIT.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/ComboBoxFocusSelectedItemIT.java
@@ -31,7 +31,6 @@ public class ComboBoxFocusSelectedItemIT extends AbstractComboBoxIT {
     @Before
     public void init() {
         open();
-        waitForElementPresent(By.tagName("vaadin-combo-box"));
     }
 
     @Test
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/ComboBoxFocusSelectedItemIT.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/ComboBoxFocusSelectedItemIT.java
@@ -20,7 +20,6 @@ import java.util.List;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
-import org.openqa.selenium.By;
 
 import com.vaadin.flow.component.combobox.testbench.ComboBoxElement;
 import com.vaadin.flow.testutil.TestPath;
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/ComboBoxFocusSelectedItemIT.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/ComboBoxFocusSelectedItemIT.java
@@ -1,0 +1,136 @@
+/*
+ * Copyright 2000-2026 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component.combobox.test;
+
+import java.util.List;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.openqa.selenium.By;
+
+import com.vaadin.flow.component.combobox.testbench.ComboBoxElement;
+import com.vaadin.flow.testutil.TestPath;
+
+@TestPath("vaadin-combo-box/focus-selected-item")
+public class ComboBoxFocusSelectedItemIT extends AbstractComboBoxIT {
+
+    @Before
+    public void init() {
+        open();
+        waitForElementPresent(By.tagName("vaadin-combo-box"));
+    }
+
+    @Test
+    public void lazyWithProvider_open_scrollsToSelected() {
+        openAndAssertContains("lazy-with-provider", "Item 5000");
+    }
+
+    @Test
+    public void lazyWithProvider_setValue_open_scrollsToNewSelected() {
+        clickButton("lazy-with-provider-set-9000");
+        openAndAssertContains("lazy-with-provider", "Item 9000");
+    }
+
+    @Test
+    public void lazyWithProvider_reopen_stillScrollsToSelected() {
+        ComboBoxElement combo = openAndAssertContains("lazy-with-provider",
+                "Item 5000");
+        combo.closePopup();
+        openAndAssertContains("lazy-with-provider", "Item 5000");
+    }
+
+    @Test
+    public void lazyWithProvider_clearValue_open_doesNotScroll() {
+        clickButton("lazy-with-provider-clear");
+        openAndAssertContains("lazy-with-provider", "Item 0");
+    }
+
+    @Test
+    public void lazyWithProvider_toggleOff_open_doesNotScroll() {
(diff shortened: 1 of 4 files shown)


vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -368,6 +373,81 @@ public T getEmptyValue() {
...
+            return;
+        }
+        Integer index;
+        try {
+            if (dataProvider.isInMemory()) {
+                index = getListDataView().getItemIndex(getValue()).orElse(null);
+            } else {
+                index = getLazyDataView().getItemIndex(getValue()).orElse(null);
+            }
+        } catch (UnsupportedOperationException e) {
```
vursen: Why does the exception need to be suppressed?

DiegoCardoso: At first it seemed to be reasonable to state in the docs that the silent is no-op if the `ItemIndexProvider` is not configured, but we can remove it and change the javadoc in the `setFocusSelectedItem` method to mention that.

vursen: Yeah, I think throwing the exception is useful since it will make it clear when you have forgotten to provide an item index provider.

DiegoCardoso: Fixed in 97ce73163e — exception now propagates, javadoc updated, IT page/test trimmed, and a unit test in `ComboBoxTest` locks in the throw contract.

vursen:
I'm a bit unsure whether the selected item is expected to scroll into view when the filter is cleared after a new one was entered. Shouldn't the selected item only be scrolled into view when you open the dropdown?

https://github.com/user-attachments/assets/2a95a59e-8dfe-4ca9-83a9-4eb4fa3434fc

DiegoCardoso:
> Shouldn't the selected item only be scrolled into view when you open the dropdown?

I was not sure either. The reason I picked this behavior was that if you cleared the selection and pressed <kbd>Tab</kbd>, the value would also get cleared. Checking now the behavior in ComboBox, from the docs page, it seems to already behave like this, so it might not be a good idea to scroll back to the selected item when the filter is cleared. It also removes some code I wasn't really happy about.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -164,6 +166,9 @@ public ComboBox(int pageSize) {
                 refreshValue();
             }
         });
+
+        getElement().addEventListener("vaadin-combo-box-dropdown-opened",
+                event -> focusSelectedItemOnOpen());
```
vursen: How about a small restructuring here:

```suggestion
        getElement().addEventListener("vaadin-combo-box-dropdown-opened",
                event -> {
                if (focusSelectedItem) {
                    scrollToSelectedItem();
                });
```

The current `focusSelectedItemOnOpen()` name feels a bit misleading, because it doesn't actually focus the item – it only scrolls to it. The focusing itself is handled by the web component, if I understand correctly

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -368,6 +373,75 @@ public T getEmptyValue() {
...
+    private void focusSelectedItemOnOpen() {
+        if (!focusSelectedItem || getValue() == null) {
+            return;
+        }
+        String filter = getFilter();
+        if (filter != null && !filter.isEmpty()) {
+            // The user is filtering — don't override their navigation by
+            // scrolling to the previously selected item.
+            return;
+        }
```
vursen: Is this still necessary?

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
DiegoCardoso: Done.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
DiegoCardoso: Yes. This is to prevent the ComboBox from scrolling while the user is filtering. Without this change, the ComboBox might scroll to the wrong item, as seen in this example:

https://github.com/user-attachments/assets/3cc04ed7-5553-4a89-ba62-e552a823815a

This is also aligned with the feedback shared by Sascha in the previous attempt to implement this feature (https://github.com/vaadin/flow-components/pull/9194#discussion_r3147361335).

Additionally, it seems that the query variable in the setItemIndexProvider callback does not include the client-side filter (for some reason), so it is not reliable to scroll to the selected item while the user is filtering the ComboBox.

DiegoCardoso (commented):
Review comments addressed.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
vursen: > Without this change, the ComboBox might scroll to the wrong item, as seen in this example:

Hmm, I wonder if this actually happens because the `vaadin-combo-box-dropdown-opened` event currently fires on every filter request:

https://github.com/user-attachments/assets/2947416a-d41c-4e55-8db2-a695d1970e5d

Have you considered listening to the `opened-changed` event instead? It shouldn't have this issue... 

...although this one, on the other hand, might fire too early, before the data is received, right? If so, would it make sense to improve the web component's scrollToIndex method so that it waits until all data is loaded, as it's done in Grid?

vursen: That said, I gave it a try and it actually seems to work just fine, without needing any additional changes in the web component:

```diff
diff --git a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
index b3df39d9d8..c5120d5a68 100644
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
@@ -167,12 +167,12 @@ public class ComboBox<T> extends ComboBoxBase<ComboBox<T>, T, T>
             }
         });

-        getElement().addEventListener("vaadin-combo-box-dropdown-opened",
-                event -> {
-                    if (focusSelectedItem) {
-                        scrollToSelectedItem();
-                    }
-                });
+        getElement().addPropertyChangeListener("opened", event -> {
+            var isOpened = (boolean) event.getValue();
+            if (isOpened && focusSelectedItem) {
+                scrollToSelectedItem();
+            }
+        });
     }

     /**
@@ -424,12 +424,6 @@ public class ComboBox<T> extends ComboBoxBase<ComboBox<T>, T, T>
         if (getValue() == null) {
             return;
         }
-        String filter = getFilter();
-        if (filter != null && !filter.isEmpty()) {
-            // The user is filtering — don't override their navigation by
-            // scrolling to the previously selected item.
-            return;
-        }
         DataProvider<T, ?> dataProvider = getDataProvider();
         if (dataProvider == null) {
             return;
```

DiegoCardoso: Right. Makes sense to use the "opened" event instead. Changed it as proposed. Thanks.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/ComboBoxFocusSelectedItemIT.java
```diff
@@ -0,0 +1,125 @@
...
+import com.vaadin.flow.component.combobox.testbench.ComboBoxElement;
+import com.vaadin.flow.testutil.TestPath;
+
+@TestPath("vaadin-combo-box/focus-selected-item")
+public class ComboBoxFocusSelectedItemIT extends AbstractComboBoxIT {
+
+    @Before
+    public void init() {
+        open();
+        waitForElementPresent(By.tagName("vaadin-combo-box"));
```
vursen: This shouldn't be needed:

```suggestion
```

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -368,6 +377,66 @@ public T getEmptyValue() {
...
+     * {@link com.vaadin.flow.component.combobox.dataview.ComboBoxLazyDataView#setItemIndexProvider(com.vaadin.flow.data.provider.ItemIndexProvider)
+     * getLazyDataView().setItemIndexProvider(...)} so that the selected item's
+     * index can be resolved against the current sorting. Opening the dropdown
+     * throws {@link UnsupportedOperationException} otherwise.
+     * <p>
+     * The setting is a silent no-op when:
+     * <ul>
+     * <li>no value is currently selected, or</li>
+     * <li>the resolved index cannot be determined.</li>
+     * </ul>
```
vursen: Not sure this is worth mentioning. That's just how it's supposed to work, IMO. Maybe remove?

```suggestion
```

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/ComboBoxTest.java
```diff
@@ -238,6 +238,34 @@ void setOverlayWidth() {
...
+        ComboBox<String> comboBox = new ComboBox<>();
+        ui.add(comboBox);
+        comboBox.setItems(query -> Stream.of("a", "b", "c")
+                .skip(query.getOffset()).limit(query.getLimit()), query -> 3);
+        comboBox.setFocusSelectedItem(true);
+        comboBox.setValue("a");
+
+        Assertions.assertThrows(UnsupportedOperationException.class,
+                () -> comboBox.getElement().setProperty("opened", true));
+    }
```
vursen: Can also be removed IMO, since throwing `UnsupportedOperationException` is Flow's responsibility

```suggestion
```

vursen (approved):

PR #9241, opened 2026-05-07
fix: Make SpreadsheetOverlay native popover

### LINKED ISSUE #9240: Spreadsheet overlays do not work when Spreadsheet is in Dialog.
### Description

Vaadin 25 Dialog, Popover etc. are now native popovers. Spreadsheet has some integrated overlays e.g. popup for filter table, comment, context menu etc. Those have not been updated to use popover="manual" yet. Hence when placing Spreadsheet into Dialog will result those overlays to open behind the Dialog and not visible. 

### Expected outcome

Spreadsheet overlays should work.

Trivial. Add Spreadsheet to Dialog.

### Steps to reproduce

Trivial. Add Spreadsheet to Dialog.

### Environment

Vaadin version(s): Vaadin 25+

### Browsers

Issue is not browser related

TatuLund:
Fixes #9240 

Note1: Spreadsheet tabs initial alignment is wrong if the first tab is not the selected one when Spreadsheet is placed on Dialog. This is not regression from this PR, but a different bug that needs to be addressed separately.

Note2: When Spreadsheet is in Dialog, inserting comment does not work. This is not regression from this PR, as inserting comment after this PR still works, when Spreadsheet is not in Dialog. This is different bug that needs to be address separately.

- Actually these bugs happen also with Vaadin 24.10, so they are not regressions from Dialog's native popover refactor.

### DIFF
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/resources/META-INF/resources/frontend/vaadin-spreadsheet/vaadin-spreadsheet-styles.js
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/resources/META-INF/resources/frontend/vaadin-spreadsheet/vaadin-spreadsheet-styles.js
@@ -1520,6 +1520,7 @@ export const spreadsheetOverlayStyles = css`
     white-space: pre-wrap;
   }
   #spreadsheet-overlays .v-contextmenu {
+    border: unset;
     padding: 4px 4px;
     border-radius: 4px;
     background-color: white;
@@ -1574,6 +1575,7 @@ export const spreadsheetOverlayStyles = css`
   }
 
   #spreadsheet-overlays .v-spreadsheet-comment-overlay {
+    border: unset;
     padding: 4px 4px;
     border-radius: 4px;
     background-color: white;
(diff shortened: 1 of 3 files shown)


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/resources/META-INF/resources/frontend/vaadin-spreadsheet/vaadin-spreadsheet-styles.js
```diff
@@ -1520,6 +1520,7 @@ export const spreadsheetOverlayStyles = css`
     white-space: pre-wrap;
   }
   #spreadsheet-overlays .v-contextmenu {
+    border: unset;
```
web-padawan, line 1523: Apart from the border, there could be other styles. I see that `padding` and `background-color` are already overridden - please see https://github.com/vaadin/web-components/issues/9703 for default user agent CSS.

TatuLund: By observing SpreadsheetDialogPage demo view no other CSS seemed necessary, I think the styles already set in Spreadsheet cover that.

manolo:
@TatuLund since this is a fork in your account you need to configure TB in your fork

PR #9248, opened 2026-05-07
feat: collect all test results in a single Collect tests job

manolo:
Currently unit test results are published immediately when the `unit-tests` job finishes, appearing as a separate check before IT and WTR results are available. This change moves all test reporting to the `Collect tests` job so that Unit Tests, ITs and WTR Tests all appear together at the end of the run.

**Unit tests:** The `Publish unit test results` step is removed from `unit-tests`. Instead, surefire XML files are uploaded as a `surefire-reports` artifact and published by `Collect tests`.

**WTR tests:** The `wtr.js` script installs `@web/test-runner-junit-reporter` and generates a per-component `wtr-results.xml` using a temporary CI config. The `wtr-tests` job uploads these as a `wtr-reports` artifact, which `Collect tests` downloads and publishes as "WTR Tests".

The IT matrix job is renamed from `it` to `its` and displayed as `ITs (N)` in the UI. The aggregator job is renamed from "Collect IT results" to "Collect tests".

All intermediate artifacts (`surefire-reports`, `wtr-reports`, `failsafe-reports-*`, `error-screenshots-*`) are cleaned up at the end of `Collect tests`.


.github/workflows/validation-sharded.yml
vursen:
```suggestion
    name: Collect results
```

vursen (approved):

PR #9250, opened 2026-05-08
chore: replace blur() helper with Keys.TAB/ENTER in tests

vursen:
The `blur()` helper in `AbstractComponentIT` ran `document.activeElement.blur()` through JavaScript to commit pending value changes in field tests. Triggering blur out-of-band from the typed input is fragile and hides what the test does.

Replace it with `Keys.TAB` or `Keys.ENTER` sent through `sendKeys`. The focus change now happens in the same input stream as the value. As a side benefit, the helper can be dropped from `AbstractComponentIT`, keeping the shared base class smaller.


vursen:
@copilot rebase the PR from main

web-padawan (approved):

PR #9254, opened 2026-05-09
fix: remove explicit ol @NpmPackage, let @vaadin/map own it

manolo:
Removing the explicit `ol` `@NpmPackage` declaration that conflicted with the version declared by `@vaadin/map` itself.

`@vaadin/map@25.2.0-alpha10` already declares `ol: "10.6.0"` as a direct dependency. Declaring `ol: "10.6.1"` here created a version conflict that was silently masked by pnpm `--shamefully-hoist=true`, but broke with the switch to `node-linker=hoisted` in vaadin/flow: Vite ended up bundling two separate OpenLayers instances, causing `synchronize()` to operate on a different `Map` object than the one exposed via `mapElement.configuration`.

Related to vaadin/flow#24307


vursen (approved):

Artur-:
So now we are back with #8113 again. Can we revert this so any build can work?

PR #9256, opened 2026-05-10
fix: restore NpmPackage annotation with ol dep, downgrade ol dep to 10.6.0

Artur-:
Reverts vaadin/flow-components#9254

### DIFF
--- a/vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/Map.java
+++ b/vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/Map.java
@@ -82,6 +82,7 @@ import tools.jackson.databind.node.ObjectNode;
  */
 @Tag("vaadin-map")
 @NpmPackage(value = "@vaadin/map", version = "25.2.0-alpha10")
+@NpmPackage(value = "ol", version = "10.6.1")
 @NpmPackage(value = "proj4", version = "2.17.0")
 @JsModule("@vaadin/map/src/vaadin-map.js")
 @JsModule("./vaadin-map/mapConnector.js")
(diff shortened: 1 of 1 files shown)


vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/Map.java
```diff
@@ -82,6 +82,7 @@
  */
 @Tag("vaadin-map")
 @NpmPackage(value = "@vaadin/map", version = "25.2.0-alpha10")
+@NpmPackage(value = "ol", version = "10.6.0")
```
manolo, line 85: Well sounds like this is not a proper revert but a downgrade from 10.6.1 to 10.6.0 

Why is it better to keep here the dependency if it comes transitively from "@vaadin/map" ?

manolo: I see, it's needed to no https://github.com/vaadin/flow-components/pull/8113

manolo (approved):

vaadin-map-flow-parent/vaadin-map-flow/src/main/java/com/vaadin/flow/component/map/Map.java
vursen: I discussed this with the team and we agreed that we will:

1. Restore the annotation with a temporarily downgraded version
2. Upgrade the ol version in the web component
3. Upgrade the ol version in the flow component once the WC release lands
4. Discuss how to avoid having to maintain the ol dependency in the flow component separately

vursen (approved):

PR #9257, opened 2026-05-11
chore: invalidate WAR cache on component source change

vursen:
WAR cache key only hashed integration-tests sources. Component production sources (`vaadin-*/src/main/**`) were not part of the key, so rebasing onto main commits that touched component code did not invalidate the WAR cache. 

Build cache (jars in `~/.m2/repository/com/vaadin`) does hash component sources and rebuilds correctly on miss, but the WAR cache hit on the unchanged key restored a WAR packaged against the previous jars. ITs then ran against a stale WAR. Only workaround was deleting the cache manually.

The PR extends WAR key globs to match the build cache so both invalidate together

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -324,7 +324,7 @@ jobs:
       - name: Compute WAR cache key
         id: war-key
         env:
-          KEY: ${{ runner.os }}-war-${{ hashFiles('**/pom.xml', '**/src/main/java/**', '**/src/main/resources/**', 'integration-tests/src/**') }}
+          KEY: ${{ runner.os }}-war-${{ hashFiles('**/pom.xml', '**/src/main/java/**', '**/src/main/resources/**', 'integration-tests/src/test/java/**') }}
         run: echo "key=$KEY" >> "$GITHUB_OUTPUT"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/validation.yml
```diff
@@ -324,7 +324,7 @@ jobs:
       - name: Compute WAR cache key
         id: war-key
         env:
-          KEY: ${{ runner.os }}-war-${{ hashFiles('integration-tests/pom.xml', 'integration-tests/src/**') }}
```
manolo, line 327: The fix is correct. One minor redundancy: `integration-tests/src/**` includes `integration-tests/src/main/java/**` and `integration-tests/src/main/resources/**`, which are already matched by the leading  `**/src/main/java/**` and `**/src/main/resources/**` patterns (the ** prefix matches any directory including integration-tests/). The content ends up hashed twice.

You could narrow it to just that:
```
hashFiles('**/pom.xml', '**/src/main/java/**', '**/src/main/resources/**', 'integration-tests/src/test/java/**')
```

vursen: Done.

manolo (approved):

PR #9270, opened 2026-05-12
fix: Make SpreadsheetOverlay native popover (#9241)

### LINKED ISSUE #9240: Spreadsheet overlays do not work when Spreadsheet is in Dialog.

TatuLund:
Fixes #9240 

Note1: Spreadsheet tabs initial alignment is wrong if the first tab is not the selected one when Spreadsheet is placed on Dialog. This is not regression from this PR, but a different bug that needs to be addressed separately.

Note2: When Spreadsheet is in Dialog, inserting comment does not work. This is not regression from this PR, as inserting comment after this PR still works, when Spreadsheet is not in Dialog. This is different bug that needs to be address separately.

- Actually these bugs happen also with Vaadin 24.10, so they are not regressions from Dialog's native popover refactor.

### DIFF
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/SpreadsheetOverlay.java
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/SpreadsheetOverlay.java
@@ -18,34 +18,65 @@ import com.vaadin.client.ui.VOverlay;
  * A VOverlay Implementation that attaches the overlay to the container added by
  * vaadin-spreadsheet webcomponent
  */
-@SuppressWarnings("deprecation")
+@SuppressWarnings({ "deprecation", "java:S1699" })
 public class SpreadsheetOverlay extends VOverlay {
+
+    private static final String POPOVER_ATTRIBUTE = "popover";
 
     /**
      * A VContextMenu Implementation that attaches the overlay to the container
      * added by vaadin-spreadsheet webcomponent
      */
     public static class SpreadsheetContextMenu extends VContextMenu {
+
         public SpreadsheetContextMenu() {
             DOM.setElementProperty(getElement(), "id", "PID_VAADIN_CM");
+            setPopover(getElement());
         }
 
         @Override
         public Element getOverlayContainer() {
             return getOverlayContainerElement();
         }
+
+        @Override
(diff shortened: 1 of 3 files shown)


DiegoCardoso:
There's one issue I found and it relates with the modal dialogs. `Spreadsheet` creates a `<div id="spreadsheet-overlays>` in the `<body>` and place all the overlay inside this element (similar to the overlays in Vaadin before the V25 refactor). So, when `Spreadsheet` is placed inside a modal `Dialog`, the options are not reachable. That explains the comment in `Note2` (I haven't tested in V24, but it could be the same issue there). 

Perhaps, with the usage of native popover, there's no need to have the `<div>` mentioned above inside the `<body>` and move it to the `<vaadin-spreadsheet>` instead.

#### Modal Dialog (doesn't work)
https://github.com/user-attachments/assets/0ce4ab12-81d6-4d9e-b562-02da65b78f9e

#### Modaless Dialog (works)
https://github.com/user-attachments/assets/74511a56-8f49-4d5b-8d22-3c749ee894a5

There are also a lot of errors logged in the client console when the `Dialog` opens:
image

It seems to be unrelated to these changes, though, as I verified it in `main` too.

TatuLund:
> It seems to be unrelated to these changes, though, as I verified it in main too.

@DiegoCardoso Exactly. Thus I would prefer to do separate PR about that, which could be more clearly backported to Vaadin 24 as well. The native popover implementation is Vaadin 25 specific.

DiegoCardoso:
I've created [PR #9303](https://github.com/vaadin/flow-components/pull/9303) based on this (still in draft). The main change is to render the overlays as children of the Spreadsheet instances instead of appending them to <body>, which fixes the issue with modal dialogs. This also resolves the issue mentioned above regarding adding comments to cells when a Spreadsheet is inside a dialog.

yuriy-fix:
Closing in favor of https://github.com/vaadin/flow-components/pull/9303

PR #9273, opened 2026-05-12
feat: Add text selection API for text input fields [3 days]

### LINKED ISSUE #1377: Selection API
A Java API for [setSelectionRange](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setSelectionRange) would be very useful for all the Vaadin text fields, e.g.

```java
TextField textField = new TextField();
new Button("Select all", event -> textField.selectAll());
```

Also access to [execCommand](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand) could open advanced use cases for the Java API, e.g.

```java
TextField textField = new TextField();
new Button("Copy", event -> {
    textField.selectAll();
    textField.copyToClipboard();
    Notification.show("Copied to your clipboard!");
});
```

See [Clipboard.js](https://clipboardjs.com/) for a related JavaScript API.

Artur-:
## Summary                                                                                                                                                                                 

- Add `HasSelection` mixin with `selectAll`, `deselect`, `setSelectionRange`, `setCursorPosition`, and a reactive `Signal<SelectionRange> selectionSignal()` so applications can drive the 
browser text selection from the server.                         
- Implement on `TextField`, `TextArea`, `PasswordField`, and `BigDecimalField`. `EmailField`, `NumberField`, and `IntegerField` are deliberately excluded because their underlying `<input 
type="email">` / `type="number"` elements throw `InvalidStateError` for these APIs per the HTML spec.                                                                                      

Fixes #1377
                                                                                                                                                                      
## Details                                                      
                                                                                                                                                                                           
### Reactive read                                               
`selectionSignal()` returns a `Signal<SelectionRange>` lazily backed by a `ValueSignal`. On first call it installs a client-side listener bundle (`select`, `keyup`, `mouseup`, `input`, `focus`) on the inner `inputElement`; each event dispatches a `vaadin-selection-change` event on the host, which the server picks up via `Element.addEventListener(...).addEventData(...)` and pushes into the signal. Subsequent calls return the same cached instance, surviving detach/re-attach.                                                                                  
                                                                
This replaces the async `getSelectionRange(callback)` shape from the earlier [PR #3194](https://github.com/vaadin/flow-components/pull/3194) attempt: a server-side click handler reads the latest selection synchronously via `signal.peek()` without an extra roundtrip, so timing races between user cursor moves and server reads no longer apply.
                                                                                                                                                                                           
### Imperative methods                                          
`SelectionRange` is a `record(int start, int end, String content)` with `length()`, `isEmpty()`, and `empty()` helpers. Indices follow the `HTMLInputElement.setSelectionRange()` convention (zero-based, end-exclusive).
                                                                                                                                                                                           
`selectAll`, `setSelectionRange`, and `setCursorPosition` focus the field by default so the highlight is rendered in the active color rather than the browser's faded inactive state. Each accepts a `boolean focus` overload to opt out. `deselect` is focus-neutral.
                                                                                                                                                                                           
The generated JS applies the selection before focusing — `inputElement.focus()` scrolls to the current cursor position, so focusing first would scroll to the stale cursor (often end-of-value after a preceding `setValue`) and the subsequent range update would not re-scroll. `selectAll` uses `setSelectionRange(0, length)` rather than `HTMLInputElement.select()`
because the latter always implicitly focuses per the WHATWG spec, which would make the `focus=false` overload a lie. All mutators are wrapped in `setTimeout(...,0)` so they run after pending value/focus reflection on the web component, matching the workaround used in `viritin/flow-viritin`.


yuriy-fix:
We will start with manual testing to see if previous concern of not working consistently between the browsers still persists.

sissbruecker:
Did some manual testing, results below.

### Various issues
- Clicking into an existing selection collapses it, but the selection signal does not update _(Video 1)_
- Clicking into a field and holding the mouse button first resets selection signal to 0:0, then when releasing mouse button updates selection to correct value. Depending on what the app does with the info it would first do an update for an empty selection, while the user is still selecting, and then update afterwards again after receiving the final selection. _(Video 2)_
-  Setting selection range without focusing the input
	- On pretty much all browsers I tested, clicking into the field or tabbing into it clears the selection, making this seemingly useless
	- Only Edge on Win 11 preserves selection when navigating to field with keyboard selection

### iOS / Safari
- Focus does not work, seems to be a technical limitation
	- When selecting with focus, Safari 18 does not show the selection range, Safari 26 does
- Selection signal update works sporadically:
	- Works when just tapping into the field and editing using virtual keyboard, and with the cut / paste / select / select all actions
	- Does not work when using long press to adjust cursor position or when adjusting selection range

### Android / Chrome
- Focus works here, opens virtual keyboard
- Selection signal update has similar issues as with iOS _(Video 3)_

### UX
- Due to syncing selection on every keypress, just quickly typing or moving the cursor with the arrow keys results in the loading indicator showing almost permanently. This is more noticeable when testing with a network delay.

**Video 1 (Selection collapse on click)**

https://github.com/user-attachments/assets/a4d57771-6439-466b-85a5-791ce07952a1

**Video 2 (Duplicate selection update when clicking into field)**

https://github.com/user-attachments/assets/02ef3ee3-bce2-47bb-a016-b2092e1c58c0

**Video 3 (Android selection update issues)**

https://github.com/user-attachments/assets/d07d3721-3867-4a7f-9e76-13f022143215

PR #9274, opened 2026-05-12
feat: add response failed and request start hooks to ai controller

ugur-vaadin:
## Description

This PR:
- Adds the following to `AIController`:
  - `void onRequestStart()`: Fires synchronously on the UI thread before the LLM stream opens. Throwing aborts the turn and routes to `onResponseFailed`. Default does nothing.
  - `void onResponseFailed(Throwable)`: Fires on stream error, timeout, or any pre-stream throw. Mutually exclusive with `onResponseComplete`, every turn fires exactly one of the two. Default does nothing.
- Updates `AIOrchestrator` to uses the new API. Exceptions thrown from either hook are caught and logged; Errors propagate.

Fixes https://github.com/orgs/vaadin/projects/103/views/1?filterQuery=&pane=issue&itemId=186537285

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/orchestrator/AIController.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/orchestrator/AIController.java
@@ -42,15 +42,19 @@ public interface AIController {
     List<LLMProvider.ToolSpec> getTools();
 
     /**
-     * Called synchronously on the UI thread at the start of every LLM turn,
-     * before the orchestrator opens the LLM stream. Implementations can prepare
-     * for the turn — locking UI surfaces, snapshotting state the tool
-     * definitions depend on, and so on.
+     * Called synchronously on the UI thread just before the LLM stream opens.
+     * By the time this method fires, the user message and an empty assistant
+     * placeholder are already in the message list and the user message is in
+     * the conversation history. Implementations can prepare for the turn —
+     * locking UI surfaces, snapshotting state the tool definitions depend on,
+     * and so on.
      * <p>
      * The default does nothing. Throwing from this method aborts the turn: the
-     * LLM stream is not opened, {@link #onResponseFailed(Throwable)} fires with
-     * the thrown exception so per-turn state captured before the throw can
-     * still be released, and the user sees a generic error message.
+     * LLM stream is not opened, the assistant placeholder is updated to a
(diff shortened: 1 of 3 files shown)


vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/orchestrator/AIController.java
```diff
@@ -39,30 +35,55 @@
...
+     * definitions depend on, and so on.
      * <p>
-     * This method is invoked after all tool executions for a given user request
-     * have finished and the LLM has generated its final response. Controllers
-     * can use this callback to perform deferred operations, such as rendering
-     * UI updates or committing state changes.
+     * The default does nothing. Throwing from this method aborts the turn: the
+     * LLM stream is not opened, {@link #onResponseFailed(Throwable)} fires with
+     * the thrown exception so per-turn state captured before the throw can
+     * still be released, and the user sees a generic error message.
```
tomivirkki: "the user sees a generic error message", I think this is incorrect. By the time `onRequestStart` is ran, no messages have been added to the Message List and an exception doesn't produce one either.

ugur-vaadin: Reordered `processUserInput` and wrapped the "pre-stream" code in a try/catch so that every failure (including the ones in `onRequestStart`) is displayed as a generic error in chat. Also updated the JavaDoc for `onRequestStart` to match the new behaviour and added a couple of unit tests.

tomivirkki: `onRequestStart` throwing leaves the user message in conversationHistory without it ever reaching the LLM, so `getHistory()` diverges from the provider's chat memory. The same applies to `onAttachmentSubmit`: it has been told to persist data for a submission that never happened.

Should we move both the `conversationHistory.add` and the `onAttachmentSubmit` call to right before `streamResponseToMessage`? A pre-stream throw then leaves no orphan state behind.

ugur-vaadin: Moved both in the try-catch block, extracted building request to a method, added a few tests.

tomivirkki (approved):

PR #9275, opened 2026-05-12
perf: cache expensive test setup objects per class

manolo:
`MockUIExtension` and `EnableFeatureFlagExtension` were recreating Mockito mocks before every `@Test` method across all test classes that use them. Analysis of CI surefire reports showed SignalTest classes taking 5-6s each in CI (e.g. `BadgeSignalTest` 6s, `SliderSignalTest` 5s) despite having simple, fast test logic.

The expensive operations per test method:
- `Mockito.mock(VaadinService.class)` + `Mockito.mock(DeploymentConfiguration.class)`
- `Mockito.spy(new AlwaysLockedVaadinSession(service))`
- `Mockito.mockStatic(FeatureFlags.class)` (in `EnableFeatureFlagExtension`)

These objects are stateless across tests within the same class, so reusing them is safe.

**`MockUIExtension`**: uses `ExtensionContext.Store` at the CLASS scope to compute service and session on the first test and reuse them for subsequent ones. `UI` is still created fresh per test to avoid state leakage. The `context` null-check preserves compatibility with `replaceUI()` / `clearUI()`.

**`EnableFeatureFlagExtension`**: same pattern for `MockedStatic<FeatureFlags>`. The mock holder implements `CloseableResource` so JUnit 5 closes it automatically after all tests in the class. `enableFeature()` is called in `beforeEach` to reset state, since some tests call `disableFeature()`.

**`SVGGeneratorTest`**: each of the 14 tests was creating a new `SVGGenerator` which copies a 3.5 MiB JS bundle to a temp dir. Switch to `@TestInstance(PER_CLASS)` + `@BeforeAll` to share one generator. The 2 tests that close the generator use a local instance.

Measured on local Mac (same parallelism flags as CI: `-T 4 -Dsurefire.parallel=classes -Dsurefire.threadCount=2`):

| | Wall time |
|--|--|
| Before | 37.1s |
| After | 33.4s |

The improvement is larger on CI where Mockito initialization is more expensive due to slower JVM startup and byte-buddy agent loading.

### DIFF
--- a/vaadin-charts-flow-parent/vaadin-charts-flow-svg-generator/src/test/java/com/vaadin/flow/component/charts/export/SVGGeneratorTest.java
+++ b/vaadin-charts-flow-parent/vaadin-charts-flow-svg-generator/src/test/java/com/vaadin/flow/component/charts/export/SVGGeneratorTest.java
@@ -24,9 +24,10 @@ import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
-import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
 
 import com.vaadin.flow.component.charts.model.AnnotationItemLabel;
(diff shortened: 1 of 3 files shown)


manolo:
Closing: the optimization did not produce a meaningful improvement in CI. The 1s granularity in surefire reporting masks sub-second gains, and the actual bottlenecks (DashboardTest, SerializableTests, SVGGeneratorTest) require deeper changes.

PR #9279, opened 2026-05-13
ci: skip svg charts npm build when running unit tests

manolo:
The `vaadin-charts-flow-svg-generator` module runs `npm install` + webpack on every Maven build via the `validate` phase. This regenerates `jsdom-exporter-bundle.js` even though the bundle is already committed to git and the build cache guarantees nothing has changed.

Adding `skipSvgChartsBuild` to the `npm-install` and `generator-build` exec-maven-plugin executions allows skipping these steps. The unit test CI job passes `-DskipSvgChartsBuild` since the build cache is always present there (enforced by `fail-on-cache-miss: true`).

Reduces unit test job duration by ~5s locally, more in CI where `npm install` is slower.

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -178,6 +178,10 @@ jobs:
         if: matrix.type != 'unit'
         run: npm ci --ignore-scripts
 
+      - name: Merge ITs
+        if: matrix.type == 'war'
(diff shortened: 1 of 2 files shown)


manolo:
Approach doesn't work: the bundle is gitignored so it must be generated by webpack on every build. Skipping npm/webpack causes SVGGeneratorTest to fail with NPE.

PR #9290, opened 2026-05-13
feat: add FormAIController field hints API

tomivirkki:
## Summary
- Lets callers attach per-field hints (`describe`, `allowedValues`, `ignore`) so the LLM gets business semantics that aren't implied by the field's label or type.
- Hints are keyed by an opaque id attached to the field component, so they survive removing and re-adding the field within a session.

Closes https://github.com/orgs/vaadin/projects/103/views/3?pane=issue&itemId=186540736

### DIFF
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
@@ -15,7 +15,10 @@
  */
 package com.vaadin.flow.component.ai.form;
 
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.UUID;
 
@@ -58,6 +61,7 @@ public class FormAIController implements AIController {
     private static final String FIELD_ID_KEY = "vaadin.ai.form.fieldId";
 
     private final Component form;
+    private final Map<String, FormFieldHints> hintsById = new HashMap<>();
 
     /**
      * Creates a new form AI controller for the given container. Fields are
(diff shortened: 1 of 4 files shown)


vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
```diff
@@ -92,22 +149,30 @@ public void onRequestStart() {
     public void onResponseComplete() {
     }
 
+    private FormFieldHints requireFillable(HasValue<?, ?> field) {
```
ugur-vaadin: Would it make sense to rename `requireFillable` to something like `hintsFor` to make what it does more obvious?

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/test/java/com/vaadin/flow/component/ai/form/FormAIControllerTest.java
```diff
@@ -167,4 +167,43 @@ void formWithNoFieldsProducesEmptyList() {
                     FormFieldDiscovery.collectFields(form));
         }
     }
+
+    @Nested
+    class HintApi {
+
+        @Test
+        void allHintMethodsReturnThisForChaining() {
```
ugur-vaadin: I'm not sure if this test is necessary.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
```diff
@@ -76,6 +80,59 @@ public <T extends Component & HasComponents> FormAIController(T form) {
...
+     * @param field
+     *            the field to constrain, not {@code null}
+     * @param values
+     *            the allowed values, not {@code null}; a defensive copy is
+     *            taken
+     * @param <T>
+     *            the field's value type
+     * @return this controller, for chaining
+     */
+    public <T> FormAIController allowedValues(HasValue<?, T> field,
```
ugur-vaadin, line 116: I wonder if we should accept a `Collection` instead of a `List`. Is there any reason we should not accept a `Set` as allowed values?

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
```diff
@@ -76,6 +80,59 @@ public <T extends Component & HasComponents> FormAIController(T form) {
...
+            List<? extends T> values) {
+        requireFillable(field).allowedValues = new ArrayList<>(values);
```
ugur-vaadin: Maybe we can assert `requireNonNull` for `values` here similar to the method `describe`.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
```diff
@@ -92,22 +149,30 @@ public void onRequestStart() {
...
      */
     private void attachIds() {
         FormFieldDiscovery.collectFields(form)
-                .forEach(FormAIController::attachId);
+                .forEach(FormAIController::getOrCreateId);
     }
 
-    private static void attachId(HasValue<?, ?> field) {
+    private static String getOrCreateId(HasValue<?, ?> field) {
         // Fields come from Component.getChildren(), so the cast is safe.
```
ugur-vaadin: Since now the user also provides fields, bringing back the check for whether the field is a `Component` might make sense.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/test/java/com/vaadin/flow/component/ai/form/FormAIControllerTest.java
tomivirkki: Well, we don't have any code other than this that makes sure the API is chainable. It's kind of weird as a test though, done.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
tomivirkki: Let's just stick to the spec and do follow-up API redesign systematically. This will actually get removed by a follow-up PR which adds (`valueOptions` hint API)

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
tomivirkki: There was already a test case this throws an NPE for null values, but maybe it's nicer to have it explicit and with a clear error message. Done.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java

ugur-vaadin (approved):

PR #9293, opened 2026-05-15
fix!: remove not working helper-above variants from CheckboxVariant

web-padawan:
## Description

Fixes https://github.com/vaadin/web-components/issues/11750

The only variant in `CheckboxVariant` is `helper-above-field` and it does nothing. Let's remove it as agreed.

## Type of change

- Behavior altering change

### DIFF
--- a/vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/CheckboxVariant.java
+++ b/vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/CheckboxVariant.java
@@ -21,13 +21,7 @@ import com.vaadin.flow.component.shared.ThemeVariant;
  * Set of theme variants applicable for {@code vaadin-checkbox} component.
  */
 public enum CheckboxVariant implements ThemeVariant {
-    LUMO_HELPER_ABOVE_FIELD("helper-above-field"),
-    /**
-     * @deprecated Use {@link #HELPER_ABOVE} instead.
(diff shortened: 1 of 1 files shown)


jouni (approved):

vaadin-checkbox-flow-parent/vaadin-checkbox-flow/src/main/java/com/vaadin/flow/component/checkbox/CheckboxVariant.java
```diff
@@ -1,46 +0,0 @@
...
- * Set of theme variants applicable for {@code vaadin-checkbox} component.
- */
-public enum CheckboxVariant implements ThemeVariant {
-    LUMO_HELPER_ABOVE_FIELD("helper-above-field"),
-    /**
-     * @deprecated Use {@link #HELPER_ABOVE} instead.
-     */
-    @Deprecated
-    AURA_HELPER_ABOVE_FIELD("helper-above-field"),
-    HELPER_ABOVE("helper-above-field");
```
knoobie, line 30: Suggestion: Instead of deleting the whole enum and possible create compile problems for people that rely on the fact that typically fields have `HasThemeVariant<?>` - just delete the enum variants leaving an empty class for later use-cases.

web-padawan: Thanks, done. Also added `AURA_SMALL` which is actually supported to avoid empty enum.

vursen (approved):

PR #9298, opened 2026-05-15
feat: lock form fields during AI fill turn

tomivirkki:
## Summary
- Prevent the user from typing into a field the AI is about to overwrite by switching every non-ignored, currently-editable field to read-only for the duration of a fill turn, releasing on success, failure, or timeout.
- Document the limitation that application code which toggles a field's read-only state mid-turn (e.g. a value-change listener reacting to the LLM's writes) will be overridden when the controller releases its locks — applications should avoid mid-turn toggling, or reapply state after the turn completes.

Related to https://github.com/orgs/vaadin/projects/103/views/3?pane=issue&itemId=186543521

### DIFF
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
@@ -44,6 +44,17 @@ import com.vaadin.flow.component.ai.provider.LLMProvider;
  * </p>
  *
  * <p>
+ * <b>Field locking:</b> while a fill is in progress, every non-ignored field
+ * that wasn't already read-only is set to read-only so the user cannot type
+ * into a field the AI is about to overwrite. Locks are released when the turn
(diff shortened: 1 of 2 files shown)


vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
```diff
@@ -144,10 +156,65 @@ public void onRequestStart() {
...
+     */
+    private void unlockFields() {
+        for (var field : lockedFields) {
+            field.setReadOnly(false);
+        }
+        lockedFields.clear();
+    }
+
+    private boolean isIgnored(HasValue<?, ?> field) {
+        if (!(field instanceof Component component)) {
```
ugur-vaadin: Nit: this check seems to be unnecessary since it is currently not possible to trigger it.

ugur-vaadin (approved):

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java

PR #9301, opened 2026-05-19
refactor: simplify grid connector viewport range handling

vursen:
The grid connector's fetch range used to depend on both the rendered rows and the page the grid asked for. With fast scrolling, the requested page could end up outside the viewport e.g. when quickly scrolling to the bottom and back to the top. In that case, `requestPage` collapsed the range to just the last page, which was requsted at the bottom, because otherwise the DataCommunicator wouldn't return anything when the first page was requested again at the top. The server would then respond, and `confirm` could clear that pending callback. `confirm` had a matching branch that resolved out-of-range callbacks with an empty array and forced a content update.

This workaround is no longer needed: pending callbacks are now always resolved after the server responds to `setViewportRange`, even when no new data is sent. A stale callback for a page outside the current viewport can't keep the grid in a loading state regardless of what range was requested, so the fetch range can be derived purely from the rendered rows.

**Summary of changes in this PR:**

- Drop the `page` argument from the request flow; replace `requestPage(page)` with async `fetchCurrentRange()` that derives the range from rendered rows only, skips duplicate requests, and awaits `setViewportRange`.
- Split the old `getViewportRange` into `getRenderedRange` (rendered rows) and `getFetchRange` (rendered range plus buffer, aligned to page boundaries — the inclusive end rounds up to the end of its page so the last rendered row's page is always fetched).
- Extract `resolvePendingCallbacks`; call it after `setViewportRange` resolves with no new data, and from `confirm`. Removes the three-branch logic in `confirm`.
- Track `requestedRange` as `[startIndex, endIndex]` instead of `lastRequestedRange` in pages.
- Update `treeGridConnector` to use the renamed `getRenderedRange`.

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
@@ -151,9 +151,10 @@ window.Vaadin.Flow.gridConnector.initLazy = (grid) => {
     range[0] = Math.max(range[0] - buffer, 0);
     range[1] = Math.min(range[1] + buffer, grid.size);
 
-    // Align the range to page boundaries
+    // Align the range to page boundaries. range[1] is inclusive of the last
+    // rendered row, so round it up to the end of that row's page.
     range[0] = Math.floor(range[0] / grid.pageSize) * grid.pageSize;
-    range[1] = Math.ceil(range[1] / grid.pageSize) * grid.pageSize;
+    range[1] = (Math.floor(range[1] / grid.pageSize) + 1) * grid.pageSize;
 
     if (requestedViewportRange[0] !== range[0] || requestedViewportRange[1] !== range[1]) {
       grid.$server.setViewportRange(range[0], range[1] - range[0]);
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
@@ -156,7 +156,7 @@ window.Vaadin.Flow.gridConnector.initLazy = (grid) => {
     range[1] = Math.ceil(range[1] / grid.pageSize) * grid.pageSize;
 
     if (requestedViewportRange[0] !== range[0] || requestedViewportRange[1] !== range[1]) {
-      grid.$server.setViewportRange(range[0], range[1] - range[0] + grid.pageSize);
+      grid.$server.setViewportRange(range[0], range[1] - range[0]);
       requestedViewportRange = range;
     }
   };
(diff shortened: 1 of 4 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
```diff
@@ -5,6 +5,10 @@ import { Grid } from '@vaadin/grid/src/vaadin-grid.js';
 import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
 import { GridFlowSelectionColumn } from './vaadin-grid-flow-selection-column.js';
 
+function isRangeEqual(range1, range2) {
+  return new Set(range1).difference(new Set(range2)).size === 0;
```
tomivirkki: Should this be `return range1?.[0] === range2?.[0] && range1?.[1] === range2?.[1];`? The current approach is a bit exotic and `isRangeEqual([5,5],[5,9])` (even while non-reachable state) returns `true`

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
```diff
@@ -138,47 +141,73 @@ window.Vaadin.Flow.gridConnector.initLazy = (grid) => {
...
+    if (isRangeEqual(range, requestedRange)) {
+      // Skip duplicate requests for the same range.
+      return;
+    }
+
+    requestedRange = range;
+
+    await grid.$server.setViewportRange(range[0], range[1] - range[0]);
+
+    if (isRangeEqual(range, requestedRange)) {
```
tomivirkki: Could we add a test to cover the case in which `range !== requestedRange` here?

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
```diff
@@ -138,47 +141,73 @@ window.Vaadin.Flow.gridConnector.initLazy = (grid) => {
...
+
+    return range;
+  };
+
+  grid.$connector.fetchCurrentRange = async () => {
+    const range = grid.$connector.getFetchRange();
+
+    if (isRangeEqual(range, requestedRange)) {
+      // Skip duplicate requests for the same range.
+      return;
```
tomivirkki, line 171: In which scenario could another fetch for the same range occur? There's no test for it

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
vursen: Good point. Done.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
vursen: Added a test.

tomivirkki (approved):

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
vursen: This check isn't really required, and we can call `$connector.resolvePendingRequests` unconditionally. Calling it after it was already called by `$connector.confirm` is harmless –  it will just cause the grid to synchronously recreate page callbacks once more if the viewport still contains items that need to be loaded. So I think I'll just remove it.

PR #9303, opened 2026-05-19
fix: make spreadsheet overlays native popover, per instance

DiegoCardoso:
The spreadsheet attaches a `<div id="spreadsheet-overlays">` to host its overlay widgets (context menu, tooltips, comment overlays, popup buttons). It has historically lived in `<body>` to escape clipping by ancestor `overflow`/`transform` rules and as a singleton shared by all spreadsheet instances on the page. Switching the overlays to native popover changes the picture: every overlay enters the top layer regardless of DOM position, so neither the body-attachment nor the singleton are load-bearing anymore.

This PR includes the native popover changes proposed in #9270 and adds the structural changes that the popover approach enables:

- the container moves into the `<vaadin-spreadsheet>` light DOM, projected into the shadow root via a new `overlays` slot, with one container per instance;
- the reference is threaded from JS into the GWT widget chain instead of being looked up by id;
- `SpreadsheetConnector` — not `ApplicationConnection` — now owns the `SpreadsheetContextMenu`.

Tooltips and the cell-comment overlay are created in `SheetWidget`'s constructor (before the host is known), so the container is wired in via a setter rather than a constructor argument.

A side benefit: spreadsheet overlays inside a modal `Dialog` are interactive again. The modal sets `body { pointer-events: none }` to disable everything outside the dialog overlay; with the container now inside the dialog's slotted content, the overlays inherit `pointer-events: auto` from the dialog's overlay part instead of `pointer-events: none` from `<body>`.

Intended to replace #9270.

Related to #9270

---

🤖 Generated with Claude Code


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/resources/META-INF/resources/frontend/vaadin-spreadsheet/vaadin-spreadsheet-styles.js
sissbruecker, line 1: `.v-tooltip` also needs a reset for the user agent styles, the tooltip that shows when resizing a column for example shows a 3px border.

vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/SpreadsheetOverlay.java
```diff
@@ -8,53 +8,111 @@
...
             DOM.setElementProperty(getElement(), "id", "PID_VAADIN_CM");
+            setPopover(getElement());
+        }
+
+        @Override
+        public com.google.gwt.user.client.Element getOverlayContainer() {
+            return asUserElement(overlayContainer);
+        }
+
+        public void setOverlayContainer(Element overlayContainer) {
```
sissbruecker: `SpreadsheetContextMenu` has the same method, which is unused.

vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/resources/META-INF/resources/frontend/vaadin-spreadsheet/vaadin-spreadsheet-styles.js
DiegoCardoso: Good catch — added `border: unset;` to `.v-tooltip` as well. The base rule covers the Lumo variant too, so one line does it.

vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/SpreadsheetOverlay.java
DiegoCardoso: Removed it — the context menu now gets its container from the constructor, so the setter was dead. The outer one on `SpreadsheetOverlay` stays since the no-arg overlays still rely on it.

sissbruecker (approved):

PR #9305, opened 2026-05-19
fix: Added https to moreInfoLink for feature flag

mgarciavaadin:
## Description

Added https to moreInfoLink for the feature flag provider, otherwise the link is opened in localhost


yuriy-fix:
No longer needed since badge flag was removed (alongside MDL and planning to remove Slider flag as well)

PR #9307, opened 2026-05-19
feat: add get_form_state tool to FormAIController

tomivirkki:
## Summary
- The form filler needs the LLM to know which fields exist, what each one means, and what each one currently holds before it can decide what to fill — `get_form_state` is the controller-side contract that delivers that view as JSON.
- Schema rendering handles the awkward edges the LLM would otherwise stumble on: `BigInteger` without overflow, non-finite doubles as JSON null, `BigDecimal` in plain (non-scientific) form, multi-select shape via `items`, per-field error isolation so one bad field doesn't kill the whole payload.
- Classification recognises real Vaadin selection components (`HasItems`, `HasDataView`, `HasLazyDataView`, `HasListDataView`) and the full integer/number value-type families, so `ComboBox`/`Select` and custom `Long`/`Short`/`Float` fields all land in the right schema bucket.


vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/test/java/com/vaadin/flow/component/ai/form/FormAIControllerTest.java
tomivirkki, line 1: This test file was getting way too large, so needed to split it in https://github.com/vaadin/flow-components/pull/9307/changes/fb70e78722c3f9f754c33eddf06a4f9aad3d7a40

ugur-vaadin (approved):

PR #9310, opened 2026-05-20
ci: disable pnpm minimumReleaseAge when @NpmPackage versions changed

manolo:
When bumping @NpmPackage versions, the new alpha packages are published minutes before CI runs and pnpm/npm reject them with minimumReleaseAge.

For both the WAR build (pnpm) and WTR tests (npm via flow:build-frontend): fetch the PR target branch, diff against HEAD, and if any Java file containing @NpmPackage changed, set minimumFrontendPackageAgeDays=0 via MAVEN_OPTS so the Flow plugin disables the age check for that run.

No GH token needed: pure git fetch + diff + grep.

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -188,6 +188,15 @@ jobs:
         if: matrix.type != 'unit'
         run: npm ci --ignore-scripts
 
+      - name: Detect @NpmPackage version changes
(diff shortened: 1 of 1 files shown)


vursen:
This was already merged together with https://github.com/vaadin/flow-components/pull/9311

PR #9319, opened 2026-05-21
fix: correct false positive messages in validation summary

manolo:
When a WTR job fails due to a session-level error (e.g. browser timeout), the `wtr-results.xml` accurately reflects the individual test results (all passed) but gives no indication of why the job failed. The summary then showed WTR as green while incorrectly firing the "Compilation failed. Check the Package WAR job" message.

Two fixes:

**`scripts/wtr.js`**: when `web-test-runner` exits with a non-zero code and the XML contains no real test failures, a synthetic failing `testsuite` is appended with the session error message. This makes dorny publish the WTR report as failed with the error visible inline, the same as unit tests and IT failures.

**`.github/workflows/validation.yml`**: the "IT compilation error" condition now also requires `needs.fast-tests.outputs.has-tests == ''`. That output is only set after WAR packaging succeeds, so an empty value reliably means WAR failed rather than another job being cancelled first. The redundant fallback messages for unit and WTR (which fired incorrectly on cancellation) have been removed.

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -1,8 +1,6 @@
 name: Validation
 
 on:
-  push:
-    branches: ['fix/validation-summary-false-positives']
   schedule:
     - cron: '30 23 * * *'
   pull_request:
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -534,7 +534,7 @@ jobs:
           fail-on-error: 'false'
 
       - name: Unit test compilation error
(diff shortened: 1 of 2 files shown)


.github/workflows/validation.yml
```diff
@@ -1,6 +1,8 @@
 name: Validation
 
 on:
+  push:
+    branches: ['fix/validation-summary-false-positives']
```
vursen:
```suggestion
```

vursen: Highlighting this so that we don't forget to remove it.

vursen (approved):

PR #9308, opened 2026-05-19
feat: add binder integration to form ai controller

ugur-vaadin:
## Description
This PR:
- Adds a `FormAIController(container, Binder<?>)` overload.                                                                                                                
  - Property-named bindings (`bind("name")`, `bindInstanceFields(this)`, `@PropertyId(...)`) contribute the bean property name as a default for the field's description — surfaced to the LLM alongside the field's label and helper text. The field's tool-call id stays an opaque UUID assigned at discovery time; the property name is description-side metadata only.
  - Seeding only fills nulls: an explicit `describe(field, "...")` always wins. Lambda-bound bindings and unbound fields contribute no default.
- Adds a package-private `findBinding(field)` returning the live binding for any field on the supplied binder, or an empty optional for unbound fields and for controllers constructed without a binder. 

Fixes https://github.com/orgs/vaadin/projects/103/views/1?filterQuery=&pane=issue&itemId=186542732

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [x] I have read the contribution guide: https://vaadin.com/docs/latest/contributing/overview
- [x] I have added a description following the guideline.
- [x] The issue is created in the corresponding repository and I have referenced it.
- [x] I have added tests to ensure my change is effective and works as intended.
- [x] New and existing tests are passing locally with my change.
- [x] I have performed self-review and corrected misspellings.
- [x] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.

PR #9320, opened 2026-05-21
feat: add fill form tool to form ai controller

ugur-vaadin:
## Description

This PR:

- Adds the `fill_form` tool to `FormAIController.getTools()`. The two existing tools only let the LLM read the form; this one lets it write, closing the loop on the form-filler workflow. Payload is keyed by the same field UUIDs `get_form_state` emits; ignored and unsupported fields are filtered from both the schema and the write path.
- Returns a `Written:` / `Rejected:` write-summary, not the full form state. Excludes ignored fields.
- Adds the JSON-to-typed-value write path in `FormValueConverter#convert` for the `FormFieldType` values. Failed conversions throw `RejectedValueException` with a curated LLM-facing reason that the `Rejected:` block surfaces verbatim. One field failing doesn't abort the rest of the fill.

Fixes https://github.com/orgs/vaadin/projects/103/views/1?filterQuery=&pane=issue&itemId=186545282

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
@@ -508,10 +508,10 @@ public class FormAIController implements AIController {
             for (var id : arguments.propertyNames()) {
                 var field = byId.get(id);
                 if (field == null) {
-                    rejected.add(new RejectedEntry(id,
-                            "Unknown field id '" + id
-                                    + "'. Call get_form_state to refresh the "
-                                    + "id list and retry only this entry."));
+                    rejected.add(new RejectedEntry(id, "Unknown field id '" + id
+                            + "'. Call get_form_state to refresh the "
+                            + "id list and retry only entries that are "
+                            + "rejected with the reason unknown field id."));
                     continue;
                 }
                 applyValue(field, arguments.get(id), written, rejected);
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
@@ -18,6 +18,7 @@ package com.vaadin.flow.component.ai.form;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -34,15 +35,14 @@ import org.slf4j.LoggerFactory;
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.ComponentUtil;
 import com.vaadin.flow.component.HasComponents;
-import com.vaadin.flow.component.HasLabel;
 import com.vaadin.flow.component.HasValue;
-import com.vaadin.flow.component.UI;
 import com.vaadin.flow.component.ai.form.FormAITools.FormFieldDescriptor;
 import com.vaadin.flow.component.ai.form.FormValueConverter.RejectedValueException;
 import com.vaadin.flow.component.ai.orchestrator.AIController;
 import com.vaadin.flow.component.ai.orchestrator.AIOrchestrator;
 import com.vaadin.flow.component.ai.provider.LLMProvider;
 import com.vaadin.flow.data.binder.Binder;
+import com.vaadin.flow.internal.JacksonUtils;
 
 import tools.jackson.databind.JsonNode;
 
@@ -189,6 +189,16 @@ public class FormAIController implements AIController {
      * options as the labels the LLM should see; the {@code toValue} function
      * converts a chosen label back to the field's value type. Later calls for
      * the same field overwrite earlier ones.
+     * <p>
+     * <b>Multi-select fields:</b> a multi-select's value type is
+     * {@code Set<Item>}, so the typed signature forces {@code toValue} to
+     * return {@code Set<Item>}. Write {@code toValue} to return a single-item
+     * set per label (e.g. {@code label -> Set.of(itemByLabel.get(label))}); the
+     * orchestrator aggregates the per-label sets into the final selection
+     * before calling {@link HasValue#setValue setValue}. A
+     * {@code Function<String, Item>} also works through an unchecked cast on
+     * the {@code field} argument, dropping the {@code Set} wrap; the
+     * orchestrator collects single items into the aggregate set the same way.
      *
      * @param field
      *            the field whose options the LLM may query, not {@code null}
@@ -328,14 +338,11 @@ public class FormAIController implements AIController {
      * for any bound field that does not already have an explicit description.
      * Called at the start of every turn so bindings added or removed between
      * turns are reflected; an explicit {@link #describe(HasValue, String)}
-     * always wins because the seeding only fills in nulls.
+     * always wins because the seeding only fills in nulls. Safe to call when
+     * the controller was built without a binder — {@link BinderReflection}
+     * returns an empty map for a {@code null} binder.
      */
     private void seedDescriptionsFromBinder() {
-        if (binder == null) {
-            // Controllers created with the no-binder constructor have
-            // nothing to seed.
-            return;
-        }
         var propertyNames = BinderReflection.collectPropertyNames(binder);
         for (var entry : propertyNames.entrySet()) {
             var field = entry.getKey();
@@ -455,14 +462,18 @@ public class FormAIController implements AIController {
 
         @Override
         public String executeFill(JsonNode arguments) {
-            var ui = form.getUI().orElse(null);
-            // Run inline when there's no UI to hop to (unit tests, detached
-            // form) or when the calling thread is already that UI's thread
-            // (e.g. a click handler). Only background reactor threads need
-            // the ui.access + wait hop.
-            if (ui == null || UI.getCurrent() == ui) {
-                return doFill(arguments);
-            }
+            // The fill always hops through ui.access so writes land on the
+            // UI thread with CurrentInstance bound, regardless of whether
+            // the LLM provider invoked the tool from a reactor scheduler
+            // thread (the production path) or directly from the UI thread
+            // itself (in which case ui.access runs the lambda synchronously
+            // and future.get() returns immediately, so the hop is a no-op).
+            // A controller whose form isn't attached to a UI is a
+            // configuration error — fail fast rather than write silently
+            // to a detached state tree.
+            var ui = form.getUI().orElseThrow(() -> new IllegalStateException(
+                    "fill_form invoked on a controller whose form is not "
+                            + "attached to a UI"));
             var future = new CompletableFuture<String>();
             ui.access(() -> {
                 try {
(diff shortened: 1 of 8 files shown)


vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAITools.java
```diff
@@ -247,11 +258,81 @@ public String execute(JsonNode arguments) {
...
+                for (var d : callbacks.visibleFields()) {
+                    try {
+                        properties.set(d.id(), FormFieldSchema.build(d.id(),
+                                d.field(), d.type(), d.hints()));
+                    } catch (Exception ex) {
+                        LOGGER.warn("fill_form schema build failed for {}",
+                                d.id(), ex);
+                    }
+                }
+                return schema.toString();
```
tomivirkki: The `fill_form` schema is rebuilt per `getParametersSchema()` call, with one property per visible field. We'd agreed in the meeting a couple of days ago to go with a static, open-keyed `values: object` schema and let the form's shape flow through `get_form_state` results rather than the tool definition — looks like this PR went the other way.

Two things the dynamic shape loses worth flagging:

**Tool-definition cache invalidation.** LLM providers cache prompt prefixes including tool definitions; a schema regenerated per turn misses that cache on every request. With the static shape, the tool definition warms up on the first request of a session and stays warm.

**Mid-turn structural changes.** `getParametersSchema()` fires once when the stream opens, so the per-field schema is frozen for the duration of the turn. A value-change listener that swaps options between `fill_form` calls, or a follow-up that adds/removes a field, can't be reflected — the schema is stuck on the turn's opening snapshot. With the static shape, `get_form_state` re-reads on every call, so structural changes show up on the next read.

ugur-vaadin: Updated the PR to use a static open-keyed schema. Added a few tests for the behaviour.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
```diff
@@ -317,6 +331,11 @@ public void onResponseFailed(Throwable error) {
      * always wins because the seeding only fills in nulls.
      */
     private void seedDescriptionsFromBinder() {
+        if (binder == null) {
```
tomivirkki: Looks like Claude restored this https://github.com/vaadin/flow-components/pull/9308#discussion_r3273581655

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
tomivirkki, line 1: The PR includes some untested lines. Please make sure to cover all code.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAITools.java
```diff
@@ -247,11 +258,105 @@ public String execute(JsonNode arguments) {
         };
     }
 
+    /**
+     * Guidance prepended to the {@code fill_form} tool description. Pulled into
+     * a constant so the LLM-facing text can be reviewed in one place.
+     */
+    private static final String FILL_FORM_GUIDANCE = """
```
tomivirkki: Why not inline the description and schema to the tool, like the other tools do?

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
```diff
@@ -433,5 +452,110 @@ public List<String> queryFieldOptions(String fieldId, String filter,
...
+        private String doFill(JsonNode arguments) {
+            var written = new ArrayList<String>();
+            var rejected = new ArrayList<String>();
+            for (var field : visibleFields()) {
+                if (!arguments.has(field.id())) {
+                    continue;
+                }
+                applyValue(field, arguments.get(field.id()), written, rejected);
+            }
+            return formatWriteSummary(written, rejected);
```
tomivirkki: I'm suggesting returning a JSON node instead of the human-readable summary. Example shape:

```json
{"success": true, "rejected": [{"id": "<uuid>", "reason": "..."}]}
```

The current contract loses per-id attribution: a follow-up turn that wants to retry only the rejected entries has to text-match `displayName` lines back to ids, which the LLM cannot reliably do (labels collide, descriptions get truncated).

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
```diff
@@ -433,5 +452,110 @@ public List<String> queryFieldOptions(String fieldId, String filter,
...
+            }
+        }
+
+        private String doFill(JsonNode arguments) {
+            var written = new ArrayList<String>();
+            var rejected = new ArrayList<String>();
+            for (var field : visibleFields()) {
+                if (!arguments.has(field.id())) {
+                    continue;
+                }
```
tomivirkki: This iterates the **form's** fields and silently drops payload entries whose id doesn't match a visible field. An LLM sending a stale id gets no feedback — the write looks like a no-op.

Consider iterating `arguments.properties()` (or `fieldNames()`) instead and adding a rejection entry for any id that isn't in `byId`, with a reason that points back to `get_form_state` so the model knows to refresh, e.g.:

> `"Unknown field id 'X'. Call get_form_state to refresh the id list and retry only this entry."`

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormValueConverter.java
```diff
@@ -149,4 +158,152 @@ private static Optional<Method> lookupMethod(MethodKey key) {
...
+        case BIG_DECIMAL -> convertBigDecimal(value);
+        case NUMBER -> convertNumber(value);
+        case INTEGER -> convertInteger(value);
+        case BOOLEAN -> convertBoolean(value);
+        case DATE -> convertDate(value);
+        case DATE_TIME -> convertDateTime(value);
+        case TIME -> convertTime(value);
+        default -> throw new RejectedValueException(
+                "Unsupported field type: " + field.type());
+        };
```
tomivirkki, line 230: Multi-select writes currently die here with `"Unsupported field type"` even though `valueOptions(field, query, toValue)` is a public API on `FormAIController`.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
```diff
@@ -433,5 +452,110 @@ public List<String> queryFieldOptions(String fieldId, String filter,
...
+            }
+            return formatWriteSummary(written, rejected);
+        }
+
+        @SuppressWarnings({ "unchecked" })
+        private void applyValue(FormFieldDescriptor field, JsonNode value,
+                List<String> written, List<String> rejected) {
+            Object converted;
+            try {
+                converted = FormValueConverter.convert(field, value);
```
tomivirkki, line 527: `valueOptionsToValue` is registered on hints (line 212) but **never invoked on writes** — `FormValueConverter.convert` only does type-driven parsing.

The result: an application registering

```java
controller.valueOptions(combobox,
    (filter, limit) -> projectsByName.values().stream().map(Project::name).toList(),
    projectsByName::get);
```

will have the LLM send `"Apollo Redesign"` as a JSON string, `convert` returns it as a `String`, and `setValue` on the `ComboBox<Project>` rejects it because the field expects a `Project`.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
```diff
@@ -433,5 +452,110 @@ public List<String> queryFieldOptions(String fieldId, String filter,
...
+        @Override
+        public String executeFill(JsonNode arguments) {
+            var ui = form.getUI().orElse(null);
+            // Run inline when there's no UI to hop to (unit tests, detached
+            // form) or when the calling thread is already that UI's thread
+            // (e.g. a click handler). Only background reactor threads need
+            // the ui.access + wait hop.
+            if (ui == null || UI.getCurrent() == ui) {
+                return doFill(arguments);
+            }
```
tomivirkki:
```suggestion
            var ui = form.getUI().orElseThrow(
                    () -> new IllegalStateException("fill_form invoked on a "
                            + "controller whose form is not attached to a UI"));
```

Let's migrate FillFormToolTest to MockUIExtension so this block, which only exists for testing purposes, wouldn't be needed

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
ugur-vaadin: Fixed the converter to use `toValue` when it's available.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
ugur-vaadin: Added the "try again" rejection message for unknown fields.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
ugur-vaadin: Updated the tool to return `{"success": bool, "written": [field-ids], "rejected": [{"id": field-id, "reason": "..."}]}`.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAITools.java
ugur-vaadin: Inlined.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
ugur-vaadin: Added more tests.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
ugur-vaadin: I added it back to avoid spamming warnings when there is no binder. However, I moved it now to `BinderReflection`. Now it returns an empty map if that is the case, rather than logging a warn on every prompt.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
ugur-vaadin: Done. Updated the tests.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormValueConverter.java
ugur-vaadin: `convert` now handles `MULTI_SELECT` explicitly. Added a paragraph to the JavaDoc of `valueOptions` in order to be able to use `toValue` since it either requires an overload that handles multi-select or parameter change to `Function<String, ?> toValue` otherwise.

tomivirkki (approved):

PR #9341, opened 2026-05-22
feat: surface validation in fill_form response

tomivirkki:
## Summary
- Wires `Binder.Binding` / `HasValidator` rejections into the `fill_form` tool result so the LLM can self-correct in the same turn instead of silently leaving the form in an invalid state. Validation runs at `fill_form` execution time only, per the RFC's validation-feedback flow.
- Per RFC, `fill_form` now mirrors `get_form_state`'s shape — returning the full post-write `fields` block plus `rejected` — so value-change listener cascades and other downstream shifts are visible without an extra round-trip. (Breaking: the previous `success` / `written` keys are gone; derive success from `rejected.isEmpty()`.)
- The schema is otherwise intentionally kept shape-only (type / format / value / enum / queryable / array / items); JSR-303 and component-side *constraints* reach the LLM through the validator feedback path on `fill_form`. Surfacing those constraints proactively in `get_form_state` (e.g. via `jakarta.validation.Validator` introspection, paired with `BeanValidationBinder`) is deferred until later.

Closes https://github.com/orgs/vaadin/projects/103/views/3?pane=issue&itemId=186557546


ugur-vaadin (approved):
We can maybe add a paragraph about validation to `FormAIController` class-level Javadocs.

PR #9357, opened 2026-05-25
fix: resolve eager-items selects without explicit valueOptions

tomivirkki:
## Summary
- An eager `setItems(...)` ComboBox/Select/RadioButtonGroup/CheckboxGroup/MultiSelectComboBox was unwritable from `fill_form` unless the app also called `FormAIController.valueOptions(...)`, even though the schema already surfaces those items as `enum`.
- The converter now resolves an LLM-supplied label against the field's in-memory items via `renderItem` for both single- and multi-select, so both halves of the tool protocol agree on the label set without forcing a second registration.


vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormValueConverter.java
ugur-vaadin: Maybe we can update this part to reflect the changes.

tomivirkki: Done

ugur-vaadin (approved):

PR #9370, opened 2026-05-26
refactor: key FormAIController hints by field

tomivirkki:
## Summary
- `FormAIController` previously stored hints in `HashMap<String, FormFieldHints>` keyed by an id. Nothing removed entries when fields left the form, so a long-lived controller across many add/remove cycles (dynamic forms, tabs, lists) grew the map without bound.
- Switched storage to `WeakHashMap<HasValue<?, ?>, FormFieldHints>`. When a field is removed from the form and unreferenced elsewhere, its hint entry becomes GC-eligible.

https://github.com/user-attachments/assets/d4a3b247-7913-4fa4-9903-4e5a2a214a14

Fixes https://github.com/orgs/vaadin/projects/103/views/3?pane=issue&itemId=189450046

### DIFF
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
@@ -17,13 +17,13 @@ package com.vaadin.flow.component.ai.form;
 
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.UUID;
+import java.util.WeakHashMap;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.function.BiFunction;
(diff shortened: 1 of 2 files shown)


vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
```diff
@@ -113,7 +113,13 @@ public class FormAIController implements AIController {
...
     private final Component form;
     private final Binder<?> binder;
-    private final Map<String, FormFieldHints> hintsById = new HashMap<>();
+    /**
+     * Per-field hint state, keyed by the field itself. {@link WeakHashMap} lets
+     * the entry become collectable once the field is no longer reachable from
+     * the form tree or application code, so a long-lived controller across many
+     * add/remove cycles does not accumulate stale hints.
+     */
+    private final Map<HasValue<?, ?>, FormFieldHints> hintsByField = new WeakHashMap<>();
```
ugur-vaadin, line 122: I wonder if we should wrap the hints in a `WeakReference`. It seems likely that the value options lambdas will contain a reference to the component itself, which will prevent the entry from being garbage collected.

ugur-vaadin, line 122: I wonder if there is a good way to move the id into the hints and drop the use of `ComponentUtil.setData` as mentioned in the issue description. Should we abandon that idea, what do you think?

tomivirkki: The map entry is the only thing referencing the hints. If `WeakReference` was used, wouldn't the hints get GC'd unexpectedly?

tomivirkki: I did try this approach initially, [here's the proto commit](https://github.com/vaadin/flow-components/commit/bf6402628f5df6b439af16d48c24bbad82c49b8f). But it requires adding [package-private helper](https://github.com/vaadin/flow-components/commit/bf6402628f5df6b439af16d48c24bbad82c49b8f#diff-b063e15e577933ab65cc5a01f1e9e58a4f66fda4278ce03d860e818797e0fb8fR433) for testing purposes only, which isn't ideal. Keeping the String id attached to the field feels like the lesser evil.

EDIT: Well, maybe there could be a helper in the test package that just uses reflection. Should be fine

PR #9381, opened 2026-05-28
fix: defer Spreadsheet.selectCellRange until sheet widget layout is ready

DiegoCardoso:
Follow-up to #9333. When a Spreadsheet inside a closed Dialog is opened, Flow
flushes the property writes and the `setSelectedCellAndRange` RPC in the same
task. The api receives the call before the deferred init scheduled in
`SheetWidget.resetFromModel` has run — so `SheetWidget.getRowHeight` reads an
uninitialized `definedRowHeights` and throws

```
Uncaught TypeError: Cannot read properties of undefined (reading 'length')
```

That throw drops the selection and trips the `*_noErrors`/`*_noConsoleErrors`
console-log assertions.

#9333's connector-side guard masked this by probing `definedRowHeights`
through the api's private GWT fields — names that only survive
`-Dgwt.style=pretty` (local/dev). CI compiles the GWT client with
`-Dgwt.style=OBF`, where those private names are renamed and `api.spreadsheetWidget`
is `undefined`. The guard itself then threw, producing the same SEVERE log and
the same ~20 spreadsheet IT failures on the cherry-pick PRs.

Move the layout-readiness handling into the widget that owns the operation:

- `SheetWidget.isLoaded()` (package-private) — exposes the already-existing
  `loaded` flag that flips at the end of `resetFromModel`'s deferred command.
- `SpreadsheetWidget.selectCellRange` — if `!sheetWidget.isLoaded()`, store the
  latest args and `Scheduler.scheduleDeferred(this::applyPendingSelectCellRange)`.
  The deferred re-checks, applies, or re-schedules. Latest-wins (rapid successive
  calls collapse to the final selection). Bounded at 100 hops as a safety net;
  in the typical case the layout init has already been scheduled before us, so
  GWT's FIFO scheduler resolves us in 0–1 hops.
- `vaadin-spreadsheet.js` — `setSelectedCellAndRange` goes back to a one-liner
  (`this._flush(); this.api.setSelectedCellAndRange(...args)`). No try/catch,
  no pending state, no constants.

Only `selectCellRange` is deferred — it's the only RPC that touches row/column
metrics. Other RPCs (`cellsUpdated`, popup ops, etc.) don't need this and
didn't fail in the original CI runs.

Reproduce locally by compiling the client with `-Dgwt.style=OBF` (matching CI)
and running the merged ITs in production mode:

```
mvn clean install -DskipTests -Dgwt.style=OBF \
  -pl vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client,vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow
node scripts/mergeITs.js spreadsheet
mvn verify -Drun-it -pl integration-tests -Dvaadin.productionMode=true \
  -Dit.test='NavigationIT,SpreadsheetDialogIT' -DskipUnitTests
```

- With this PR → NavigationIT 26/0, SpreadsheetDialogIT 1/0.
- Reverting only the GWT-side change → SpreadsheetDialogIT fails with the
  underlying `Cannot read properties of undefined (reading 'length')` in
  `FlowClient-…js` — the actual `definedRowHeights` NPE.

Follow-up to #9333.

---


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/SpreadsheetWidget.java
```diff
@@ -2045,15 +2049,59 @@ public void selectCell(String name, int col, int row, String value,
...
+        }
+        if (!sheetWidget.isLoaded()) {
+            if (++selectCellRangeRetries > MAX_SELECT_CELL_RANGE_RETRIES) {
+                pendingSelectCellRangeArgs = null;
+                selectCellRangeRetries = 0;
+                return;
+            }
+            selectCellRangeRetryScheduled = true;
+            Scheduler.get().scheduleDeferred(this::applyPendingSelectCellRange);
+            return;
```
vursen: I'm not sure the retry logic is actually necessary. Couldn't we just do one scheduleDeferred directly in `selectCellRange` if `isLoaded` returns `false`?

DiegoCardoso: Makes sense. I was being a bit conservative, but doing just one `scheduleDeferred` should be enough for most if not all cases.

vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/SpreadsheetWidget.java
```diff
@@ -2051,10 +2051,27 @@ public void selectCell(String name, int col, int row, String value,
...
 
+    /**
+     * Sheet layout state is populated by a {@code scheduleDeferred} command in
+     * {@link SheetWidget#resetFromModel}, so {@code selectCellRange} can arrive
+     * before {@link SheetWidget#isLoaded()} flips {@code true} — e.g. when a
+     * Spreadsheet inside a Dialog opens and Flow flushes property writes and
+     * RPCs in the same task. In that window, {@code scrollAreaIntoView} would
+     * dereference an uninitialized {@code definedRowHeights}. Defer the call
+     * via the same scheduler so it runs after the layout init command. Only
+     * this RPC touches row/column metrics, so only it needs deferring.
```
vursen: **nit:** The current description reads more like an inline code comment. Could we first describe what the method does, and then mention on a separate line that it may defer the call if the widget hasn't loaded yet?

vursen (approved):

PR #9386, opened 2026-05-28
fix: hide already-filtered values from other spreadsheet column filters

### LINKED ISSUE #9329: Spreadsheet: Inclusion of hidden values in filter dropdown is confusing
### Description

In a multi-column table with filtering, the filter dropdown for any column includes values that have already been hidden via filters in other columns.

As an example, if you have the following table:

| Column A   | Column B    | Column C       |
| ------ | ----- | ------- |
| Alpha | Foo | Alice |
| Beta | Bar | Bob |
| Gamma | Baz | Carol |

If you filter out `Beta` in Column A, then open the filter-dropdown for Column B, it will show the corresponding value `Bar` unchecked (i.e. already filtered-out), but checking the checkbox (to unfilter it) does nothing, since the Column A filter still applies.

To make things worse, applying additional filters in another column seems to implicitly add the automatically-unchecked (i.e. filtered in other columns) values to that column's filter, so that even when you re-check them in the original column's filter (to bring them back), they don't re-appear, because the second column _also_ filters them.

As an example, after filtering out `Beta` in Column A, if you then filter out `Foo` in column B, and go back to Column A's dropdown and un-filter `Beta`, that row will not re-appear.

### Expected outcome

Values hidden by filters on other columns should be omitted from the filter list entirely.

https://vaadin.com/docs/latest/components/spreadsheet#tables-filters

### Steps to reproduce

Filter some values in Column 1, then open the dropdown for Column 2.

### Environment

Vaadin version(s):  25.1

### Browsers

Issue is not browser related

vursen:
## Description

In a `SpreadsheetFilterTable` with filters on multiple columns, opening any column's filter dropdown showed values from rows already hidden by other columns as unchecked, suggesting they were filtered by this column. Re-checking them did nothing because another column still hid the row.

Applying a second filter then implicitly added those already-hidden rows to the second column's `filteredRows`, so clearing the first filter did not bring the rows back — the second column now also filtered them.

Root cause: `ItemFilter` read `spreadsheet.isRowHidden(r)` for both building option lists and deciding which rows belong in its own `filteredRows`. That property reflects the union of all column filters, so each filter could not tell its own work from a sibling's.

Each `ItemFilter` now operates only on its own `filteredRows`. The dropdown option list omits values whose rows are all hidden by other columns via a new `SpreadsheetFilterTable.getRowsHiddenByOtherFilters(self)` helper. `clearAllFilters` does a second pass to refresh option lists so cleared siblings are reflected immediately.

Fixes #9329

## Reproduction

```java
@Route("spreadsheet-filter-cross-column")
public class SpreadsheetFilterCrossColumnPage extends VerticalLayout {

    public SpreadsheetFilterCrossColumnPage() {
        Spreadsheet spreadsheet = new Spreadsheet();
        spreadsheet.setTheme(SpreadsheetTheme.LUMO);
        spreadsheet.setHeight("400px");

        spreadsheet.createCell(0, 0, "Column A");
        spreadsheet.createCell(0, 1, "Column B");
        spreadsheet.createCell(0, 2, "Column C");

        spreadsheet.createCell(1, 0, "Alpha");
        spreadsheet.createCell(1, 1, "Foo");
        spreadsheet.createCell(1, 2, "Alice");

        spreadsheet.createCell(2, 0, "Beta");
        spreadsheet.createCell(2, 1, "Bar");
        spreadsheet.createCell(2, 2, "Bob");

        spreadsheet.createCell(3, 0, "Gamma");
        spreadsheet.createCell(3, 1, "Baz");
        spreadsheet.createCell(3, 2, "Carol");

        CellRangeAddress range = new CellRangeAddress(0, 3, 0, 2);
        SpreadsheetFilterTable table = new SpreadsheetFilterTable(spreadsheet,
                range);
        spreadsheet.registerTable(table);
        spreadsheet.refreshAllCellValues();

        add(spreadsheet);
    }
}
```

| Column A | Column B | Column C |
| -------- | -------- | -------- |
| Alpha    | Foo      | Alice    |
| Beta     | Bar      | Bob      |
| Gamma    | Baz      | Carol    |

Steps to reproduce:

- Filter `Beta` in column A.
- Open the filter dropdown in column B → `Bar` shows up unchecked (before the fix), suggesting column B filtered it.
- Filter `Foo` in column B.
- Re-check `Beta` in column A → the row does not reappear (before the fix), because column B now also filters it.

🤖 Generated with Claude Code


DiegoCardoso (commented):
There's an inconsistent behavior with this implementation. The unchecked values only appear at the last filtered column:

https://github.com/user-attachments/assets/56dbe4b1-8f38-4fdb-a21c-3f2db8dec7f1

In the example above, I have set up the following table:

| Column A | Column B | Column C | Column D | Column E |
| --- | --- | --- | --- | --- |
| A2 | B2 | C2 | D2 | E2 |
| A3 | B3 | C3 | D3 | E3 |
| A4 | B4 | C4 | D4 | E4 |
| A5 | B5 | C5 | D5 | E5 |
| A6 | B6 | C6 | D6 | E6 |
| A7 | B7 | C7 | D7 | E7 |
| A8 | B8 | C8 | D8 | E8 | 
| A9 | B9 | C9 | D9 | E9 | 

Here's what is done in the video:
- I started by filtering out the "A2" and "A3" values from "Column A"
- Upon closing and opening the filter popup in "Column A", the filtered out values appear unchecked
- Then, I filtered out the "B4" value in "Column B"
- Now, when I close and open the filter popup in "Column B", the "B4" values appears unchecked
- BUT, when the filter popup in "Column A" is opened, no unchecked values appear

The only way to show the rows again is by clicking on the "Clear filters" button.

Perhaps, the values that have been filtered out by each column should remain available for checking back again?

vursen:
> Perhaps, the values that have been filtered out by each column should remain available for checking back again?

Yes, good catch, thanks. Fixed.

vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-integration-tests/src/main/java/com/vaadin/flow/component/spreadsheet/tests/SpreadsheetFilterPage.java
DiegoCardoso, line 1: NIT: There seems to not be a consensus of where reproductions should live. We have many that uses the fixtures and open the main `/vaadin-spreadsheet` demo page and loads the fixture needed to reproduce the tested behavior. We also have been adding some views lately, so it could be just a matter of preference. Still, I wanted to point out here that this could use a fixture instead, like [SpreadsheetTableFixture](https://github.com/vaadin/flow-components/blob/main/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-integration-tests/src/main/java/com/vaadin/flow/component/spreadsheet/tests/fixtures/SpreadsheetTableFixture.java) (which seems to be unused by any ITs, so it could be updated to have a more meaningful table than this one).

vursen: Oh, good point, done. I also found there was already `SheetFilterTableIT` for these kinds of tests, so I merged my new tests into it.

DiegoCardoso (approved):

PR #9391, opened 2026-05-29
feat: add Markdown.bindContent(Signal<String>)

### LINKED ISSUE #9388: Markdown should provide a bindContent(Signal) method
### Describe your motivation

I tried to bind a value signal to the `Markdown` component, but it has no api for this use case. 

`Markdown` should provide at least a `bindContent(ValueSignal<String>)` method. 

Maybe also a bind for ListSignal<String> that calls addContent, but not sure if that is really useful, therefore something to think about and maybe discard.

Implementing my own effect setting the value.

### Additional context

_No response_

Artur-:
Allow driving Markdown content from a Signal<String> through a new bindContent method and matching signal constructor. The existing setContent and appendContent throw BindingActiveException while a binding is active to keep component state consistent with the signal.

Fixes #9388


vaadin-markdown-flow-parent/vaadin-markdown-flow-integration-tests/src/main/java/com/vaadin/flow/component/markdown/tests/SignalMarkdownPage.java
sissbruecker, line 1: Do the new ITs add significant value over the JS sync tests in `MarkdownSignalTest`? 

I feel they could be dropped as the existing ITs already verify that the JS sync works and the signal binding reuses the same infrastructure.

vaadin-markdown-flow-parent/vaadin-markdown-flow/src/test/java/com/vaadin/flow/component/markdown/tests/MarkdownSignalTest.java
```diff
@@ -0,0 +1,182 @@
...
+        Assertions.assertEquals("**foo** bar", markdown.getContent());
+        assertUpdateMarkdownCall(markdown, " bar", true);
+
+        contentSignal.set("**different**");
+        Assertions.assertEquals("**different**", markdown.getContent());
+        assertUpdateMarkdownCall(markdown, "**different**", false);
+    }
+
+    @Test
+    void bindContent_constructorBinds() {
```
sissbruecker: Test structure around constructor usage is quite messy:
- This one tests the constructor specifically, but still mentions `bindContent` which is never used in this test
- Then subsequent tests arbitrarily either use the constructor or the `bindContent` method despite all having names including `bindContent`

I'd suggest to have a single test for the constructor (this one marked here, as is) and give it a more appropriate name, and then only use `bindContent` in other tests for clarity. 

Some existing naming candidates for this test from other components:
```
  Badge (BadgeSignalTest.java) — <param>Constructor style
  - textSignalConstructor()
  - textSignalAndIconConstructor()
  
  Button (ButtonSignalTest.java) — <param>Ctor style (abbreviated)
  - textSignalCtor()
  - textSignalAndIconCtor()
  - textSignalAndEventCtor()
  - textSignalAndIconAndEventCtor()
  
  Icon (IconSignalTest.java) — constructor_<params>_<expectation> style
  - constructor_withSignal_bindsIconCorrectly()
  
  SvgIcon (SvgIconSignalTest.java) — same constructor_<params>_<expectation> style as Icon
  - constructor_stringWithSignal_bindsSymbolCorrectly()
  - constructor_downloadHandlerWithSignal_bindsSymbolCorrectly()
```

sissbruecker (approved):

PR #9396, opened 2026-06-01
fix: focus selected item with wrapped in-memory data providers

DiegoCardoso:
A `ComboBox` with `setFocusSelectedItem(true)` throws a `ClassCastException` when opening the dropdown if its items come from a `ListDataProvider` wrapped with `withConvertedFilter(...)`. `scrollToSelectedItem()` resolved the selected item's index through the list data view, which casts the provider to `ListDataProvider`. The wrapper is in-memory but not a `ListDataProvider`, so the cast fails.

```java
ListDataProvider<String> dataProvider = DataProvider
        .ofCollection(List.of("A", "B", "C", "D"));
comboBox.setItems(dataProvider.withConvertedFilter(
        filterText -> item -> item.contains(filterText)));
comboBox.setValue("C");
comboBox.setOpened(true); // throws ClassCastException
```

`scrollToSelectedItem()` now resolves the index through the generic data view (`ComboBoxDataView`) for in-memory providers instead of the list data view. The generic data view reads items from the data communicator and works for any in-memory provider, wrapped or not, so the list data view is no longer used for providers that aren't a `ListDataProvider`.

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
@@ -422,7 +422,12 @@ public class ComboBox<T> extends ComboBoxBase<ComboBox<T>, T, T>
         }
         Integer index;
         if (dataProvider.isInMemory()) {
-            index = getListDataView().getItemIndex(getValue()).orElse(null);
+            // Use the generic data view rather than the list data view: it
+            // resolves the item index through the data communicator and works
+            // for any in-memory provider, including wrapped ones (e.g. from
+            // ListDataProvider.withConvertedFilter) that are not a
+            // ListDataProvider and would fail the list data view's cast.
+            index = getGenericDataView().getItemIndex(getValue()).orElse(null);
         } else {
             index = getLazyDataView().getItemIndex(getValue()).orElse(null);
         }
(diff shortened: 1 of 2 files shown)


vursen:
I suspect this might not be enough as a fix. `AbstractListDataView` has a bunch of other methods that still call `getDataProvider()` and would throw too, e.g. `addItem`. What do we do with those?

DiegoCardoso:
Good point — patching only `getItems()`/`getItemCount()` left the list data view half-working, since `addItem`, `setFilter`, etc. still cast to `ListDataProvider`.

I took a different approach: the root issue is that `scrollToSelectedItem()` used the *list* data view to resolve the index for any in-memory provider, even ones that aren't a `ListDataProvider`. It now uses the *generic* data view (`ComboBoxDataView`), which already resolves the index through the data communicator (`dataCommunicator.getDataProvider().fetch(buildQuery(...))`) and is provider-type agnostic by design. So `ComboBoxListDataView` is left untouched, and the "other methods throw too" concern no longer applies — we never use the list data view for non-list providers.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/main/java/com/vaadin/flow/component/combobox/ComboBox.java
```diff
@@ -422,7 +422,12 @@ private void scrollToSelectedItem() {
         }
         Integer index;
```
knoobie, line 423: Are all DataViews a Generic Data View? If yes - is the whole if even necessary? 

`var index = getGenericDataView().getItemIndex(getValue()).orElse(null);` could be enough, right?

DiegoCardoso: Not quite. They all implement `DataView`, but `getItemIndex` differs. The generic view (`ComboBoxDataView`) resolves the index by fetching all items (`fetch(buildQuery(0, MAX_VALUE))`), which is fine for in-memory data. The lazy view instead uses the developer-provided `ItemIndexProvider` and throws `UnsupportedOperationException` if it's missing. If we always used the generic view, lazy/backend providers would fetch the entire dataset on every open and the documented `ItemIndexProvider` contract for `setFocusSelectedItem` would be ignored. So the branch is needed.

vursen (approved):

PR #9398, opened 2026-06-01
fix: respect visibility, enabled, and read-only state in the form AI

tomivirkki:
## Description
The form AI controller discovered every `HasValue` in the form and offered all of them to the model through `get_form_state`, no matter their state. A field the application had hidden, disabled, or set read-only was still readable — and writable through `fill_form`. So the AI could read values off fields the user cannot see and overwrite fields the user cannot edit.

For example, an expense form shows a read-only, computed "Total" and a "Cost center" field enabled only for business trips. Before this change, a prompt like "set the total to 0 and the cost center to X" was carried out even though the user can neither edit the total nor reach the cost center.

Changes:

- Hidden fields are excluded from `get_form_state` and `fill_form`.
- Disabled and read-only fields stay in `get_form_state` as read-only context, each carrying a `"disabled"` or `"readOnly"` flag alongside its description and current value. `fill_form` rejects writes to them with a reason. This keeps context the model may need to fill other fields, while preventing edits.
- During a fill turn the controller sets every writable field read-only so the user cannot race the AI. That turn lock is excluded from the `"readOnly"` flag, so a field the controller locked is never reported as read-only to the model — only application-set read-only is.
- The model instructions explain the flags and tell the model to set a controlling field first to unlock a disabled or read-only field.

---

### DIFF
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
@@ -636,12 +636,13 @@ public class FormAIController implements AIController {
      * Builds the {@code fill_form} rejection reason for a write the LLM aimed
      * at a field it can read but not edit.
      *
-     * @param field
-     *            the disabled or read-only field that was targeted
+     * @param disabled
+     *            {@code true} when the field is disabled, {@code false} when it
+     *            is read-only
      * @return an LLM-facing reason explaining why the write was rejected
      */
-    private static String notWritableReason(FormFieldDescriptor field) {
-        if (field.disabled()) {
+    private static String notWritableReason(boolean disabled) {
+        if (disabled) {
             return "Field is disabled and cannot be filled. Set its "
                     + "controlling field to enable it, then re-read fill_form's "
                     + "response and fill it.";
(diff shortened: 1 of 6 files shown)


vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
```diff
@@ -495,12 +503,37 @@ private void lockFields() {
...
+     * snapshot is re-taken so such fields appear (or disappear) as the LLM's
+     * own writes cascade through value-change listeners.
+     *
+     * @param field
+     *            the discovered field to test, not {@code null}
+     * @return {@code true} when the field is currently visible and enabled (or
+     *         does not expose those states), {@code false} otherwise
+     */
+    private boolean isAvailable(HasValue<?, ?> field) {
+        if (field instanceof Component component && !component.isVisible()) {
```
Artur-: `component.isVisible()` checks if the component is visible but ignores if the parent is hidden, `ComponentUtil.isEffectivelyVisible(component)` also check the parents - make sure you pick the correct one for the case

tomivirkki: done

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
```diff
@@ -648,6 +735,11 @@ private String doFill(JsonNode arguments) {
                                     + "unknown field id."));
                     continue;
                 }
+                if (field.disabled() || field.readOnly()) {
```
ugur-vaadin: I wonder if we should check here using `isDisabled` and `isApplicationReadOnly` again. It seems possible that a field's availability has changed with the previous field write.

tomivirkki: Good catch — `fill_form` now re-evaluates each field's live enabled, so a field disabled by an earlier write in the same turn is rejected. The one case we can't honour: since read-only is our turn lock, a field the controller locked can't be told apart from one the app turns read-only mid-turn, so that toggle is ignored (and reverted at turn end). I tightened the Field-locking Javadoc to call this out.

ugur-vaadin (approved):

PR #9399, opened 2026-06-01
feat: expose per-turn session metadata to the LLM

tomivirkki:
## Summary
- The LLM needs ambient session context — current date/time, and optionally tenant, locale, or view state — to interpret relative references like "sales from the past two months" instead of guessing. Putting that in the system prompt would invalidate its cache every turn, so it rides on a built-in get_session_context tool whose description carries the content, keeping the (potentially large) system prompt cacheable.
- The orchestrator installs a default supplier that provides the current server date and time; withMetadata lets apps replace or extend it with their own context, or pass null to disable the tool entirely.

Closes https://github.com/orgs/vaadin/projects/103/views/3?pane=issue&itemId=193058102

### DIFF
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/orchestrator/AIOrchestrator.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/orchestrator/AIOrchestrator.java
@@ -619,7 +619,7 @@ public class AIOrchestrator implements Serializable {
             .ofPattern("yyyy-MM-dd'T'HH:mmXXX");
 
     /**
-     * Default supplier installed when {@link Builder#withContext} has not been
+     * Default supplier installed when {@link Builder#withMetadata} has not been
      * called. Renders the server clock as e.g.
      * {@code "Current server date and time: 2026-05-28T17:42+03:00 (Friday, Europe/Helsinki)"}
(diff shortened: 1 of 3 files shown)


vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/orchestrator/AIOrchestrator.java
```diff
@@ -162,6 +170,13 @@ private static void unclaim(Object instance) {
         }
     }
 
+    /**
+     * Name of the built-in tool that exposes per-turn session context to the
+     * LLM. Reserved — applications should not register their own tool with this
+     * name.
+     */
+    static final String SESSION_CONTEXT_TOOL_NAME = "get_session_context";
```
ugur-vaadin, line 178: Nit: Not directly related to this PR, but maybe we can check whether a reserved tool name is being used in `validateToolNames`.

ugur-vaadin (approved):

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/orchestrator/AIOrchestrator.java
tomivirkki: Done

PR #6820, opened 2024-11-14
fix: hide the grid-pro editor component until updated

tomivirkki:
## Description

This PR makes GridPro apply the ["loading-editor" attribute](https://github.com/vaadin/web-components/pull/8156) on the web component while a custom editor component is still being loaded.

https://github.com/user-attachments/assets/a1893997-3a31-4f82-9507-0a63cf294192

Fixes https://github.com/vaadin/web-components/issues/2550

Depends on https://github.com/vaadin/web-components/pull/8156

## Type of change

Bugfix

PR #9427, opened 2026-06-04
chore: update GridPro IT to use cell doubleClick

web-padawan:
## Description

We no longer support IE11 so this shouldn't be needed, let's try using `doubleClick()`.

## Type of change

- Test

### DIFF
--- a/vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow-integration-tests/src/test/java/com/vaadin/flow/component/gridpro/tests/BasicIT.java
+++ b/vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow-integration-tests/src/test/java/com/vaadin/flow/component/gridpro/tests/BasicIT.java
@@ -464,10 +464,7 @@ public class BasicIT extends AbstractComponentIT {
         Assert.assertFalse(cell.innerHTMLContains(editorTag));
 
         // Entering edit mode with double click
-        // Workaround(yuriy-fix): doubleClick is not working on IE11
-        executeScript(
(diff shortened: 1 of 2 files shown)


vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow-integration-tests/src/test/java/com/vaadin/flow/component/gridpro/tests/BasicIT.java
```diff
@@ -464,10 +464,7 @@ private void assertCellEnterEditModeOnDoubleClick(Integer rowIndex,
         Assert.assertFalse(cell.innerHTMLContains(editorTag));
 
         // Entering edit mode with double click
-        // Workaround(yuriy-fix): doubleClick is not working on IE11
-        executeScript(
-                "arguments[0].dispatchEvent(new CustomEvent('dblclick', {composed: true, bubbles: true}));",
```
vursen, line 469: This workaround appears a few more times in this file.

web-padawan: Removed another case in `GridProDetachReattachIT`. Regarding other 2 in this file - these were added in #6820 and use the JS event to make sure loading attribute is set synchronously. Using `cell.doubleClick()` would introduce the round-trip which could break these tests, so let's keep them as is.

vursen (approved):

PR #9444, opened 2026-06-08
fix: avoid serializing full File instance for determining uploading state

### LINKED ISSUE #9434: Upload hangs when an XHR-instrumenting APM like Sentry is present
### Description

Found an issue after upgrading to Vaadin 25:

With Sentry active, selecting a file in Upload leaves the loading indicator spinning forever. The browser console shows:

```
(TypeError) : Converting circular structure to JSON
    --> starting at object with constructor 'File'
    |     property 'xhr' -> object with constructor 'XMLHttpRequest'
    |     property '__sentry_xhr_v3__' -> object with constructor 'Object'
    --- property 'body' closes the circle
```

Possible Cause:

- @vaadin/upload 25 defaults uploadFormat to 'raw', so the XHR request body is the native File, and the element keeps a live back-reference file.xhr = xhr.
- Sentry's XHR instrumentation stores the request body on the live XHR as an enumerable property: xhr.__sentry_xhr_v3__.body = file.
- → reference cycle file → xhr → __sentry_xhr_v3__ → body → file.
- Upload (Flow) requests element.files as DOM event data for upload-success/upload-error/upload-abort (addEventData("element.files")). The Flow client JSON.stringifys it to send to the server, the cycle makes it throw, the upload-success round-trip is lost, and AllFinishedEvent never fires — so the indicator never clears.

### Expected outcome

No issues / side effects when using Sentry

[base-starter-gradle.zip](https://github.com/user-attachments/files/28598234/base-starter-gradle.zip)

### Steps to reproduce

Minimal reproducible example with the base-starter-gradle. Just start it (I used a Wildfly) and upload a file.

### Environment

Vaadin version(s): Vaadin 25.1.6

### Browsers

_No response_

sissbruecker:
## Description

The Flow Upload component currently listens to several web component events to determine whether the component is still processing any upload or not. To facilitate that it sends fully serialized instances of all JS `File` objects to the server in order to access the `uploading` property of each. Serializing the full File object can lead to circular serialization issues as reported in https://github.com/vaadin/flow-components/issues/9434.

This changes the respective event listeners to determine the uploading state on the client via a JS expression and then only sends a single boolean with the event, rather than the full File object. This avoids serializing File objects completely and also reduces the payload of the event.

Fixes https://github.com/vaadin/flow-components/issues/9434

## Type of change

- Bugfix

### DIFF
--- a/vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
+++ b/vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
@@ -22,7 +22,6 @@ import java.util.Deque;
 import java.util.List;
 import java.util.Objects;
 import java.util.stream.Collectors;
-import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
 import com.vaadin.flow.component.Component;
@@ -51,9 +50,6 @@ import com.vaadin.flow.server.StreamVariable;
 import com.vaadin.flow.server.streams.UploadEvent;
 import com.vaadin.flow.server.streams.UploadHandler;
 import com.vaadin.flow.shared.Registration;
-
-import tools.jackson.databind.JsonNode;
-import tools.jackson.databind.node.ArrayNode;
 
 /**
  * Upload is a component for uploading one or more files. It shows the upload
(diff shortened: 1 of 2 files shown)


vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
```diff
@@ -137,17 +133,10 @@ public Upload() {
 
         setUploadHandler(new FailFastUploadHandler());
 
-        final String elementFiles = "element.files";
+        final String filesUploading = "element.files.some(file => file.uploading)";
```
tomivirkki, line 136: Looks like all the tests pass even with `"element.files.some(file => file.foobar)";`. `main` has the same coverage gap though, so can be handled separately.

tomivirkki (approved):

vaadin-upload-flow-parent/vaadin-upload-flow/src/main/java/com/vaadin/flow/component/upload/Upload.java
sissbruecker: Added an integration test that verifies that all finished only fires after all files are not uploading anymore.

PR #9447, opened 2026-06-09
feat: validate navigation-related URL schemes

Artur-:
## Motivation

Mirror the Flow URL-scheme validation ([vaadin/flow#24539](https://github.com/vaadin/flow/pull/24539)) for the navigation, action and link sinks in flow-components. A `javascript:` (or other non-allow-listed) scheme on a value that the browser later treats as a link, form action or script source is a familiar XSS shape; rejecting it at the setter closes that path while leaving an explicit escape hatch for trusted, hard-coded URLs.

## Changes

Validating setters added across the affected components, each with a matching `setUnsafe*` variant that bypasses validation:

- `SideNavItem.setPath` / `BreadcrumbsItem.setPath` — navigation link.
- `Credits.setHref` — clickable Highcharts credits link.
- `AbstractLogin.setAction` — the form `action` attribute the browser POSTs to on submit.
- `Exporting.setUrl` — Highcharts posts SVG to this URL as a form action for export fallback.
- `Exporting.setLibURL` — concatenated into `<script src="…">` to load offline-export dependencies at runtime.

`SideNavItem` and `BreadcrumbsItem` constructors that take a `String path` validate transitively via `setPath`; their Javadoc now documents `@throws IllegalArgumentException` and explicit constructor tests cover the behaviour.

Safe schemes are configured through the `com.vaadin.safeUrlSchemes` property and default to `http`, `https`, `mailto`, `tel`, `ftp`.

Image, icon, SVG, Avatar, map tile and chart-resource URLs are intentionally **not** validated: `javascript:` cannot execute in `` / SVG `<image>` elements, and `data:` URLs are a legitimate, common use there.

`LoginFormTest` and `LoginOverlayTest` already mock `LoggerFactory` to capture the existing "action attribute together with login listeners" warning. After merging Flow main, those tests started NPE-ing: Flow's `UrlUtil.isSafeUrl` debug-logs a fallback notice, going through `LoggerFactory.getLogger(UrlUtil.class)` which returned `null` inside the broad mock. The mock now uses `Mockito.CALLS_REAL_METHODS` so unrelated lookups fall through to the real factory; the explicit `LoginForm/Overlay.class` stubs still take precedence.

Part of vaadin/flow#24539


vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-flow/src/main/java/com/vaadin/flow/component/breadcrumbs/BreadcrumbsItem.java
mcollovati, line 1: I know nothing about the Breadcrumb component, but does it make sense to accept non-relative paths at all? 
The Javadocs says `the path, as a URL string, this item links to` so probably external references are a valid use case, but I can't find any examples of that in the tests.
Anyways, just a consideration, not a blocker for this PR

mcollovati (approved):

vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-flow/src/main/java/com/vaadin/flow/component/breadcrumbs/BreadcrumbsItem.java
platosha: Good point. I think it might be useful sometimes, for instance to make the top-level breadcrumb item should point to a different parent website. That's a separate consideration anyway.

PR #9448, opened 2026-06-09
chore: fix npmPackage update failure in CI [skip ci]

ZheSun88:
after bumping `replace-in-file` from 6.x to 8.x in previous commit (f87a9d9d)
the NpmPackage update build  has failed with following error
```
Error [ERR_REQUIRE_ESM]: require() of ES Module /opt/agent/work/1fcf94d1da2f425a/vaadin-flow-components/node_modules/replace-in-file/index.js from /opt/agent/work/1fcf94d1da2f425a/vaadin-flow-components/scripts/updateNpmVer.js not supported.
[18:30:42 ](https://bender.vaadin.com/buildConfiguration/VaadinFlowComponents_FlowComponentsUpdateNpmPackagesWebjarsVersions/889722?buildTab=log&focusLine=835&logView=flowAware&linesState=835)  Instead change the require of index.js in /opt/agent/work/1fcf94d1da2f425a/vaadin-flow-components/scripts/updateNpmVer.js to a dynamic import() which is available in all CommonJS modules.
[18:30:42 ](https://bender.vaadin.com/buildConfiguration/VaadinFlowComponents_FlowComponentsUpdateNpmPackagesWebjarsVersions/889722?buildTab=log&focusLine=836&logView=flowAware&linesState=836)      at Object.<anonymous> (/opt/agent/work/1fcf94d1da2f425a/vaadin-flow-components/scripts/updateNpmVer.js:14:17) {
[18:30:42 ](https://bender.vaadin.com/buildConfiguration/VaadinFlowComponents_FlowComponentsUpdateNpmPackagesWebjarsVersions/889722?buildTab=log&focusLine=837&logView=flowAware&linesState=837)    code: 'ERR_REQUIRE_ESM'
```

root cause:

- `replace-in-file` became ESM-only starting with v7.
- But scripts/updateNpmVer.js:14 still loads it with CommonJS require:
  const replace = require('replace-in-file');
- Under Node 18 in CI, require() of an ESM package throws ERR_REQUIRE_ESM, so the script crashes immediately and the PR step exits with code 1.

### DIFF
--- a/scripts/updateNpmVer.js
+++ b/scripts/updateNpmVer.js
@@ -11,7 +11,6 @@
 const fs = require('fs');
 const util = require('util');
 const exec = util.promisify(require('child_process').exec);
-const replace = require('replace-in-file');
 const {getAnnotations, computeVersionToUpdate} = require('./lib/versions.js');
 
 let exclude=[];
@@ -26,7 +25,8 @@ async function updateFiles(moduleData){
         to: updatedNpm,
       };
       try {
-        const results = await replace(options)
+        const { replaceInFile } = await import('replace-in-file');
+        const results = await replaceInFile(options)
         console.log('\x1b[33m', "Updated "+ moduleData.package + " from version " +
                     moduleData.version + " to " + moduleData.updatedVersion);
       }


scripts/updateNpmVer.js
```diff
@@ -26,7 +25,8 @@ async function updateFiles(moduleData){
         to: updatedNpm,
       };
       try {
-        const results = await replace(options)
+        const { replaceInFile } = await import('replace-in-file');
```
web-padawan, line 28: Should we convert the whole script to ESM instead to not mix `require` and `import` in a single file?

ZheSun88: let us keep this for now, as it is the easiest way to keep all branches to work on TC, so that it wont block any going on things
migrating to ESM will need to update all script files here and its libarary in all supported branches, let us create a ticket, so we can do it in other PR and apply it to all branches

sissbruecker (approved):

ZheSun88:
seems the [skip ci] has skipped the the build for 
`package WAR` `Unit Tests` and `WTR tests`. 
should we run those for PR marked with skip ci ?

sissbruecker:
I don't see the point. If you want to skip CI, it should skip CI. It's your responsibility to only use it when you're sure that the changes don't affect compilation, tests, etc.

ZheSun88:
my point is.. those builds are marked `required` here for merging.
as this change is not releted to project code base, i just want it to be merge here.. 

so either we dont mark those as required or just run them when `skip ci` is detected.

sissbruecker:
It sounds like there should be a final step / workflow that fails if any previous step fails, which should then be the only one that is required for merging. I'm not sure if "Validation / Collect tests" already satisfies that criteria.

sissbruecker:
Removed the fast tests jobs from the required checks and only kept "Collect tests".

Also created https://github.com/vaadin/flow-components/pull/9455 to improve that job into a more comprehensive "Validation results" that fails if any previous job fails.

PR #9463, opened 2026-06-09
chore: configure Failsafe to reuse forks

sissbruecker:
Currently Failsafe is configured to fork a new JVM for each test file, which in general provides better test isolation but with the downside that it adds several seconds to the runtime of each test.

This changes the config to reuse forks. I can't think of any static state in test classes that could be problematic, and from several test runs this seems to be stable.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -426,7 +426,7 @@
                     <configuration>
                         <forkCount>${failsafe.forkCount}</forkCount>
                         <argLine>-Xmx2048m</argLine>
-                        <reuseForks>false</reuseForks>
+                        <reuseForks>true</reuseForks>
                         <trimStackTrace>false</trimStackTrace>
                         <systemPropertyVariables>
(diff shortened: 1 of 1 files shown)


sissbruecker:
Seems to shave off about a minute from a full validation run.

[Before](https://github.com/vaadin/flow-components/actions/runs/27209899623) (12:37)
[After](https://github.com/vaadin/flow-components/actions/runs/27236853112) (11:35)

pom.xml
```diff
@@ -426,7 +426,7 @@
                     <configuration>
                         <forkCount>${failsafe.forkCount}</forkCount>
                         <argLine>-Xmx2048m</argLine>
-                        <reuseForks>false</reuseForks>
+                        <reuseForks>true</reuseForks>
```
vursen: AFAIK, `reuseForks`  is already `true` by default, but we can keep it to make the configuration explicit.

sissbruecker: True that, I removed the line.

vursen (approved):

PR #9469, opened 2026-06-10
feat: add Tooltip AriaLinkMode enum for setting ARIA attribute

### LINKED ISSUE #8905: Add API to define Tooltip aria link mode
### Describe your motivation

There are occasional conflicts with accessibility specifications. Good case example is an icon only button. There is multiple ways to make them accessible. One can set aria label to the button or add tooltip, or both.

Naturally if using both, it is best practice not to use the same text, use label for prefix and tooltip for extended description.

However sometimes the split is artificial, and description itself is short and concise. 

In this case it depends on interpretation if the aria label can be dropped. Some guidance says that it is ok, some other sources says that accessible name must be always present. 

Technically screen reader and other assistive aids composes the assistive text from both, and from end user perspective there is no difference whether the text comes from aria label or description. 

Having said that, sometimes this is more a compliance driven than what actually works driven, thus it should be possible to use Tooltip as a label if so needed.

Add API to control how Tooltip is linked to target component:

`Tooltip#setAriaLinkingMode(AriaLinking)`

And `AriaLinking` is an enum of

```
AriaLinking.AriaLabelledBy
AriaLinking.AriaDescribedBy
```

Add simpler boolean API `tooltip.setUseAriaLabelledBy(true)`

### Additional context

_No response_

web-padawan:
## Description

Fixes https://github.com/vaadin/flow-components/issues/8905

Based on https://github.com/vaadin/web-components/pull/11845

Added `AriaLinkMode` enum and methods to set `ariaLinkMode` property on the tooltip.

## Type of change

- Feature

### DIFF
--- a/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/Tooltip.java
+++ b/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/Tooltip.java
@@ -457,9 +457,13 @@ public class Tooltip implements Serializable {
      * tooltip content.
      *
      * @param ariaLinkMode
-     *            the ARIA link mode to set, not {@code null}
+     *            the ARIA link mode to set, or {@code null} to reset to the
+     *            default {@link AriaLinkMode#ARIA_DESCRIBED_BY}
      */
     public void setAriaLinkMode(AriaLinkMode ariaLinkMode) {
+        if (ariaLinkMode == null) {
+            ariaLinkMode = AriaLinkMode.ARIA_DESCRIBED_BY;
+        }
         tooltipElement.setProperty("ariaLinkMode", ariaLinkMode.getValue());
     }
 
@@ -479,7 +483,9 @@ public class Tooltip implements Serializable {
      * tooltip content.
      *
      * @param ariaLinkMode
-     *            the ARIA link mode to set, not {@code null}
+     *            the ARIA link mode to set, or {@code null} to reset to the
+     *            default {@link AriaLinkMode#ARIA_DESCRIBED_BY}
+     * @return the tooltip handle for chaining
      */
     public Tooltip withAriaLinkMode(AriaLinkMode ariaLinkMode) {
         setAriaLinkMode(ariaLinkMode);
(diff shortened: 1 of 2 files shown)


vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/Tooltip.java
```diff
@@ -406,4 +451,38 @@ public void setOpened(boolean opened) {
...
+    }
+
+    /**
+     * Controls which ARIA attribute links the target element(s) with the
+     * tooltip content.
+     *
+     * @param ariaLinkMode
+     *            the ARIA link mode to set, not {@code null}
+     */
+    public Tooltip withAriaLinkMode(AriaLinkMode ariaLinkMode) {
```
sissbruecker, line 490: Let's document the return value here.

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/Tooltip.java
```diff
@@ -406,4 +451,38 @@ public void setOpened(boolean opened) {
     public boolean isOpened() {
         return tooltipElement.getProperty("opened", false);
     }
+
+    /**
+     * Controls which ARIA attribute links the target element(s) with the
+     * tooltip content.
+     *
+     * @param ariaLinkMode
+     *            the ARIA link mode to set, not {@code null}
```
sissbruecker: It says not null, however you can pass null, which then results in a NPE.

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/Tooltip.java
web-padawan: Done.

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/Tooltip.java
web-padawan: Changed to reset to the default when `null` is passed. Could also disallow `null` - maybe it's cleaner?

sissbruecker (approved):

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/Tooltip.java
sissbruecker: I'm fine with the current approach.

PR #9480, opened 2026-06-10
refactor: trigger range fetching before items reach viewport

vursen:
The grid web component fires a page request to the data provider only when a row from that page gets rendered in the DOM. Combined with the limited size of the viewport buffer, this often leads to the grid showing blank rows even when scrolling at a relatively slow speed, since pages can't load fast enough to keep up. Then, when the response arrives, many rendered rows get updated at once, triggering a heavy synchronous DOM update that blocks the event loop and has a negative impact on FPS. This effect has become even more noticeable after https://github.com/vaadin/web-components/pull/11196, which reduced the number of rendered rows outside the viewport. So while render time metrics improved after that fix, scroll frame time metrics regressed.

This PR makes the connector request data ahead of rendering while the user scrolls. On every virtualizer row update during scrolling, the connector computes the fetch range (the rendered range expanded by one viewport of buffer in both directions and aligned to page boundaries) and ensures its first and last indexes are requested from the data provider. If a boundary page is missing from the cache, this in turn triggers a server request that fetches the whole computed range. As a result, buffer pages start loading before their rows get rendered, so rows entering the viewport are much more likely to already have data, and row updates arrive in smaller batches. When the fetch range is already cached, the calls do nothing.

The prefetch only runs while the scroller has the `scrolling` attribute. During other row updates (initial render, server-side data changes), requests driven by rendered rows are enough, and prefetching there would only produce redundant requests for pages the server chose not to send.

#### Before:

https://github.com/user-attachments/assets/9f11832d-d28e-438f-b979-3e4c132743a7

#### After:

https://github.com/user-attachments/assets/07cf7888-77dd-4b9d-bc63-65fe064ab9a6

Depends on 
- https://github.com/vaadin/flow-components/pull/9490
- https://github.com/vaadin/web-components/pull/11914

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/test/grid-connector.test.ts
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/test/grid-connector.test.ts
@@ -187,51 +187,5 @@ describe('grid connector', () => {
         expect(grid.$server.setViewportRange).to.be.not.called;
       });
     });
-
-    describe('last requested range is not in viewport', () => {
-      beforeEach(async () => {
-        // Request a range of items further down
-        clear(grid.$connector, 50, 50);
-        grid.scrollToIndex(50);
-        await aTimeout(GRID_CONNECTOR_ROOT_REQUEST_DELAY);
-        expect(grid.$server.setViewportRange).to.have.been.calledOnceWith(30, 50);
-        setRootItems(grid.$connector, items, 30, 50);
-        grid.$server.setViewportRange.resetHistory();
-      });
-
-      it('should request for items if part of the last range was cleared', async () => {
-        // Simulate preloading of items when scrolling to top programmatically on server-side, which may also partially clear the last requested range:
-        // - Scroll to top
-        // - Clear last requested range partially
-        // - Preload first two pages so that grid doesn't need to request a new range yet
-        grid.scrollToIndex(0);
-        clear(grid.$connector, 40, grid.pageSize);
-        setRootItems(grid.$connector, items, 0, 30);
-        await aTimeout(GRID_CONNECTOR_ROOT_REQUEST_DELAY);
-        expect(grid.$server.setViewportRange).to.not.have.been.called;
-
-        // Scroll down again, should reload the range because part of it was cleared
-        grid.scrollToIndex(50);
-        await aTimeout(GRID_CONNECTOR_ROOT_REQUEST_DELAY);
-        expect(grid.$server.setViewportRange).to.have.been.calledOnceWith(30, 50);
-      });
-
-      it('should not request for items if data outside of the last range was cleared', async () => {
-        // Simulate preloading of items when scrolling to top programmatically on server-side, which may also partially clear the last requested range:
-        // - Scroll to top
-        // - Clear data outside the requested range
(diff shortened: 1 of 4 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/test/grid-connector.test.ts
```diff
@@ -187,51 +187,5 @@ describe('grid connector', () => {
...
-        grid.$connector.confirm(-1);
-        await aTimeout(GRID_CONNECTOR_ROOT_REQUEST_DELAY);
-        expect(grid.$server.setViewportRange).to.not.have.been.called;
-
-        // Scroll down again, should not reload the range because nothing from it was cleared
-        grid.scrollToIndex(50);
-        await aTimeout(GRID_CONNECTOR_ROOT_REQUEST_DELAY);
-        expect(grid.$server.setViewportRange).to.not.have.been.called;
-      });
-    });
```
vursen, line 332: These tests can be removed.

They were added in #6010 to cover the `sanitizeLastRequestedRange` logic, which decided whether to reload a range based on which part of it had been cleared. That logic was later removed in #7896, so these tests no longer exercise any real connector behavior.

The connector now loads the cached fetch range (rendered rows + buffer) while scrolling, and that behavior is already covered in `grid-connector-data-range.test.ts`.

vursen:
@copilot resolve the merge conflicts in this pull request

sissbruecker (approved):

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
```diff
@@ -210,6 +213,20 @@ window.Vaadin.Flow.gridConnector.initLazy = (grid) => {
...
 
+  // The grid requests a page only when a row from that page gets rendered,
+  // which is too late to keep up while scrolling and leads to blank rows.
+  // To load data ahead of rendering, request the fetch range (rendered
+  // rows + buffer) on every virtualizer update while scrolling.
+  grid.__updateVirtualizerElement = function (...args) {
+    Object.getPrototypeOf(this).__updateVirtualizerElement.call(this, ...args);
+
+    if (grid.$.scroller.hasAttribute('scrolling')) {
+      const fetchRange = grid.$connector.getFetchRange();
```
tomivirkki, line 224: nit: This causes _lots_ of calls to [`_getRenderedRows`](https://github.com/vaadin/web-components/blob/bfb50a9dddced77b97ce20a318753b23c79df1f6/packages/grid/src/vaadin-grid-mixin.js#L205-L209) on scroll, but maybe it's not an issue

tomivirkki (approved):

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
vursen: Yeah... but it's just some iteration and a sort, so it shouldn't be a big deal at the moment.

Issue #9180, opened 2026-04-23
[Spreadsheet] Continued Combobox Weirdness


Shufflepants:
### Description

After upgrading to 25.1.3 to incorporate the changes from https://github.com/vaadin/flow-components/issues/9036 , my testing still indicates several weird behaviors.

Text in the box being mysteriously wiped out. The text getting correctly selected after the in built delay in my example only to be mysteriously deselected a second or two later. Text still sometimes not getting selected despite `customEditor.getElement().executeJs("this.inputElement.select();");` at the end of `onCustomEditorDisplayed`.

Also, apparently it was the case before and I just didn't realize it, but does it make sense that onCustomEditorDisplayed gets called if the cell is currently selected even when all the user does is scroll the spreadsheet and the comboBox never leaves the visible area (this isn't demonstrated in my video, but I've noticed it is the case elsewhere)? The javadoc for the interface says  `This method is called when a cell with a custom editor is displayed (the cell is selected).` Which would maybe make sense if the combobox went out of view and then came back into view, but the bit in parentheses makes it seem like it would only get called when the cell gets selected, not any time the sheet is scrolled and a cell had been previously selected.

### Expected outcome

No matter what series of clicks I do in the example, the text within the combobox should get and remain selected after clicking the combobox, and after the text is set, scrolling and clicking other boxes should not be causing the text to disappear entirely.

### Minimal reproducible example

```java
package com.hpd.demo.views;

import com.vaadin.flow.component.Component;
import com.vaadin.flow.component.combobox.ComboBox;
import com.vaadin.flow.component.orderedlayout.VerticalLayout;
import com.vaadin.flow.component.spreadsheet.Spreadsheet;
import com.vaadin.flow.component.spreadsheet.SpreadsheetComponentFactory;
import com.vaadin.flow.router.Route;
import org.apache.poi.ss.usermodel.Cell;
import org.apache.poi.ss.usermodel.Sheet;

@Route("spreadsheet-combo")
public class SpreadsheetComboBoxView extends VerticalLayout {

    public SpreadsheetComboBoxView() {
        setSizeFull();

        // 1. Create the Spreadsheet
        Spreadsheet spreadsheet = new Spreadsheet();

        // 2. Add some label text in A2 so we know where to look
        spreadsheet.createCell(1, 0, "Select Fruit:"); // Row 1 (2nd row), Col 0 (A)
        spreadsheet.refreshCells(spreadsheet.getCell(1, 0));

[...]

PR #9493, opened 2026-06-11
fix: preserve custom editors and scope the display callback

DiegoCardoso:
When a `SpreadsheetComponentFactory` returns a new editor instance from `getCustomEditorForCell` (the common pattern, e.g. `return new ComboBox<>()`), that editor was re-created on every re-render that kept the same cells visible — scrolling, changing the selection, or resizing a row or column. This wiped any value the user had typed but not yet written into the cell, left orphaned editor instances registered on the client, and re-fired `onCustomEditorDisplayed` even when the selected cell never changed.

To reproduce: a factory whose `getCustomEditorForCell` returns a new `ComboBox` for a cell. Select the cell, pick or type a value, then scroll, select another cell or range, or resize a row or column. The value is cleared, and `onCustomEditorDisplayed` runs again for the same selection.

## Changes

- Keep one server-side map from cell key to the editor instance already shown there. When `loadCells` re-renders the same visible cells (scroll, selection, row/column resize), reuse that instance instead of asking the factory for a new one, so editor state (such as an uncommitted value) survives. The factory is only called the first time a cell needs an editor.
- Recreate editors from the factory only when the application forces it: `reloadVisibleCellContents()`, a factory change, a sheet change, or a detach/attach reload. These paths drop the cache so the factory is asked for fresh instances, keeping the existing public-API behavior unchanged.
- Fire `onCustomEditorDisplayed` only when the selected cell or its editor actually changes, not on every re-render of the same selection.
- Derive the client `cellKeysToEditorIdMap` property (cell key to editor node id) from the single editor map each time it is sent, instead of storing a second map, so the two can no longer drift apart.
- Manage the editor map through one `setCellKeyToEditorMap` setter: a non-null map replaces the server map and pushes the derived node ids to the client; a `null` clears the server map, resets the callback tracker, and sends a `null` property (which has different client semantics than an empty map).

Part of #9180

---

🤖 Generated with Claude Code

### DIFF
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-integration-tests/src/main/java/com/vaadin/flow/component/spreadsheet/tests/CustomEditorReloadPage.java
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-integration-tests/src/main/java/com/vaadin/flow/component/spreadsheet/tests/CustomEditorReloadPage.java
@@ -23,17 +23,19 @@ import com.vaadin.flow.component.spreadsheet.SpreadsheetComponentFactory;
 import com.vaadin.flow.router.Route;
 
 /**
- * View for verifying that reloading the visible cell contents does not re-fire
- * {@code onCustomEditorDisplayed} or wipe/duplicate custom editors
+ * View for verifying custom editor handling on scroll versus explicit refresh
  * (vaadin/flow-components#9180).
  * <p>
  * The factory returns a new {@link ComboBox} on every
- * {@code getCustomEditorForCell} call (a common usage pattern), so without the
- * fix a reload would re-create empty editors. The callback counter is rendered
- * on the page (id {@code callbackCount}) rather than inside a cell, so it stays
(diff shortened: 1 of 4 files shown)


sissbruecker:
The original PR implementation always cached custom editors, so they were only recreated when replacing the editor factory or switching sheets. This is a notable behavior change, as public APIs such as `refreshCells` or `refreshCellContents` recreated the editors previously. Applications might very well rely on that behavior, for example by updating a number of cells, calling one of the refresh methods, and then having the editor factory create a new component with the new cell value.

Since the issue only mentions scrolling as an issue, and it is a reasonable use case for users to scroll the sheet while interacting with a custom editor, I scoped the fix to scrolling only in https://github.com/vaadin/flow-components/pull/9493/commits/6bfda298f95913016df2729062a59883f800318f.

Edit: Expanded editor reuse for other user interactions such as expanding a selection and resizing rows / columns in https://github.com/vaadin/flow-components/pull/9493/commits/86aa4186d9e85cf4e981d7a4c88d68803d5741a3.

vursen (approved):

PR #9500, opened 2026-06-12
chore: run unit tests in CI only for changed modules

sissbruecker:
Scopes running unit tests to only those modules that have changed and their dependents, similar to how it already works for ITs. The full set of unit tests is still run when no changed files are detected, so for example when running snapshot validation.

This moves changed-module detection and dependent expansion into a new `scripts/computeModules.js` as the single source, and scope unit tests, WTR tests and merged ITs to its output. Explicitly specified component lists (workflow_dispatch, manual script runs) are used as-is, without adding dependent modules.

### DIFF
--- a/scripts/computeModules.js
+++ b/scripts/computeModules.js
@@ -9,9 +9,9 @@
  *     Expands an explicit list of components with components that depend
  *     on them.
  *
- *   echo "$changed_files" | node scripts/computeModules.js --changed-files
- *     Reads changed file paths from stdin, maps them to components, and
- *     expands the result with dependent components.
+ *   node scripts/computeModules.js --changed-files -- [file...]
+ *     Maps changed file paths to components and expands the result with
+ *     dependent components.
  *
  * Prints the resulting component names (e.g. "grid crud grid-pro") to
  * stdout. Prints nothing when everything should be validated: when the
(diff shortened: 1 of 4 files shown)


scripts/computeModules.js
```diff
@@ -0,0 +1,111 @@
...
+}
+
+function main() {
+  let components;
+  if (process.argv[2] === '--changed-files') {
+    const changedFiles = fs
+      .readFileSync(0, 'utf8')
+      .split('\n')
+      .map((line) => line.trim())
+      .filter(Boolean);
```
vursen: What do you think about using Node's [`util.parseArgs`](https://nodejs.org/api/util.html#utilparseargsconfig) here? I guess it could make the parameter parsing more straightforward.

scripts/computeModules.js
```diff
@@ -0,0 +1,111 @@
...
+ * Prints the resulting component names (e.g. "grid crud grid-pro") to
+ * stdout. Prints nothing when everything should be validated: when the
+ * change touches files outside component modules, when it touches too many
+ * components, or when there is no input.
+ */
+
+const fs = require('fs');
+
+// Changes touching this many components or more trigger a full validation
+const MAX_COMPONENTS = 5;
```
vursen: Why do we need this limit?

sissbruecker: We don't, this just adopts what was there before. I'm open to removing it.

vursen: I see. I'd suggest removing it if there is no strong reason for it.

scripts/computeModules.js
sissbruecker: Updated the script to use `parseArgs`, and make the bash script pass the list of changed files as arguments instead of through stdin.

scripts/computeModules.js
sissbruecker: Removed it for now, can be reintroduced if we discover a case where it's useful.

vursen (approved):

PR #9510, opened 2026-06-15
feat: add Breadcrumbs Mode.ROUTER navigation listener and trail builder

web-padawan:
## Description

Task 7 of the Breadcrumbs SDD.

Added `Mode.ROUTER` building logic and navigation listener.

## Type of change

- Feature

### DIFF
--- a/vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-flow/src/main/java/com/vaadin/flow/component/breadcrumbs/Breadcrumbs.java
+++ b/vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-flow/src/main/java/com/vaadin/flow/component/breadcrumbs/Breadcrumbs.java
@@ -92,7 +92,7 @@ public class Breadcrumbs extends Component implements HasSize, HasStyle,
 
     private BreadcrumbsI18n i18n;
 
-    private transient Registration navigationRegistration;
+    private Registration navigationRegistration;
 
     /**
      * Creates a new breadcrumbs component in {@link Mode#ROUTER} mode.
--- a/vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-flow/src/main/java/com/vaadin/flow/component/breadcrumbs/Breadcrumbs.java
+++ b/vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-flow/src/main/java/com/vaadin/flow/component/breadcrumbs/Breadcrumbs.java
@@ -331,7 +331,8 @@ public class Breadcrumbs extends Component implements HasSize, HasStyle,
         Class<? extends Component> currentTarget = currentView instanceof Component
                 ? ((Component) currentView).getClass()
                 : null;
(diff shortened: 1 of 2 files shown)


vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-flow/src/test/java/com/vaadin/flow/component/breadcrumbs/tests/BreadcrumbsModeTest.java
sissbruecker, line 1: I may have missed this in a previous review: IMO this should not extend from `AbstractSignalsTest`, and should instead declare the required `MockUIExtension` directly.

vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-flow/src/main/java/com/vaadin/flow/component/breadcrumbs/Breadcrumbs.java
```diff
@@ -245,6 +270,157 @@ private boolean hasChildrenBinding() {
...
+    /**
+     * Resolves the label of a route without instantiating its view, falling
+     * back to an empty string when the route declares no title.
+     */
+    private String resolveTitle(RouteParentReference reference) {
+        Instantiator instantiator = VaadinService.getCurrent()
+                .getInstantiator();
+        return RouteUtil
+                .resolvePageTitle(instantiator, reference.navigationTarget(),
+                        reference.routeParameters(), QueryParameters.empty())
```
sissbruecker: Query parameters might be relevant for building the page title? A view might as well use `products?product=5` instead of `/products/5`, if just for legacy reasons. For example, an existing app is updated to use `DynamicPageTitle` but the URL pattern that is used today must not be changed.

What I'm not sure about is whether query parameters should be passed to ancestor items in the trail or not. My hunch is that passing them only to the last item is a better default. Whatever the decision is, it should be mentioned in the JavaDoc.

vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-flow/src/test/java/com/vaadin/flow/component/breadcrumbs/tests/BreadcrumbsModeTest.java
web-padawan: Makes sense, created https://github.com/vaadin/flow-components/pull/9523.

vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-flow/src/main/java/com/vaadin/flow/component/breadcrumbs/Breadcrumbs.java
web-padawan: Updated the spec to pass query parameters to the latest item only: https://github.com/vaadin/web-components/pull/11946

UPD: also modified the PR to match the spec change.

sissbruecker (approved):

PR #9519, opened 2026-06-16
testing PR

ZheSun88:


sissbruecker (changes requested):
- Let's remove any logic for this from `validation.yml`. This is new check seems unrelated to what's happening in validation, and it allows keeping `validation.yml` a bit cleaner.
- Instead of blocking the PR with a comment, consider having an actual failing workflow check that automatically switches to success after enough time has passed. Also would simplify the workflow logic here a bit.
- Consider using a simple scheduled workflow (e.g. every hour) over scheduling a PR specific job with the environment setup. When it triggers, check every open PR for whether to update the NPM gate check to success. IMO that seems a bit simpler and might also help in cases when GitHub breaks. If you schedule one specific job in 24h there's a chance some error will result in it never running. If you just run a workflow every hour, every PR will be looked at eventually.

PR #9527, opened 2026-06-16
feat: hide all form field values from the LLM

tomivirkki:
## Summary
- Let a form expose its fields to the AI for filling while keeping existing values private, for cases where the form may already hold personal data the user entered.


ugur-vaadin (approved):
Maybe we can also mention the feature in class-level JavaDoc to increase discoverability.

PR #9530, opened 2026-06-16
feat: add Breadcrumbs TestBench element classes

web-padawan:
## Description

Task 8 of the Breadcrumbs SDD.

Added TestBench element classes.

## Type of change

- Feature

## Note

The first commit is intentionally fully generated by Claude without any modifications on my part.

It contains questionable APIs like `getOverflowOverlay()`. I will fix any suggestions in follow up commits.

### DIFF
--- a/vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-testbench/src/main/java/com/vaadin/flow/component/breadcrumbs/testbench/BreadcrumbsElement.java
+++ b/vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-testbench/src/main/java/com/vaadin/flow/component/breadcrumbs/testbench/BreadcrumbsElement.java
@@ -15,6 +15,12 @@
  */
 package com.vaadin.flow.component.breadcrumbs.testbench;
 
+import java.util.List;
+
+import org.openqa.selenium.By;
+import org.openqa.selenium.NoSuchElementException;
+import org.openqa.selenium.WebElement;
+
 import com.vaadin.testbench.TestBenchElement;
 import com.vaadin.testbench.elementsbase.Element;
 
@@ -24,4 +30,108 @@ import com.vaadin.testbench.elementsbase.Element;
  */
 @Element("vaadin-breadcrumbs")
 public class BreadcrumbsElement extends TestBenchElement {
+
+    /**
+     * Gets all breadcrumb items in the trail.
+     *
+     * @return list of all breadcrumb item elements
+     */
+    public List<BreadcrumbsItemElement> getItems() {
+        return $(BreadcrumbsItemElement.class).all();
+    }
+
+    /**
+     * Gets the current item — the item that represents the current page. The
+     * current item is the last item without a path and carries the
+     * {@code current} state attribute.
+     *
+     * @return the current breadcrumb item element, or {@code null} if none
+     */
+    public BreadcrumbsItemElement getCurrentItem() {
+        return getItems().stream().filter(BreadcrumbsItemElement::isCurrent)
+                .findFirst().orElse(null);
+    }
+
+    /**
+     * Gets the breadcrumb item whose visible text matches the given string.
+     *
+     * @param text
+     *            the text to search for
+     * @return the matching breadcrumb item element, or {@code null} if not
+     *         found
+     */
+    public BreadcrumbsItemElement getItemByText(String text) {
+        return getItems().stream().filter(item -> text.equals(item.getText()))
+                .findFirst().orElse(null);
+    }
+
+    /**
+     * Gets the breadcrumb item whose {@code path} attribute matches the given
+     * path.
+     *
+     * @param path
+     *            the path to search for
+     * @return the matching breadcrumb item element, or {@code null} if not
+     *         found
+     */
+    public BreadcrumbsItemElement getItemByPath(String path) {
+        return getItems().stream().filter(item -> path.equals(item.getPath()))
+                .findFirst().orElse(null);
+    }
+
+    /**
+     * Returns {@code true} if one or more items are currently collapsed into
+     * the overflow overlay. Reads the {@code has-overflow} state attribute.
+     *
+     * @return {@code true} if the breadcrumbs has overflow, {@code false}
+     *         otherwise
+     */
+    public boolean hasOverflow() {
+        return hasAttribute("has-overflow");
+    }
+
+    /**
+     * Gets the overflow button element from the shadow DOM.
+     *
+     * @return the shadow-DOM element with {@code part="overflow-button"}
+     * @throws NoSuchElementException
+     *             if the overflow button is not found in the shadow DOM
+     */
+    public TestBenchElement getOverflowButton() {
+        WebElement button = getWrappedElement().getShadowRoot()
+                .findElement(By.cssSelector("[part='overflow-button']"));
+        return (TestBenchElement) wrapElement(button, getCommandExecutor());
+    }
+
+    /**
+     * Opens the overflow overlay by clicking the overflow button and waits
+     * until the overlay element reports that it is open.
+     */
+    public void openOverflowOverlay() {
+        WebElement button = getWrappedElement().getShadowRoot()
+                .findElement(By.cssSelector("[part='overflow-button']"));
+        // click() on elements in shadow DOM does not work with Chrome driver
+        executeScript("arguments[0].click();", button);
+        waitUntil(driver -> getOverflowOverlay().hasAttribute("opened"));
+    }
+
+    /**
+     * Gets the {@code <vaadin-breadcrumbs-overlay>} element.
+     *
+     * @return the overlay element
+     */
+    public TestBenchElement getOverflowOverlay() {
+        return $("vaadin-breadcrumbs-overlay").first();
+    }
+
+    /**
+     * Gets the breadcrumb item elements that are currently shown inside the
+     * open overflow overlay (i.e. items with {@code slot="overlay"}).
+     *
+     * @return list of breadcrumb item elements inside the open overlay
+     */
+    public List<TestBenchElement> getOverflowItems() {
+        return $(BreadcrumbsItemElement.class).withAttribute("slot", "overlay")
+                .all().stream().map(el -> (TestBenchElement) el).toList();
+    }
 }
--- a/vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-testbench/src/main/java/com/vaadin/flow/component/breadcrumbs/testbench/BreadcrumbsElement.java
+++ b/vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-testbench/src/main/java/com/vaadin/flow/component/breadcrumbs/testbench/BreadcrumbsElement.java
@@ -37,11 +37,9 @@ public class BreadcrumbsElement extends TestBenchElement {
     }
 
     /**
-     * Gets the current item — the item that represents the current page. The
-     * current item is the last item without a path and carries the
-     * {@code current} state attribute.
+     * Gets the item that represents the current page.
      *
-     * @return the current breadcrumb item element, or {@code null} if none
+     * @return the current breadcrumb item, or {@code null} if none
      */
     public BreadcrumbsItemElement getCurrentItem() {
         return getItems().stream().filter(BreadcrumbsItemElement::isCurrent)
@@ -53,8 +51,7 @@ public class BreadcrumbsElement extends TestBenchElement {
      *
      * @param text
      *            the text to search for
-     * @return the matching breadcrumb item element, or {@code null} if not
-     *         found
+     * @return the matching breadcrumb item, or {@code null} if not found
      */
     public BreadcrumbsItemElement getItemByText(String text) {
         return getItems().stream().filter(item -> text.equals(item.getText()))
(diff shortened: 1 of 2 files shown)


vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-testbench/src/main/java/com/vaadin/flow/component/breadcrumbs/testbench/BreadcrumbsItemElement.java
```diff
@@ -24,4 +30,89 @@
...
+    }
+
+    /**
+     * Gets the element slotted into the {@code prefix} slot of this breadcrumb
+     * item.
+     *
+     * @return the prefix slot element, or {@code null} if no prefix is set
+     */
+    public TestBenchElement getPrefixSlotContent() {
+        WebElement prefixSlot = getWrappedElement().getShadowRoot()
```
web-padawan: This could be simplified using `$("*").withAttribute("slot", "new-button")`.

vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-testbench/src/main/java/com/vaadin/flow/component/breadcrumbs/testbench/BreadcrumbsElement.java
```diff
@@ -24,4 +30,108 @@
...
+    }
+
+    /**
+     * Gets the {@code <vaadin-breadcrumbs-overlay>} element.
+     *
+     * @return the overlay element
+     */
+    public TestBenchElement getOverflowOverlay() {
+        return $("vaadin-breadcrumbs-overlay").first();
+    }
```
web-padawan: The overlay component is an implementation detail, in reality we only need `getOverflowItems()` so IMO let's remove this from public API for now.

vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-testbench/src/main/java/com/vaadin/flow/component/breadcrumbs/testbench/BreadcrumbsElement.java
```diff
@@ -24,4 +30,102 @@
...
+
+    /**
+     * Gets the overflow button element from the shadow DOM.
+     *
+     * @return the shadow-DOM element with {@code part="overflow-button"}
+     * @throws NoSuchElementException
+     *             if the overflow button is not found in the shadow DOM
+     */
+    public TestBenchElement getOverflowButton() {
+        WebElement button = getWrappedElement().getShadowRoot()
```
web-padawan: Let's replace `getWrappedElement().getShadowRoot()` with something like `$(TestBenchElement.class).withAttribute("part", "overflow-button")`.

vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-testbench/src/main/java/com/vaadin/flow/component/breadcrumbs/testbench/BreadcrumbsElement.java
```diff
@@ -24,4 +30,102 @@
...
+                .findElement(By.cssSelector("[part='overflow-button']"));
+        return (TestBenchElement) wrapElement(button, getCommandExecutor());
+    }
+
+    /**
+     * Opens the overflow overlay by clicking the overflow button and waits
+     * until the collapsed items become visible.
+     */
+    public void openOverflowOverlay() {
+        WebElement button = getWrappedElement().getShadowRoot()
```
web-padawan: Let's reuse `getOverflowButton()` here.

vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-testbench/src/main/java/com/vaadin/flow/component/breadcrumbs/testbench/BreadcrumbsElement.java
```diff
@@ -24,4 +30,102 @@
...
+        executeScript("arguments[0].click();", button);
+        waitUntil(driver -> {
+            List<TestBenchElement> items = getOverflowItems();
+            return !items.isEmpty() && items.get(0).isDisplayed();
+        });
+    }
+
+    /**
+     * Gets the breadcrumb item elements that are currently shown inside the
+     * open overflow overlay (i.e. items with {@code slot="overlay"}).
```
web-padawan: Let's remove `(i.e. items with {@code slot="overlay"})` - slot is an implementation detail.

vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-testbench/src/main/java/com/vaadin/flow/component/breadcrumbs/testbench/BreadcrumbsElement.java
```diff
@@ -24,4 +30,102 @@
...
+     *         found
+     */
+    public BreadcrumbsItemElement getItemByPath(String path) {
+        return getItems().stream().filter(item -> path.equals(item.getPath()))
+                .findFirst().orElse(null);
+    }
+
+    /**
+     * Returns {@code true} if one or more items are currently collapsed into
+     * the overflow overlay. Reads the {@code has-overflow} state attribute.
```
web-padawan: Let's remove "Reads the {@code has-overflow} state attribute" - it's an implementation detail.

vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-testbench/src/main/java/com/vaadin/flow/component/breadcrumbs/testbench/BreadcrumbsElement.java
sissbruecker, line 1: I would question the value of retrieving overflow items separately. Whether items overflow or not depends on logic in the Breadcrumbs component, not necessarily some specific application code that wants to enforce (and thus test) the overflow. I'd say we can drop all the overflow stuff from a first version, make item retrieval / interaction work regardless whether an item is in the overflow or not, and then wait and see if someone really wants to test whether items end up overflowing.

vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-testbench/src/main/java/com/vaadin/flow/component/breadcrumbs/testbench/BreadcrumbsElement.java
```diff
@@ -24,4 +26,97 @@
...
+        });
+    }
+
+    /**
+     * Gets the breadcrumb item elements that are currently shown inside the
+     * open overflow overlay.
+     *
+     * @return list of breadcrumb item elements inside the open overlay
+     */
+    public List<TestBenchElement> getOverflowItems() {
```
sissbruecker: As mentioned, I'd suggest to drop this. But apart from that it's not clear why this casts to a generic element instead of using `BreadcrumbsItemElement`.

vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-testbench/src/main/java/com/vaadin/flow/component/breadcrumbs/testbench/BreadcrumbsItemElement.java
```diff
@@ -24,4 +29,80 @@
...
+        return prefix.exists() ? prefix.first() : null;
+    }
+
+    /**
+     * Clicks the anchor element in the shadow DOM of this breadcrumb item. This
+     * only applies to items that have a {@code path} set; for current
+     * (non-link) items the click is a no-op.
+     */
+    @Override
+    public void click() {
```
sissbruecker: For consistency, maybe this should be in a separate `navigate` method, similar to how it works with Side Nav. That one also throws when the item does not contain an anchor, which is more reasonable when the method says "navigate" and can help preventing broken / useless test setups.

vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-testbench/src/main/java/com/vaadin/flow/component/breadcrumbs/testbench/BreadcrumbsItemElement.java
```diff
@@ -24,4 +29,80 @@
...
+        return hasAttribute("has-prefix");
+    }
+
+    /**
+     * Gets the element slotted into the {@code prefix} slot of this breadcrumb
+     * item.
+     *
+     * @return the prefix slot element, or {@code null} if no prefix is set
+     */
+    public TestBenchElement getPrefixSlotContent() {
```
sissbruecker: Let's not use the slot terminology, neither in the method name, nor in the JavaDoc.

There's multiple other methods that reference slots / shadow DOM, etc.. I feel there's too much technical detail for a consumer of those APIs, just mentioning what these APIs do / return would be enough, instead of explaining how it is accomplished technically.

vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-testbench/src/main/java/com/vaadin/flow/component/breadcrumbs/testbench/BreadcrumbsItemElement.java
```diff
@@ -24,4 +29,80 @@
...
+        ElementQuery<TestBenchElement> prefix = $("*").withAttribute("slot",
+                "prefix");
+        return prefix.exists() ? prefix.first() : null;
```
sissbruecker: Let's use `single` over the deprecated `first`.

vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-testbench/src/main/java/com/vaadin/flow/component/breadcrumbs/testbench/BreadcrumbsElement.java
web-padawan: Removed everything related to overflow for now.

vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-testbench/src/main/java/com/vaadin/flow/component/breadcrumbs/testbench/BreadcrumbsItemElement.java
web-padawan: Renamed to `navigate` as suggested.

vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-testbench/src/main/java/com/vaadin/flow/component/breadcrumbs/testbench/BreadcrumbsItemElement.java
web-padawan: Renamed to `getPrefixComponent()`.

vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-testbench/src/main/java/com/vaadin/flow/component/breadcrumbs/testbench/BreadcrumbsItemElement.java
web-padawan: Done.

vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-testbench/src/main/java/com/vaadin/flow/component/breadcrumbs/testbench/BreadcrumbsItemElement.java
```diff
@@ -24,4 +29,74 @@
...
+    public void navigate() {
+        WebElement anchor;
+        try {
+            anchor = getWrappedElement().getShadowRoot()
+                    .findElement(By.cssSelector("a"));
+        } catch (NoSuchElementException e) {
+            throw new NoSuchElementException("Item does not contain an anchor",
+                    e);
+        }
+        // click() on elements in shadow DOM does not work with Chrome driver
```
sissbruecker: Small nit: From my testing this also makes clicking items that are in the overflow overlay work, even though they are not be visible on the page when the overflow is not opened. Could extend the comment to include that bit.

sissbruecker (approved):

vaadin-breadcrumbs-flow-parent/vaadin-breadcrumbs-testbench/src/main/java/com/vaadin/flow/component/breadcrumbs/testbench/BreadcrumbsItemElement.java
web-padawan: Thanks, updated the comment accordingly.

PR #9563, opened 2026-06-18
refactor: clean up grid connector

tomivirkki:
## Summary
- Simplify and modernize `gridConnector.ts` without behavior changes: extract shared `singleTimeRenderer`/`renderContent`/`updateItemDetails` helpers to remove duplicated detail-toggle and content-render logic.
- Tidy idioms throughout: strict equality, optional chaining/nullish coalescing, `Array.fill`, `Set`-based sorter dedup, and a side-effect import for the grid module since the `Grid` symbol is no longer referenced.

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
@@ -1,7 +1,6 @@
 // @ts-nocheck
 import { Debouncer } from '@vaadin/component-base/src/debounce.js';
 import { timeOut } from '@vaadin/component-base/src/async.js';
-import '@vaadin/grid/src/vaadin-grid.js';
 import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
 import { GridFlowSelectionColumn } from './vaadin-grid-flow-selection-column.js';
 
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
@@ -1,12 +1,29 @@
 // @ts-nocheck
 import { Debouncer } from '@vaadin/component-base/src/debounce.js';
-import { timeOut, animationFrame } from '@vaadin/component-base/src/async.js';
-import { Grid } from '@vaadin/grid/src/vaadin-grid.js';
+import { timeOut } from '@vaadin/component-base/src/async.js';
+import '@vaadin/grid/src/vaadin-grid.js';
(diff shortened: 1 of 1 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
```diff
@@ -257,15 +274,9 @@ window.Vaadin.Flow.gridConnector.initLazy = (grid) => {
...
-        const sorters = Array.from(grid.querySelectorAll('vaadin-grid-sorter'));
-
         // Sorters for hidden columns are removed from DOM but stored in the web component.
         // We need to ensure that all the sorters are reset when using `grid.sort(null)`.
-        grid._sorters.forEach((sorter) => {
-          if (!sorters.includes(sorter)) {
-            sorters.push(sorter);
-          }
-        });
+        const sorters = [...new Set([...grid.querySelectorAll('vaadin-grid-sorter'), ...grid._sorters])];
```
vursen, line 278: **nit:** `Set` already has a `forEach` method, so we don't technically need to convert it back to an array:

```suggestion
        const sorters = new Set([...grid.querySelectorAll('vaadin-grid-sorter'), ...grid._sorters]);
```

vursen (approved):

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/resources/frontend/gridConnector.ts
```diff
@@ -1,14 +1,31 @@
 // @ts-nocheck
 import { Debouncer } from '@vaadin/component-base/src/debounce.js';
-import { timeOut, animationFrame } from '@vaadin/component-base/src/async.js';
-import { Grid } from '@vaadin/grid/src/vaadin-grid.js';
+import { timeOut } from '@vaadin/component-base/src/async.js';
+import '@vaadin/grid/src/vaadin-grid.js';
```
vursen:
```suggestion
```

tomivirkki: I'm not sure if this should have been removed. Having the explicit import in place guaranteed vaadin-grid to be defined by the time initLazy is called. Now we're relying on import order of external functionality.

vursen: Mmm, I see... I'll revert that, thanks

vursen: But we don't have such imports in the other connectors, so this must be an issue for them too?

tomivirkki: It's probably not a technical issue at all right now, as long as the external timings are correct (which they most probably are).

PR #9564, opened 2026-06-18
feat!: switch form filler value options to items-based resolution

ugur-vaadin:
## Description

In form filler DX test sessions, the return type of value options APIs was a pain point. Even after updating examples and clarifying it in the documentation, it still was not intuitive: the developer had to think in labels (strings) when the rest of their codebase thought in domain objects, and had to supply a `toValue` converter that re-encoded the label-to-value mapping the field's `ItemLabelGenerator` already expressed.

This PR replaces that shape with an items-based resolution path:

- Registrations now carry the domain objects, not pre-rendered labels. Aligned with how the rest of the codebase models selection options.
- Labels for the LLM are derived through one `ItemLabelGenerator`. By default, the controller reflects the field's own `setItemLabelGenerator(...)` so common cases need no extra wiring; an explicit one can be supplied on `ValueOptions` for custom components or cases where the LLM should see a different label than the UI.
- When the LLM picks a label, the controller walks the registration's items, applies the same generator per item, and writes the first matching item to the field. The bijection lives in one place — the label generator — and the `toValue` converter is no longer needed and has been removed.

Changed in `ValueOptions`:
- `options(Collection<String>)` → `options(Collection<V>)`
- `options(BiFunction<String, Integer, List<String>>)` → `options(BiFunction<String, Integer, List<V>>)`

New in `ValueOptions`:
- `itemLabelGenerator(ItemLabelGenerator<V>)` — optional; falls back to the field's own labeler, then to `String.valueOf(item)`.

Changed in `FormAIController`:
- `fieldValueOptions(ValueOptions<String>)` → `fieldValueOptions(ValueOptions<V>)`
- `fieldValueOptions(ValueOptions<V>, Function<String, V> toValue)` removed.

Other behavior:
- Items with duplicate labels resolve to the first in registration order; a fixed-options registration logs a warning at registration so the developer notices.
- A query-mode `fill_form` that arrives before any `query_field_options` call is rejected with a reason naming `query_field_options` so the LLM can recover on the next turn.

> [!WARNING]
> The two-argument `fieldValueOptions(ValueOptions, Function<String, V>)` overload and the `String`-only generics on the single-argument variant are gone.

Based on DX test findings.

No related issue.

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
@@ -27,6 +27,7 @@ import java.util.Objects;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
+import java.util.function.BiFunction;
 import java.util.function.Function;
 
 import org.slf4j.Logger;
@@ -37,6 +38,7 @@ import com.vaadin.flow.component.ComponentUtil;
 import com.vaadin.flow.component.HasComponents;
 import com.vaadin.flow.component.HasEnabled;
 import com.vaadin.flow.component.HasValue;
+import com.vaadin.flow.component.ItemLabelGenerator;
 import com.vaadin.flow.component.ai.form.FormAITools.FormFieldDescriptor;
 import com.vaadin.flow.component.ai.form.FormValueConverter.RejectedValueException;
 import com.vaadin.flow.component.ai.orchestrator.AIController;
@@ -87,7 +89,10 @@ import tools.jackson.databind.JsonNode;
  * {@link #fieldValueOptions(ValueOptions, Function) two-argument overload} also
  * accepts a label-to-value converter, which the controller applies per label;
  * for multi-select fields the resolved elements are then aggregated into a
- * {@link LinkedHashSet} before {@link HasValue#setValue}.
+ * {@link LinkedHashSet} before {@link HasValue#setValue}. LLM-facing labels for
+ * the registered items are derived from the field's
+ * {@code setItemLabelGenerator(...)} by default; see {@link ValueOptions} for
+ * the full resolution chain.
  * </p>
  *
  * <p>
@@ -339,10 +344,10 @@ public class FormAIController implements AIController {
     }
 
     /**
-     * Registers a known set of labels for a {@link String}-typed field. The
-     * labels are presented to the LLM as the field's choices. No converter is
-     * needed — the chosen label is itself the value written to the field. For
-     * any non-{@link String} value type, use
+     * Registers a known set of options for a {@link String}-typed field. Each
+     * item is also the label the LLM sees — the chosen label is itself the
+     * value written to the field, with no converter needed. For any
+     * non-{@link String} value type, use
      * {@link #fieldValueOptions(ValueOptions, Function) the two-argument
      * overload} to supply a converter; this is enforced at compile time.
      * <p>
@@ -353,14 +358,14 @@ public class FormAIController implements AIController {
      *
      * @param config
      *            the field's options registration, not {@code null}; must have
-     *            its label source set via either
+     *            its item source set via either
      *            {@link ValueOptions#options(Collection)} or
      *            {@link ValueOptions#options(BiFunction)}
      * @return this controller, for chaining
      * @throws NullPointerException
      *             if {@code config} is {@code null}
      * @throws IllegalArgumentException
-     *             if the registration has no label source set; if the developer
+     *             if the registration has no item source set; if the developer
      *             routed a {@code MultiSelect} field through the single-value
      *             {@code forField} overload (upcast reference); or if the
      *             field's value type is a Collection but the field does not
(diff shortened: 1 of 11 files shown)


vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
sissbruecker: The JavaDoc says:
> Each item is also the label the LLM sees — the chosen label is itself the value written to the field, with no converter needed

But it runs the same `applyValueOptions` method that applies a potential item labeler. So when you provide a label function through the value options (and presumably also when through the field), the labels and values differ.

From my understanding this method should not use an item labeler, and potentially log a warning if one is configured in the options.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
```diff
@@ -372,13 +377,17 @@ public FormAIController fieldValueOptions(ValueOptions<String> config) {
...
+     * {@link ValueOptions#itemLabelGenerator(ItemLabelGenerator)} if set,
+     * otherwise the field's own {@code setItemLabelGenerator(...)} (read
+     * reflectively), otherwise {@link String#valueOf(Object)} as a last resort.
+     * When the LLM picks a label, the controller calls {@code toValue} on it
+     * and writes the result to the field. If {@code toValue} returns
+     * {@code null} or throws — for example because the LLM picked a label the
+     * converter does not recognize — the write is rejected back to the LLM with
+     * a reason and the model can correct on the next turn.
      * <p>
      * A typical converter delegates to a service or repository that looks the
```
sissbruecker: It might be worth pointing out explicitly that `toValue` must produce the reverse of whatever an item label generator produced. Currently this is connection is only implicit.

I feel this surfaces a gap in the design where you configure model -> label and label -> model converters in separate places, which has a higher chance for a misconfiguration. It's also problematic that this uses the label generator from the field as a fallback: That one can be easily changed in the UI code for whatever reason, but then you also need to remember to update the converter in the code that configures the form controller to resolve the model from the changed label.

I wonder if `toValue` is even necessary? We already keep references to the items in the value options now, so could we also add a lookup for the reverse mapping? Or was there some concern about stale data?

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
ugur-vaadin: Dropped the `Function.identity()` along with the two-arg overload entirely. For duplicate labels: first registered item wins, and the registration logs a warning including the duplicates and pointing at `ItemLabelGenerator`.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
ugur-vaadin: Dropped `toValue`. We iterated on this API quite a bit and I agree that it doesn't provide much value anymore, and it creates more confusion.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
```diff
@@ -465,24 +411,59 @@ private FormAIController applyValueOptions(ValueOptions<?> config,
         var labeler = resolveItemLabeler(config.field(),
                 config.itemLabelGenerator());
         var hints = hintsFor(config.field());
-        hints.valueOptionsToValue = toValue;
         hints.itemLabelGenerator = labeler;
+        // Items the converter resolves chosen labels against. Pre-populated
+        // with the fixed list, or appended on each query-callback invocation.
+        var observedItems = new ArrayList<>();
```
sissbruecker: Would it make sense to use a map to automatically dedupe, instead of keeping lots of object references that will never be used because we only pick the first one? Also the lookups would become easier / cheaper.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
```diff
@@ -465,24 +411,59 @@ private FormAIController applyValueOptions(ValueOptions<?> config,
...
         } else {
             @SuppressWarnings({ "unchecked", "rawtypes" })
             BiFunction<String, Integer, List<Object>> rawQuery = (BiFunction) query;
-            hints.valueOptionsQuery = (filter, limit) -> rawQuery
-                    .apply(filter, limit).stream().map(labeler).toList();
+            hints.valueOptionsQuery = (filter, limit) -> {
+                var batch = rawQuery.apply(filter, limit);
+                observedItems.addAll(batch);
+                return batch.stream().map(labeler).toList();
+            };
```
sissbruecker: This doesn't look correct, as it keeps adding potentially the same instances over and over if multiple queries match them. A map could fix this part as well. 

Other than that it's questionable if this should keep previous entries, in case the application has removed data or updated it. Maybe it should clear on every query, or there might be some other point in the workflow where that is sensible.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
ugur-vaadin: Done. We not use a `LinkedHashMap` that uses insertion order.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
ugur-vaadin: Updated the code to use the map. It should fix the potential inflation issue. For the second part, the controller now refreshes every value options binding at the start of each turn. It re-reads the field's labeler, repopulates the fixed options map from its item list, and clears the query options map.

sissbruecker (approved):

PR #9568, opened 2026-06-19
fix: sync editor value to server when same value is typed again

### LINKED ISSUE #9437: GridPro: Custom editColumn does not register consecutive changes with same value
### Description

I add a custom edit-column to GridPro (e.g., a `BigDecimalField`). When I enter the **same value** at least two times, the second time the value is not updated when leaving the edit-field.
This is only a problem in **Firefox** (works in Chrome and Safari).

### Expected outcome

I would expect the value to be updated each time.

```java
package com.example;

import java.math.BigDecimal;
import java.util.List;

import com.vaadin.flow.component.gridpro.GridPro;
import com.vaadin.flow.component.textfield.BigDecimalField;

@Route("")
    public MainView() {
        var grid = new GridPro<DataDTO>();
        grid.setSizeFull();

        grid.setItems(List.of(
                new DataDTO(new BigDecimal("1.1")),
                new DataDTO(new BigDecimal("2.2"))
        ));

        var inputField = new BigDecimalField();

        grid.addEditColumn(DataDTO::getValue1)
            .custom(inputField, DataDTO::setValue1)
            .setHeader("Value 1");

        add(grid);
    }

    static class DataDTO {
        private BigDecimal value1;

        DataDTO(BigDecimal value1) {
            this.value1 = value1;
        }

        public BigDecimal getValue1() {
            return value1;
        }

        public void setValue1(BigDecimal value1) {
            this.value1 = value1;
        }
    }
}
```

### Steps to reproduce

https://github.com/user-attachments/assets/951e05b7-6790-428f-bac3-37df4d09945b

### Environment

Vaadin version(s): 25
OS: MacOS

### Browsers

Firefox

sissbruecker:
## Description

Since a recent change, Firefox treats input fields that have the same value on blur as they had on focus as unchanged, even if the value was changed programmatically in between. This is problematic for custom editors in GridPro, where the value is set programmatically from the server after the client-side component has already focused the editor. If, after starting editing, a user types exactly the same value that the input had before the server updated the editor value, no change event is fired and the editor value is never synced to the server where it would be used to update the item data:
- First edit:
  - User edits cell, types "foo", commits
  - "foo" is different from the value the editor had before focus (empty value), property is synced to the server
  - `item-property-change` fires, server detects change in the editor's value, item is updated on server
- Second edit (same column, different cell)
  - User starts editing, web component focuses editor component immediately
  - Initially the client-side value is still "foo", the server only updates the editor's value once it receives the `cell-edit-started` event, which eventually results in a client-side update of the editor value, let's say to "bar"
  - User types "foo" again, commits
  - "foo" is not different from the value the editor had before focusing it, no change event is fired for the editor, it's value is not synced back to the server
  - `item-property-change` fires, server does not detect a change in the editor's value, no item update

This change addresses the issue by always firing a synthetic change event whenever an input element loses focus. The fix is scoped only Firefox, and only to custom editors. For built-in editors no such fix is needed as the edited value is applied by the web component before focusing. The change is not covered by tests, as the issue only manifests in Firefox whereas tests only run in Chrome.

Fixes https://github.com/vaadin/flow-components/issues/9437

## Type of change

- Bugfix

### DIFF
--- a/vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/resources/META-INF/resources/frontend/gridProConnector.js
+++ b/vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/resources/META-INF/resources/frontend/gridProConnector.js
@@ -7,6 +7,7 @@
  * See <https://vaadin.com/commercial-license-and-service-terms> for the full
  * license.
  */
+import { isFirefox } from '@vaadin/component-base/src/browser-utils.js';
 import { iterateRowCells, updatePart } from '@vaadin/grid/src/vaadin-grid-helpers.js';
 
 function isEditedRow(grid, rowData) {
@@ -26,21 +27,10 @@ window.Vaadin.Flow.gridProConnector = {
 
     grid.toggleAttribute(LOADING_EDITOR_CELL_ATTRIBUTE, false);
 
-    const active = editor.getRootNode().activeElement;
-    const candidates = [editor.contains(active) ? active : null, editor, editor.focusElement];
-    const focusElement = candidates.find((c) => c instanceof HTMLInputElement);
-
-    if (focusElement) {
-      // The editor is focused before its value is set from the server. Firefox
-      // treats an input as unchanged when its value on blur equals the value it
-      // had on focus, ignoring programmatic changes in between, so typing the
-      // value from the previous edit again would not fire a change event and
-      // the cell would not update. Blurring and re-focusing the input after the
-      // value has been set (this runs after the property sync) recaptures the
-      // baseline, so a later change is detected.
-      focusElement.blur();
-      focusElement.focus();
-      focusElement.select();
+    if (editor instanceof HTMLInputElement) {
+      editor.select();
+    } else if (editor.focusElement && editor.focusElement instanceof HTMLInputElement) {
+      editor.focusElement.select();
     }
   },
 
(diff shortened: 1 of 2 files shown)


vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/resources/META-INF/resources/frontend/gridProConnector.js
sissbruecker, line 1: An alternative fix would have been to add a `focusout` listener on the editor, and then fire a synthetic `change` event if the event came from an input. That seems to work for Vaadin field components (which then fire `value-changed`) and native inputs. I have a hunch that the re-focus approach might cover more cases, though it's arguably also more invasive.

sissbruecker: One small issue with the current approach is that the focus outline of Vaadin field components is lost.

sissbruecker: Discussed this, and decided to go with the change event for now.

vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/resources/META-INF/resources/frontend/gridProConnector.js
```diff
@@ -16,7 +17,7 @@ function isEditedRow(grid, rowData) {
 const LOADING_EDITOR_CELL_ATTRIBUTE = 'loading-editor';
 
 window.Vaadin.Flow.gridProConnector = {
-  selectAll: (editor, itemKey, grid) => {
+  handleCustomEditorValueUpdate: (editor, itemKey, grid) => {
```
sissbruecker, line 20: Kept this refactoring as it includes other logic than just selecting the value.

vaadin-grid-pro-flow-parent/vaadin-grid-pro-flow/src/main/resources/META-INF/resources/frontend/gridProConnector.js
```diff
@@ -61,6 +62,21 @@ window.Vaadin.Flow.gridProConnector = {
...
+    if (isFirefox && !component.__firefoxChangeWorkaroundInstalled) {
+      // The editor is focused before its value is set from the server. Firefox
+      // treats an input as unchanged when its value on blur equals the value it
+      // had on focus, ignoring programmatic changes in between. So typing the
+      // value from the previous edit again would not fire a change event and
+      // the cell would not update. As a workaround, always fire a synthetic
+      // change event when an input loses focus.
+      component.addEventListener('focusout', (e) => {
+        if (e.target instanceof HTMLInputElement) {
+          e.target.dispatchEvent(new Event('change'));
```
vursen: The change event should bubble in case `component !== input`

sissbruecker: Done

vursen (approved):

PR #9594, opened 2026-06-23
refactor: improve instructions for ambiguous values

ugur-vaadin:
## Description

There is an issue with `FormAIController` ignoring per-field description directives like *"let the user pick when the query returns multiple items"* on queryable fields. When the query returned multiple matches, the LLM picked one (or all) of them rather than asking the user, regardless of what the developer wrote in the description.

This PR updates the form-fill workflow text in `INSTRUCTIONS_TEXT` step 3 so the LLM checks the field's `description` for an ambiguity directive before writing a queryable field whose query returned multiple matches. When the directive is present and the user's prompt does not single out the subset to write, the LLM asks for clarification instead. Prompts like *"Add both X and Y"* or *"Pick all of them"* already name the subset, so they proceed without asking. Without a directive, the LLM picks the most reasonable match and writes it.

Based on Form Filler DX test session findings.

No related issue.

## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] Refactor

## Checklist


vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAIController.java
```diff
@@ -201,7 +201,15 @@ public class FormAIController implements AIController {
             and pick a returned label. Fields with an inline "enum" array \
             carry their full option set already — pick from it directly.
             3. Call fill_form({"values": {<id>: <value>}}) with every value \
-            you mean to set this turn. Skip fields the user did not mention.
+            you mean to set this turn. Skip fields the user did not mention. \
+            Before writing a queryable field whose query returned multiple \
```
tomivirkki: Field descriptions are free-form text; developers will write arbitrary things in them ("always uppercase", "never guess, confirm with the user", "prefer the newest matching item", ...). Baking one specific directive pattern (an ambiguity directive on a multi-match queryable field) into the workflow doesn't generalize.

Rather state the general rule, e.g.:

> A field's description may contain handling instructions from the developer (e.g. when to ask the user instead of guessing) — follow them; they take precedence over the defaults in this workflow.

That covers this use-case and every other arbitrary one, and keeps step 3 compact.

tomivirkki (approved):

PR #9595, opened 2026-06-23
refactor: update session context tool description

ugur-vaadin:
## Description

There is an issue with relative date phrases like *"tomorrow"*, *"yesterday"*, *"two weeks from now"*, or *"next Friday"* in user prompts. The LLM either left date fields empty or wrote the phrase verbatim rather than resolving it into an ISO date.

This PR updates the `SessionContextTool` description in `AIOrchestrator` to lead with explicit guidance for resolving relative phrases against the date/time included in the session context, into ISO date / date-time / time strings. The orchestrator stays controller-agnostic. What to do with the resolved value is left to whichever controller is wired in.

Based on Form Filler DX test session findings.

No related issue.

## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] Refactor

## Checklist

### DIFF
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/orchestrator/AIOrchestrator.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/orchestrator/AIOrchestrator.java
@@ -649,9 +649,13 @@ public class AIOrchestrator implements Serializable {
         SessionContextTool(String content) {
             this.content = content;
             this.description = """
-                    Read for current session context (e.g. date and time, user \
-                    locale). The content below is captured at the start of \
-                    this turn:
+                    Read for current session context. If a date/time is \
+                    included below, use it to resolve relative phrases in \
+                    the user's prompt — "today", "tomorrow", "yesterday", \
+                    "next Friday", "in two weeks", "end of next month", \
+                    etc. — into ISO date / date-time / time strings.
+
+                    Captured at the start of this turn:
 
                     """ + content;
         }
--- a/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/test/java/com/vaadin/flow/component/ai/orchestrator/AIOrchestratorTest.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/test/java/com/vaadin/flow/component/ai/orchestrator/AIOrchestratorTest.java
@@ -2509,6 +2509,34 @@ class AIOrchestratorTest {
     }
 
     @Test
+    void sessionContextToolDescription_carriesRelativeDateGuidance() {
+        // Real LLMs often leave date fields empty on the first turn when the
+        // user writes "tomorrow" or "next Friday" because nothing in the
+        // tool surface tells them to anchor relative phrases against the
+        // date that the session-context tool carries. Pin the load-bearing
+        // phrases that close that gap; a regression here re-opens it.
+        stubAddMessage();
+        Mockito.when(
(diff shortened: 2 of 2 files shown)


vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/test/java/com/vaadin/flow/component/ai/orchestrator/AIOrchestratorTest.java
```diff
@@ -2508,6 +2508,34 @@ void prompt_byDefault_includesSessionContextToolWithCurrentDateTime() {
                         + contextTool.getDescription());
     }
 
+    @Test
+    void sessionContextToolDescription_carriesRelativeDateGuidance() {
```
sissbruecker, line 2512: This test seems fairly pointless as it can not really assert any behavior.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/orchestrator/AIOrchestrator.java
```diff
@@ -649,9 +649,13 @@ private static final class SessionContextTool
         SessionContextTool(String content) {
             this.content = content;
             this.description = """
-                    Read for current session context (e.g. date and time, user \
-                    locale). The content below is captured at the start of \
-                    this turn:
+                    Read for current session context. If a date/time is \
```
sissbruecker, line 652: Should this still mention what's in the context, so an LLM knows when to access it? 

Maybe I don't understand well enough what this is for. It starts by "Read for current session context.". But then it already includes all the context in the description. Calling the tool also returns the same context. So there is never any point in calling the tool, because the LLM already has all the data just from reading the tool's description?

ugur-vaadin: This tool is more of a special case that is practically an appendix to the system prompt. However, since a dynamic system prompt will cause it to not be cached, we have this tool where all the data already in the description. I agree that technically it is awkward, but practically it works well.

Regarding the update in this PR, I think it is a small and generic enough update to be included in this tool description. Also, this context is global and not controller-specific. I tried it with different prompts and descriptions many times to be statistically relevant, and this version seems to increase the success ratio considerably with weaker models.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/test/java/com/vaadin/flow/component/ai/orchestrator/AIOrchestratorTest.java
ugur-vaadin: Since we cannot test the texts intended for LLMs behaviourally, we at least include guards against prompt drift when there is a specific section that serve a specific purpose that might be hard to keep track otherwise. I don't have a strong opinion though, we can rename or remove the test.

sissbruecker: If such tests already exist then I won't object. Personally I feel they do not provide value.

vaadin-ai-components-flow-parent/vaadin-ai-components-flow/src/main/java/com/vaadin/flow/component/ai/orchestrator/AIOrchestrator.java
sissbruecker: OK, let's leave as is then.

sissbruecker (approved):

PR #9607, opened 2026-06-24
refactor: simplify grid item details state handling

vursen:
The grid connector used to sync the open/closed state of item details via the `openedItemDetails` array by manually calling `openItemDetails`/`closeItemDetails` on each item update, set, or clear. This PR refactors that in favor of overriding the `_isDetailsOpened` method exposed by the web component to read the state directly from the item, removing duplicate state bookkeeping. It also makes this consistent with the way treeGridConnector manages the expanded state.

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/test/grid-connector-row-details.test.ts
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/test/grid-connector-row-details.test.ts
@@ -70,7 +70,6 @@ describe('grid connector - row details', () => {
   describe('updateFlatData', () => {
     it('should open details for updated items', async () => {
       grid.$connector.updateFlatData([{ key: '0', name: 'foo', detailsOpened: true }]);
-      await nextFrame();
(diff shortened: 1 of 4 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/test/grid-connector-row-details.test.ts
```diff
@@ -57,30 +63,20 @@ describe('grid connector - row details', () => {
...
-  });
-
-  it('should close details for cleared items', async () => {
-    setRootItems(grid.$connector, [
-      { key: '0', name: 'foo', detailsOpened: true },
-      { key: '1', name: 'bar' }
-    ]);
-
-    grid.$connector.clear(0, 2);
-    expect(grid.detailsOpenedItems).to.be.empty;
```
vursen, line 70: `$connector.clear` only removes items from the cache, so they are marked as loading on the next render, but their content isn't re-rendered until new data is available in the cache – this test can be safely removed.

web-padawan (approved):

PR #9612, opened 2026-06-25
fix!: decouple grid row details and selection from activeItem

vursen:
Row details and single-selection both relied on `activeItem` property observers. Routing both through `activeItem` caused several bugs, which the updated WTR tests now cover:

- With deselect disallowed, clicking a row with open details failed to close them. The selection observer reverted `activeItem`, which prevented the details observer from running.
- Selecting an item from server data opened its details, because the programmatic `activeItem` change triggered the details observer as if the row had been clicked.

The PR replaces the two `activeItem` observers with `cell-activate` and `row-activate` listeners that read `detailsOpened` and `selected` from the event. These events only fire on user interaction, so programmatic changes don't trigger them. There's also no longer any need to mirror selection onto `activeItem` when selection changes come from the server. As a result, details and selection are now fully independent of `activeItem` and only react to real clicks.

> [!WARNING]
> This should be a relatively safe change, since rows were always rendered as selected via the `selected` attribute/part rather than `active`, and the same applies to `details-opened`. That said, the `activeItem` property is no longer populated, which could affect apps that did something based on it in their custom logic.

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/test/grid-connector-row-details.test.ts
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/test/grid-connector-row-details.test.ts
@@ -87,16 +87,16 @@ describe('grid connector - row details', () => {
   });
 
   it('should set details hidden on selected item click', () => {
-    getBodyCellContent(grid, 0, 0)!.click();
+    grid.$connector.updateFlatData([{ key: '0', name: 'foo', detailsOpened: true }]);
     getBodyCellContent(grid, 0, 0)!.click();
     expect(grid.$server.setDetailsVisible).to.be.calledWith(null);
   });
 
-  it('should not set details hidden on selected item click when deselect is disallowed', () => {
+  it('should set details hidden on selected item click when deselect is disallowed', () => {
     grid.__deselectDisallowed = true;
+    grid.$connector.updateFlatData([{ key: '0', name: 'foo', detailsOpened: true }]);
     getBodyCellContent(grid, 0, 0)!.click();
-    getBodyCellContent(grid, 0, 0)!.click();
-    expect(grid.$server.setDetailsVisible).not.to.be.calledWith(null);
+    expect(grid.$server.setDetailsVisible).to.be.calledWith(null);
   });
 
   it('should not set details visible on click when details on click is disallowed', () => {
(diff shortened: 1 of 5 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/frontend/gridConnector.ts
```diff
@@ -116,45 +109,32 @@ window.Vaadin.Flow.gridConnector.initLazy = (grid) => {
...
-          // cache.
-          oldVal = dataProviderController.getItemContext(oldVal).item;
-          grid.$connector.doDeselection([oldVal], true);
-        }
+  function onItemActivate(event) {
+    const item = event.detail.model?.item;
+
+    if (selectionMode === 'SINGLE') {
+      if (item && item.selected && grid.__deselectDisallowed) {
+        return;
```
sissbruecker: The early return here prevents the details open / closing logic from running. Thus when `__deselectDisallowed` is active that means clicking on a selected row still prevents details from closing.

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/test/grid-connector-row-details.test.ts
```diff
@@ -87,16 +87,16 @@ describe('grid connector - row details', () => {
...
 
   it('should set details hidden on selected item click', () => {
-    getBodyCellContent(grid, 0, 0)!.click();
+    grid.$connector.updateFlatData([{ key: '0', name: 'foo', detailsOpened: true }]);
     getBodyCellContent(grid, 0, 0)!.click();
     expect(grid.$server.setDetailsVisible).to.be.calledWith(null);
   });
 
-  it('should not set details hidden on selected item click when deselect is disallowed', () => {
+  it('should set details hidden on selected item click when deselect is disallowed', () => {
```
sissbruecker, line 96: The clicked item doesn't appear to be selected, at least I can't find any test setup for it.

The test above also doesn't match its description.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/frontend/gridConnector.ts
vursen: Good catch. Fixed.

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/test/grid-connector-row-details.test.ts
vursen: Fixed these tests to actually select items.

sissbruecker (approved):

PR #9631, opened 2026-06-29
fix!: update multi-select-combo-box to synchronize on change

### LINKED ISSUE #9611: MultiSelectComboBox: Spurious client-initiated value-change event resets value when ui.push() is called before setValue()
### Description

When ui.push() is called before setValue() on a MultiSelectComboBox, a spurious client-initiated value-change event fires after the second push(). This event carries an empty value, corrupting server-side state — while the component continues to visually display the correct value.

As a result, there is a mismatch between the displayed value and the actual server-side value, and subsequent client interactions report the wrong oldValue (empty instead of the previously set value).

Removing the first push() (called before setValue()) resolves the issue entirely.

### Expected outcome

No spurious value-change event should fire. After ui.push(), the combo box should retain its value ("1") on both server and client.

```java
UI ui = UI.getCurrentOrThrow();
var comboBox = new MultiSelectComboBox<String>();
	System.out.println("value-change event" + (e.isFromClient() ? " from client" : ""));
	System.out.println("    old value: " + String.join(",", e.getOldValue()));
	System.out.println("    value: " + String.join(",", e.getValue()));
	System.out.println("    value after value changed: " + String.join(",", comboBox.getValue()));
});

comboBox.setItems(List.of("1", "2", "3"));
ui.add(comboBox);
ui.push(); 			// first push (before setValue)

comboBox.setValue(Set.of("1"));
ui.push();			// second push (after setValue)

System.out.println("# value: " + String.join(",", comboBox.getValue()));
```

**Actual Output**
```
value-change event
    old value: 
    value: 1
    value after value changed: 1
# value: 1
value-change event from client
    old value: 1
    value: 
    value after value changed: 
```

### Steps to reproduce

Create a MultiSelectComboBox, call ui.push() before setting a value via setValue(), then call ui.push() again — see the minimal reproducible example below. The issue reproduces immediately on every page load.

### Environment

Vaadin version(s): 25.1.8
Spring Boot: 4.0.7
Java version: 21

### Browsers

Chrome

web-padawan:
## Description

Fixes #9611

With `@Push` and two push frames around a server-side `setValue`, MultiSelectComboBox could reset its value to empty because the `selected-items-changed` notification also fires for programmatic changes and gets echoed back as a spurious client value change.

- Synchronized the value on the web component `change` event instead of `selected-items-changed`, so only genuine user gestures are treated as client changes
- Updated TestBench helpers to dispatch `change` (and `deselectAll` to use `clear()`)
- Added integration tests for the push scenario and user value commits

> [!WARNING]
> Client-side changes to the web component `selectedItems` property no longer synchronize to the server unless a `change` event is dispatched. Server-side `setValue` and end-user interactions are unaffected.

## Type of change

- Behavior altering fix

---

🤖 Generated with Claude Code


vaadin-combo-box-flow-parent/vaadin-combo-box-testbench/src/main/java/com/vaadin/flow/component/combobox/testbench/MultiSelectComboBoxElement.java
```diff
@@ -137,7 +139,13 @@ public void deselectByText(String label) {
      * Deselects all items, effectively clearing the value.
      */
     public void deselectAll() {
-        String script = "const combobox = arguments[0]; combobox.selectedItems = [];";
+        //@formatter:off
+        String script =
+                "const combobox = arguments[0];" +
+                "if (combobox.selectedItems.length) {" +
+                "  combobox.clear();" +
+                "}";
```
vursen: **nit:** Let's use Java text blocks

vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/main/java/com/vaadin/flow/component/combobox/test/MultiSelectComboBoxUserValueChangePage.java
```diff
@@ -0,0 +1,63 @@
...
+        comboBox.setClearButtonVisible(true);
+
+        Span eventValue = new Span();
+        eventValue.setId("event-value");
+        Span eventOrigin = new Span();
+        eventOrigin.setId("event-origin");
+        Span eventCount = new Span("0");
+        eventCount.setId("event-count");
+
+        comboBox.addValueChangeListener(e -> {
```
vursen: I'm actually not sure if we need this many tests. How about just adding a unit test that ensures the change event is used, and a few ITs verifying that the change event is fired by the TestBench element correctly? It seems that the change event is now being adopted by the majority of field components, so I'd say it was rather an omission that MultiSelectComboBox didn't use it.

vaadin-combo-box-flow-parent/vaadin-combo-box-testbench/src/main/java/com/vaadin/flow/component/combobox/testbench/MultiSelectComboBoxElement.java
web-padawan: Done.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/main/java/com/vaadin/flow/component/combobox/test/MultiSelectComboBoxUserValueChangePage.java
web-padawan: Good point, thanks. Added unit tests and reduced ITs to focus on covering TestBench elements.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/MultiSelectComboBoxTest.java
```diff
@@ -49,6 +51,23 @@ void initialValue() {
...
+        String syncExpression = JsonConstants.SYNCHRONIZE_PROPERTY_TOKEN
+                + "selectedItems";
+
+        Assertions.assertTrue(
+                listeners.getExpressions("change").contains(syncExpression),
+                "value should synchronize on the change event");
+        Assertions.assertFalse(
+                listeners.getExpressions("selected-items-changed")
+                        .contains(syncExpression),
+                "value should not synchronize on selected-items-changed");
```
vursen, line 68: Could this be tested by actually firing a synthetic change event like here

https://github.com/vaadin/flow-components/blob/main/vaadin-date-time-picker-flow-parent/vaadin-date-time-picker-flow/src/test/java/com/vaadin/flow/component/datetimepicker/validation/BasicValidationTest.java#L282-L295

vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/main/java/com/vaadin/flow/component/combobox/test/MultiSelectComboBoxPushValueChangePage.java
vursen, line 1: Is this issue specific to MultiSelectComboBox or can it also happen with other components?

web-padawan: It can likely affect other components too, would need verification. I'll try to reproduce it for ComboBox.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow/src/test/java/com/vaadin/flow/component/combobox/MultiSelectComboBoxTest.java
web-padawan: I agree that current test is not ideal. But looks like `fireEvent("change")` without updating `selectedItems` property in the test would do nothing, so it's probably not the right way to test this.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/main/java/com/vaadin/flow/component/combobox/test/MultiSelectComboBoxPushValueChangePage.java
web-padawan: Confirmed that ComboBox is not affected. The issue is specific to MSCB and `selectedItems` logic.

vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/MultiSelectComboBoxUserValueChangeIT.java
```diff
@@ -0,0 +1,70 @@
...
+import com.vaadin.flow.testutil.TestPath;
+import com.vaadin.testbench.TestBenchElement;
+import com.vaadin.tests.AbstractComponentIT;
+
+/**
+ * Verifies that the TestBench element helpers, which change the selection
+ * through the {@code change} event, propagate the value to the server.
+ */
+@TestPath("vaadin-multi-select-combo-box/user-value-change")
+public class MultiSelectComboBoxUserValueChangeIT extends AbstractComponentIT {
```
vursen: Looks like this suite is a complete duplicate of `MultiSelectComboBoxValueChangeIT`. I think we can drop it.

vursen: @web-padawan I removed this suite and moved the deselectAll test to `MultiSelectComboBoxValueChangeIT`.

vursen (approved):

Issue #1603, opened 2021-05-21
Add Column visibility toggler for the grid


jcgueriaud1:
The column visibility toggler is a common feature for datatable.

Is it possible to add this feature in vaadin-grid or add an example of code to create a column toggler?
There are few problems or missing features: (if you have a simple solution for these, please comment this ticket)

- no hideable attribute for a column (you cannot configure this easily for your custom toggler)
- no getHeaderText or getHeader to display a label for each column in the column visibility toggler
- a non visible column is different than having a column removed. (from the Column.setVisible Note that column related data is sent to the client side even if the column is invisible. Use {@link Grid#removeColumn(Column)} to remove column (or don't add the column all) and avoid sending extra data.). So if you have a lot of hidden columns, the grid could be slow down because of the hidden columns.

Here one forum post: https://vaadin.com/forum/thread/17131429/set-hidable-and-hidden-columns-in-grid

mvysny:
Also see https://github.com/vaadin/web-components/issues/986

jcgueriaud1:
There is another example of how to implement this in the cookbook: https://cookbook.vaadin.com/grid-column-toggle

jouni:
The documentation for column visibility, which includes an example of this: https://vaadin.com/docs/latest/ds/components/grid/#column-visibility

rolfsmeds:
The initial implementation of this could be just the column toggle dropdown itself, with an API for triggering it from anywhere. That way each application could implement the trigger in whatever makes the most sense there (custom Grid toolbar, right-click context menu, whatever).

We could then later provide an option to show a standard trigger somewhere somehow.

rolfsmeds:
Design detail: the button for opening the column selector should provide some kind of indication when one or more columns are hidden.

mstahv:
I missed this feature (again) today while trying to clean up one of our start.vaadin.com templates: https://github.com/mstahv/enhancing-master-detail-view/commits/main

mstahv:
I modified the workaround shared previously to be triggered via "column selector icon", like in previous Vaadin versions, and it is now shared in in.virit:viritin:2.0.1. See [this change](https://github.com/viritin/flow-viritin/commit/799a92d895940613315b3aa5c9634470839cbaa9) how it is down and how to use it.

TatuLund:
Just as concept testing and feeling I have had column selector menu in BeanTable for a while

https://v-herd.eu/tatulund-addons/beantable

[...]

PR #9660, opened 2026-07-02
feat: add hideable columns and skip data for hidden columns

marcushellberg:
Adds the Flow side of the Grid column toggle (issue #1603) and stops sending
per-item data for hidden columns.

The toggle is a built-in grid feature: calling `setHideable(true)` on at least
one column makes the grid show its column toggle button. There is no separate
component to create or position.

Changes:
- `Grid.Column.setHideable(boolean)` / `isHideable()` (default `false`),
  mapped to the web component `hideable` property. The value also reaches
  columns that are currently hidden: Flow does not send property changes to
  invisible elements, so `AbstractColumn` pushes the value with JS through the
  grid element (same approach as the `_flowId` handling in `Grid#addColumn`).
  For the same reason `isHideable()` has no `@Synchronize`: a sync listener
  makes the client revert that pushed value (client updates to invisible
  elements are rejected), and hideable is a server-controlled flag.
- When the user shows or hides a column through the toggle menu, the
  `<vaadin-grid>` element fires `column-visibility-changed`. `Grid` listens
  with an internal `@DomEvent` and applies the change on the server, so
  `Column.isVisible()` stays correct. `event.detail.column._flowId` maps the
  toggled element back to the server column.
- `Column.addVisibilityChangedListener(...)` and `ColumnVisibilityChangedEvent`,
  fired for both programmatic and client-driven visibility changes.
- Performance: a hidden column no longer sends its per-item cell, tooltip or
  part data to the client. The column's data generator only runs while the
  column is visible, and `generateTooltipTextData` / `generatePartData` skip
  hidden columns. When a column becomes visible again, the data communicator is
  reset so already-loaded rows receive its data.
- Updated the `AbstractColumn.setVisible` / `addColumn` javadoc, which
  previously said data is always sent for invisible columns.

Draft until the sibling web-components PR is merged and released: the
`GridColumnToggleIT` runs against the published `@vaadin/grid` package, which
does not contain the column toggle yet. To test against local web-component
changes before that, overlay the local sources into the integration tests'
`node_modules/@vaadin/grid` and clear the Vite cache (`node_modules/.vite`).

Part of #1603
Sibling PR (web-components): https://github.com/vaadin/web-components/pull/12020

---

🤖 Generated with Claude Code

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridColumnTogglePage.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/src/main/java/com/vaadin/flow/component/grid/it/GridColumnTogglePage.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2000-2026 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.vaadin.flow.component.grid.it;
+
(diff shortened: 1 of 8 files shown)


vursen:
> Performance: a hidden column no longer sends its per-item cell, tooltip or
part data to the client. The column's data generator only runs while the
column is visible, and generateTooltipTextData / generatePartData skip
hidden columns. When a column becomes visible again, the data communicator is
reset so already-loaded rows receive its data.

Please note that this is already implemented in: https://github.com/vaadin/flow-components/pull/6798

PR #9667, opened 2026-07-03
refactor: deprecate DataCommunicatorBuilder and its constructors

vursen:
This PR deprecates DataCommunicatorBuilder together with the constructors that accept it in Grid and TreeGrid, and introduces a `createDataCommunicator()` method to Grid as a new override point.

Existing constructors keep their current behavior unchanged: the deprecated builder-based constructors store the builder in a field, and `createDataCommunicator()` uses it, which preserves backward compatibility for anyone with a custom DataCommunicatorBuilder. 

A new `getCompositeDataGenerator()` getter is added as well: it keeps the grid's composite data generator accessible for `createDataCommunicator()` overrides, since with DataCommunicatorBuilder deprecated, there would otherwise be no way to pass it into DataCommunicator.

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
@@ -1354,10 +1354,10 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
 
     private final GridArrayUpdater arrayUpdater;
 
-    private DataCommunicatorBuilder dataCommunicatorBuilder;
-
     private final CompositeDataGenerator<T> gridDataGenerator;
     private final DataCommunicator<T> dataCommunicator;
+    @SuppressWarnings("rawtypes")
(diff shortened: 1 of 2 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -1725,6 +1734,18 @@ protected void initConnector() {
                         getElement());
     }
 
+    /**
+     * Creates the {@link DataCommunicator} that this Grid uses to handle all
+     * data communication.
+     *
+     * @return the new data communicator
+     */
```
DiegoCardoso, line 1744: Should we add a `@since 25.3` mention here and in the other method below?

vursen: Thanks, added.

vursen:
@copilot Please re-format the code

DiegoCardoso (approved):

PR #9699, opened 2026-07-08
chore: add script to point NpmPackage at a feature snapshot

Artur-:
updateNpmFeatureSnapshot.js writes the dev-<feature> dist-tag published by the web-components publishFeatureSnapshot.sh into every @vaadin NpmPackage annotation, so flow tracks the latest snapshot build like a -SNAPSHOT dependency. Packages without the tag are skipped.


yuriy-fix:
We will bring it up within the team to decide how it should behave.

Artur-:
Merged the script into `updateNpmVer.js`

yuriy-fix:
Discussed with the team and decided not to proceed with this convinience at the moment. Keeping the branch open in case we would want to introduce it later.

PR #9705, opened 2026-07-09
chore: clean up release notes generation output

manolo:
The generated GitHub release notes were hard to read. The Web-Component
version bumps produced one repeated line per commit, and each PR entry pasted
the whole PR body: the template sections, checklists, HTML comment hints and
trailing separators. This made the notes long and noisy, and it also published
internal template text on the public release pages.

Changes in `scripts/generateChangeLog.js`:

- Web-Component version bumps are collapsed into a single bullet that links
  every commit, instead of one identical line per commit.
- Each PR entry shows only the first paragraph of its Description or Summary
  section, followed by a `(more)` link to the PR when content was cut.
- Template sections (`Type of change`, `Checklist`), checkbox items, HTML
  comments and horizontal-rule separators are removed from the output.
- Footer references that are not issue numbers are no longer listed as tickets.

### DIFF
--- a/scripts/generateChangeLog.js
+++ b/scripts/generateChangeLog.js
@@ -187,6 +187,111 @@ function parseLog(log) {
   return commits;
 }
 
+// Lines that must never reach the release notes (template noise, AI refs)
+function isNoise(line) {
+  return /^- \[[ xX]\]/.test(line)
+    || /^🤖/.test(line)
+    || /generated with .*claude/i.test(line)
+    || /i have read the contribution guide/i.test(line);
+}
+// A line starting a non-prose markdown block (list, table, code fence, quote)
+function isBlockStart(line) {
+  return /^[-*+]\s/.test(line) || /^\d+\.\s/.test(line)
+    || /^\|/.test(line) || /^```/.test(line) || /^>/.test(line);
+}
+function isHeading(line) {
+  return /^#{1,6}\s/.test(line);
+}
+// A standalone footer reference (e.g. "Fixes #123", "Part of #9489").
+// Prose that merely starts with such a keyword (e.g. "Follow-up to #9333. When a
+// Spreadsheet...") is not treated as a footer.
+function isFooter(line) {
+  const m = /^(fixes|closes|resolves|related to|connected to|part of|follow-up to|warranty)\b[:\s]*(.*)$/i.exec(line);
+  if (!m) return false;
+  const rest = m[2]
+    .replace(/https?:\/\/\S+/g, '')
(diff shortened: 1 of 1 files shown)


scripts/generateChangeLog.js
```diff
@@ -187,6 +187,111 @@ function parseLog(log) {
   return commits;
 }
 
+// Lines that must never reach the release notes (template noise, AI refs)
+function isNoise(line) {
+  return /^- \[[ xX]\]/.test(line)
+    || /^🤖/.test(line)
+    || /generated with .*claude/i.test(line)
+    || /i have read the contribution guide/i.test(line);
```
web-padawan: question: should we remove commit message body altogether? IMO the message and PR link are enough, see e.g. release notes for [25.2.0](https://github.com/vaadin/flow-components/releases/tag/25.2.0) which I formatted manually.

manolo: Sorry for coming back after a while to this, not a clear preference whether a short description might help to users or just the title, I guess, a short description might help users to understand in a quick view, although the page with descriptions might become huge in the case of dozens of commits in the list like in 25.2.0 

But in sum, we cannot change the presentation for small/long lists, so probably the best option is to remove the description. If you agree, I'll update the PR

web-padawan: Sounds good to me. Yes, IMO let's remove it then.

manolo: Done in fe5ab99, the commit message body is no longer printed. Entries now show only the commit link, title, tickets and components, plus the PR link.

manolo:
validation fails, even after rebasing, what's wrong? this PR should not break that

sissbruecker:
Looks like there was a network issue or something on the first attempt:

> [ERROR] No plugin found for prefix 'failsafe' in the current project and in the plugin groups [com.diffplug.spotless, org.sonatype.plugins, org.eclipse.jetty.ee10, org.eclipse.m2e, biz.aQute.bnd, com.vaadin, org.apache.maven.plugins, org.codehaus.mojo] available from the repositories [local (/home/runner/.m2/repository), central (https://repo.maven.apache.org/maven2), vaadin-prereleases (https://maven.vaadin.com/vaadin-prereleases)] -> [Help 1]

The second attempt only re-ran IT (4), which does not work with the current workflow, as it deletes the build artifact that the job needs to download at the end. So re-running individual jobs isn't possible in general.

Anyway, seems to be fine now after another full validation run.

sissbruecker (approved):

PR #9713, opened 2026-07-09
fix: style spreadsheet overlays when nested in a shadow root

DiegoCardoso:
Since #9303, spreadsheet overlays (context menu, popup-button filters, cell comments, tooltips) live in a `#spreadsheet-overlays` container in the component's light DOM, styled by a global `<style>` in `document.head`. When `<vaadin-spreadsheet>` is nested in another element's shadow root — for example embedded as a web component, as in the docs filtering example — the container ends up inside that shadow tree, which the document-level stylesheet cannot cross, so the overlays render unstyled.

To reproduce: open the [docs "spreadsheet filtering" example](https://vaadin.com/docs/latest/example?embed=spreadsheet-filtering-wc.js&import=component/spreadsheet/spreadsheet-imports.ts) (the spreadsheet is embedded as a web component) and open a column filter — the popup shows with no rounded corners, background, or shadow.

On attach, also adopt the overlay stylesheet onto the container's own root (`this._overlays.getRootNode()`) when that root is a `ShadowRoot`, so the scoped `#spreadsheet-overlays …` rules apply inside the shadow tree. Top-level usage is unchanged: the root is the document, already covered by the existing `document.head` stylesheet.

Follow-up to #9303

---

https://claude.ai/code/session_01QjNdPXWEUciMEXQe9hMgMt

### DIFF
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/resources/META-INF/frontend/vaadin-spreadsheet/vaadin-spreadsheet.js
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/resources/META-INF/frontend/vaadin-spreadsheet/vaadin-spreadsheet.js
@@ -187,7 +187,7 @@ export class VaadinSpreadsheet extends LitElement {
     // container's root so the scoped rules apply there too.
     const root = this._overlays.getRootNode();
     if (root instanceof ShadowRoot && !root.adoptedStyleSheets.includes(spreadsheetOverlayStyles.styleSheet)) {
-      root.adoptedStyleSheets = [...root.adoptedStyleSheets, spreadsheetOverlayStyles.styleSheet];
+      root.adoptedStyleSheets.push(spreadsheetOverlayStyles.styleSheet);
     }
   }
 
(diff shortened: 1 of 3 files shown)


vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow/src/main/resources/META-INF/frontend/vaadin-spreadsheet/vaadin-spreadsheet.js
```diff
@@ -179,6 +179,16 @@ export class VaadinSpreadsheet extends LitElement {
...
     }
+
+    // The overlay container lives in light DOM, so the overlay styles are
+    // injected into `document.head` (see constructor). That does not reach the
+    // container when `<vaadin-spreadsheet>` is nested inside another element's
+    // shadow root. In that case, also adopt the overlay styles onto the
+    // container's root so the scoped rules apply there too.
+    const root = this._overlays.getRootNode();
+    if (root instanceof ShadowRoot && !root.adoptedStyleSheets.includes(spreadsheetOverlayStyles.styleSheet)) {
+      root.adoptedStyleSheets = [...root.adoptedStyleSheets, spreadsheetOverlayStyles.styleSheet];
```
web-padawan: suggestion: let's use `adoptedStyleSheets.push()` which is supported in all browsers for V25.

DiegoCardoso: Thanks, done.

web-padawan (approved):

Issue #6181, opened 2024-04-09
Tabs throws IllegalArgumentException when client side edits the html 


knoobie:
### Description

When I add another `<vaadin-tab>` to the frontend via developer tools and click it, the server throws an exception:

```java


java.lang.IllegalArgumentException: The 'index' argument should not be greater than or equals to the number of children tabs. It was: 14
	at com.vaadin.flow.component.tabs.Tabs.lambda$getTabAt$10(Tabs.java:792)
	at java.base/java.util.Optional.orElseThrow(Optional.java:403)
	at com.vaadin.flow.component.tabs.Tabs.getTabAt(Tabs.java:792)
	at com.vaadin.flow.component.tabs.Tabs.getSelectedTab(Tabs.java:558)
	at com.vaadin.flow.component.tabs.Tabs.updateSelectedTab(Tabs.java:667)
	at com.vaadin.flow.component.tabs.Tabs.lambda$new$c53a20e4$1(Tabs.java:96)
	at com.vaadin.flow.internal.nodefeature.ElementPropertyMap.lambda$fireEvent$2(ElementPropertyMap.java:465)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at com.vaadin.flow.internal.nodefeature.ElementPropertyMap.fireEvent(ElementPropertyMap.java:465)
	at com.vaadin.flow.internal.nodefeature.ElementPropertyMap$PutResult.run(ElementPropertyMap.java:170)
	at com.vaadin.flow.server.communication.ServerRpcHandler.lambda$handleInvocations$1(ServerRpcHandler.java:436)
	at com.vaadin.flow.server.communication.ServerRpcHandler.runMapSyncTask(ServerRpcHandler.java:452)
	at com.vaadin.flow.server.communication.ServerRpcHandler.lambda$handleInvocations$3(ServerRpcHandler.java:446)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at com.vaadin.flow.server.communication.ServerRpcHandler.handleInvocations(ServerRpcHandler.java:446)
	at com.vaadin.flow.server.communication.ServerRpcHandler.handleRpc(ServerRpcHandler.java:324)
	at com.vaadin.flow.server.communication.UidlRequestHandler.synchronizedHandleRequest(UidlRequestHandler.java:114)
	at com.vaadin.flow.server.SynchronizedRequestHandler.handleRequest(SynchronizedRequestHandler.java:40)
[...]

PR #9734, opened 2026-07-13
fix: ignore sortersChanged for non-existent columns

### LINKED ISSUE #1201: Removing column does not clear the client-side state
After the column is removed using the grid.removeColumn() method, some garbage remains on the client side which affects server-side errors  e.g. while sorting (especially in multi sort vaadin/vaadin-grid-flow#813).

In this situation, the event listener on the grid component (sortersChanged method) gets data in JSON also from the removed columns (which were sorted), as a consequence the exception 

> "Received a sorters changed call from the client for a non-existent column"

is thrown.

Details:
- Vaadin: 14.0.15 / Flow 2.0.19
- Grid: 5.4.12 / Flow 4.0.11
- Java: 1.8 202 x64
- OS: Win10 PRO x64 PL
- Browser: FF 72.0b7 (64-bit), Chrome 79.0.3945.88 (x64)

vursen:
Following up on the AI reproduction, `Grid.sortersChanged` no longer throws when a sorter references a column that was removed on the server between the client emitting the event and the server handling it. The stale sorter is now skipped and any remaining valid sorters are still applied.

Fixes #1201

---
🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
@@ -4028,8 +4028,10 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
             JsonNode sorter = sorters.get(i);
             Column<T> column = idToColumnMap.get(sorter.get("path").asString());
             if (column == null) {
-                throw new IllegalArgumentException(
-                        "Received a sorters changed call from the client for a non-existent column");
+                // The column may have been removed on the server between the
+                // client emitting the event and the server handling it. Skip
+                // the stale sorter instead of failing the whole call.
+                continue;
             }
             if (sorter.has("direction") && sorter.get("direction")
                     .getNodeType() == JsonNodeType.STRING) {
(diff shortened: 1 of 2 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
```diff
@@ -4028,8 +4028,10 @@ private void sortersChanged(ArrayNode sorters) {
             JsonNode sorter = sorters.get(i);
             Column<T> column = idToColumnMap.get(sorter.get("path").asString());
             if (column == null) {
-                throw new IllegalArgumentException(
-                        "Received a sorters changed call from the client for a non-existent column");
+                // The column may have been removed on the server between the
+                // client emitting the event and the server handling it. Skip
+                // the stale sorter instead of failing the whole call.
```
vursen:
```suggestion
                // the stale sorter.
```

knoobie:
Here is a similar issue for Tabs, if you wanna tackle this as well https://github.com/vaadin/flow-components/issues/6181

web-padawan (approved):

PR #9738, opened 2026-07-14
fix: reset grid column width when setWidth(null) is called

### LINKED ISSUE #6811: Cannot reset column width of a grid
### Description of the bug

If I set the width of a column, either programmatic or by user (resizable=true) and I want programmatically to reset the width to apply defaults by calling _Column.setWidth(null)_, the column is shown with a very large width.
If I call _Column.getElement().removeProperty("width")_ instead , the column's width is changes to its default (as width has never been set explicitly).
The problem seems to be, that _Column.setWidth(String)_ sets the width property to null instead of removing the property.
Instead of 
```
        public Column<T> setWidth(String width) {
            getElement().setProperty("width", width);
            return this;
        }
```
it should be 
```
        public Column<T> setWidth(String width) {
            if ( width != null ) {
                getElement().setProperty("width", width);
            } else {
                getElement().removeProperty("width");
            }
            return this;
        }

```
Although the workaround is easy, I think it should be fixed.

### Versions

- Vaadin / Flow version: 24.5.2
- Java version: 17
- OS version: Windows
- Browser version (if applicable): Chrome
- Application Server (if applicable): Tomcat 10
- IDE (if applicable): Eclipse

web-padawan:
Calling `Column.setWidth(null)` set the web component's `width` property to `null` instead of removing it, causing the column to render with a very large width rather than its default. Now a `null` width removes the property.

Fixes #6811

---

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/java/com/vaadin/flow/component/grid/Grid.java
@@ -625,12 +625,16 @@ public class Grid<T> extends Component implements HasStyle, HasSize,
          * @see #setFlexGrow(int)
          *
          * @param width
-         *            the width to set this column to, as a CSS-string, not
-         *            {@code null}
+         *            the width to set this column to, as a CSS-string, or
+         *            {@code null} to reset the width to the default
          * @return this column, for method chaining
          */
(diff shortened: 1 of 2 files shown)


sissbruecker (approved):

web-padawan:
Looks like this might not be a proper fix. The issue is that `width` on the web component is not reset to `100px`.
We might want to modify WC to fall back to the default value when `null` or `undefined` is set. Closing.

Issue #9384, opened 2026-05-28
[dialog] Refactor DialogHeader and DialogFooter to use slots instead of renderers


web-padawan:
### Describe your motivation

We added support for `vaadin-dialog` header and footer slots in https://github.com/vaadin/web-components/pull/10624.
This should be relatively easy to integrate in the Flow component so IMO we should do it.

### Describe the solution you'd like

Update `DialogHeaderFooter` to not use `headerRenderer` / `footerRenderer` function, remove all the logic related to renderer state and instead add elements to the `vaadin-dialog-content` slotted element.

web-padawan:
Done in #9743.

PR #9743, opened 2026-07-15
refactor: use header and footer slots in Dialog

web-padawan:
The `vaadin-dialog` web component gained native `header-content` and `footer` slots in https://github.com/vaadin/web-components/pull/10624. This moves the Flow `DialogHeader`/`DialogFooter` off the `headerRenderer`/`footerRenderer` renderer functions and onto those slots, matching the earlier Popover cleanup (#7791).

The old approach set the renderer via `executeJs`, tracked renderer state (`rendererCreated`), added the wrapper as a virtual child, and re-created the renderer on every attach. All of that is removed: the header/footer wrapper is now a real slotted child, so its content is preserved across reattach (including `@PreserveOnRefresh`) without any re-init code.

## Changes

- `DialogHeaderFooter` sets `slot="header-content"` / `slot="footer"` on its wrapper and attaches it as a real child; the wrapper is removed from the DOM when it has no content, so the web component does not keep the empty header/footer visible.
- The main dialog content still renders through the existing `renderer`. The slotted wrappers are excluded from `virtualChildNodeIds` (via an identity check) so they are not relocated into the content renderer root.
- Because the wrappers are now real children, `getChildren()` is overridden to return only the main content, and `addComponentAtIndex` maps the content index to the element index that skips the wrappers. Otherwise `removeAll()` and index-based inserts would see the slotted wrappers.
- Added unit tests (slotting, `virtualChildNodeIds` exclusion, empty-wrapper detach, `getChildren`/index behavior) and integration tests asserting the `has-header`/`has-footer` state attributes toggle on add and remove.

Part of #9384

---

### DIFF
--- a/vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
+++ b/vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import com.vaadin.flow.component.AttachEvent;
 import com.vaadin.flow.component.ClientCallable;
@@ -594,9 +595,60 @@ public class Dialog extends Component implements HasComponents, HasSize,
      */
     @Override
     public void addComponentAtIndex(int index, Component component) {
-        HasComponents.super.addComponentAtIndex(index, component);
+        Objects.requireNonNull(component, "Component should not be null");
+        if (index < 0) {
+            throw new IllegalArgumentException(
+                    "Cannot add a component with a negative index");
+        }
+        // The header/footer wrappers are also element children, so the content
+        // index must be translated to the matching element index that skips
+        // them. Otherwise a wrapper preceding the position would offset it.
+        HasComponents.super.addComponentAtIndex(toElementIndex(index),
+                component);
 
         updateVirtualChildNodeIds();
+    }
+
+    /**
+     * Translates an index into the main content (as seen by
+     * {@link #getChildren()}) to the corresponding index into the dialog
+     * element's children, which also include the slotted header/footer
+     * wrappers.
+     */
+    private int toElementIndex(int contentIndex) {
+        List<Element> children = getElement().getChildren()
+                .collect(Collectors.toList());
+        int contentSeen = 0;
+        for (int elementIndex = 0; elementIndex < children
+                .size(); elementIndex++) {
+            if (contentSeen == contentIndex) {
+                return elementIndex;
+            }
+            if (isContentChild(children.get(elementIndex))) {
+                contentSeen++;
+            }
+        }
+        if (contentSeen < contentIndex) {
+            throw new IllegalArgumentException(String.format(
+                    "Cannot add a component at index %d, there are only %d content components",
+                    contentIndex, contentSeen));
+        }
+        return children.size();
+    }
+
+    /**
+     * {@inheritDoc}
+     * <p>
+     * Only the main content components are returned. Components added to the
+     * header or footer via {@link #getHeader()} / {@link #getFooter()} live in
+     * separate slotted wrapper elements and are excluded.
+     */
+    @Override
+    public Stream<Component> getChildren() {
+        return super.getChildren().filter(child -> {
+            Element parent = child.getElement().getParent();
+            return parent == null || isContentChild(parent);
+        });
     }
 
     @Override
@@ -605,6 +657,7 @@ public class Dialog extends Component implements HasComponents, HasSize,
         // innerHTML of the dialog element. This results in removing the content
         // elements created by the web component for slotting contents into its
         // overlay. To avoid this, we manually remove all children instead.
+        // getChildren() excludes header/footer content, so those are preserved.
         getChildren().forEach(this::remove);
     }
 
@@ -932,7 +985,7 @@ public class Dialog extends Component implements HasComponents, HasSize,
      */
     final public static class DialogHeader extends DialogHeaderFooter {
         private DialogHeader(Dialog dialog) {
-            super("headerRenderer", dialog);
+            super("header-content", dialog);
         }
     }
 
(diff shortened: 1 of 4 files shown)


vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
```diff
@@ -1270,26 +1283,39 @@ public void onDetach(ElementDetachEvent e) {
...
-        });
+        getElement().getChildren().filter(this::isContentChild)
+                .forEach(child -> {
+                    if (!childDetachListenerMap.containsKey(child)) {
+                        childDetachListenerMap.put(child,
+                                child.addDetachListener(childDetachListener));
+                    }
+                });
 
         this.getElement().setPropertyList("virtualChildNodeIds",
```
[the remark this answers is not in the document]
web-padawan: Fixed.

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
```diff
@@ -594,17 +595,69 @@ public void add(Collection<Component> components) {
...
+     * {@inheritDoc}
+     * <p>
+     * Only the main content components are returned. Components added to the
+     * header or footer via {@link #getHeader()} / {@link #getFooter()} live in
+     * separate slotted wrapper elements and are excluded.
+     */
+    @Override
+    public Stream<Component> getChildren() {
+        return super.getChildren().filter(child -> {
+            Element parent = child.getElement().getParent();
```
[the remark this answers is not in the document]
web-padawan: Fixed.

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
```diff
@@ -594,17 +595,69 @@ public void add(Collection<Component> components) {
...
         updateVirtualChildNodeIds();
     }
 
+    /**
+     * Translates an index into the main content (as seen by
+     * {@link #getChildren()}) to the corresponding index into the dialog
+     * element's children, which also include the slotted header/footer
+     * wrappers.
+     */
+    private int toElementIndex(int contentIndex) {
```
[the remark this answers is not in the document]
web-padawan: Fixed.

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
```diff
@@ -594,17 +595,53 @@ public void add(Collection<Component> components) {
...
 
         updateVirtualChildNodeIds();
     }
 
+    /**
+     * {@inheritDoc}
+     * <p>
+     * Only the main content components are returned. Components added to the
+     * header or footer via {@link #getHeader()} / {@link #getFooter()} live in
+     * separate slotted wrapper elements and are excluded.
```
sissbruecker: Let's not mention "slotted wrappers". It's questionable whether this needs a new JavaDoc at all, as the whole header / footer are element children is not something that a developer should ever be concerned about.

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/test/java/com/vaadin/flow/component/dialog/DialogChildrenTest.java
```diff
@@ -164,15 +164,88 @@ void relocateChild_detachListenerRemoved() {
     }
 
     @Test
-    void headerAndFooterAddedAsVirtualChildren() {
+    void headerAndFooterAddedAsSlottedChildren() {
         dialog.getHeader().add(new Div());
         dialog.getFooter().add(new Div());
 
-        Assertions.assertEquals(0, dialog.getElement().getChildCount());
+        // The header/footer wrappers are real, slotted children of the dialog
```
sissbruecker: I think it's obvious from the test name and assertions.

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
```diff
@@ -594,17 +595,53 @@ public void add(Collection<Component> components) {
...
+     */
+    @Override
+    public Stream<Component> getChildren() {
+        return super.getChildren().filter(
+                child -> isContentChild(child.getElement().getParent()));
```
sissbruecker: The logic here looks a bit brittle and confusing. It passes a **parent** of a child to `isContentChild`. So for any actual content child it passes the dialog element to the method. Which could potentially also fail in edge cases if the dialog itself gets a slot attribute for whatever reason.

The other issue is that for the header and footer wrappers this only works because those are not registered as components, which makes `super.getChildren()` then recurse into their children and returns those. That is not obvious at all.

I think this might be a bit more obvious:
```java
    private boolean isHeaderFooterWrapper(Element element) {
        return (dialogHeader != null && dialogHeader.root.equals(element))
                || (dialogFooter != null && dialogFooter.root.equals(element));
    }

    @Override
    public Stream<Component> getChildren() {
        // Collect into a list so the returned stream is a snapshot, matching
        // Component.getChildren() and staying safe when callers such as
        // removeAll() remove children while iterating.
        return getElement().getChildren()
                .filter(element -> !isHeaderFooterWrapper(element))
                .flatMap(element -> element.getComponent().stream()).toList()
                .stream();
    }
```

The one thing that's a bit opaque here is how `flatMap` turns an `Optional` into either something that gets added to the stream or not, but that's still better IMO.

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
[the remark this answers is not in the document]
sissbruecker: IMO we should consider using an approach that relies on index calculation and using `super.addComponentAtIndex` (what you had before), as that includes additional checks for whether signal bindings are active. But for this PR it's fine as is and it would make sense to apply the approach to Card as well.

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
web-padawan: Removed.

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/test/java/com/vaadin/flow/component/dialog/DialogChildrenTest.java
web-padawan: Removed.

vaadin-dialog-flow-parent/vaadin-dialog-flow/src/main/java/com/vaadin/flow/component/dialog/Dialog.java
web-padawan: Thanks, updated accordingly.

sissbruecker (approved):

PR #9745, opened 2026-07-15
chore: add initial set of coding conventions [skip ci]

sissbruecker:
Adds an initial set of code conventions for AI agents. The goal is to:
- Support AI code reviews with conventional rules that should be followed in this repo
- Support AI coding agents with the same rules, so that hopefully review comments for basic things are not needed in the first place

Most rules are extracted from code reviews of the last 2.5 years, plus some additional things that I could think of. Feel free to suggest more.


DiegoCardoso (approved):

web-padawan (approved):
Overall LGTM, we should probably somehow align this with the guidelines folder later.

Issue #1165, opened 2021-05-21
Validate that editor's components are non-empty


tarekoraby:
Validating that one of the editor's components is not empty seems problematic because CRUD clears the content of these components after every save, cancel, or delete operation. This means that if one's use case depend on validating that an editor component is non-empty, then the validation will always fail after every save, cancel, or delete (even if the user has properly saved, canceled, or delete).

To reproduce, change the binding in one of the TextFields in the CRUD example at https://vaadin.com/components/vaadin-crud/java-examples. For example, replace the binding of the firstName TextField as such:

```
binder.forField(firstName).withValidator(name -> !name.isEmpty(), "name cannot be empty")
				.withValidationStatusHandler(status -> {
					if (status.isError()) {
						Dialog errorDialog = new Dialog();
						errorDialog.add(new Label("name cannot be empty"));
						errorDialog.open();
					}
				}).bind(Person::getFirstName, Person::setFirstName);
```

Then hit cancel, for example, on any of the opened (and valid) editor's records. It will then be observed that the error dialog will popup (which is not a desirable behavior given that the user didn't actually perform an invalid operation).

web-padawan:
> [!WARNING]
> **Automated reproduction — produced by the Claude Code `repro` skill. Needs human verification.**
> The steps, verdict, and root-cause pointer below were generated automatically and must be confirmed by a human before being treated as authoritative.

- **Verdict:** reproduced
- **Hypothesis tested:** The bug is a spurious validation error, triggered by CRUD clearing the editor fields on cancel/save/delete, observable as a `withValidationStatusHandler` firing with `isError()` after a valid cancel of a non-empty-validated field.
- **Regression?:** not a regression (broken since introduction — the field-clearing has been in `BinderCrudEditor.clear()` since the binder-based editor was first added)
- **Branch:** `repro/1165` — pushed to `vaadin/flow-components`
- **Reproduced on:** vaadin/flow-components @ `main` (bdd9b85, Flow 25.3-SNAPSHOT)
- **Present on main?:** yes (still broken)
- **Theme / Browser:** Lumo / Chromium (playwright-cli)
- **Screenshot:** ![CRUD cancel triggers a spurious validation error](https://raw.githubusercontent.com/vaadin/flow-components/b85ebddf9e2b610ab58ce7bdca6a39af713b748f/repro-1165.png)

## Observed behavior

[...]

PR #9752, opened 2026-07-16
chore: add repro skill for triaging Flow component issues

web-padawan:
## Description

Added new skill `.claude/skills/repro` that reproduces a bug from a `vaadin/flow-components` or `vaadin/flow` issue URL: builds a minimal integration-test View, confirms the bug in a running browser via playwright-cli, points at the likely root cause, pushes a shareable `repro/<issue>` branch and (after confirmation) posts a summary comment on the issue.

```sh
/repro https://github.com/vaadin/flow-components/issues/1165
```

## Type of change

- Internal change

## Note

- Requires [`playwright-cli`](https://github.com/microsoft/playwright-cli#installation) to be installed (including `ffmpgeg` to record reproduction videos)
- Example output: https://github.com/vaadin/flow-components/issues/1165#issuecomment-4991357817

### DIFF
--- a/.claude/skills/repro/SKILL.md
+++ b/.claude/skills/repro/SKILL.md
@@ -1,0 +1,59 @@
+---
+name: repro
+description: Reproduce a Vaadin Flow component bug from a GitHub issue in vaadin/flow-components or a component-specific issue in vaadin/flow. Builds a minimal integration-test View, confirms the bug in a running browser with playwright-cli, points at the likely root cause, pushes a shareable repro/<issue> branch, and (after confirmation) posts a verification-pending summary comment on the issue.
+argument-hint: <issue-url>
+allowed-tools: Read, Write, Edit, Glob, Grep, Bash(gh:*), Bash(mvn:*), Bash(playwright-cli:*), Bash(npx:*), Bash(git:*), Bash(curl:*), Bash(lsof:*), Bash(kill:*)
+---
+
+<!-- In-repo Flow-only port of the ds-tools `repro` skill (v1.4.0). -->
+
+You are a tester reproducing a Vaadin Flow component bug. Input `$0` is a GitHub issue URL in `vaadin/flow-components` or `vaadin/flow`. Work the phases in order. **Never claim a bug is reproduced until you have seen it in a running browser.**
+
+Detailed instructions live in references — read each one the **first time** a phase needs it. They stay in context: on later runs in the same session, do not re-read them.
+
+| Reference | Covers |
+| --- | --- |
+| [`references/issue-analysis.md`](references/issue-analysis.md) | Phase 1 in full |
+| [`references/reproduce.md`](references/reproduce.md) | Phases 2–3: View scaffold, running the server, maintenance branches |
+| [`references/verify.md`](references/verify.md) | browser-observation discipline |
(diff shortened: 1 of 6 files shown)


.claude/skills/repro/SKILL.md
```diff
@@ -0,0 +1,59 @@
+---
+name: repro
+description: Reproduce a Vaadin Flow component bug from a GitHub issue in vaadin/flow-components or a component-specific issue in vaadin/flow. Builds a minimal integration-test View, confirms the bug in a running browser with playwright-cli, points at the likely root cause, pushes a shareable repro/<issue> branch, and (after confirmation) posts a verification-pending summary comment on the issue.
+argument-hint: <issue-url>
+allowed-tools: Read, Write, Edit, Glob, Grep, Bash(gh:*), Bash(mvn:*), Bash(playwright-cli:*), Bash(npx:*), Bash(git:*), Bash(curl:*), Bash(lsof:*), Bash(kill:*)
+---
```
sissbruecker, line 7: Would it make sense to restrict this to explicit calls? The reason is that the `claude` workflow would not be able to use this as is for various reasons, so if someone asks it to "reproduce" it should not automatically pick up this skill and then flail around because of permission issues or lack of tools.

.claude/skills/repro/SKILL.md
```diff
@@ -0,0 +1,59 @@
...
+
+<!-- In-repo Flow-only port of the ds-tools `repro` skill (v1.4.0). -->
```
sissbruecker: This doesn't add any meaningful context.

.claude/skills/repro/SKILL.md
```diff
@@ -0,0 +1,59 @@
...
+| --- | --- |
+| [`references/issue-analysis.md`](references/issue-analysis.md) | Phase 1 in full |
+| [`references/reproduce.md`](references/reproduce.md) | Phases 2–3: View scaffold, running the server, maintenance branches |
+| [`references/verify.md`](references/verify.md) | browser-observation discipline |
+| [`references/share.md`](references/share.md) | Phases 4–6 and cleanup |
+
+## Phase 0 — Setup (once per session)
+
+- Resolve `<ROOT>` = `git rev-parse --show-toplevel` (this flow-components checkout). Prefix every later command with absolute paths (`git -C <ROOT> …`) — env vars don't persist between shell calls.
+- `<FLOW_CORE>` = `<dirname of ROOT>/flow`, optional — only for root-cause search on `vaadin/flow` issues. If absent, note it and continue.
```
sissbruecker: Should we also provide the web components repo as source? The instructions below only say to point at "vaadin/web-components" source, but not where to find those.

.claude/skills/repro/SKILL.md
web-padawan: Added `disable-model-invocation: true` to align with cherry-pick skill.

.claude/skills/repro/SKILL.md
web-padawan: Removed.

.claude/skills/repro/SKILL.md
web-padawan: Added.

sissbruecker (approved):

PR #9757, opened 2026-07-17
refactor: allow grid header and footer renderers to run multiple times

vursen:
The grid web component is moving to rendering its header and footer declaratively with Lit (vaadin/web-components#12134). With that change, header and footer cells can be re-created, which re-runs the column renderers. The connector wrapped these renderers in `singleTimeRenderer`, so a re-created cell would stay empty.

The sorter element is now cached between renderer runs: re-creating it on every run would reset the sort direction and send a needless `sorters-changed` event to the server.

Part of vaadin/web-components#10789

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/frontend/gridConnector.ts
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/frontend/gridConnector.ts
@@ -10,7 +10,11 @@ function isRangeEqual(range1, range2) {
 
 function renderContent(root, content) {
   if (content instanceof Node) {
-    root.appendChild(content);
+    // Skip if the content is already in place: re-appending would detach
+    // and reattach it, making header components lose state such as focus.
+    if (content.parentNode !== root) {
+      root.appendChild(content);
+    }
   } else {
     root.textContent = content;
   }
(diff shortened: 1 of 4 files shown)


vursen:
@claude Investigate why single time renderers were added in the first place.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/frontend/gridConnector.ts
```diff
@@ -430,11 +421,15 @@ window.Vaadin.Flow.gridConnector.initLazy = (grid) => {
...
+    // The renderer can run multiple times, e.g. when the grid re-creates
+    // header cells. The sorter is cached between runs: re-creating it would
+    // reset its direction and fire a sorters-changed round-trip to the server.
+    let sorter;
+
+    column.headerRenderer = (root) => {
       let contentRoot = root;
       if (showSorter) {
-        const sorter = document.createElement('vaadin-grid-sorter');
+        sorter ??= document.createElement('vaadin-grid-sorter');
```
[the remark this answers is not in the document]
vursen: Thanks, I've improved the test coverage.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/frontend/gridConnector.ts
```diff
@@ -443,7 +438,7 @@ window.Vaadin.Flow.gridConnector.initLazy = (grid) => {
       }
 
       renderContent(contentRoot, content);
```
tomivirkki, line 444: This will now result in the header content getting detached+reattached on render, seems to happen initially also:

Image

vursen: Good catch! Fixed it and added tests to cover the case

tomivirkki (approved):

PR #9781, opened 2026-07-23
test: wait for loaded items count in LazyLoadingIT page-size tests

web-padawan:
Two LazyLoadingIT tests fail randomly on CI and locally: `loadItems_changePageSize` and `disabledLazyLoading_reducePageSize_enablesLazyLoading`. Both assert an exact loaded-items count right after reopening the overlay following a page-size change. The count is updated asynchronously, and the connector fetches the visible range plus a prefetch buffer, so one or two full pages may load on open. Depending on timing, the same assertion sees 0, 100 or 200 items.

Observed failures for the same assertion across runs:

```
expected:<100> but was:<0>
expected:<100> but was:<200>
expected:<50> but was:<100>
```

The tests now wait until the loaded count settles on a non-zero multiple of the new page size instead of asserting an exact count immediately. A new `waitForLoadedItemsCount` helper in `AbstractComboBoxIT` supports both exact counts and predicates. The exact-count assertion after scrolling in `disabledLazyLoading_reducePageSize_enablesLazyLoading` is removed because the count depends on when items outside the active range are dropped; the existing `assertRendered` check already verifies that lazy loading works.

With this change the suite passed twice in a row locally (31/31); before, every run had 1-3 failures.

Follow-up to #9238

---

🤖 Generated with Claude Code

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/AbstractComboBoxIT.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/test/java/com/vaadin/flow/component/combobox/test/AbstractComboBoxIT.java
@@ -19,6 +19,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.Function;
+import java.util.function.IntPredicate;
 import java.util.stream.Collectors;
 
 import org.junit.Assert;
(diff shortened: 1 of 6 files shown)


vursen:
There seem to be more usages of assertLoadedItemsCount, are those not affected?

web-padawan:
> There seem to be more usages of assertLoadedItemsCount, are those not affected?

Updated a few other ITs to use `waitForLoadedItemsCount` after `openPopup()` for consistency. 
They shouldn't be affected though - only tests where page size changed were actually flaky.

vursen (approved):

PR #9789, opened 2026-07-24
refactor: convert grid connector to a class

tomivirkki:
## Summary
- The grid connector was one large init function that assembled the `$connector` object out of closures, which made it hard to follow what is state, what is public API and what is internal wiring.
- It is now a `GridConnector` class: closure variables become private fields, the server-facing API is public methods, and the grid patching and event listeners live in named private setup methods. The class also replaces the hand-written `GridConnector` interface as the single source of truth for the `$connector` type.
- Based on #9784 and should be merged after it.

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/frontend/gridConnector.ts
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/frontend/gridConnector.ts
@@ -2,9 +2,19 @@ import { Debouncer } from '@vaadin/component-base/src/debounce.js';
 import { timeOut } from '@vaadin/component-base/src/async.js';
 import { isFocusable } from '@vaadin/grid/src/vaadin-grid-active-item-mixin.js';
 import { GridFlowSelectionColumn } from './vaadin-grid-flow-selection-column.ts';
+import type { GridColumn } from '@vaadin/grid/src/vaadin-grid-column.js';
 import type { GridSorter } from '@vaadin/grid/src/vaadin-grid-sorter.js';
+import type { GridSorterDirection } from '@vaadin/grid/src/vaadin-grid-data-provider-mixin.js';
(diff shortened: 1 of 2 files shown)


vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/frontend/gridConnector.ts
```diff
@@ -489,88 +370,305 @@ function initLazy(grid: FlowGrid) {
...
-        detail: {
-          resizedColumnKey: e.detail.resizedColumn._flowId
+    grid.scrollToIndex(...args);
+  }
+
+  /**
+   * Overrides and extends the grid's methods and properties to integrate with
+   * the server-side data communicator and selection state.
+   */
+  private patchGrid(): void {
```
vursen:
```suggestion
  private #patchGrid(): void {
```

tomivirkki: Done. The class conversion actually weakened some of the encapsulation because the private state was previously stored in the closures. Using native `#` instead of `private` resolves this issue.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/frontend/gridConnector.ts
```diff
@@ -611,71 +709,28 @@ function initLazy(grid: FlowGrid) {
...
-    });
-    if (targetRow && isRowFullyInViewport(targetRow)) {
-      return;
-    }
-
-    grid.scrollToIndex(...args);
-  };
+function initLazy(grid: FlowGrid) {
+  // Init the connector only once for the grid
+  grid.$connector ||= new GridConnector(grid);
```
vursen:
```suggestion
  grid.$connector ??= new GridConnector(grid);
```

vursen (approved):

PR #9825, opened 2026-08-05
refactor: small cleanups in spreadsheet custom editor code

DiegoCardoso:
Three independent cleanups in the custom editor code, follow-ups to #9814. They
were opened as a stack (#9826, #9827) whose bases merged into this branch, so
they now travel together.

## Changes

**Mount editors once per state change** (`SpreadsheetConnector`)

`loadStateChangeDataToWidget` had three sibling branches that each ended in
`showCellCustomEditors(...)`. The server sets `componentIDtoCellKeysMap` and
`cellKeysToEditorIdMap` together in `loadCustomComponents()`, so the common case
walked the whole editor map twice per round trip. Enabling
`showCustomEditorOnFocus` was worse: the first two branches mounted the editors
and the third removed them again.

The order was also wrong. The `componentIDtoCellKeysMap` branch called
`showCellCustomEditors(...)` before the `cellKeysToEditorIdMap` branch had run
`setupCustomEditors()`, so it could mount editors before the factory was set up.

Now the three changed properties are recorded first, then the editors are
mounted or removed once, after `setupCustomEditors()`. The end state is the same
in every combination of changed properties.

**Drop duplicate cell comment load** (`Spreadsheet.updateMarkedCells`)

The method loaded the cell comments itself and then again through
`reloadVisibleCellContents()` -> `updateRowAndColumnRangeCellData()`, so every
`refreshCells()` call scanned all comments twice. Application code calls
`refreshCells()` from value change listeners, so this runs per committed cell.
Only the first call is removed; the second still shows comments as soon as they
are added.

**Enable shared custom editor test** (`CustomEditorIT`)

`customEditorShared_persistsValuesCorrectly` had no `@Test` annotation, so JUnit
never ran it. It is the only test covering `CustomEditorSharedFixture`, which
therefore had no executed coverage at all while looking covered in the file. The
test passes as written, so this only adds the annotation.

Follow-up to #9814

---

🤖 Generated with Claude Code

### DIFF
--- a/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/SpreadsheetConnector.java
+++ b/vaadin-spreadsheet-flow-parent/vaadin-spreadsheet-flow-client/src/main/java/com/vaadin/addon/spreadsheet/client/SpreadsheetConnector.java
@@ -400,7 +400,13 @@ public class SpreadsheetConnector extends AbstractHasComponentsConnector
             StateChangeEvent stateChangeEvent) {
         final SpreadsheetWidget widget = getWidget();
         SpreadsheetState state = getState();
-        if (stateChangeEvent.hasPropertyChanged("componentIDtoCellKeysMap")) {
+        boolean componentMapChanged = stateChangeEvent
+                .hasPropertyChanged("componentIDtoCellKeysMap");
+        boolean editorMapChanged = stateChangeEvent
+                .hasPropertyChanged("cellKeysToEditorIdMap");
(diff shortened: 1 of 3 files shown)


vursen (approved):

web-padawan (approved):
LGTM. I merged PRs targeting this branch, maybe it should have been a stack 🙈

DiegoCardoso:
> LGTM. I merged PRs targeting this branch, maybe it should have been a stack 🙈

I tried to create them as stack, but it seems it's yet [not available in repos with merge queue enabled](https://github.blog/changelog/2026-07-30-stacked-pull-requests-are-now-in-public-preview/#find-out-more-and-share-your-feedback).

PR #9865, opened 2026-08-11
feat: add date metadata provider to DatePicker

web-padawan:
`setDisabledDates` and `setDisabledWeekdays` cover rules the server knows in full and can send to the browser once. They do not cover rules that depend on data the server has to look up, such as which days are already fully booked, where sending every date up front is not practical.

This adds a callback that the calendar asks for the dates it is about to show:

```java
datePicker.setDateMetadataProvider(range -> bookingService
        .findFullyBooked(range.start(), range.end()).stream()
        .map(date -> new DateMetadata(date, true))
        .toList());
```

The callback is called for a range that always covers whole months and may cover several, so a single query can answer for all of them. Return an entry only for the dates that have metadata. Call `refreshDateMetadata()` when the data behind the callback changes.

## Changes

The callback maps onto the web component's `dateMetadataProvider`, so the timing is the web component's own: while a range is being fetched the affected dates render in a loading state but **stay selectable**, and nothing is disabled before the callback has reported it. A slow callback therefore does not make the calendar unusable.

Server-side validation does not rely on that. It calls the callback with the whole month that contains the value, so a disabled date cannot be committed even if the browser was never told about it.

The connector keeps a single provider function and reuses it for every configuration update. The web component compares the provider by reference and drops its whole month cache when a new function is assigned, so building a fresh one each time would refetch every visible range on every update. Dropping the cache on purpose is what `refreshDateMetadata()` does, through the web component's `clearCache()`.

Months are zero-based in both directions on the wire, matching what the web component uses, so the connector converts nothing. The one conversion lives on the server.

The request is a non-void `@ClientCallable`, which already returns a promise on the client, so there is no request id, no pending map and no cancellation bookkeeping — the web component's own cache decides which answers are still wanted. It uses `DisabledUpdateMode.ALWAYS` because a dropped request would never settle its promise, leaving the calendar showing that month as loading forever.

## Scope

Custom part names for dates, the other half of `dateMetadataProvider`, follow in a separate PR. Then the same API on Date Time Picker.

Part of vaadin/platform#2867

---

### DIFF
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/test/shared.ts
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/test/shared.ts
@@ -26,15 +26,8 @@ export type DatePickerConnector = {
 };
 
 export type DatePickerServer = {
-  /** Months are 0-based, matching what the web component passes to the provider. */
-  requestDateMetadata: (
-    startYear: number,
-    startMonth: number,
-    startDay: number,
-    endYear: number,
-    endMonth: number,
-    endDay: number
-  ) => Promise<DatePickerDateMetadata[]>;
+  /** The range is given as ISO 8601 dates, the entries come back with a 0-based month. */
+  requestDateMetadata: (start: string, end: string) => Promise<DatePickerDateMetadata[]>;
 };
 
 export type FlowDatePicker = DatePicker & {
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/test/shared.ts
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/test/shared.ts
@@ -39,7 +39,7 @@ export type DatePickerServer = {
 
 export type FlowDatePicker = DatePicker & {
   $connector: DatePickerConnector;
-  // Optional, so that tests can cover the connector running before the server connection is ready.
+  // Assigned by Flow when the element is bound, and stubbed by the tests.
   $server?: DatePickerServer;
 };
 
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/test/shared.ts
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/test/shared.ts
@@ -1,7 +1,7 @@
 import './env-setup.js';
 import '@vaadin/date-picker/src/vaadin-date-picker.js';
 import '../frontend/generated/jar-resources/datepickerConnector.js';
-import type { DatePicker } from '@vaadin/date-picker/src/vaadin-date-picker.js';
+import type { DatePicker, DatePickerDateMetadata } from '@vaadin/date-picker/src/vaadin-date-picker.js';
 export { extractDateParts } from '@vaadin/date-picker/src/vaadin-date-picker-helper.js';
 import type {} from '@web/test-runner-mocha';
 
@@ -15,6 +15,8 @@ export type FlowDatePickerDateMetadataConfig = {
   disabledDates?: [number, number, number][];
   /** Disabled weekdays as ISO weekday numbers, Monday = 1 ... Sunday = 7. */
   disabledWeekdays?: number[];
+  /** Whether a date metadata provider is set on the server. */
+  hasProvider?: boolean;
 };
 
 export type DatePickerConnector = {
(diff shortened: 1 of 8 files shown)


vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
```diff
@@ -611,13 +624,81 @@ public void setDisabledWeekdays(Collection<DayOfWeek> weekdays) {
...
+     * selectable before may not be anymore. Does nothing if no provider is set.
+     *
+     * @see #setDateMetadataProvider(DateMetadataProvider)
+     * @since 25.3
+     */
+    public void refreshDateMetadata() {
+        if (dateMetadataProvider == null) {
+            return;
+        }
+        requestDateMetadataUpdate(true);
```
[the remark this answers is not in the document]
web-padawan: Fixed.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
```diff
@@ -631,12 +712,78 @@ public boolean isDateDisabled(LocalDate date) {
...
+        }
+        if (dateMetadataProvider == null) {
+            return false;
+        }
+        // The provider's contract is that a range always covers whole months,
+        // so ask for the month holding the date rather than for the single day,
+        // then filter.
+        DateRange month = new DateRange(date.withDayOfMonth(1),
+                date.withDayOfMonth(date.lengthOfMonth()));
+        Collection<DateMetadata> metadata = dateMetadataProvider
```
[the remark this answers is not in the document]
web-padawan: This logic will be changed in the follow-up PR so let's keep it for now.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/frontend/datepickerConnector.js
```diff
@@ -179,6 +179,19 @@ window.Vaadin.Flow.datepickerConnector.initLazy = (datepicker) => {
...
 
+  // STABLE reference — created once per connector, never reassigned. Assigning a new function
+  // to `dateMetadataProvider` clears the web component's cache and re-fetches every visible
+  // range, so the same function object is reused for every update.
+  const dateMetadataProvider = ({ start, end }) => {
+    if (!datepicker.$server) {
+      // Throwing lets the web component drop the months and retry on the next navigation,
+      // instead of caching "nothing is disabled" as authoritative.
+      throw new Error('Date metadata requested before the server connection was ready');
+    }
```
vursen: Is this scenario possible in practice?

web-padawan: Not really. Removed this logic and the corresponding WTR test.

vaadin-date-picker-flow-parent/vaadin-date-picker-flow-integration-tests/test/shared.ts
```diff
@@ -23,8 +25,22 @@ export type DatePickerConnector = {
...
+export type DatePickerServer = {
+  /** Months are 0-based, matching what the web component passes to the provider. */
+  requestDateMetadata: (
+    startYear: number,
+    startMonth: number,
+    startDay: number,
+    endYear: number,
+    endMonth: number,
+    endDay: number
+  ) => Promise<DatePickerDateMetadata[]>;
```
vursen: Could we send the start and end as ISO date strings instead of separate year/month/day parameters, so they match the value/min/max format?

```suggestion
  requestDateMetadata: (start: string, end: string) => Promise<DatePickerDateMetadata[]>;
```

web-padawan: Changed.

vursen: Also created a ticket to update WC to use ISO format:  https://github.com/vaadin/web-components/issues/12423

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DatePicker.java
```diff
@@ -631,12 +711,72 @@ public boolean isDateDisabled(LocalDate date) {
...
+        if (disabledDates.contains(date)
+                || disabledWeekdays.contains(date.getDayOfWeek())) {
+            return true;
+        }
+        if (dateMetadataProvider == null) {
+            return false;
+        }
+        // The provider's contract is that a range always covers whole months,
+        // so ask for the month holding the date rather than for the single day,
+        // then filter.
```
vursen: Does this contract have to apply to the server side? I know that the web component batches requests this way, but I don't see why requesting a single date would be a problem on the server.

web-padawan: Updated to request a single date, removed this comment and updated JavaDoc.

vursen (approved):

PR #9866, opened 2026-08-11
feat: add custom date part names to DatePicker

web-padawan:
The date metadata provider can say that a date cannot be selected, but not that a date is worth pointing out. Applications often want to show a day as busy or almost fully booked while still letting the user pick it, which is a styling concern rather than a constraint.

Metadata entries can now carry CSS part names, which are added to the date cell so a theme can style it with `::part()`:

```java
datePicker.setDateMetadataProvider(range -> availability.forRange(range).stream()
        .map(day -> day.isFull()
                ? new DateMetadata(day.date(), true)
                : new DateMetadata(day.date(), "almost-full"))
        .toList());
```

```css
vaadin-date-picker-overlay::part(almost-full) {
  color: var(--lumo-warning-text-color);
}
```

## Changes

`DateMetadata` gains a `partName` component. The two-argument `(date, disabled)` form stays available as a convenience constructor, and a new `(date, partName)` form covers the styling-only case so it does not have to pass a meaningless `false`.

A part name never affects whether a date can be selected, so an entry is now sent to the browser when it is disabled **or** has a part name, instead of only when it is disabled. The `disabled` and `part` keys are each written only when they apply, which keeps the payload the same as before for callers that do not use part names. A blank part name counts as none.

The connector needs no change: it already forwards the server's answer to the web component untouched, and the entry shape is the web component's own.

The Javadoc warns against reusing the built-in part names (`date`, `disabled`, `selected`, `focused`, `today`, `past`, `future`, `loading`). Those come with the theme's own rules — Lumo removes pointer events from `[part~='disabled']`, so a date borrowing that name could not be clicked even though the component still counts it as selectable.

## Scope

This completes the Date Picker side of the feature. TestBench additions with integration tests, and the same API on Date Time Picker, follow separately.

Part of vaadin/platform#2867

---

### DIFF
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DateMetadataProvider.java
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DateMetadataProvider.java
@@ -33,6 +33,10 @@ import java.util.Collection;
  * unusable. A date selected before the answer arrives is reported invalid by
  * server-side validation.
  * <p>
+ * Entries can also carry custom CSS part names, so that a theme can style
+ * particular dates. A part name only affects styling, never whether a date can
+ * be selected. See {@link DateMetadata} for the details.
+ * <p>
  * Results are cached per month in the browser, but not on the server, so an
  * expensive implementation should cache its own results. Call
  * {@link DatePicker#refreshDateMetadata()} when the data behind the callback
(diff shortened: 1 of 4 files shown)


vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/java/com/vaadin/flow/component/datepicker/DateMetadataProvider.java
```diff
@@ -33,6 +33,10 @@
  * unusable. A date selected before the answer arrives is reported invalid by
  * server-side validation.
  * <p>
+ * Entries can also carry custom CSS part names, so that a theme can style
+ * particular dates. A part name only affects styling, never whether a date can
+ * be selected. See {@link DateMetadata} for the details.
```
vursen: > A part name only affects styling, never whether a date can be selected.

Overly defensive statement, IMO :)

```suggestion
 * Entries can also carry custom CSS part names, so that a theme can style
 * particular dates. See {@link DateMetadata} for the details.
```

vursen (approved):

Issue #9870, opened 2026-08-12
TimePicker parses digits typed without separator wrong in fi-FI, da-DK


web-padawan:
### Description

In locales that use a dot as the time separator, such as `fi-FI` and `da-DK`, TimePicker parses digits typed without a separator into the wrong time. Typing `1234` gives `3.04` instead of `12.34`, and the wrong time is also sent to the server, so `getValue()` returns `03:04`. The same input works in locales that use a colon, such as `en-US` and `de-DE`.

The cause is in the client-side connector. It builds a regular expression by putting the localized separator straight into the pattern:

```js
const numbersRegExp = new RegExp('([\\d\\u0660-\\u0669]){1,2}(?:' + separator + ')?', 'g');
```

For a dot separator the pattern becomes `(?:.)?`, which matches any character instead of a literal dot, so the group takes the third digit with it: `1234` is read as hours `123` and minutes `4`. `formatTime` then passes the out-of-range hour to `Date.setHours`, which rolls it over into a valid but different time, so the field never shows the invalid state. `helpers.js` already exports `escapeRegExp` for the AM/PM token search, and it is not applied here.

Input that contains a separator, such as `13.00`, parses correctly, which is why the existing localization tests do not catch this.

### Expected outcome

Typing `1234` in a `fi-FI` TimePicker shows `12.34` and produces the value `12:34`, the same as `en-US` shows `12:34 PM` for the same input.

### Minimal reproducible example

```java
TimePicker timePicker = new TimePicker("Time");
timePicker.setLocale(new Locale("fi", "FI"));
timePicker.addValueChangeListener(event -> System.out.println("Value: " + event.getValue()));
add(timePicker);
```

### Steps to reproduce

1. Add the snippet above to a view
2. Click the time picker field and type `1234` without a separator
3. Press Enter
4. Notice the field shows `3.04` and the console logs `Value: 03:04`, instead of `12.34` and `12:34`

Measured in Chrome with the localization test view, for the three inputs below:

| Locale | Typed | Input shows | Server value | Expected value |
| --- | --- | --- | --- | --- |
| `fi-FI` | `1234` | `3.04` | `03:04` | `12:34` |
| `fi-FI` | `2359` | `19.09` | `19:09` | `23:59` |
| `fi-FI` | `130` | `10.00` | `10:00` | `13:00` |
| `da-DK` | `1234` | `03.04` | `03:04` | `12:34` |
[...]

PR #9871, opened 2026-08-12
refactor: drop manual reformatting after time picker locale change

web-padawan:
The connector reformatted the input itself after a locale change: it captured the current value, waited for the element, and wrote the newly formatted text into the input. The web component does this on its own when `i18n` changes, so the block only duplicated work.

It also worked in ways it should not have:

- it wrote to `timepicker.inputElement.value`, which a `FIXME` in the code already flagged as private API
- it assigned a localized string to `timepicker.value`, which expects ISO 8601. The component rejected it and restored the previous value, so the reformatting was a side effect of that rejection.
- its `parseISO()` helper read `timeObject.hours` without checking that `parseISOTime` returned an object, so a value that is not ISO 8601 would throw

Removing the block also removes the `when()` helper, which had no other caller, and the deep import from `@vaadin/time-picker/src`, so the connector no longer reaches into the web component sources.

This adds a web-test-runner suite for the connector, since the module had none. It covers formatting and parsing for a 12 hour and a 24 hour locale, seconds and milliseconds granularity, and the reformat on locale change that the removed code used to do. `TimePickerLocalizationIT` and `TimePickerDetachAttachPageIT` also pass.

Related to #9870

---

🤖 Generated with Claude Code

### DIFF
--- a/vaadin-time-picker-flow-parent/vaadin-time-picker-flow-integration-tests/test/time-picker-connector.test.ts
+++ b/vaadin-time-picker-flow-parent/vaadin-time-picker-flow-integration-tests/test/time-picker-connector.test.ts
@@ -1,0 +1,124 @@
+import { expect } from 'chai';
+import { fixtureSync, nextUpdate } from '@vaadin/testing-helpers';
+import { getInputValue, init, timepickerConnector, type FlowTimePicker } from './shared.js';
+
+describe('time-picker connector', () => {
+  let timePicker: FlowTimePicker;
+
+  beforeEach(async () => {
+    timePicker = fixtureSync('<vaadin-time-picker></vaadin-time-picker>');
+    init(timePicker);
+    await nextUpdate(timePicker);
+  });
+
+  it('should not reinitialize the connector', () => {
+    const connector = timePicker.$connector;
+    timepickerConnector.initLazy(timePicker);
+    expect(timePicker.$connector).to.equal(connector);
+  });
+
+  describe('12 hour clock locale', () => {
+    beforeEach(() => {
+      timePicker.$connector.setLocale('en-US');
+    });
+
+    it('should format time with the AM/PM token', () => {
+      expect(timePicker.i18n.formatTime!({ hours: 13, minutes: 30, seconds: 0, milliseconds: 0 })).to.equal(
+        '1:30 PM'
+      );
+    });
+
+    it('should parse time with the AM/PM token', () => {
+      expect(timePicker.i18n.parseTime!('1:30 PM')).to.eql({
+        hours: 13,
+        minutes: 30,
+        seconds: 0,
+        milliseconds: 0
+      });
+    });
+
+    it('should parse time without the AM/PM token', () => {
+      expect(timePicker.i18n.parseTime!('1:30')).to.eql({
+        hours: 1,
+        minutes: 30,
+        seconds: 0,
+        milliseconds: 0
+      });
+    });
+
+    it('should not parse an empty string', () => {
+      expect(timePicker.i18n.parseTime!('')).to.be.undefined;
+    });
+
+    it('should format seconds when the step is below a minute', () => {
+      timePicker.step = 1;
+      expect(timePicker.i18n.formatTime!({ hours: 13, minutes: 30, seconds: 15, milliseconds: 0 })).to.equal(
+        '1:30:15 PM'
+      );
+    });
+
+    it('should format milliseconds when the step is below a second', () => {
+      timePicker.step = 0.5;
+      expect(timePicker.i18n.formatTime!({ hours: 13, minutes: 30, seconds: 15, milliseconds: 250 })).to.equal(
+        '1:30:15.250 PM'
+      );
+    });
+
+    it('should parse milliseconds when the step is below a second', () => {
+      timePicker.step = 0.5;
+      expect(timePicker.i18n.parseTime!('1:30:15.250 PM')).to.eql({
+        hours: 13,
+        minutes: 30,
+        seconds: 15,
+        milliseconds: 250
+      });
+    });
+  });
+
+  describe('24 hour clock locale', () => {
+    beforeEach(() => {
+      timePicker.$connector.setLocale('de-DE');
+    });
+
+    it('should format time without the AM/PM token', () => {
+      expect(timePicker.i18n.formatTime!({ hours: 13, minutes: 30, seconds: 0, milliseconds: 0 })).to.equal('13:30');
+    });
+
+    it('should parse time without the AM/PM token', () => {
+      expect(timePicker.i18n.parseTime!('13:30')).to.eql({
+        hours: 13,
+        minutes: 30,
+        seconds: 0,
+        milliseconds: 0
+      });
+    });
+  });
+
+  describe('locale change', () => {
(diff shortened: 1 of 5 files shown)


vaadin-time-picker-flow-parent/vaadin-time-picker-flow-integration-tests/test/time-picker-connector.test.ts
```diff
@@ -0,0 +1,128 @@
...
+  });
+
+  describe('12 hour clock locale', () => {
+    beforeEach(() => {
+      timePicker.$connector.setLocale('en-US');
+    });
+
+    it('should format time with the AM/PM token', () => {
+      expect(timePicker.i18n.formatTime!({ hours: 13, minutes: 30, seconds: 0, milliseconds: 0 })).to.equal(
+        '1:30 PM'
```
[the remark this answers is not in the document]
web-padawan: > These tests are red on every current browser, so the new suite fails in CI.

It does not fail in CI though 🤷‍♂️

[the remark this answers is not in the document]
web-padawan: Not reproducible — the narrow no-break space does not appear here, so these assertions pass as written.

Measured by codepoint in the Playwright Chromium the suite actually runs on (`HeadlessChrome/148.0.7778.96`):

```
en-US  "1:30:15 PM"  ->  31 3a 33 30 3a 31 35 20 50 4d
```

`0x20`, a plain ASCII space. I also swept 51 locales against both option sets the connector uses (`hour+minute` and `hour+minute+second`), testing each character for "whitespace but not `0x20`" rather than matching a literal character, and got zero hits — including `en-CA`, `es-PA`, `ko-KR` and `ar-EG`.

The suite is green: 15 of 15 tests pass on this branch via `node scripts/wtr.js time-picker`.

The normalization in `shared.ts` `getInputValue` is there for the input element text, where the existing `TimePickerLocalizationIT` does the same for char 160, so it is kept as a precaution rather than because a current locale needs it.

vaadin-time-picker-flow-parent/vaadin-time-picker-flow-integration-tests/test/time-picker-connector.test.ts
```diff
@@ -0,0 +1,128 @@
...
+      timePicker.step = 1;
+      expect(timePicker.i18n.formatTime!({ hours: 13, minutes: 30, seconds: 15, milliseconds: 0 })).to.equal(
+        '1:30:15 PM'
+      );
+    });
+
+    it('should format milliseconds when the step is below a second', () => {
+      timePicker.step = 0.5;
+      expect(timePicker.i18n.formatTime!({ hours: 13, minutes: 30, seconds: 15, milliseconds: 250 })).to.equal(
+        '1:30:15.250 PM'
```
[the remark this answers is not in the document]
web-padawan: This test passes, since it depends on the same premise as the previous comment: `toLocaleTimeString` returns a plain ASCII space before `PM` in the Chromium the suite runs on, measured by codepoint (`... 31 35 20 50 4d`). So `formatMilliseconds` strips ` PM` as intended and returns `1:30:15.250 PM`, and there is no mangled output to fix.

The general point is fair though: `formatMilliseconds` hardcodes `' ' + amString`, so it would produce `1:30:15 PM.250 PM` for any locale whose time string used a non-ASCII space before the token. I swept 51 locales against both option sets the connector uses and none do, so it is latent rather than a live bug, and unrelated to this change. Happy to file it separately if you think it is worth guarding against a future ICU change.

vaadin-time-picker-flow-parent/vaadin-time-picker-flow/src/main/resources/META-INF/frontend/vaadin-time-picker/timepickerConnector.js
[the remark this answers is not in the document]
web-padawan: This is correct but not related to logic modified in the current PR. Will fix separately.

vaadin-time-picker-flow-parent/vaadin-time-picker-flow-integration-tests/test/time-picker-connector.test.ts
```diff
@@ -0,0 +1,128 @@
...
+      timePicker.$connector.setLocale('en-US');
+      timePicker.value = '13:00';
+      await nextUpdate(timePicker);
+    });
+
+    it('should format the value using the initial locale', () => {
+      expect(getInputValue(timePicker)).to.equal('1:00 PM');
+    });
+
+    it('should reformat the value using the new locale', async () => {
```
vursen:
```suggestion
    it('should reformat the value when the locale changes', async () => {
```

vaadin-time-picker-flow-parent/vaadin-time-picker-flow-integration-tests/test/time-picker-connector.test.ts
```diff
@@ -0,0 +1,128 @@
...
+  describe('locale change', () => {
+    beforeEach(async () => {
+      timePicker.$connector.setLocale('en-US');
+      timePicker.value = '13:00';
+      await nextUpdate(timePicker);
+    });
+
+    it('should format the value using the initial locale', () => {
+      expect(getInputValue(timePicker)).to.equal('1:00 PM');
+    });
```
vursen: This is already covered in the `12 hour clock locale` describe block.

vaadin-time-picker-flow-parent/vaadin-time-picker-flow-integration-tests/test/time-picker-connector.test.ts
```diff
@@ -0,0 +1,128 @@
...
+      expect(getInputValue(timePicker)).to.equal('13:00');
+    });
+
+    it('should keep the value when the locale changes', async () => {
+      timePicker.$connector.setLocale('de-DE');
+      await nextUpdate(timePicker);
+      expect(timePicker.value).to.equal('13:00');
+    });
+
+    it('should not reformat an empty value', async () => {
```
vursen:
```suggestion
    it('should not reformat an empty value when the locale changes', async () => {
```

vaadin-time-picker-flow-parent/vaadin-time-picker-flow-integration-tests/test/time-picker-connector.test.ts
web-padawan: Done.

vaadin-time-picker-flow-parent/vaadin-time-picker-flow-integration-tests/test/time-picker-connector.test.ts
web-padawan: Removed.

vaadin-time-picker-flow-parent/vaadin-time-picker-flow-integration-tests/test/time-picker-connector.test.ts
web-padawan: Done.

vursen (approved):

PR #9873, opened 2026-08-12
fix: name the unsupported locale in the time picker error

web-padawan:
The connector checks whether the browser supports the locale, and reports an error when it does not. The `catch` block reassigned `locale` to `en-US` before building the message, so the error always read `The locale en-US is not supported` and never named the locale that actually failed.

The reassignment was also dead, since the `throw` exits before anything reads `locale` again. No fallback ever happened, so the message no longer promises one.

```java
// before: "vaadin-time-picker: The locale en-US is not supported, falling back to default locale setting(en-US)."
// after:  "vaadin-time-picker: The locale en_US is not supported."
timePicker.setLocale(...); // with a locale the browser rejects
```

Extracted from #9871 to keep that PR about the connector cleanup only. Based on the same branch, since the test uses the web-test-runner suite added there.

Making the fallback real, which `TimePicker.setLocale` javadoc promises, is a behavior change and is left out of this PR.

---

🤖 Generated with Claude Code

### DIFF
--- a/vaadin-time-picker-flow-parent/vaadin-time-picker-flow/src/main/resources/META-INF/frontend/vaadin-time-picker/timepickerConnector.js
+++ b/vaadin-time-picker-flow-parent/vaadin-time-picker-flow/src/main/resources/META-INF/frontend/vaadin-time-picker/timepickerConnector.js
@@ -23,11 +23,8 @@ window.Vaadin.Flow.timepickerConnector.initLazy = (timepicker) => {
       // Check whether the locale is supported by the browser or not
       TEST_PM_TIME.toLocaleTimeString(locale);
     } catch (e) {
-      locale = 'en-US';
       // FIXME should do a callback for server to throw an exception ?
-      throw new Error(
-        'vaadin-time-picker: The locale ' + locale + ' is not supported, falling back to default locale setting(en-US).'
-      );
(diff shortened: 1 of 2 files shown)


vaadin-time-picker-flow-parent/vaadin-time-picker-flow/src/main/resources/META-INF/frontend/vaadin-time-picker/timepickerConnector.js
```diff
@@ -23,11 +23,8 @@ window.Vaadin.Flow.timepickerConnector.initLazy = (timepicker) => {
       // Check whether the locale is supported by the browser or not
       TEST_PM_TIME.toLocaleTimeString(locale);
     } catch (e) {
-      locale = 'en-US';
       // FIXME should do a callback for server to throw an exception ?
-      throw new Error(
-        'vaadin-time-picker: The locale ' + locale + ' is not supported, falling back to default locale setting(en-US).'
-      );
+      throw new Error('vaadin-time-picker: The locale ' + locale + ' is not supported.');
```
vursen:
```suggestion
      throw new Error(`vaadin-time-picker: The locale ${locale} is not supported.`);
```

vursen (approved):

PR #9890, opened 2026-08-17
fix: focus selected item correctly when filtering

### LINKED ISSUE #9883: [combo-box] Wrong item focused when filtering on the client with setFocusSelectedItem
### Description

`ComboBox.setFocusSelectedItem(true)` resolves the selected item's index on the server and sends it to the client when the dropdown opens. The index is resolved against the **unfiltered** set of items, but the client applies it to the **filtered** list of items shown in the dropdown.

When the user types a filter to open the dropdown, the two lists no longer line up and a wrong item gets focused. Since the combo box commits the focused item when the dropdown is closed without an explicit selection (outside click or blur), the value silently changes to an item the user never picked.

This happens with in-memory items that fit into a single page (item count <= page size), where filtering is done on the client and the server never learns about the filter. Lazy data providers resolve the index through a filtered query, so they are not affected.

### Expected outcome

Typing a filter and clicking outside keeps the value unchanged. If an item is focused while a filter is active, it should be the selected item, matched by identity and not by index.

```java
List<String> items = new ArrayList<>();
IntStream.range(0, 10).forEach(i -> items.add("Apple " + i));
IntStream.range(0, 20).forEach(i -> items.add("Banana " + i));

comboBox.setItems(items);
// Index 15 in the full list, index 5 among the "Banana" items
comboBox.setValue("Banana 5");

comboBox.addValueChangeListener(e -> System.out.println("Value: " + e.getValue()));
```

Reproduction branch: https://github.com/vaadin/flow-components/tree/repro/9883

### Steps to reproduce

1. Add the snippet above to a view and open it
2. Focus the combo box input using <kbd>Tab</kbd> (without clicking it) and select all its text
3. Press `B`, so that the dropdown opens with the 20 "Banana" items
4. Notice that "Banana 15" is focused (highlighted) instead of "Banana 5": the server sent index 15, which is the index of "Banana 5" in the full 30-item list
5. Click outside the combo box to close the dropdown
6. Notice that the value changed to "Banana 15" and a value change event was fired

Observed in the browser after step 3:

```js
comboBox._focusedIndex;                              // 15
comboBox._dropdownItems.length;                      // 20
comboBox._dropdownItems[comboBox._focusedIndex];     // "Banana 15"
```

With `setFocusSelectedItem(false)` (the default) the same steps keep `_focusedIndex` at `-1` and the value stays "Banana 5".

Dropdown filtered to the 20 "Banana" items with "Banana 15" highlighted while the selected value is "Banana 5"

### Environment

Vaadin version(s): 25.2, 25.3-SNAPSHOT (`setFocusSelectedItem` was added in 25.2)
OS: macOS 15.6

### Browsers

Issue is not browser related

### Additional information

Root cause: `ComboBox.focusOnSelectedItem()` resolves the index via `getGenericDataView().getItemIndex(getValue())`. `ComboBoxDataView.getItems()` builds its query from the data communicator, which by design does not know about the client-side filter. The resulting index is then passed to the web component's `__focusIndex(index)`, which indexes into `_dropdownItems`, i.e. the filtered list.

The web component only guards partially: `__clearPendingFocusOnFilter` drops a *pending* index when the filter changes, and `__focusIndex` ignores indexes past the end of the filtered list. An index that happens to be in range is applied as is, and `_commitValue()` then commits `_dropdownItems[_focusedIndex]` as the new `selectedItem`.

The existing IT `ComboBoxFocusSelectedItemIT.inMemory_filterActive_doesNotErrorOut` does not catch this, because its filter leaves fewer items than the selected index, so `__focusIndex` returns early on the out-of-range check.

Related to #9239 (feature), #9396 (earlier fix).

web-padawan:
`setFocusSelectedItem(true)` resolves the selected item's index on the server when the dropdown opens, and the client applies it to the items the dropdown shows. Those are not always the same set of items: with in-memory items that fit a single page, the combo box filters on the client and the server never learns about the typed filter. The index then points at an unrelated item, and since closing the dropdown without an explicit selection commits the focused item, the value changed to an item the user never picked.

```java
List<String> items = new ArrayList<>();
IntStream.range(0, 10).forEach(i -> items.add("Apple " + i));
IntStream.range(0, 20).forEach(i -> items.add("Banana " + i));

comboBox.setItems(items);
// Index 15 in the full list, index 5 among the "Banana" items
comboBox.setValue("Banana 5");
```

Focus the combo box with <kbd>Tab</kbd>, select all its text and type `B`: "Banana 15" is focused instead of "Banana 5", and clicking outside changes the value to "Banana 15".

### Changes

- The connector resolves the selected item by its key among the items the dropdown shows. The server index is used only when the filter it was resolved against is still the one in effect, which is what a lazy data provider needs when the selected item is outside the loaded pages.
- A focus request that arrives before any items do is retried once they arrive. It is dropped when the dropdown closes or the value changes, so that a stale index cannot be applied to a later interaction.
- The call to the connector uses optional chaining, because opening the combo box from the server before it is attached schedules the call before the connector is initialized.

Fixes #9883

---

### DIFF
--- a/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/main/java/com/vaadin/flow/component/combobox/test/ComboBoxFocusSelectedItemPage.java
+++ b/vaadin-combo-box-flow-parent/vaadin-combo-box-flow-integration-tests/src/main/java/com/vaadin/flow/component/combobox/test/ComboBoxFocusSelectedItemPage.java
@@ -25,6 +25,7 @@ import com.vaadin.flow.component.combobox.dataview.ComboBoxLazyDataView;
 import com.vaadin.flow.component.html.Div;
 import com.vaadin.flow.component.html.NativeButton;
 import com.vaadin.flow.component.html.Paragraph;
+import com.vaadin.flow.component.html.Span;
 import com.vaadin.flow.dom.Element;
 import com.vaadin.flow.router.Route;
 
@@ -38,6 +39,15 @@ public class ComboBoxFocusSelectedItemPage extends Div {
     private static final String LAZY_WITH_PROVIDER = "lazy-with-provider";
     private static final String IN_MEMORY = "in-memory";
     private static final String LAZY_TOGGLE_OFF = "lazy-toggle-off";
+    private static final String CLIENT_FILTER = "client-filter";
+
+    // Fits into a single page, so that the combo box filters on the client and
+    // the server never learns about the filter. "Banana 5" is at index 15 of
+    // all items, and at index 5 among the items matching a "B" filter.
+    private static final List<String> CLIENT_FILTER_ITEMS = Stream
+            .concat(IntStream.range(0, 10).mapToObj(i -> "Apple " + i),
+                    IntStream.range(0, 20).mapToObj(i -> "Banana " + i))
+            .toList();
 
     public ComboBoxFocusSelectedItemPage() {
         ComboBox<String> withProvider = lazyCombo(LAZY_WITH_PROVIDER);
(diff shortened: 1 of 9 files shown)


web-padawan:
Closing since this is not the right fix that adds significant complexity.

Issue #9280, opened 2026-05-13
Card excludes slot children while accessing children


abdullahtellioglu:
### Description

In Copilot, all components are gathered through traversing the element tree starting from the UI element. The issue comes with the Card element, where slotted elements are explicitly filtered out. 

The ambiguity here is that you can access the parent from the child, but you cannot access the child from the parent using the same semantic methods. 

Additionally, it prevents iterating the component tree from top to bottom for us. There is another method in Card called `getFooterComponents` to access such components; however, the method is not provided by an interface.

To conclude, overriding `getChildren` method seems problematic because it breaks accessing children using Component and Element APIs. s  

### Expected outcome

Three solutions comes to my mind: 

- The first one is removing the overriding getChildren method. 
- The second one  is keeping it but implementing it from an interface, something like `HasFooterChildren` so that user of the Component API would know there is another method for accessing the slotted element
 - The third one is `getChildren` to return all components, including those in footers and headers; there could be another explicit method for returning non-slotted elements. 

### Minimal reproducible example

The example code: 

```java
Card card = new Card();
Button myButton = new Button("My Button");
card.addToFooter(myButton);
myButton.addAttachListener(event -> {
    Component parentComponent = myButton.getParent().get();
    System.out.println(parentComponent.getClass().getSimpleName() + " is the parent of myButton");
    String collect = parentComponent.getChildren().map(child -> child.getClass().getSimpleName()).collect(Collectors.joining(","));
    System.out.println("Card's children count is "+parentComponent.getChildren().toList().size() +", and class names = ["+collect+"]");
});
add(card);
```
Output of this code:
```java
Card is the parent of myButton
Card's children count is 0, and class names = []
```

### Steps to reproduce

1. Copy the snippet above to a view
2. Start the application
3. Open the view
4. Checkout console output.

### Environment

Flow: 25.2.0-alpha11
Vaadin: 25.2.0-alpha6
Spring Boot: 4.1.0-RC1
Spring: 7.0.7
Copilot: 25.2-SNAPSHOT
Frontend Hotswap: Vite
OS: aarch64 Mac OS X 26.2
Java: JetBrains s.r.o. 21.0.6
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/147.0.0.0 Safari/537.36
Spring Security: true
Spring Data JPA: true
Java Hotswap: Java Hotswap is enabled
[...]

PR #9893, opened 2026-08-18
fix: run signal-binding checks in Card index-based add

web-padawan:
## Description

Related to https://github.com/vaadin/flow-components/issues/9280

`Card.addComponentAtIndex` wrote to the element directly, so the checks in the default `HasComponents` implementation never ran. With a `bindChildren` binding active on the default slot, the inherited `Card.add(...)` already threw `BindingActiveException` while index-based adds silently changed the DOM and left it out of sync with the signal.

- Translated the default-slot index to an element index and delegated to `HasComponents.super.addComponentAtIndex(...)`, so the text- and children-binding checks run
  - Insert positions are unchanged: `appendChild` is `insertChild(getChildCount(), …)`, and `insertChild` already adjusts the index when the component is already a child
- Covered `addComponentAsFirst` by the same change, as the default implementation routes it through `addComponentAtIndex`
- Added `CardSignalTest` covering `bindChildren` on the default slot, slotted content surviving a binding, and `BindingActiveException` from both index-based add methods

## Type of change

- Behavior altering fix

## Note

With this change, application code that calls `addComponentAtIndex` or `addComponentAsFirst` while a children binding is active now gets a `BindingActiveException` instead of a silent DOM change (previous behavior).

### DIFF
--- a/vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
+++ b/vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
@@ -336,7 +336,6 @@ public class Card extends Component implements HasSize, HasAriaLabel,
 
     @Override
     public void addComponentAtIndex(int index, Component component) {
-        Objects.requireNonNull(component, "Component should not be null");
         if (index < 0) {
             throw new IllegalArgumentException(
                     "Cannot add a component with a negative index");
--- a/vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
+++ b/vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
@@ -352,13 +352,14 @@ public class Card extends Component implements HasSize, HasAriaLabel,
                             + children.size() + ").");
         }
 
-        if (index == children.size()) {
-            getElement().appendChild(component.getElement());
-        } else {
-            var reference = children.get(index);
-            var actualIndex = getElement().indexOfChild(reference.getElement());
-            getElement().insertChild(actualIndex, component.getElement());
-        }
+        // The default-slot index is relative to getChildren(), which excludes
+        // the slotted children (header, footer, ...) that are interleaved in
+        // the element tree. Translate it to the element index and delegate to
+        // the default implementation so its signal-binding checks still run.
+        var elementIndex = index == children.size()
+                ? getElement().getChildCount()
+                : getElement().indexOfChild(children.get(index).getElement());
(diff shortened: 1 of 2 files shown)


vaadin-card-flow-parent/vaadin-card-flow/src/test/java/com/vaadin/flow/component/card/CardSignalTest.java
```diff
@@ -0,0 +1,88 @@
...
+import com.vaadin.tests.AbstractSignalsTest;
+
+class CardSignalTest extends AbstractSignalsTest {
+
+    private Card card;
+
+    @BeforeEach
+    void setup() {
+        card = new Card();
+        UI.getCurrent().add(card);
```
sissbruecker:
```suggestion
        ui.add(card);
```

vaadin-card-flow-parent/vaadin-card-flow/src/test/java/com/vaadin/flow/component/card/CardSignalTest.java
```diff
@@ -0,0 +1,88 @@
...
+        Assertions.assertSame(header, card.getHeader());
+        Assertions.assertEquals(1, card.getChildren().count());
+    }
+
+    @Test
+    void childrenBindingActive_addComponentAtIndex_throws() {
+        var textSignal = new ValueSignal<>("Item 1");
+        var listSignal = new ValueSignal<>(List.of(textSignal));
+        card.bindChildren(listSignal, Span::new);
+
```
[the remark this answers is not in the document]
web-padawan: Improved tests to cover the fix, and removed redundant test that covered the default Flow behavior.

vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
sissbruecker, line 1: `Objects.requireNonNull(component, "Component should not be null");` at the start of the method could be removed, as the component is not referenced anymore in the override logic.

sissbruecker (approved):

vaadin-card-flow-parent/vaadin-card-flow/src/test/java/com/vaadin/flow/component/card/CardSignalTest.java
web-padawan: Done.

vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
web-padawan: Removed.

PR #9896, opened 2026-08-18
fix: reject Card removeAll while a children binding is active

web-padawan:
## Description

Follow-up to #9893

`Card` cannot delegate to the default `removeAll()`, which calls `Element.removeAllChildren()` and would clear the header, footer, title and media slots too, so it removes the default-slot children one by one instead. That loop does not run when the default slot is empty, so `removeAll()` silently did nothing while a children binding was active — and when the slot was not empty, the rejection came from `remove(...)` and named that method instead.

- Rejected `removeAll()` with `BindingActiveException` while a children binding is active, matching the contract the default implementation defines
- Added a private `hasChildrenBinding()` helper that reads the binding state from the element node, since Flow's own `throwIfChildrenBindingIsActive` and `hasChildrenBinding` are private to `HasComponentsOfType` — the same approach `Breadcrumbs` and `Button` already use
- Added tests for a non-empty bound list, asserting the message names `removeAll`, and for an empty bound list with a header set, where nothing was thrown before

## Type of change

- Bugfix

> [!NOTE]
> Only the children binding is checked, not the text binding that the default `removeAll()` also rejects: `Card` implements no `HasText` and `HasComponentsOfType` exposes no `bindText`, so a text binding cannot be established through Card's API.

### DIFF
--- a/vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
+++ b/vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
@@ -32,6 +32,8 @@ import com.vaadin.flow.component.html.Span;
 import com.vaadin.flow.component.shared.HasThemeVariant;
 import com.vaadin.flow.component.shared.SlotUtils;
 import com.vaadin.flow.dom.Element;
+import com.vaadin.flow.internal.nodefeature.SignalBindingFeature;
+import com.vaadin.flow.signals.BindingActiveException;
 
 /**
  * Card is a visual content container for creating a card-based layout.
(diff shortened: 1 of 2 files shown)


vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
sissbruecker, line 1: I was presented this issue too in a local review. IMO there's very little value in this change. Mutations are prevented in any case, otherwise I don't see much of a point to throw the exception if it doesn't result in faulty behavior otherwise. So I guess I'm asking: What is the concrete use case / scenario where this is useful?

sissbruecker (approved):
Anyway, the PR is correct, so I have no objections to merging. I guess my point is more about what we chose to spend time on and what value we get out of it.

vaadin-card-flow-parent/vaadin-card-flow/src/main/java/com/vaadin/flow/component/card/Card.java
web-padawan: I agree it's rather minor, decided to fix this as Claude flagged it as a problem while working on https://github.com/vaadin/flow-components/pull/9893.

PR #9911, opened 2026-08-19
fix: restore sub menu when a hidden menu item becomes visible

### LINKED ISSUE #1042: Sub-submenu does not construct when visibility changes
If a nested submenu item (sub-submenu) is not visible at the time its root item is made visible, then it won't be displayed after making the sub-submenu item visible. I guess the culprit could be a lack of `resetContent()` or something like that.

I know adding and removing the MenuItem instead of making it in/visible could solve the issue, but this is not feasible, because the use case for this is a deep hierarchical submenu, where the visibility of items changes dynamically and the parent item's visibility depends on the visibility of all child items in the submenu (i.e. hierarchical progressively constructed visibility predicate) & the order of items must be retained.

A workaround I found is to add and remove a dummy menu item, which forces to call `resetContent()`, but this causes a flicker of the MenuBar. This is annoying, because we need to check the visibility predicate frequently (on grid selection change or menu item click), and thus the MenuBar flickers each time a user clicks.

Code to reproduce
```java
final var menuBar = new MenuBar();
menuBar.addItem("Dummy");
final var rootItem = menuBar.addItem("Root");
rootItem.add(VaadinIcon.CHEVRON_DOWN_SMALL.create());

final var subMenu = rootItem.getSubMenu();
final var sub1 = subMenu.addItem("Sub 1");
final var subSub1 = sub1.getSubMenu().addItem("Sub sub 1");

final var sub2 = subMenu.addItem("Sub 2");
final var subSub2 = sub2.getSubMenu().addItem("Sub sub 2");
sub2.setVisible(false);
subSub2.setVisible(false);

final var btn1 = new Button("Toggle visibility vaadin/vaadin-menu-bar#1", event -> {
    subSub1.setVisible(!subSub1.isVisible());
    // hierarchical visibility
    sub1.setVisible(subSub1.isVisible());
    rootItem.setVisible(sub1.isVisible() || sub2.isVisible());
});
final var btn2 = new Button("Toggle visibility vaadin/vaadin-menu-bar#2", event -> {
    subSub2.setVisible(!subSub2.isVisible());
    // hierarchical visibility
    sub2.setVisible(subSub2.isVisible());
    rootItem.setVisible(sub1.isVisible() || sub2.isVisible());
});

add(menuBar, new HorizontalLayout(btn1, btn2));
```

web-padawan:
## Description

Fixes #1042
Depends on #9906

Flow withholds property changes for invisible elements, so a menu item that is hidden when
the items array is generated never receives the `_containerNodeId` that links it to its sub
menu. Showing the item later delivers the property, but the connector only re-syncs the
hidden and disabled state from the existing items tree, so the sub menu stayed lost forever.

- Regenerated the menu content when a parent item becomes visible, so the client rebuilds its
  items array and reads the `_containerNodeId` it previously had to skip
- Ran the same regeneration when visibility comes from a `bindVisible` signal, including the
  binding's initial application
  - A hidden item bound to a signal that is already `true` is how such an item is typically
    shown, and the server cannot tell that apart from binding an already visible item
- Skipped leaf items, which have no sub menu link to restore
- Added unit tests for the imperative and reactive paths, the leaf-item skip, repeat hide/show
  cycles, and several items shown in one round trip regenerating once
- Added integration tests for a menu-bar root item, a nested menu-bar sub item, and a context
  menu item hidden at first render

## Type of change

- Bugfix

## How to test

1. Open `vaadin-menu-bar/sub-menu-visibility`, click **Show root item**, then click the
   **Root item** button — the sub menu opens with `Sub item 1` and `Sub item 2`.
2. In the second menu bar, click **Show sub item**, open **Root item**, then hover
   **Sub item** — its sub menu opens with `Sub sub item`.
3. Open `vaadin-context-menu/sub-menu-visibility`, click **Show item**, right-click
   **target**, then open the sub menu of **Item** — it lists both sub items.

> [!NOTE]
> Showing a parent item while a menu overlay is open leaves that overlay empty until it is
> reopened. That is a pre-existing web-components defect — assigning `items` does not re-run
> the overlay renderer, and `menu-bar` never re-points an open sub menu at the new children —
> reproducible on `main` without this change by calling `addItem` from a `keepOpen` item's
> click handler. This change adds one more trigger for it. Fixing it belongs upstream:
> vaadin/web-components#5400 covers the context-menu half.

### DIFF
--- a/vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/java/com/vaadin/flow/component/contextmenu/MenuItemBase.java
+++ b/vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/java/com/vaadin/flow/component/contextmenu/MenuItemBase.java
@@ -315,6 +315,43 @@ public abstract class MenuItemBase<C extends ContextMenuBase<C, I, S>, I extends
         return HasComponents.super.bindEnabled(enabledSignal);
     }
 
+    @Override
+    public void setVisible(boolean visible) {
+        boolean becomingVisible = visible && !isVisible();
+        super.setVisible(visible);
+        if (becomingVisible) {
+            resetContentIfParentItem();
+        }
+    }
+
+    @Override
+    public SignalBinding<Boolean> bindVisible(Signal<Boolean> visibleSignal) {
+        // This also runs when the binding is first applied. Binding a hidden
+        // item to a signal that is already true is how such an item is
+        // typically shown, and that case is indistinguishable from binding an
+        // already visible item, so the content is regenerated for both.
+        return super.bindVisible(visibleSignal).onChange(context -> {
+            if (Boolean.TRUE.equals(context.getNewValue())) {
+                resetContentIfParentItem();
+            }
+        });
+    }
+
+    /**
+     * Regenerates the menu content so that the client rebuilds its items array.
+     * Flow withholds property changes for invisible elements, so an item that
+     * was invisible when the content was last generated ended up in that array
+     * without the sub menu that its {@code _containerNodeId} property points
+     * to. Showing the item delivers the property, but only a regeneration makes
+     * the client read it again.
+     */
+    private void resetContentIfParentItem() {
+        // Checking the field first avoids creating a sub menu for leaf items
+        if (subMenu != null && isParentItem()) {
+            contentReset.run();
+        }
+    }
+
     /**
      * Adds one or more theme names to this item. Multiple theme names can be
      * specified by using multiple parameters.
--- a/vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/test/java/com/vaadin/flow/component/contextmenu/MenuItemSignalTest.java
+++ b/vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/test/java/com/vaadin/flow/component/contextmenu/MenuItemSignalTest.java
@@ -15,10 +15,13 @@
  */
 package com.vaadin.flow.component.contextmenu;
 
+import java.util.List;
+
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
+import com.vaadin.flow.component.internal.PendingJavaScriptInvocation;
 import com.vaadin.flow.signals.local.ValueSignal;
 import com.vaadin.tests.AbstractSignalsTest;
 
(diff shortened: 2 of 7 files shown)


vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/java/com/vaadin/flow/component/contextmenu/MenuItemBase.java
```diff
@@ -315,6 +315,43 @@ public SignalBinding<Boolean> bindEnabled(Signal<Boolean> enabledSignal) {
...
+     * Regenerates the menu content so that the client rebuilds its items array.
+     * Flow withholds property changes for invisible elements, so an item that
+     * was invisible when the content was last generated ended up in that array
+     * without the sub menu that its {@code _containerNodeId} property points
+     * to. Showing the item delivers the property, but only a regeneration makes
+     * the client read it again.
+     */
+    private void resetContentIfParentItem() {
+        // Checking the field first avoids creating a sub menu for leaf items
+        if (subMenu != null && isParentItem()) {
```
sissbruecker, line 350: As an alternative, it seems reasonable to update `isParentItem` to check for existence of the sub menu first before creating one.

vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/java/com/vaadin/flow/component/contextmenu/MenuItemBase.java
sissbruecker: Nit: Comments are a bit too verbose for me.

vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/java/com/vaadin/flow/component/contextmenu/MenuItemBase.java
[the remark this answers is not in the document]
web-padawan: This is valid finding but not directly related to the current PR. Created https://github.com/vaadin/flow-components/pull/9926 to fix it.

vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/test/java/com/vaadin/flow/component/contextmenu/MenuItemSignalTest.java
```diff
@@ -61,4 +64,80 @@ void setDisableOnClickFalse_enabledBindingActive_doesNotThrow() {
...
+                .stream().filter(invocation -> invocation.getInvocation()
+                        .getExpression().contains("$connector.generateItems"))
+                .toList();
+    }
+
+    private void flushPendingInvocations() {
+        getPendingInvocations();
+    }
+
+    private List<PendingJavaScriptInvocation> getPendingInvocations() {
```
sissbruecker, line 139: This utility is not necessary, as `dumpPendingJavaScriptInvocations` already calls `fakeClientCommunication`

vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/test/java/com/vaadin/flow/component/contextmenu/MenuItemVisibilityTest.java
```diff
@@ -0,0 +1,160 @@
...
+                .stream().filter(invocation -> invocation.getInvocation()
+                        .getExpression().contains("$connector.generateItems"))
+                .toList();
+    }
+
+    private void flushPendingInvocations() {
+        getPendingInvocations();
+    }
+
+    private List<PendingJavaScriptInvocation> getPendingInvocations() {
```
sissbruecker, line 156: Same here.

vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/java/com/vaadin/flow/component/contextmenu/MenuItemBase.java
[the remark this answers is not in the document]
web-padawan: UPD: `generateItemsTree` will be fixed by #9943

vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/java/com/vaadin/flow/component/contextmenu/MenuItemBase.java
```diff
@@ -315,6 +315,43 @@ public SignalBinding<Boolean> bindEnabled(Signal<Boolean> enabledSignal) {
...
+            resetContentIfParentItem();
+        }
+    }
+
+    @Override
+    public SignalBinding<Boolean> bindVisible(Signal<Boolean> visibleSignal) {
+        // This also runs when the binding is first applied. Binding a hidden
+        // item to a signal that is already true is how such an item is
+        // typically shown, and that case is indistinguishable from binding an
+        // already visible item, so the content is regenerated for both.
```
sissbruecker, line 332: IMO the use case (binding hidden item to a signal) is not a common one, and the comment itself seems mostly unnecessary.

vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/test/java/com/vaadin/flow/component/contextmenu/MenuItemSignalTest.java
```diff
@@ -61,4 +64,80 @@ void setDisableOnClickFalse_enabledBindingActive_doesNotThrow() {
...
+        Assertions.assertEquals(0, getGenerateItemsInvocations().size());
+    }
+
+    @Test
+    void bindVisible_visibleParentItem_signalTurnsFalse_menuContentNotGenerated() {
+        item.getSubMenu().addItem("sub item");
+        var visibleSignal = new ValueSignal<>(true);
+        item.bindVisible(visibleSignal);
+        // Discards the regeneration from applying the binding, which is
+        // asserted by bindVisible_visibleParentItem_signalAlreadyTrue_*
```
sissbruecker, line 120: Unnecessary comment

web-padawan:
Closing in favor of https://github.com/vaadin/flow-components/pull/9947 since this can be also fixed in the connector.

PR #9926, opened 2026-08-21
fix: keep menu item tooltip when the item was hidden

web-padawan:
## Description

Related to #9911

The connector copies the tooltip off the item element into the generated items array, and
the web component then reads it from that array instead of from the element. Flow does not
send property changes for invisible elements, so an item that was hidden when the array was
generated ends up in it without a tooltip, and showing the item does not put one there.

- Added a `setTooltip` function to the context menu connector that writes the tooltip and
  its position into the generated items array, next to the existing `setChecked`,
  `setKeepOpen` and `setTheme` functions
- Called it from `setTooltipText` and `setTooltipPosition` through the item element, so Flow
  holds the call back for as long as the item is invisible and runs it once the item is shown
- Cancelled a call that has not been sent yet before scheduling the next one, so several
  tooltip changes in one round trip result in a single client call
- Added unit tests for a visible item, the deferral for a hidden one, the tooltip position,
  and repeated calls in one round trip
- Added a hidden item with a tooltip to `MenuBarTooltipPage` and an integration test that
  shows it and checks the tooltip

## Type of change

- Bugfix

## How to test

1. Open `vaadin-menu-bar/tooltip`
   (`vaadin-menu-bar-flow-parent/vaadin-menu-bar-flow-integration-tests/src/main/java/com/vaadin/flow/component/menubar/tests/MenuBarTooltipPage.java`).
2. Click **Show Item 3**.
3. Hover **Item 3** — the tooltip `Item 3 / Tooltip` is shown. Without this change no
   tooltip appears.

### DIFF
--- a/vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/java/com/vaadin/flow/component/contextmenu/MenuItemBase.java
+++ b/vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/java/com/vaadin/flow/component/contextmenu/MenuItemBase.java
@@ -27,6 +27,7 @@ import com.vaadin.flow.component.HasComponents;
 import com.vaadin.flow.component.HasEnabled;
 import com.vaadin.flow.component.HasText;
 import com.vaadin.flow.component.Tag;
+import com.vaadin.flow.component.page.PendingJavaScriptResult;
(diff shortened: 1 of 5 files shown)


web-padawan:
Closing in favor of https://github.com/vaadin/flow-components/pull/9943.

PR #9931, opened 2026-08-21
refactor: convert remaining connectors to classes

tomivirkki:
## Description

Related to #9789, #9850

Each of these connectors was one init function that assembled the `$connector`
object out of closures, which made it hard to tell what is state, what is the API
the server calls, and what is internal wiring. As classes those are separate. The
files stay plain JavaScript so that a later typing pass can move them as a rename
git can detect, the same way #9850 set up #9855.

- Converted `virtualListConnector` to a `VirtualListConnector` class, and removed a `requestAnimationFrame` callback that only read a function reference without ever calling it
- Converted `datepickerConnector` to a `DatePickerConnector` class, moving `_lastParseStatus` and `_lastParsedDate` off the public connector object into private fields, and the format-string helpers to module level
- Converted `timepickerConnector` to a `TimePickerConnector` class, holding the locale-derived values and the parse cache as private fields that `setLocale` resets
- Converted `menubarConnector` to a `MenuBarConnector` class, moving `__generatedItems` off the element into a private field
- Converted `contextMenuTargetConnector` to a `ContextMenuTargetConnector` class, storing the open event on the connector instead of resolving it from the target on every event
  - `updateOpenOn` now skips registering the listener when the connector was removed while waiting for `vaadin-context-menu` to be defined, which previously left a listener on the target with no way to remove it
- Converted `mapConnector` to a `MapConnector` class and split the event wiring into named private setup methods. `init` stays non-lazy, so the connector and its OpenLayers listeners are still re-created on every attach
- Kept `$connector.placeholderItem` and `list.__requestDebounce` reachable from outside the virtual list connector, because `VirtualListViewIT` and `VirtualListElement` read them

## Type of change

- Refactor

### DIFF
--- a/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/frontend/datepickerConnector.js
+++ b/vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/frontend/datepickerConnector.js
@@ -1,11 +1,7 @@
 import dateFnsFormat from 'date-fns/format';
 import dateFnsParse from 'date-fns/parse';
 import dateFnsIsValid from 'date-fns/isValid';
-import {
-  createDate,
-  extractDateParts,
-  parseDate as _parseDate
-} from '@vaadin/date-picker/src/vaadin-date-picker-helper.js';
+import { createDate, extractDateParts, parseDate } from '@vaadin/date-picker/src/vaadin-date-picker-helper.js';
 
 function createLocaleBasedDateFormat(locale) {
   try {
@@ -142,7 +138,7 @@ class DatePickerConnector {
   }
 
   #formatDate(dateParts, format) {
-    const date = _parseDate(`${dateParts.year}-${dateParts.month + 1}-${dateParts.day}`);
+    const date = parseDate(`${dateParts.year}-${dateParts.month + 1}-${dateParts.day}`);
 
     return dateFnsFormat(date, format);
   }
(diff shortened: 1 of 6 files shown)


vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/frontend/datepickerConnector.js
```diff
@@ -196,15 +116,110 @@ window.Vaadin.Flow.datepickerConnector.initLazy = (datepicker) => {
...
+    }
+
+    return {
+      formatDate: (dateParts) => this.#formatDate(dateParts, formats[0]),
+      parseDate: (dateString) => this.#parseDate(dateString, formats)
+    };
+  }
+
+  #formatDate(dateParts, format) {
+    const date = _parseDate(`${dateParts.year}-${dateParts.month + 1}-${dateParts.day}`);
```
web-padawan: suggestion: replace `parseDate as _parseDate` with `parseDate` as there is no more naming conflict.

web-padawan (approved):

vaadin-date-picker-flow-parent/vaadin-date-picker-flow/src/main/resources/META-INF/frontend/datepickerConnector.js
tomivirkki: Done

PR #9932, opened 2026-08-24
refactor: fully type remaining connector frontend files

tomivirkki:
## Summary
Follow-up to #9931, Related to https://github.com/vaadin/flow-components/pull/9855

- Converted the virtual list, date picker, time picker, menu bar, context menu target and map connectors from plain JavaScript to TypeScript, following the pattern #9855 set up for the combo box: changes to how a connector uses the web component API are now checked at compile time, and the connectors that still sat at the `META-INF/frontend` root moved into component-named subfolders as renames git can detect
- Private web component API each connector relies on is declared in type-only `.d.ts` files that ship with the connector, with small augmentations for API missing from the upstream typings (proj4); the map's synchronization modules stay JavaScript for a later typing pass
- `scripts/wtr.js` now type-checks a module's frontend files even when it has no WTR tests, so all the new module tsconfigs are enforced in CI

PR #9942, opened 2026-08-24
refactor: move grid frontend files into component subfolder

tomivirkki:
## Summary
Follow-up to #9932

- Moved `gridConnector.ts`, `treeGridConnector.ts` and `vaadin-grid-flow-selection-column.ts` from the `META-INF/frontend` root into the `vaadin-grid` subfolder, where the conventions say frontend files belong and where `vaadin-grid-types.d.ts` already lives, aligning grid with the other typed connector modules
- Updated the `@JsModule` paths, the relative type imports and the WTR test imports accordingly; the moves are renames git can detect


vursen:
I'm afraid this can break add-ons. I'm pretty sure I've seen some that import these connectos in their custom components.

web-padawan (approved):

tomivirkki:
> I'm afraid this can break add-ons. I'm pretty sure I've seen some that import these connectos in their custom components.

Good catch. See https://github.com/vaadin/flow-components/pull/9944

PR #9943, opened 2026-08-24
refactor: expose menu item state to items array via live getters

web-padawan:
## Description

The items array that the connector generates for the web component was a snapshot of the item elements' properties. Flow does not send property changes for invisible elements, so the snapshot could go stale, and each case needed its own connector call to patch the array. Reading the properties through getters keeps the array always current and removes the need for those calls.

- Changed `generateItemsTree` in the context menu connector to define the item state properties (`checked`, `keepOpen`, `disabled`, `className`, `theme`, `tooltip`, `tooltipPosition`) as getters reading from the item element
- Removed the connector functions that patched the generated items (`setKeepOpen`, `setTheme`, `menubarConnector.setClassName`) and the Java code calling them from `MenuItemBase`, `MenuBarItem` and `MenuBarRootItem`
- Reduced `setChecked` to toggling the `menu-item-checked` attribute, still needed because a menu kept open by a `keepOpen` item does not re-render its items on selection
- Removed the `_rootItem` reference and the `disabled` state copy from the menu bar connector; root buttons read fresh values every time the buttons re-render
- Fixed a tooltip set while a menu item is hidden not being shown after the item becomes visible, and added an integration test for it

## Type of change

- Refactor

## How to test

1. Open the integration test view `MenuBarTooltipPage.java` (`/vaadin-menu-bar/tooltip`)
2. Click the "Show Item 3" button
3. Hover over the "Item 3" button in the menu bar — the "Item 3 / Tooltip" tooltip is shown


web-padawan:
> Comment in updateButtons() still describes the removed connector behavior

Updated the comment.

vaadin-menu-bar-flow-parent/vaadin-menu-bar-flow/src/main/resources/META-INF/frontend/vaadin-menu-bar/menubarConnector.ts
```diff
@@ -76,13 +77,6 @@ export class MenuBarConnector {
...
     this.#generatedItems.forEach((item) => {
-      // Propagate disabled state from items to parent buttons
-      item.disabled = item.component.disabled;
-
-      // Saving item to component because `_item` can be reassigned to a new value
-      // when the component goes to the overflow menu
-      item.component._rootItem = item;
-
       this.#observer.observe(item.component, {
         attributeFilter: ['hidden', 'disabled'],
```
vursen, line 81: It seems we don't need to listen to disabled prop changes anymore.

```suggestion
        attributeFilter: ['hidden']
```

vaadin-menu-bar-flow-parent/vaadin-menu-bar-flow/src/main/java/com/vaadin/flow/component/menubar/MenuBarRootItem.java
```diff
@@ -116,8 +116,6 @@ public void removeClassNames(String... classNames) {
     }
 
     private void updateClassName() {
```
sissbruecker: Nit: Could call `menuBar.updateButtons()` directly, which might be a bit easier to follow.

sissbruecker (approved):

vaadin-menu-bar-flow-parent/vaadin-menu-bar-flow/src/main/resources/META-INF/frontend/vaadin-menu-bar/menubarConnector.ts
sissbruecker: Probably still needs to happen for root items? Though I could be missing something.

web-padawan: The filter still needs `disabled`: root buttons snapshot `item.disabled` in the lit template (`.disabled="${this.disabled || item.disabled}"`), which is only re-evaluated when `menuBar.items` is re-assigned, and this observer is the only remaining trigger for that on server-side `setEnabled` (nothing calls `updateButtons()` for enabled changes). Verified: with `['hidden']` only, `MenuBarPageIT.disable_overflow_openAndClose_unOverflow_buttonDisabled` fails. It also covers `setDisableOnClick`, which disables the element client-side without a round trip.

vaadin-menu-bar-flow-parent/vaadin-menu-bar-flow/src/main/java/com/vaadin/flow/component/menubar/MenuBarRootItem.java
web-padawan: Thanks, done.

PR #9947, opened 2026-08-25
fix: read menu item sub menu through a getter

### LINKED ISSUE #1042: Sub-submenu does not construct when visibility changes

web-padawan:
## Description

Fixes #1042
Follow-up to #9943

Flow does not send property changes for invisible elements, so an item that was hidden when
the items were generated was left without the container node id that links it to its sub
menu. Showing the item delivers the id, but the items were not generated again, so the sub
menu stayed missing. #9943 made the item state properties getters for the same reason; the
sub menu was the one field still read eagerly.

- Read the sub menu of a menu item through a getter in `generateItemsTree`, so the container
  node id is read whenever the web component renders instead of only when the items are
  generated
  - The result is cached, as the web component reads `children` on every render and the node
    id only changes when the items are generated anew
- Added integration tests for a menu bar root item, a nested menu bar sub item and a context
  menu item hidden at first render

## Type of change

- Bugfix

## How to test

1. Open `vaadin-menu-bar/sub-menu-visibility` (`MenuBarSubMenuVisibilityPage.java`), click
   **Show root item**, then click the **Root item** button — the sub menu opens with
   `Sub item 1` and `Sub item 2`.
2. In the second menu bar, click **Show sub item**, open **Root item**, then hover
   **Sub item** — its sub menu opens with `Sub sub item`.
3. Open `vaadin-context-menu/sub-menu-visibility` (`SubMenuVisibilityPage.java`), click
   **Show item**, right-click **target**, then open the sub menu of **Item** — it lists both
   sub items.


vaadin-context-menu-flow-parent/vaadin-context-menu-flow/src/main/resources/META-INF/frontend/contextMenuConnector.js
```diff
@@ -76,8 +76,20 @@ function generateItemsTree(appId, nodeId) {
...
+      let children;
+      Object.defineProperty(item, 'children', {
+        enumerable: true,
+        get() {
+          if (!children && child._containerNodeId) {
+            children = generateItemsTree(appId, child._containerNodeId);
+          }
+          return children;
+        }
+      });
```
vursen: Could we define this getter unconditionally and check _hasVaadinItemMixin inside it instead?

web-padawan: Done.

vaadin-context-menu-flow-parent/vaadin-context-menu-flow-integration-tests/src/test/java/com/vaadin/flow/component/contextmenu/it/SubMenuVisibilityIT.java
```diff
@@ -0,0 +1,52 @@
...
+        target = $("p").id("target");
+    }
+
+    @Test
+    public void hiddenItem_show_subMenuRenders() {
+        ContextMenuElement menu = ContextMenuElement.openByRightClick(target);
+        Assert.assertFalse("Expected the item to be hidden initially",
+                menu.getMenuItems().get(0).isDisplayed());
+        clickBody();
+        menu.waitUntilClosed();
```
sissbruecker, line 43: This test passes without the fix. It uses an auto-attached context menu which will automatically detach itself on close. Thus it will always regenerate items when it is auto-attached due to being opened again.

web-padawan: Thanks, updated the test to attach the menu manually.

sissbruecker (approved):

PR #9957, opened 2026-08-25
feat: add ToolException for LLM-safe tool errors

tomivirkki:
## Description

Part of https://github.com/orgs/vaadin/projects/103/views/3?pane=issue&itemId=231354372

Tool failures were replaced with a generic message, so the model could not learn why a call failed and was observed resubmitting the same query dozens of times in one turn.

- Added public `ToolException` to the provider package; its message is relayed verbatim to the LLM, while any other exception is logged and replaced with a generic error
- Grid, chart and form tools now use `ToolException` instead of their private validation exception types
- Both providers apply the same rule to application-supplied `ToolSpec` failures, which previously echoed every exception message to the model
- Malformed JSON arguments are still relayed, since the model produced them
- Documented the opt-in on `DatabaseProvider.executeQuery`, `ToolSpec.execute` and the tool `Callbacks` interfaces, and added tests for the relayed and generic paths

## Type of change

- Feature


tomivirkki:
> Chart configuration parse failures still reach the model as a detail-free generic error

Good catch, added a test case and fixed

guidelines/ai-components.md
```diff
@@ -106,10 +106,13 @@ The group has no `-testbench` module and, apart from `FormFieldMarker`'s
...
-- Error hygiene toward the model: forward only deliberately-safe validation
-  messages verbatim; replace any other exception with a generic string so
-  SQL, schema names, or paths never leak. Tool failures are returned as
-  strings to the model, never thrown.
+- Error hygiene toward the model: only the message of a
+  `ToolException` (public, in `com.vaadin.flow.component.ai.provider`) is
+  forwarded verbatim — throw it, from built-in tool code or an application's
+  `DatabaseProvider`/`ToolSpec`, only with deliberately-safe text. Any other
+  exception is caught, logged, and replaced with a generic string so SQL,
+  schema names, or paths never leak. Tool failures are returned as strings
```
ugur-vaadin, line 114: For annotated tools the leak still might be possible. In `LangChain4JLLMProvider` we use a `DefaultToolExecutor` constructor that sets `propagateToolExecutionExceptions` to `false`, so it catches the tool exception and returns its raw message as the result text. I think it is also similar for Spring AI. What do you think? Should we keep them out of scope?

tomivirkki: Confirmed: the `(Object, Method)` constructor hardcodes `propagateToolExecutionExceptions=false` and returns the raw message of any exception, and Spring AI's default `ToolExecutionExceptionProcessor` does the same.

Agreed on keeping vendor-annotated tools out of scope: their execution and error handling belong to the vendor framework. Documented the boundary in the javadocs and guideline.

ugur-vaadin (approved):
LGTM

PR #9971, opened 2026-08-27
fix: give parameterless AI tools a non-empty parameter schema

ugur-vaadin:
## Description

- Added `LLMProvider.ToolSpec.NO_PARAMETERS_SCHEMA`, an object schema with a
  single optional `reason` property that the tool ignores
  - A tool with no declared parameter schema makes models disagree on what to
    send as arguments, and Amazon Bedrock rejects the request that replays
    such a tool call with a 400
- Changed the built-in parameterless tools to return it instead of `null` or
  an empty `properties` object: `get_session_context`, `get_database_schema`,
  `get_grid_instructions`, `get_chart_instructions`, `get_form_instructions`
  and `get_form_state`
- Changed `SpringAILLMProvider` and `LangChain4JLLMProvider` to substitute
  `NO_PARAMETERS_SCHEMA` for a `null` or blank schema; a declared schema is
  passed through unchanged
- Changed `LangChain4JLLMProvider` to fall back to `NO_PARAMETERS_SCHEMA`
  instead of an empty schema when a schema string fails to parse
- Updated the `ToolSpec#getParametersSchema()` JavaDoc, which told
  implementors to return `null` for a tool with no parameters
- Added unit tests for the tool schemas, the provider substitution, and the
  pass-through of user-declared schemas

Fixes https://github.com/orgs/vaadin/projects/103/views/1?filterQuery=&pane=issue&itemId=231353831

## Type of change

- Bugfix

### DIFF
--- a/vaadin-ai-components-flow-parent/vaadin-ai-core-flow/src/main/java/com/vaadin/flow/component/ai/provider/LangChain4JLLMProvider.java
+++ b/vaadin-ai-components-flow-parent/vaadin-ai-core-flow/src/main/java/com/vaadin/flow/component/ai/provider/LangChain4JLLMProvider.java
@@ -338,9 +338,12 @@ public class LangChain4JLLMProvider implements LLMProvider {
         var builder = ToolSpecification.builder().name(tool.getName())
                 .description(tool.getDescription());
         var schema = tool.getParametersSchema();
-        if (schema != null && !schema.isBlank()) {
-            builder.parameters(parseParametersSchema(schema));
+        if (schema == null || schema.isBlank()) {
+            // A tool without a declared schema breaks some LLM APIs —
+            // see ToolSpec.NO_PARAMETERS_SCHEMA.
+            schema = LLMProvider.ToolSpec.NO_PARAMETERS_SCHEMA;
         }
+        builder.parameters(parseParametersSchema(schema));
         return builder.build();
     }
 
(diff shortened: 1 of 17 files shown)


tomivirkki (approved):

vaadin-ai-components-flow-parent/vaadin-ai-core-flow/src/main/java/com/vaadin/flow/component/ai/provider/LangChain4JLLMProvider.java
```diff
@@ -326,9 +326,12 @@ private static ToolSpecification toToolSpecification(
         var builder = ToolSpecification.builder().name(tool.getName())
                 .description(tool.getDescription());
         var schema = tool.getParametersSchema();
-        if (schema != null && !schema.isBlank()) {
-            builder.parameters(parseParametersSchema(schema));
+        if (schema == null || schema.isBlank()) {
+            // A tool without a declared schema breaks some LLM APIs —
+            // see ToolSpec.NO_PARAMETERS_SCHEMA.
+            schema = LLMProvider.ToolSpec.NO_PARAMETERS_SCHEMA;
```
tomivirkki, line 344: nit: this case seems untested

ugur-vaadin: If I didn't misunderstand the part you are referring to, they should be covered with the tests `stream_withExplicitToolNullSchema_substitutesNoParametersSchema` and `stream_withExplicitToolBlankSchema_substitutesNoParametersSchema`. If you meant something else, I can open a follow-up test coverage increase PR.

tomivirkki: Hmm...both tests are passing for me locally with this line `schema = LLMProvider.ToolSpec.NO_PARAMETERS_SCHEMA;` commented out. Not a blocker though

PR #9984, opened 2026-08-28
refactor: stop mandating get_grid_state on every request

tomivirkki:
## Summary
- `get_grid_state` returned nothing on a first question and the previous query on a follow-up, which invited the model to edit that query rather than write a new one — that is how a column keeps its name but loses its table qualifier, producing SQL that reads correctly and is wrong. It is now called only when the request changes what is already shown, with the new query written from the schema.
- The SQL examples selected from invented tables (`employees`, `sales`), which contradicts the instruction to use only tables from the schema the model was given. They are now unmistakably placeholders.


tomivirkki:
> The reworked workflow wording is not pinned by any test

We don't want unit tests asserting prompt wording. The instruction text gets tuned as models change, and pinning phrases means every tuning pass turns tests red without a defect behind it.

ugur-vaadin (approved):

Issue #9988, opened 2026-08-31
PiT 25.3: Six moved connectors lack the deprecated root entrypoints added by #9944


manolo:
## Summary

#9944 added deprecated root entrypoints for connector files that moved into per component subfolders, so existing imports of `@vaadin/flow-frontend/<name>.js` keep working. Four later migration PRs moved more connectors without adding those entrypoints, and six imports are now broken.

Separately, the connectors that were converted to TypeScript do not type check when a consuming project compiles them, because their sibling helper modules are still untyped JavaScript.

Both surfaced in `vaadin/docs`, which fails to build against `25.3.0-alpha9` and is therefore still pinned to `25.3.0-alpha8`. This blocks the docs update stage of every platform release.

## Problem 1: missing root entrypoints

Eleven connectors now live in subfolders and only five have a root entrypoint.

| Connector | Moved by | Root entrypoint |
|---|---|---|
| `gridConnector`, `treeGridConnector`, `comboBoxConnector` | #9942 | yes, added by #9944 |
| `virtualListConnector` | #9932 | no |
| `contextMenuConnector`, `contextMenuTargetConnector`, `menubarConnector` | #9955 | no |
| `uploadManagerConnector`, renamed from `vaadin-upload-manager-connector` | #9958 | no |
| `datepickerConnector` | | no |

Anything importing the old paths fails to resolve. In `vaadin/docs`, `frontend/demo/init-flow-components.ts` produces:

```
error TS2882: Cannot find module or type declarations for side-effect import
              of '@vaadin/flow-frontend/menubarConnector.js'
```

for all six. The upload one also changed name, so a plain path update is not enough for it.

## Problem 2: converted connectors do not type check in consumers

`vaadin-map/mapConnector.ts` was converted to TypeScript, but its siblings were not:

```
vaadin-map/mapConnector.ts        converted
vaadin-map/util.js                still JavaScript, no declarations
vaadin-map/synchronization/*.js   still JavaScript, no declarations
vaadin-map/vaadin-map-types.d.ts
```

A consumer compiling under `strict` gets:

```
error TS7016: Could not find a declaration file for module './synchronization/index.js'
error TS7016: Could not find a declaration file for module './util'
error TS2339: Property 'mapConnector' does not exist on type
              '{ gridConnector: ...; treeGridConnector: ... }'
```

[...]

PR #9989, opened 2026-08-31
refactor: convert map synchronization modules to TypeScript

tomivirkki:
## Description

Part of #9988

Consumers that type-check the connector in strict mode failed with TS7016 because the map connector imported plain JavaScript modules without type declarations (problem 2 in the issue).

- Converted `vaadin-map/util.js` and the `synchronization/*.js` modules to TypeScript
- Added `synchronization-types.d.ts` with types for the synchronized change objects
- Removed `allowJs` from the module's tsconfig now that no plain JavaScript modules remain

## Type of change

- Refactor


tomivirkki:
> Javadoc still points to synchronization/index.js, which the rename removed

Done

web-padawan (approved):

sissbruecker:
Let me take a second look.

sissbruecker (approved):

PR #10000, opened 2026-08-31
fix: relay ToolException raised while writing form values

tomivirkki:
## Summary
- An application marks a message safe for the model by throwing `ToolException`, but the form-fill path replaced it with `"Field rejected the value."`, so the model was told a field rejected its value without ever learning why — and resubmitted the same value.
- The fix covers both branches where application code runs during a fill, the converter and the write. Binder validators run from the value-change listener `setValue` fires, so they take the same path.


tomivirkki:
Closing: the form path already relays a validator's or converter's own message to the model verbatim, so a ToolException here would only ever mean an unexpected failure — the one case the generic message is for.

PR #10023, opened 2026-09-04
fix: respect disabled parent when syncing client disabled state

sissbruecker:
## Description

`DisableOnClickController` forces the client-side `disabled` property from `setEnabled`, because Flow does not send a change when a component is disabled and re-enabled in the same round trip. It used the value passed to `setEnabled`, so enabling a button under a disabled parent cleared `disabled` on the client while the server kept rejecting clicks.

- Changed `DisableOnClickController.onSetEnabled()` to schedule the client update with `beforeClientResponse` and read the effective enabled state at that point, so a disabled parent keeps the component disabled on the client
- Coalesced repeated `setEnabled` calls in one round trip into a single client update
- Deferred the update until the component is attached instead of queueing a JavaScript call per `setEnabled` call while detached
- Removed the unused `enabled` parameter from `onSetEnabled()` and updated `Button` and `MenuItemBase`
- Added unit tests for the disabled parent, same round trip, and detached cases

## Type of change

- Bugfix

### DIFF
--- a/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/internal/DisableOnClickController.java
+++ b/vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/internal/DisableOnClickController.java
@@ -58,6 +58,8 @@ public class DisableOnClickController<C extends Component & HasEnabled>
     public DisableOnClickController(C component) {
         this.component = Objects.requireNonNull(component);
 
+        component.addDetachListener((event) -> clientUpdateScheduled = false);
+
         ComponentUtil.addListener(component, ClickEvent.class,
                 (ComponentEventListener) (event -> {
                     if (isDisableOnClick()) {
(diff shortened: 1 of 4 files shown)


vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/internal/DisableOnClickController.java
```diff
@@ -95,18 +96,29 @@ public boolean isDisableOnClick() {
...
+        // trip, Flow will not detect any changes and the client side component
+        // would not be enabled again. The property is updated before the
+        // response so that the effective state at that point is used, for
+        // example when a parent is disabled or enabled in the same round trip.
+        if (clientUpdateScheduled) {
+            return;
+        }
+        clientUpdateScheduled = true;
+        component.getElement().getNode().runWhenAttached(
+                ui -> ui.beforeClientResponse(component, context -> {
```
tomivirkki, line 120: nit: Claude discovered a case where a button could get stuck in a disabled state:
```java
    @Test
    void detachedBeforeResponse_attachedToAnotherUi_setEnabled_clientSideDisabledPropertyUpdated() {
        component.setEnabled(false);
        component.getElement().removeFromTree();
        ui.fakeClientCommunication();

        ui.replaceUI();
        ui.add(component);
        ui.fakeClientCommunication();

        component.setEnabled(true);
        Assertions.assertEquals(List.of(false), dumpClientSideDisabledValues());
    }

```
But since it's such an unrealistic case, feel free to ignore

tomivirkki (approved):

vaadin-flow-components-shared-parent/vaadin-flow-components-base/src/main/java/com/vaadin/flow/component/shared/internal/DisableOnClickController.java
sissbruecker: Yeah, it seems unlikely. The fix was a single detach listener, so I added it.

PR #10030, opened 2026-09-04
feat: add mode to disable on click until response is handled

### LINKED ISSUE #1553: Automatically re-enable button at the end of the response when using setDisableOnClick
Can you please include a 'debounce' feature or improve the setDisableOnClick method for buttons (as I have a lot of older users who like to double-click everywhere, and the occasional user who sometimes accidentally double-clicks on buttons).

Something like:

```java
setDebounceClick(long millis);
```

or overload the existing method, to re-enable it after X millis since clicking:

```java
setDisableOnClick(long timeoutMillis);
```

or, preferably

```java
setDisabledUntilServerReturns(boolean disabled);   // Disables until componenteventlistener has returned
```

Any would provide a simply way to prevent accidental double-clicks.  setDisableOnClick(bool) as it stands isn't really that helpful.  This should really be applied to any clickable item in flow.

sissbruecker:
## Description

Closes #1553

[PR explanation](https://github.com/user-attachments/files/31849053/explanation-10030.html)

Disable on click protects against accidental double clicks, but until now the button stayed disabled until the application enabled it again from a click listener. For the common case, where the button should only be blocked while the server handles the click, that meant extra code in every listener. This adds a mode where the button is enabled again automatically once the click has been handled, so preventing double submits becomes a one-liner.

- Added `DisableOnClickMode` with two modes for how long a button stays disabled after it was disabled on click
  - `UNTIL_ENABLED` is the existing behavior and the default: the button stays disabled until the application enables it again
  - `UNTIL_RESPONSE` enables the button again automatically once the click has been handled, with the response to the click
- Added `Button.setDisableOnClick(DisableOnClickMode)` and `Button.getDisableOnClickMode()`
- In `UNTIL_RESPONSE` mode, a click listener can still decide the final state by calling `setEnabled(true)` or `setEnabled(false)`, in which case the automatic enabling is skipped for that click
- Clicks that reach the server in the same round trip after the first one are ignored, so a fast double click runs the listener only once
- The automatic enabling also works when the click listener detaches the button, for example by closing a dialog, and the button is attached again later, also in another UI after a reload with `@PreserveOnRefresh`
- Clarified in the Javadoc that the default mode requires the application to enable the button again
- Added unit tests for both modes and an integration test view and tests for `UNTIL_RESPONSE`

## Type of change

- Feature

> [!NOTE]
> `UNTIL_RESPONSE` enables the button with the next response the server sends after the click. If a click listener calls `UI.push()` while it is still running, that push is the next response and enables the button before the listener has finished. This is documented on the enum constant.

### DIFF
--- a/vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/main/java/com/vaadin/flow/component/button/tests/ButtonView.java
+++ b/vaadin-button-flow-parent/vaadin-button-flow-integration-tests/src/main/java/com/vaadin/flow/component/button/tests/ButtonView.java
@@ -29,6 +29,7 @@ import com.vaadin.flow.component.html.Paragraph;
 import com.vaadin.flow.component.icon.Icon;
 import com.vaadin.flow.component.icon.VaadinIcon;
 import com.vaadin.flow.component.orderedlayout.VerticalLayout;
+import com.vaadin.flow.component.shared.DisableOnClickMode;
 import com.vaadin.flow.component.textfield.TextField;
 import com.vaadin.flow.data.value.ValueChangeMode;
 import com.vaadin.flow.router.Route;
@@ -49,6 +50,7 @@ public class ButtonView extends Div {
         createDisabledButton();
         createButtonWithDisableOnClick();
         createButtonWithDisableOnClickThatEnablesInSameRoundTrip();
+        createButtonWithDisableOnClickUntilResponse();
         createButtonWithDisableOnClickThatIsHidden();
         createButtonWithDisableOnClickAndPointerEventsAuto();
         createButtonsWithShortcuts();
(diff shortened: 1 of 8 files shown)


DiegoCardoso (commented):
Some feedback, after reviewing the changes:

- `MenuItemBase` also has the `setDisableOnClick` feature, should it also receive the new mode API? Also, the `BeforeClientResponseAction` rewrite changes its behavior too (a `this.disabled` update now replays on re-attach). No context-menu test covers it.

**`UNTIL_RESPONSE`: only `setEnabled` cancels the pending enable**

A click listener that switches the mode instead of calling `setEnabled(false)` gets the opposite of what it asked for — the button is enabled again with the response:

```java
button.setDisableOnClick(DisableOnClickMode.UNTIL_RESPONSE);
button.addClickListener(e -> button.setDisableOnClick(DisableOnClickMode.UNTIL_ENABLED));
// after the response: button.isEnabled() == true
```

`onSetEnabled()` is the only place that calls `enable.cancel()`. Cancelling it in `setDisableOnClick(boolean)`'s else branch and in `setDisableOnClick(DisableOnClickMode)` when the mode is `UNTIL_ENABLED` fixes it without breaking any existing test.

**Coverage gaps worth one test each**

- A click listener that throws. `UNTIL_RESPONSE` recovers where `UNTIL_ENABLED` leaves the button stuck forever — arguably the strongest argument for the feature, and nothing asserts it.
- A click listener that calls `setVisible(false)`. `UIInternals` retains the `this.disabled` invocation while the owner is invisible and the server-side `false → true` flip nets to no change, so that invocation is the only carrier of the final state. It works, but this is the scenario that already needed its own test for `UNTIL_ENABLED` (`disableOnClick_hideWhenDisabled_showWhenEnabled_clientSideButtonIsEnabled`).
- Re-attach timing: after a listener detaches the button, `isEnabled()` stays `false` past the re-attach until the next response. Worth documenting on `UNTIL_RESPONSE` if intended.

sissbruecker:
- Menu items can be done as a follow up. I don't understand which context-menu test should be added as Button doesn't have dedicated tests either.
- Switching the mode seems unlikely. Also, by the same reasoning you'd end up with a disabled button if you would switch the mode back and forth in the same trip.
- Test coverage gaps seem like stuff that is mostly covered by Flow behavior. I think the current coverage is fine.

DiegoCardoso (approved):
> I don't understand which context-menu test should be added as Button doesn't have dedicated tests either.

Disregard this. I re-checked the claims from the agent, but didn't challenge this claim.

LGTM

PR #10049, opened 2026-09-08
fix: fail build when plot options schemas are not generated

tomivirkki:
## Summary
- The chart plot options schemas are generated at build time from the chart model sources and shipped in the jar. When the source directory was missing the generator only logged a warning and skipped, so a build with tests skipped could publish a jar without schemas and the tool would report every chart type as unknown. The generator now fails the build in that case, and a missing resource at runtime is logged as an error.
- Added tests that check the generated resource is on the classpath, covers every supported chart type with descriptions, and is exposed through `ChartAIController`'s tools, so a regression in generation or packaging fails the module's unit tests directly.


ugur-vaadin (approved):

ugur-vaadin:
Should we also target v25.3?

tomivirkki:
> Should we also target v25.3?

I was originally planning to exclude it from 25.3, but shouldn't hurt to have it there either. Let's keep the CP PR in draft for a while just to be sure it doesn't affect build

PR #10050, opened 2026-09-08
refactor: simplify TreeGrid connector scrollToIndex

vursen:
## Description

`scrollToIndex` asks the server to preload the range around the target. The response clears the old range while its rows are still rendered, so those rows request their pages again. The connector avoided this by blocking page loading for as long as a `scrollToIndex` call was pending. This change removes the block and instead resolves the stale requests right after the scroll, before they turn into a server request.

- Removed the `_shouldLoadCachePage` override from `treeGridConnector.ts`
- Made `scrollToIndex` call `resolvePendingCallbacks()` after scrolling to the preloaded range

## Type of change

- Refactor

### DIFF
--- a/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/frontend/vaadin-grid/treeGridConnector.ts
+++ b/vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/frontend/vaadin-grid/treeGridConnector.ts
@@ -33,20 +33,11 @@ function initLazy(grid: FlowTreeGrid) {
 
   window.Vaadin.Flow.gridConnector.initLazy(grid);
 
-  grid._dataProviderController._shouldLoadCachePage = function (cache, page) {
-    // `$server.setViewportRangeByIndexPath` sends a preloaded viewport range based on
-    // the provided index path and `padding` parameter. Applying the new range clears
-    // the old range, which is still visible because the actual scroll happens only
-    // after all connector calls in that update are processed. This check prevents
-    // the old range from being unnecessarily re-requested while the new range is
-    // still being processed, which could cause flickering.
-    return !grid.__pendingScrollToIndexes;
-  };
-
   grid.scrollToIndex = async function (...indexes) {
-    grid.__pendingScrollToIndexes = indexes;
-
     if (!grid.clientHeight || !grid._columnTree || grid._dataProviderController.isLoading()) {
+      // Not ready yet. The web component retries from __scrollToPendingIndexes
+      // once a page has loaded or the grid has become visible.
+      grid.__pendingScrollToIndexes = indexes;
       return;
     }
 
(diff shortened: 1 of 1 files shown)


vursen:
The fix can be backported to 25.3 as is, but backporting to 25.2 and 25.1 would require a slightly different solution because there is no `$connector.resolvePendingRequests()` in those versions.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/frontend/vaadin-grid/vaadin-grid-types.d.ts
```diff
@@ -96,7 +96,7 @@ export interface FlowGridInternals {
 }
 
 /** The Flow grid element */
-export type FlowGrid = Grid<Item> & FlowGridInternals;
+export type FlowGrid = FlowGridInternals & Grid<Item>;
```
vursen: **note:** Changing the order makes `FlowGridInternals` (and `FlowTreeGridInternals`) take priority over `Grid<Item>`'s properties

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/test/shared.ts
```diff
@@ -38,13 +41,13 @@ export type GridConnector = Omit<ConnectorGridConnector, 'doSelection' | 'doDese
   updateFlatData(updatedItems: Item[]): void;
 };
 
-export type FlowGrid = {
+type TestGridInternals = {
   $connector: GridConnector;
   $server: GridServer;
-} & ConnectorFlowGrid & {
-    _flatSize: number;
-    _updateItem: (index: number, item: Item) => void;
```
vursen, line 46: **note:** `_updateItem` was left out because of no usages.

vaadin-grid-flow-parent/vaadin-grid-flow/src/main/resources/META-INF/frontend/vaadin-grid/treeGridConnector.ts
```diff
@@ -55,7 +46,12 @@ function initLazy(grid: FlowTreeGrid) {
     const flatIndex = await grid.$server.setViewportRangeByIndexPath(indexes, padding);
```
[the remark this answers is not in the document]
vursen: The issue isn't serious: the expected viewport will just get re-requested after scrolling, resulting in one extra round trip. Besides, fixing that was never the goal of the original solution.

vaadin-grid-flow-parent/vaadin-grid-flow-integration-tests/test/tree-grid-connector-scroll-to-index.test.ts
```diff
@@ -0,0 +1,128 @@
...
+
+        // Scroll past the loaded page
+        grid.$.table.scrollTop = 100 * 36;
+        await nextFrame();
+        await aTimeout(GRID_CONNECTOR_ROOT_REQUEST_DELAY);
+        expect(grid.$server.setViewportRange).to.be.calledOnce;
+      });
+    });
+
+    describe('initially non-empty data set', () => {
```
sissbruecker: This does exactly the same as `should run the deferred scroll when the data set becomes non-empty`

sissbruecker (approved):

vursen:
This fix kind of solves the reported issue, but I also wonder if it's correct that the grid attempts to scroll to an index after first being rendered with an empty data set and then a non-empty one.

PR #10100, opened 2026-09-10
fix: ask for form value sources as a fill_form parameter

### LINKED ISSUE #10071: Hackathon 25.3: Source tracking is enabled and no model ever reports a source
### Description

```java
controller.setSourceTrackingEnabled(true);
```

then a real turn against OpenAI, then `controller.getFieldSource(field)` for every field the model filled. Empty, every time, with `gpt-4o-mini` and with `gpt-4o`, over more than a dozen live runs.

The plumbing works. Hand the same controller a fill that carries `confidence` and `extracts` and the value comes back and the marker's popover renders it, so the parser and the API are fine. What never happens is the model producing that shape.

The `fill_form` parameter schema is

```json
{ "values": { "type": "object", "additionalProperties": true } }
```

which says nothing about a source, so whether one arrives rests entirely on prose in the tool description.

### Why it matters

Source tracking is the headline of the 25.3 form controller and the reason an application turns it on. Without it the field marker has a badge, an explanation and a revert control, and nothing to say about where the value came from.

We also asked for it in our own system prompt, which the documentation says outranks the controller's instructions, and it changed nothing.

### Expected

Put `confidence` and `extracts` in the `fill_form` schema, so the model is asked for them by the shape of the call rather than by a sentence.

### Reproduce

Any `FormAIController` with source tracking on, a real key, and one prompt that fills a field. Then `getFieldSource` on that field.

### What a live run showed

Reproduced with the project below, `setSourceTrackingEnabled(true)`, one clean turn against `gpt-4o-mini` that filled two fields:

```
name    value=Ana Torres    source=(none)
phone   value=600123456     source=(none)
```

The field marker renders and its popover opens, and what it has to say is "This field value was modified by AI." and a Revert Value button. Nothing about where the value came from, because there is nothing to say.

### Getting the project

[`ai-live.zip`](https://github.com/vaadin/hackathon-25-3/raw/f36086c0050c7fc2fc175ed169a155c3ba418413/specs/issues/projects/ai-live.zip), 10 KB, sources only: four routes, one per finding, and the feature flag file the AI components need.

```
export OPENAI_API_KEY=...
mvn spring-boot:run
```

Then `http://localhost:8140/form`. It really calls the model, so it costs a few cents a run.

Found on 25.3.0-beta1 with `vaadin-ai-extensions-flow` and Spring AI 2.0.0.

tomivirkki:
Fixes #10071

## Summary
- Source tracking relied on prose in the `fill_form` description asking the model to wrap document-derived values in an envelope. gpt-4o follows it, gpt-4o-mini never does, so smaller models filled every field but reported no source.
- While source tracking is on, `fill_form` now declares a `sources` parameter next to `values`, keyed by the same field ids and carrying the confidence level and extracts. Asked for by the schema, gpt-4o-mini reports a source for every value read from an attachment and still leaves prompt-derived values without one.
- When source tracking is off the tool definition is byte-identical to before, and the public API (`getFieldSource()`, `ValueSource`, the field marker) is unchanged.


vaadin-ai-components-flow-parent/vaadin-ai-extensions-flow/src/test/java/com/vaadin/flow/component/ai/form/SourceTrackingTest.java
```diff
@@ -1007,13 +1144,19 @@ private static JsonNode fill(FormAIController controller,
...
-        return """
-                {"value": "%s", "confidence": "high", "extracts": [
-                  {"text": "snippet", "location": {"type": "page-region",
-                   "page": 1, "rect": [0.1, 0.2, 0.3, 0.04]}}]}"""
-                .formatted(value);
+    /**
+     * Executes {@code fill_form} with a single-field payload whose value is the
+     * given JSON text and whose source, under the same field id, is the given
+     * JSON text, and returns the parsed response.
+     */
```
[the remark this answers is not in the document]
tomivirkki: Done, tests added

vaadin-ai-components-flow-parent/vaadin-ai-extensions-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAITools.java
```diff
@@ -33,6 +37,86 @@ final class FormAITools {
...
+                  "additionalProperties": true
+                },
+                "sources": {
+                  "type": "object",
+                  "description": "For every value in \\"values\\" that was read from an attached document, where it was read from, keyed by the same field id.",
+                  "additionalProperties": {
+                    "type": "object",
+                    "properties": {
+                      "confidence": {
+                        "type": "string",
```
ugur-vaadin, line 78: I wonder if having a description here would help.

ugur-vaadin (approved):
Added one comment.

vaadin-ai-components-flow-parent/vaadin-ai-extensions-flow/src/main/java/com/vaadin/flow/component/ai/form/FormAITools.java
tomivirkki: Added the missing description

PR #10104, opened 2026-09-11
test: assert the badge renders its content, not only that its text is in the DOM

manolo:
## Summary

- `BadgeBasicIT` asserted only `BadgeElement.getText()`, which reads the light DOM text nodes with a script. The text is there even when the badge renders nothing, so the test cannot fail for the one thing this tier exists to check.
- Adds an assertion on what the browser paints: `has-content` on the host, and a non zero `offsetWidth` for `[part="content"]`.

## Context

This is the blind spot behind vaadin/web-components#12663. Flow appends an empty text node and writes its data afterwards, the badge never received `has-content`, and every Flow badge carrying text rendered as an empty pill with its text sitting in the DOM. No tier saw it: the unit and visual tests upstream set the text with `element.textContent`, which replaces the node and fires `slotchange`, and this IT read the light DOM.

The new assertion fails against the badge this module installs today, `@vaadin/badge` 25.3.0-beta2, which is the newest published version, and passes with vaadin/web-components#12664 patched into `node_modules`. That fix is merged and not published yet, so this is a draft: it is ready to merge once the badge pin moves to a version carrying it, and until then it is the check that says whether the pin is good.

### DIFF
--- a/vaadin-badge-flow-parent/vaadin-badge-flow-integration-tests/src/test/java/com/vaadin/flow/component/badge/tests/BadgeBasicIT.java
+++ b/vaadin-badge-flow-parent/vaadin-badge-flow-integration-tests/src/test/java/com/vaadin/flow/component/badge/tests/BadgeBasicIT.java
@@ -21,6 +21,7 @@ import org.junit.Test;
 
 import com.vaadin.flow.component.badge.testbench.BadgeElement;
 import com.vaadin.flow.testutil.TestPath;
+import com.vaadin.testbench.TestBenchElement;
 import com.vaadin.tests.AbstractComponentIT;
 
 @TestPath("vaadin-badge/basic")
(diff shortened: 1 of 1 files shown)


manolo:
Measured both ways on this branch, `mvn -pl vaadin-badge-flow-parent/vaadin-badge-flow-integration-tests verify -Dit.test=BadgeBasicIT`, with the module installing `@vaadin/badge` 25.3.0-beta2:

Without the fix:

```
Tests run: 2, Failures: 1, Errors: 0, Skipped: 0
BadgeBasicIT.badgeContentRendered:54 Badge should have the has-content attribute
```

With vaadin/web-components#12664 applied to `node_modules/@vaadin/badge` and the Vite dep cache cleared:

```
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0
```

`badgeDisplayed` is green in both, which is the point of the change.

vaadin-badge-flow-parent/vaadin-badge-flow-integration-tests/src/test/java/com/vaadin/flow/component/badge/tests/BadgeBasicIT.java
```diff
@@ -38,4 +39,24 @@ public void badgeDisplayed() {
...
+     * text and cannot tell the two states apart, so this asserts what the
+     * browser paints instead.
+     */
+    @Test
+    public void badgeContentRendered() {
+        var badge = $(BadgeElement.class).id("badge");
+        Assert.assertTrue("Badge should have the has-content attribute",
+                badge.hasAttribute("has-content"));
+
+        var content = badge.$(TestBenchElement.class)
```
web-padawan: nit: IMO it's enough to check for `has-content` presence - it's used to set [display](https://github.com/vaadin/web-components/blob/e4d403bd1786c6c6daaac4712746446a88f65912/packages/badge/src/styles/vaadin-badge-base-styles.js#L47-L49) property on the `part`.

manolo: Agreed, done in 11f3fb5. The base styles map the attribute to `display` one to one, so the measurement was only restating it, and the attribute gives the better failure message.

For the record on why both were there: the gap this closes is an assertion that passed while nothing was painted, so I started from the painted side. Happy to swap it back to the measurement instead of the attribute if you would rather the IT not name an internal attribute at all, but I think your version is the right one here.

The failure without vaadin/web-components#12664 is on this assertion, so the simplification changes neither of the two runs in the comment above.

PR #10116, opened 2026-09-14
feat: add typing indicator API to MessageList

tomivirkki:
## Summary
- Chat and assistant views need to show who is currently typing.
- Typing users are modelled as a `MessageListUser` type instead of flat fields, because the indicator shows an avatar: name, abbreviation, image and color index, with `DownloadHandler` images registered as a resource of the list.
- The whole API is experimental behind `com.vaadin.experimental.messageListTypingIndicator`, or the `aiComponents` umbrella flag, so its shape can still change while the AI components adopt it.


vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
```diff
@@ -388,6 +407,250 @@ public boolean isAnnounceMessages() {
...
+     *            the i18n object, not {@code null}
+     * @since 25.3
+     */
+    public void setI18n(MessageListI18n i18n) {
+        useTypingIndicator();
+        this.i18n = Objects.requireNonNull(i18n,
+                "The i18n object should not be null");
+        var typingIndicatorText = i18n.getTypingIndicatorText();
+        if (typingIndicatorText == null) {
+            getElement().removeProperty(TYPING_INDICATOR_TEXT_PROPERTY);
```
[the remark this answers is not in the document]
tomivirkki: Fixed in https://github.com/vaadin/web-components/pull/12753

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
```diff
@@ -388,6 +407,250 @@ public boolean isAnnounceMessages() {
...
+    public <S extends Signal<MessageListUser>> SignalBinding<Collection<MessageListUser>> bindTypingUsers(
+            Signal<List<S>> typingUsersSignal) {
+        useTypingIndicator();
+        Objects.requireNonNull(typingUsersSignal, "Signal cannot be null");
+        return typingUsersSupport.bind(() -> typingUsersSignal.get().stream()
+                .map(Signal::get).toList());
+    }
+
+    private void updateTypingUsers(Collection<MessageListUser> typingUsers) {
+        var users = new ArrayList<>(typingUsers);
```
ugur-vaadin, line 509: It would be good to have a null check for the users here too. I think there is no check for the signal path currently, so it might throw a NPE on `user.setHost(this)` below.

vaadin-messages-flow-parent/vaadin-messages-flow-integration-tests/src/test/java/com/vaadin/flow/component/messages/tests/MessageListIT.java
```diff
@@ -342,6 +343,216 @@ public void clickImageAttachment_eventFired() {
                 clickedAttachment.getText());
     }
```
ugur-vaadin, line 345: It might be good to have a locale change IT.

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
```diff
@@ -388,6 +407,250 @@ public boolean isAnnounceMessages() {
...
+    }
+
+    /**
+     * Gets the style in which the typing indicator is rendered.
+     *
+     * @return the typing indicator type,
+     *         {@link MessageListTypingIndicatorType#DEFAULT} by default
+     * @since 25.3
+     */
+    public MessageListTypingIndicatorType getTypingIndicatorType() {
```
ugur-vaadin, line 549: Should this and the other new methods that don't have it also get the experimental API note in their JavaDocs?

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
```diff
@@ -388,6 +407,250 @@ public boolean isAnnounceMessages() {
...
+     * This API is experimental and needs to be enabled with the
+     * {@code com.vaadin.experimental.messageListTypingIndicator} or
+     * {@code com.vaadin.experimental.aiComponents} feature flag.
+     *
+     * @param typingUsers
+     *            the users that are currently typing, not {@code null} and not
+     *            containing any {@code null} users
+     * @since 25.3
+     */
+    public void setTypingUsers(Collection<MessageListUser> typingUsers) {
```
ugur-vaadin: I wonder if we should use a set.

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
tomivirkki: Done. The null checks moved from the setter into the private update method that both the setter and the binding go through, so the signal path is covered too. Test added.

vaadin-messages-flow-parent/vaadin-messages-flow-integration-tests/src/test/java/com/vaadin/flow/component/messages/tests/MessageListIT.java
tomivirkki: Test added. The indicator is rendered in the same render pass as the messages, so a locale change re-renders it, and the test checks it survives that.

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
tomivirkki: Done.

vaadin-messages-flow-parent/vaadin-messages-flow/src/main/java/com/vaadin/flow/component/messages/MessageList.java
tomivirkki: Maybe not a Set, but the type was indeed too vague, so it takes a List now.

MessageListUser has no equals/hashCode, so a Set would only drop the exact same instance, not the same person added twice. Order is part of the contract though: the client property is an array, and the web component joins the names and renders the avatars in that order. getTypingUsers and bindTypingUsers already spoke List, so the setter was the odd one out.

ugur-vaadin (approved):

PR #10129, opened 2026-09-15
feat: show typing indicator while AIOrchestrator waits

tomivirkki:
## Summary
- An AI chat used to show an empty assistant bubble from the moment a prompt was sent until the response arrived. Building on the `MessageList` typing indicator API from #10116, `AIOrchestrator` now shows the assistant in the typing indicator and adds the assistant message when the first token arrives, so a turn without response text, such as a tool-only turn, no longer leaves an empty message behind.
- `AIMessageList` gains default no-op `showTypingIndicator` and `hideTypingIndicator` methods, so existing custom implementations keep compiling and can override them to show progress before the first token.


vaadin-ai-components-flow-parent/vaadin-ai-core-flow/src/main/java/com/vaadin/flow/component/ai/orchestrator/MessageListWrapper.java
```diff
@@ -46,6 +55,44 @@ public AIMessage addMessage(String text, String userName,
         return message;
     }
 
+    @Override
+    public void showTypingIndicator(String userName) {
+        // Only the entry of the given participant is touched, so that the
+        // typing users set by the application are kept
+        var typingUsers = messageList.getTypingUsers();
+        if (typingUsers.stream()
```
ugur-vaadin: Not sure if we want to currently support multiple users right away, but this and the `hideTypingIndicator` counterpart seems to block it. Maybe we can just remove only a single name if exists, and add can still work even though a user already exists. What do you think?

ugur-vaadin (approved):

